Issue #1, opened 2018-07-10
Template is missing source_ami_id in the variables section


rickard-von-essen:
The template reference ```{{ user `source_ami_id`}}``` but no such user variable is declared. The `Makefile` also passes this as a variable.

rickard-von-essen:
Hmm, this seems like a bug in packer...

rickard-von-essen:
Packer issue hashicorp/packer#6473

mbartsch:
It works for me, with packer 1.2.4, which version of packer are you running?

rickard-von-essen:
Yeah, I realise it works, but it shouldn't, hence I filed hashicorp/packer#6473. When that is fixed this won't work.

mbartsch:
Awesome, I'm not an expert on github, but I added the fix to my previous PR , see issue #2 for details.

PR #2, opened 2018-07-11
added nfs-utils, and fix the way docker is installed

mbartsch:
*Issue #, if available:* #1 

*Description of changes:*
Docker on Amazon Linux 2 is a amazon linux extra and not part of the 'core' distribution, to install docker you need to use amazon-linux-extras command.

Also added nfs-utils to the worker image as in my test I'm using EFS to store data

Also added the source_ami_id variable, as an empty one, as mention in Issue #1 

### DIFF
--- a/install-worker.sh
+++ b/install-worker.sh
@@ -16,6 +16,7 @@ sudo yum update -y
 sudo yum install -y \
     aws-cfn-bootstrap \
     conntrack \
+    nfs-utils \
     curl \
     socat \
     unzip \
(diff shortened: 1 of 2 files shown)


rickard-von-essen:
Great! 👍

install-worker.sh
```diff
@@ -44,7 +45,7 @@ sudo systemctl enable iptables-restore
 ################################################################################
 
 sudo yum install -y yum-utils device-mapper-persistent-data lvm2
-sudo yum install -y docker
+sudo amazon-linux-extras install -y docker
```
jhaynes: 👍 
In the interest of being explicit, I might write this as two separate lines:
```
sudo yum install -y docker
```

mbartsch: I make the changes as your comment, and fully agree is a lot more clear that way

micahhausler:
Hey thanks for the PR! We actually needed to pin docker to `17.06.x`, so we did that in #15 (and added nfs-utils).

Issue #5, opened 2018-07-12
Automatically create node labels.


dennisjohnson:
Autoscaling is a problem when using node selector, affinity or anti-affinity.
Scaling a group up would not make more pods availible, replacing unhealthy instances could leave a pod with no valid node.
A feature allowing node labels to be created when new instances were created by autoscaling would be ideal.
One possibility is to use the existing tags resource, used in a similar way, and just duplicate what's there as labels for the node inside Kubernetes. Autoscaling group tags would appear as labels on newly registered nodes.

micahhausler:
Thanks for opening this issue. We're tracking this in upstream kubernetes https://github.com/kubernetes/kubernetes/issues/63022 so I'll close this request out here.

PR #6, opened 2018-07-12
adds conforming tags as Kubernetes labels at registration time

dennisjohnson:
*Issue #, if available:*
https://github.com/awslabs/amazon-eks-ami/issues/5

*Description of changes:*
adds conforming instance tags (and therefore autoscaling group tags) as Kubernetes labels at registration time
"conforming" - regex provided by kubernetes documentation to validate tag's name as a valid label name.
Though Value has less strict requirements, I applied the same regex as a starting point to make sure values that would break it ( ',' & '=' in addition to the usual suspects) could not be included.


dennisjohnson:
have something very similar fully functional in our EKS cluster, but should not have opened w/o testing. Will reopen once I can test this.

Issue #7, opened 2018-07-12
Update Makefile default AMI to LTS AL2 AMI


jhaynes:
Today, the [Makefile](https://github.com/awslabs/amazon-eks-ami/blob/19279be9fa95dce0057474609fc28f094fa4cf6d/Makefile#L4) references the AL2 Candidate AMI as the default. We should update that to reference the latest [LTS AL2 release](https://aws.amazon.com/amazon-linux-2/release-notes/).

mbartsch:
the other issue I saw, is that the Makefile will only work in a us-west-2 region. I made a few changes to allow to pass the region as an option to make
make REGION=[us-east-1|us-west-2] ami

PR #10, opened 2018-07-13
Ami version

mbartsch:
Issue #, if available:
#7

Description of changes:
Add the latest AMI for the two supported EKS regions, also allow REGION to be passed as a parameter to the make.

Sorry for the previous pull request. I didn't realize I was pushing all the other changes too.


rickard-von-essen:
A better, general solution for this is to remove `region` altogether from the template and depend on env var `AWS_REGION`. And at the same time change from `source_ami` to `source_ami_filter`.

PR #12, opened 2018-07-19
Fixed docker installation command

toniblyx:
amazon-linux-extras install has to be used instead of yum

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/install-worker.sh
+++ b/install-worker.sh
@@ -44,7 +44,7 @@ sudo systemctl enable iptables-restore
 ################################################################################
 
 sudo yum install -y yum-utils device-mapper-persistent-data lvm2
-sudo yum install -y docker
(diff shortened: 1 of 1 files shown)


hobbsh:
This is done in #2 but merging around here seems to move slow

toniblyx:
OK thanks for let me know.

micahhausler:
Hey thanks for the PR! We actually need to pin the Docker version to `17.06.x`, and completed that in #15

hobbsh:
I was wondering why this was not done (pinning to 17.06) and could also not figure out how to install 17.06 with amazon-linux-extras - the syntax is a bit unconventional. Thanks for the update @micahhausler

PR #19, opened 2018-08-09
modify dockerd configuration

schndr:
*Issue #, if available:*

By default, `dockerd` is missing basic security parameters. I would like to add  specific `dockerd` parameters to make the daemon more secure and to build a solid and secure foundation for executing containerized workloads.

*Description of changes:*

`--no-new-privileges` Prevents LSMs like SELinux from transitioning to process labels that have access not allowed to the current process.

`--iptables=false` Docker daemon service start requires iptables rules to be enabled before it starts. Any restarts of iptables during docker daemon operation may result in losing docker-created rules.

`--live-restore` Enables full support of daemon-less containers in docker. It ensures that docker does not stop containers on shutdown or restore and properly reconnects to the container when restarted.

`--userland-proxy=false` The docker daemon starts a userland proxy service for port forwarding whenever a port is exposed. Where hairpin NAT is available, this service is generally superfluous to requirements and can be disabled.

`--log-level=info` Ensure the logging level is set to `info`.

### DIFF
--- a/install-worker.sh
+++ b/install-worker.sh
@@ -53,6 +53,8 @@ sudo yum install -y yum-utils device-mapper-persistent-data lvm2
 sudo amazon-linux-extras enable docker
 sudo yum install -y docker-17.06*
 sudo usermod -aG docker $USER
+sudo sed -i -- 's!ExecStart=/usr/bin/dockerd!ExecStart=/usr/bin/dockerd --host=fd:// $DOCKER_OPTS!g' /usr/lib/systemd/system/docker.service
+sudo mv $TEMPLATE_DIR/docker /etc/default/docker
 
 # Enable docker daemon to start on boot.
 sudo systemctl daemon-reload
--- a/files/docker
+++ b/files/docker
@@ -1,1 +1,1 @@
-DOCKER_OPTS="--pidfile /var/run/docker.pid --no-new-privileges --iptables=false --log-level=warn --live-restore --userland-proxy=false"
+DOCKER_OPTS="--pidfile /var/run/docker.pid --no-new-privileges --iptables=false --log-level=info --live-restore --userland-proxy=false"
--- a/files/docker
+++ b/files/docker
@@ -1,0 +1,1 @@
+DOCKER_OPTS="--pidfile /var/run/docker.pid --no-new-privileges --iptables=false --log-level=warn --live-restore --userland-proxy=false"


micahhausler:
Thanks for the PR! We'll review this and get back on any changes we need.

install-worker.sh
```diff
@@ -53,6 +53,8 @@ sudo yum install -y yum-utils device-mapper-persistent-data lvm2
 sudo amazon-linux-extras enable docker
 sudo yum install -y docker-17.06*
 sudo usermod -aG docker $USER
+sudo sed -i -- 's!ExecStart=/usr/bin/dockerd!ExecStart=/usr/bin/dockerd --host=fd:// $DOCKER_OPTS!g' /usr/lib/systemd/system/docker.service
```
AnakinFoxe: - `EnvironmentFile=-/etc/default/docker` is needed before `ExecStart`
- It seems `--host=fd://` does not work with the docker on the linux aws is using.

jahkeup: Rather than manipulating the text in the unit (especially as the packaged copy, edited here, could be overwritten) this could be added to a drop in unit in `/etc/systemd/system/docker.service.d/eks-customizations.conf` (see [systemd manpages for details](https://www.freedesktop.org/software/systemd/man/systemd.unit.html#id-1.14.3)).

```bash
install -m 644 -D /dev/null /etc/systemd/system/docker.service.d/eks-customizations.conf
cat > /etc/systemd/system/docker.service.d/eks-customizations.conf <<EOF
[Service]
# Unset the ExecStart for overriding
ExecStart=
ExecStart=/usr/bin/dockerd --host=fd:// $DOCKER_OPTS
EOF
```

deiwin: Could also drop the configuration into `/etc/docker/daemon.json`: https://docs.docker.com/v17.06/engine/reference/commandline/dockerd/#miscellaneous-options

micahhausler:
Thank you for the PR but we're not going to incorporate these changes. 

* `--no-new-privileges` Certain customers will have good reasons for needing additional privileges
* `--iptables=false` Docker containers not run by Kubernetes will not get proper networking. While this could be desired for certain hardened production environments, this is a breaking change in behavior we don't want to introduce
* `--live-restore` This could be a good addition, but we'd rather see this change in a scoped down PR, using the `/etc/docker/daemon.json` config file.
* `--userland-proxy=false` This would be another case of breaking behavior. Kubernetes relies on the userland-proxy for `hostPort` on container ports (See kubernetes/kubernetes#24876)
* `--log-level=info`  This is the default and unneeded

You are welcome to build your own AMI and set these options for your workers.

Issue #13, opened 2018-07-24
[discussion] Default root size - 20GB


ozbillwang:
I run EKS cluster for one week, already get "no space" issue on EKS workers.

Found the code here that the default root size is 20GB.

https://github.com/awslabs/amazon-eks-ami/blob/master/eks-worker-al2.json#L22

          "volume_size": 20,

Should we increase it to 50GB or more? 

I can update my codes to start with more space, but 20GB is too small, which I recommend to update in this AMI directly.

micahhausler:
This can be overridden when you launch a node group by editing the CloudFormation template, 

```yaml
  NodeLaunchConfig:
    Type: AWS::AutoScaling::LaunchConfiguration
    Properties:
      ...
      BlockDeviceMappings:
      - DeviceName: /dev/xvda
        Ebs:
          DeleteOnTermination: true
          VolumeType: gp2
          VolumeSize: 50
```

ozbillwang:
@micahhausler 

Thanks.

I know how to custom this option, but my question is, should we increase the default value directly from source here?

micahhausler:
I think we’ll keep the default to 20. A more advanced configuration could include mounting a second volume for docker, so the kubelet can still operate even if the primary disk is full

flegler:
@micahhausler: is there a way to increase the size of 20GB in an already deployed EKS cluster?

njfix6:
For kops, they use a much higher default (120GB). In my opinion, this is way to low to run in with any sort of production traffic.

whereisaaron:
How does network traffic increase disk usage @njfix6?

I think it matters more what the nature of the workload is. For example we have node pools where the workload uses no (local) disk at all. If the root disk is big enough for a dozen images and 10MB log/container, then it is plenty big enough. In other cases all writeable disk are separate mounts as @micahhausler mentioned. For those workloads 120 GB or even 50 GB would be a waste of many TB of unused storage.

PR #20, opened 2018-08-10
Add ability to modify node root volume size + aws stack name labels

lowerzero:
*Issue #, if available:*

By default its 20 Gb which can be too low for some workloads.

### DIFF
--- a/amazon-eks-nodegroup.yaml
+++ b/amazon-eks-nodegroup.yaml
@@ -259,6 +259,7 @@ Resources:
           !Sub |
             #!/bin/bash
             set -o xtrace
+            export KUBELET_EXTRA_ARGS="--node-labels aws-stack-name=${AWS::StackName}"
             /etc/eks/bootstrap.sh ${ClusterName}
             /opt/aws/bin/cfn-signal --exit-code $? \
                      --stack  ${AWS::StackName} \
--- a/amazon-eks-nodegroup.yaml
+++ b/amazon-eks-nodegroup.yaml
@@ -84,6 +84,11 @@ Parameters:
     Description: Maximum size of Node Group ASG.
     Default: 3
 
+  NodeVolumeSize:
+    Type: Number
+    Description: Node volume size
+    Default: 20
+
   ClusterName:
     Description: The cluster name provided when the cluster was created. If it is incorrect, nodes will not be able to join the cluster.
     Type: String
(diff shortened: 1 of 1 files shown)


lowerzero:
Also with multiple nodegroups it is handy to be able select nodes from specific cloudformation stack

amazon-eks-nodegroup.yaml
```diff
@@ -242,11 +248,18 @@ Resources:
...
+          Ebs:
+            VolumeSize: !Ref NodeVolumeSize
+            VolumeType: gp2
+            DeleteOnTermination: true
       UserData:
         Fn::Base64:
           !Sub |
             #!/bin/bash
             set -o xtrace
+            export KUBELET_EXTRA_ARGS="--node-labels aws-stack-name=${AWS::StackName}"
```
micahhausler: Can you please remove this line?

amazon-eks-nodegroup.yaml
```diff
@@ -84,6 +84,11 @@ Parameters:
     Description: Maximum size of Node Group ASG.
     Default: 3
 
+  NodeVolumeSize:
+    Type: Number
+    Description: Node volume size
+    Default: 20
```
max-rocket-internet: 20GB is very low to have as a default. For example, GKE uses 100GB.

max-rocket-internet:
> Also with multiple nodegroups it is handy to be able select nodes from specific cloudformation stack

This is a really great idea 👍

Would be useful for when you have workers with extra resources for running specific containers, or workers with fast storage, or for when running on spot instances. We could then use a `nodeSelector`.

@micahhausler how come you wanted to remove the `KUBELET_EXTRA_ARGS` line from the user-data?

amazon-eks-nodegroup.yaml
lowerzero: This is default value from AMI

max-rocket-internet: But that doesn't mean it should be 20 here. It can be set to any value.

deiwin: @micahhausler has previously mentioned that they want to stick to 20: https://github.com/awslabs/amazon-eks-ami/issues/13#issuecomment-408267434

max-rocket-internet: Seems a bit old fashioned but OK. At least with this PR we can change it now.

micahhausler:
Thanks! This is great. 

@max-rocket-internet with #23 you can add labels as a parameter to the CloudFormation stack. See [my comment](https://github.com/awslabs/amazon-eks-ami/issues/22#issuecomment-413992153) in #22, so you can add whatever labels and taints you'd like.

PR #26, opened 2018-08-23
Adds T3 Instance Types

tomweston:
**Description of changes:**
Adds T3 Instance types


tomweston:
## Adding T3 Instances failed

Running the Config Map to add the new nodes into the cluster causes several problems:

1. Nodes fail to be added to the cluster
2. All internal and external services become unresponsive, Including the dashboard.
3. Deleting the configmap and cloudformation nodes does not resolve the issue

PR #42, opened 2018-09-07
label spot instances automatically for spot-termination-handler

delgod:
*Description of changes:*
if I want to use spot-termination-handler it is needed to pass `--kubelet-extra-args --node-labels=lifecycle=Ec2Spot` string to `BootstrapArguments` parameter.
it is simple to add lifecycle label automatically in bootstrap.sh


micahhausler:
We'd rather not be opinionated about the labels/taints that are applied. You can add `--node-labels=lifecycle=Ec2Spot` as an argument to a custom user-data or as a Parameter to the CloudFormation template we provide

PR #43, opened 2018-09-07
remove packer key before shutdown

delgod:
*Description of changes:*
If I use result AMI I see ssh key which used during image creation inside `~/.ssh/authorized_keys` file.
Not a big issue, but I feel more secure without it.


rickard-von-essen:
Did you test this? I don't expect it to work without setting [`skip_cleanup`](https://www.packer.io/docs/provisioners/shell.html#skip_clean).

delgod:
Hi @rickard-von-essen,

yes, works fine, checked with packer 1.2.2 and 1.2.5 on MacOS.

micahhausler:
Thanks!

PR #48, opened 2018-09-12
Scrub temporary build-time keypair

micahhausler:
This keypair is an artifact of the build process- it's a temporary key and does not represent a secrity risk. However, it is a best practice to remove these keys before publishing, so we've added a step to remove it for all future AMI

### DIFF
--- a/install-worker.sh
+++ b/install-worker.sh
@@ -136,6 +136,9 @@ sudo rm -rf \
 
 # Clean up files to reduce confusion during debug
 sudo rm -rf \
+    /etc/machine-id \
+    /etc/ssh/ssh_host* \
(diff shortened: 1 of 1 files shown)


install-worker.sh
```diff
@@ -137,6 +137,7 @@ sudo rm -rf \
 # Clean up files to reduce confusion during debug
 sudo rm -rf \
     /home/ec2-user/.ssh/authorized_keys \
+    /root/.ssh/authorized_keys \
```
delgod: just info: this file needed to show nice message if someone try to run `ssh root@IP`
this file will be regenerated during the first instance start by cloud-init.

PR #56, opened 2018-10-01
add docker-healthcheck systemd units

neerfri:
*Issue #, if available:*
The docker daemons in our clusters occasionally freeze.
We found that all mature kubernetes deployments use some sort of health check script on the nodes to ensure docker is still running and restart it when required.

The script in this PR is taken from https://github.com/kubernetes/kops at https://github.com/kubernetes/kops/blob/e98671c010f3694b164a7eea07f6d5693256fbf3/upup/models/nodeup/docker/_systemd/_debian_family/files/opt/kubernetes/helpers/docker-healthcheck

*Description of changes:*
This adds systemd units `docker-healthcheck.service` and `docker-healthcheck.timer` that will together check docker periodically and restart it when it is not responsive.

### DIFF
--- a/files/docker-healthcheck/docker-healthcheck.service
+++ b/files/docker-healthcheck/docker-healthcheck.service
@@ -1,0 +1,9 @@
+[Unit]
+Description=Run docker-healthcheck once
+
+[Service]
+Type=oneshot
+ExecStart=/opt/docker-healthcheck/docker-healthcheck
+
+[Install]
+WantedBy=multi-user.target
--- a/files/docker-healthcheck/docker-healthcheck.sh
+++ b/files/docker-healthcheck/docker-healthcheck.sh
@@ -1,0 +1,53 @@
+#!/bin/bash
+
+# Taken from:
+# https://github.com/kubernetes/kops/blob/e98671c010f3694b164a7eea07f6d5693256fbf3/upup/models/nodeup/docker/_systemd/_debian_family/files/opt/kubernetes/helpers/docker-healthcheck
+
+# Copyright 2015 The Kubernetes Authors All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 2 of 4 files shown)


max-rocket-internet:
FYI [NPD](https://github.com/kubernetes/node-problem-detector) also checks if docker daemon has hung.

neerfri:
@max-rocket-internet that is true, and NPD also points to [draino]() for remedy but their solution is to drain the node and have it rotated. This script solves the problem in most cases much quicker and without killing the pods running on the faulty node.

This practice exists in many mature kubernetes deployments. see:

* https://github.com/kubernetes/kubernetes/blob/master/cluster/gce/gci/health-monitor.sh
* https://github.com/kubernetes/kops/blob/master/upup/models/nodeup/docker/_systemd/_debian_family/files/opt/kubernetes/helpers/docker-healthcheck
* https://github.com/apprenda/kismatic/issues/855
* https://docs.docker.com/engine/admin/live-restore/

neerfri:
@micahhausler I'd appreciate your comments on this. Thanks!

micahhausler:
Thanks for this PR! This would likely be a good addition, but we'll need to take some time to review and vet it before merging

neerfri:
Sounds great. I like to use code made by people that take the time to think :-)
Thanks for the reply!

asiragusa:
@neerfri kubernetes will soon switch to crictl to control the container runtime and this script doesn't support it. The healthcheck that runs on GCE supports it and might be a better option: https://github.com/kubernetes/kubernetes/blob/master/cluster/gce/gci/health-monitor.sh

max-rocket-internet:
Also worth mentioning [node-problem-detector](https://github.com/kubernetes/node-problem-detector). I think this is already in use on GKE and includes a check for `DockerHung` and some sort of remedy function.

Would be cool if we could unify behind a single approach for this 🙂

neerfri:
@micahhausler any plans regarding this? Thanks!

PR #61, opened 2018-10-09
Feature / Docker Log Rotation

### LINKED ISSUE #36: Does Docker do log rotation?
According to https://kubernetes.io/docs/concepts/cluster-administration/logging/, log rotation for Docker is required to not fill up the disk. GCE has it by default if kube-up.sh is used, but I can't find anything rotating `/var/log/*.log` in this repository. [Here's how it works for gce.](https://github.com/kubernetes/kubernetes/blob/master/cluster/gce/gci/configure-helper.sh#L327)

Am I correct that if I run Docker images on EKS for long enough, it will crash with full disks?

bryantbiggs:
*Issue #, if available:*
Fixes #36 

*Description of changes:*
- Add `daemon.json` file to `/etc/docker/` to enable Docker daemon log rotation using default values shown in the [Docker documentation](https://success.docker.com/article/how-to-setup-log-rotation-post-installation)

The approach above comes from the Docker documentation and is only effective against newly created containers and allows for the defaults to be overridden when launching a container:
```bash
docker run \
    --log-driver json-file \
    --log-opt max-size=10m \
    --log-opt max-file=10 \
    alpine echo hello world
```

A second approach that could also be employed would be to use logrotate (which cannot be overridden from the Docker cli):

```
/var/lib/docker/containers/*/*.log {
  rotate 10
  daily
  compress
  size=10M
  missingok
  delaycompress
  copytruncate
}
```

Either of these configs should protect instances from consuming too much disk space even if in a conservative manner.

Ref:
- https://docs.docker.com/config/containers/logging/json-file/
- https://docs.docker.com/engine/reference/commandline/dockerd/#daemon-configuration-file

### DIFF
--- a/files/daemon.json
+++ b/files/daemon.json
@@ -1,0 +1,7 @@
+{
+    "log-driver": "json-file",
+    "log-opts": {
+        "max-size": "10m",
+        "max-file": "10"
+    }
+}
--- a/install-worker.sh
+++ b/install-worker.sh
@@ -54,6 +54,10 @@ sudo amazon-linux-extras enable docker
 sudo yum install -y docker-17.06*
 sudo usermod -aG docker $USER
 
+# Enable docker daemon log rotation
+# https://success.docker.com/article/how-to-setup-log-rotation-post-installation
+sudo mv $TEMPLATE_DIR/daemon.json /etc/docker/daemon.json
+
 # Enable docker daemon to start on boot.
 sudo systemctl daemon-reload
 sudo systemctl enable docker


nickdgriffin:
Just been looking at this today, /etc/docker/daemon.json isn't picked up automatically - see [my comment](44 seconds ago) for details.

bryantbiggs:
According to the Docker documentation, adding the `daemon.json` file to `/etc/docker/` and then reloading the service will enable the configuration. 

However, I missed this section where it states that you cannot duplicate flags in the config file https://docs.docker.com/engine/reference/commandline/dockerd/#daemon-configuration-file, so this is not a good solution.

Unfortunately, it looks like the best approach might be to use the `args` in the image spec and pass in the flags like:
```yaml
args: ["--log-driver", "json-file", "--log-opt", "max-size=10m", "--log-opt" "max-file=10"]
```
https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/

Therefore I don't see value in this PR and will close.

nickdgriffin:
Yeah, the documentation is wrong - the daemon doesn't read the file without "--config-file" being specificed, at least for the version that the AMI comes with.

max-rocket-internet:
OK but then why not just change this PR to make dockerd get the correct arguments in order to enable the log rotation we need?

For example, either:
1. Add `DOCKER_OPTS` to `/etc/default/docker` with something like `--config-file /etc/docker/daemon.json`
2. Run some `sed` command against `/usr/lib/systemd/system/docker.service`
3. Or something else

> Unfortunately, it looks like the best approach might be to use the args in the image spec 

This is something completely different. These `args` are passed to the command that is run *inside* the container, not arguments passed to docker when starting the container.

bryantbiggs:
Because this is the image for node creation, I was working from the premise where users don't have to make their own custom image based off this image or manually modify the instance/node created by this image, but instead, they could configure it from within the Kubernetes config files. 

Using the docker daemon route means the whole node is affected and I don't see how that scales well. I was hoping to set a sane default for all nodes that are provisioned in the cluster, with optional arguments provided during container creation if the user wishes to use a different log rotation setting. The "default" would prevent the scenario where the node storage reaches 100% due to log storage, while the optional args override would allow users to store more/less logs based on the service they are running. Otherwise, why use Kubernetes if you are still mucking in the details of the underlying server? From my view, you generally shouldn't care where your pod/container ends up running (on which node) but if you are changing the daemons settings then you are making nodes into snowflakes or you are changing the log rotation settings for the entire cluster.

Of course, this is all very opinionated on my part, but that was the goal/idea I was working towards. Please let me know if I am incorrect in any of my assumptions as I think this solution is extremely beneficial to all EKS users.

max-rocket-internet:
> but instead, they could configure it from within the Kubernetes config files.

Ahh OK, I misunderstood this intention.

IMO it 100% needs to be solved at the node level. Perhaps I'll make a PR tomorrow.

max-rocket-internet:
> Please let me know if I am incorrect in any of my assumptions

I can't say if your assumptions are correct or not but it should be made very difficult to fill the file system with container logs with the default configuration. Currently, with no logrotate config in this repo, it's easy 🙂

dlaidlaw:
In my opinion:

- the node defaults do not allow (or practically limit) the possibility of 100% disk usage due to container logs
- the override at the service level is secondary, but highly desirable

We do not want to have nodes start failing with 100% disk usage simply because container logs are not being cleaned up.

It would also help if the images, containers, volumes are pruned regularly, again with good defaults that could be overridden.

deiwin:
> It would also help if the images, containers, volumes are pruned regularly, again with good defaults that could be overridden.

Kubelet already takes care of that: https://v1-10.docs.kubernetes.io/docs/concepts/cluster-administration/kubelet-garbage-collection/

dlaidlaw:
> Kubelet already takes care of that: https://v1-10.docs.kubernetes.io/docs/concepts/cluster-administration/kubelet-garbage-collection/

Great! Thanks!

gflarity:
@bryantbiggs 

> Unfortunately, it looks like the best approach might be to use the `args` in the image spec and pass in the flags like:
> 
> ```yaml
> args: ["--log-driver", "json-file", "--log-opt", "max-size=10m", "--log-opt" "max-file=10"]
> ```
> 
> https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/
> 
> Therefore I don't see value in this PR and will close.

I think you're misunderstanding the ARGS in the image spec. These the args to the process that runs in the container, not to docker itself.  

It's SHOCKING that this feature hasn't been implemented. I'm going to open a duplicate.

deiwin:
> It's SHOCKING that this feature hasn't been implemented. 

But it has: https://github.com/awslabs/amazon-eks-ami/pull/74

gflarity:
Looks like it's not in a public AMI yet though?

max-rocket-internet:
It is, it's included in `amazon-eks-node-1.10-v20181210`

PR #69, opened 2018-10-12
Fix kube-proxy logrotate

dpiddockcmp:
Actually rotate kube-proxy log.
Does not error if log is missing.

Current behaviour:
```
# logrotate -v /etc/logrotate.conf 
including /etc/logrotate.d
Ignoring kube-proxy because the file owner is wrong (should be root).
```

New behaviour:
```
# logrotate -v /etc/logrotate.conf
including /etc/logrotate.d 
reading config file kube-proxy

rotating pattern: /var/log/kube-proxy.log  after 1 days (5 rotations)
empty log files are rotated, old logs are removed
considering log /var/log/kube-proxy.log
  log /var/log/kube-proxy.log does not exist -- skipping
```

### DIFF
--- a/files/logrotate-kube-proxy
+++ b/files/logrotate-kube-proxy
@@ -1,4 +1,5 @@
 /var/log/kube-proxy.log {
+    missingok
     rotate 5
     daily
     compress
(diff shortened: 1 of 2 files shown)


micahhausler:
Thanks!

PR #80, opened 2018-10-31
kubelet config files should be owned by root

jpb:
Previously, the ec2-user account owned these files and had permission to write to them.

*Issue #, if available:*

*Description of changes:*

Ownership of `/etc/systemd/system/kubelet.service` and `/var/lib/kubelet/kubeconfig` will now be `root:root` instead of `ec2-user:ec2-user`.


micahhausler:
Thanks! This is a good change, we'll get this merged in the next few days once we cut the next release

PR #93, opened 2018-11-13
pre V26 updates

### LINKED ISSUE #78: Kubelet cluster-dns parameter is set incorrectly
The IP address of the `kube-dns` service is picked according to the IP address of the node: https://github.com/awslabs/amazon-eks-ami/blob/v24/files/bootstrap.sh#L105

Normally this works fine, but on one of my clusters, the `kube-dns` was running on `172.20.0.10`, and the kubelets were configured to point at `10.100.0.10`. I'm not sure how `kube-dns` decides which IP address to use, but however that works, the kubelet should use the exact same logic to pick IP address to use for `cluster-dns`.

micahhausler:
*Issue #, if available:*

Fixes #78 

*Description of changes:*

* Moved kubelet config file to `/etc/kubernetes/kubelet/`
* Correctly set DNS IP in for instances in a secondary CIDR
* Fix instance type IP limits


files/bootstrap.sh
```diff
@@ -99,19 +99,27 @@ sed -i s,CLUSTER_NAME,$CLUSTER_NAME,g /var/lib/kubelet/kubeconfig
 
 ### kubelet.service configuration
 
-INTERNAL_IP=$(curl -s http://169.254.169.254/latest/meta-data/local-ipv4)
-INSTANCE_TYPE=$(curl -s http://169.254.169.254/latest/meta-data/instance-type)
+MAC=$(curl -s http://169.254.169.254/latest/meta-data/network/interfaces/macs/ -s | head -n 1)
+CIDRS=$(curl -s http://169.254.169.254/latest/meta-data/network/interfaces/macs/$MAC/vpc-ipv4-cidr-blocks)
+set +e
+TEN_RANGE=$(echo $CIDRS | grep -c  '10\..*')
```
nckturner: Does this work if the IP address has `10.` in the middle?

ChrisCooney:
Can I ask when version 26 of the AMI will be available? We have got a cluster running in a secondary VPC and are facing this exact issue right now. The nameserver in our /etc/resolv.conf is 10.100.0.10, despite our cluster running in 100.64.0.0/16.

micahhausler:
We won't be releasing for another 2 weeks (as you can imagine, we'll be very busy with reInvent!) In the meantime, you can you build an AMI for yourself and use it with your cuslter

ChrisCooney:
Cheers Micah!

Issue #87, opened 2018-11-07
Add support for M5a instances


arjanschaaf:
I tested the latest AMI with the newly release m5a instance family. My setup works with m5 instances, but the kubelet fails to start automatically. It seems to work after restarting it by hand using `sudo systemctl start kubelet` though.

micahhausler:
I haven't had any issues starting kubelet on m5a, but the CNI provider doesn't seem to support it yet. Did you do any modification to the AMI?

arjanschaaf:
I create a launch config like so:
```
aws autoscaling create-launch-configuration --launch-configuration-name eks-test-nodes-14112018 --iam-instance-profile eks-node --image-id ami-00c3b2d35bddd4f5c --instance-type m5a.large --key-name eks --ebs-optimized --security-groups <security-group-id> --user-data file://eks-test-nodes-userdata.sh --block-device-mappings "[{\"DeviceName\": \"/dev/xvda\",\"Ebs\":{\"VolumeSize\":64,\"VolumeType\":\"gp2\"}}]"
```

The contents of the user-data file are:
```
#!/bin/bash -xe
/etc/eks/bootstrap.sh --kubelet-extra-args --node-labels=nodegroup=test-nodes my-kubernetes-cluster
```

Then I launch the autoscaling group:
```
aws autoscaling create-auto-scaling-group --auto-scaling-group-name eks-test-nodes --launch-configuration-name eks-test-nodes-14112018 --min-size 1 --desired-capacity 1 --max-size 1 --vpc-zone-identifier <subnet-id> --tags Key=Name,Value=eks-test-node,PropagateAtLaunch=true Key="kubernetes.io/cluster/my-kubernetes-cluster",Value=shared,PropagateAtLaunch=true
```

If you login to this new node and check the status of:
- docker.service : active
- kubelet.service : dead
- multi-user.target: active

To answer your questions:

- What specific instance type are you launching?
m5a.large
- What region are you launching it in?
eu-west1
- What version of the EKS AMI are you using?
ami-00c3b2d35bddd4f5c
- Are the --node-labels the only extra args you are passing to the kubelet?
No other arguments then the `--node-labels` you see in the user-data

**Please note** : If I launch an autoscaling group with the exact same configuration expect for one change: use the **m5.large** instance type instead of **m5a.large** the kubelet is started as expected. Hence my conclusion that the m5a instance family is somehow not supported yet.

ejfinkel:
[...]

PR #94, opened 2018-11-14
Gracefully handle unknown instance types

micahhausler:
*Issue #, if available:*

* Addresses #87

*Description of changes:*

Add exit trap printing line number

### DIFF
--- a/files/bootstrap.sh
+++ b/files/bootstrap.sh
@@ -4,6 +4,11 @@ set -o pipefail
 set -o nounset
 set -o errexit
 
+err_report() {
+    echo "Exited with error on line $1"
+}
+trap 'err_report $LINENO' ERR
+
 IFS=$'\n\t'
 
 function print_help {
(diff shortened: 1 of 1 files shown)


files/bootstrap.sh
```diff
@@ -117,9 +122,13 @@ INSTANCE_TYPE=$(curl -s http://169.254.169.254/latest/meta-data/instance-type)
 
 if [[ "$USE_MAX_PODS" = "true" ]]; then
     MAX_PODS_FILE="/etc/eks/eni-max-pods.txt"
+    set +o pipefail
     MAX_PODS=$(grep $INSTANCE_TYPE $MAX_PODS_FILE | awk '{print $2}')
+    set -o pipefail
     if [[ -n "$MAX_PODS" ]]; then
         echo "$(jq .maxPods=$MAX_PODS $KUBELET_CONFIG)" > $KUBELET_CONFIG
+    else
+        echo "No entry for $INSTANCE_TYPE in $MAX_PODS_FILE. Not setting max pods for kubelet"
```
mogren: So from now on, an instance type in `/etc/eks/eni-max-pods.txt`, but with no `MAX_PODS` will work and just not set a `maxPods` limit? 

And missing instances will print `"Exited with error on line 126"`?

micahhausler: No, we'll just output the message when an instance type isn't available. 

The trap will just be there to help debug any other future error case

PR #97, opened 2018-11-15
Added node ASG update policy parameters

micahhausler:
*Issue #, if available:*

N/A

*Description of changes:*

Added node ASG update parameters

### DIFF
--- a/amazon-eks-nodegroup.yaml
+++ b/amazon-eks-nodegroup.yaml
@@ -1,6 +1,6 @@
 ---
 AWSTemplateFormatVersion: '2010-09-09'
-Description: 'Amazon EKS - Node Group - Released 2018-08-30'
+Description: 'Amazon EKS - Node Group'
 
 Parameters:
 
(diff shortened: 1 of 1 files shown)


amazon-eks-nodegroup.yaml
```diff
@@ -155,6 +175,10 @@ Metadata:
           - NodeGroupName
           - NodeAutoScalingGroupMinSize
           - NodeAutoScalingGroupMaxSize
+          - NodeAutoScalingGroupDesiredCapacity
+          - NodeAutoScalingGroupUpdatePolicyMaxBatchSize
+          - NodeAutoScalingGroupUpdatePolicyMinInstancesInService
+          - NodeAutoScalingGroupUpdatePolicyPauseTime
```
micahhausler: This is getting a little dense, any opinions about moving the UpdatePolicy parameters to their own ParameterGroup for UI cleanliness?

amazon-eks-nodegroup.yaml
```diff
@@ -107,8 +107,28 @@ Parameters:
...
+
+  NodeAutoScalingGroupUpdatePolicyMinInstancesInService:
+    Type: Number
+    Description: MinInstancesInService for the NodeGroup UpdatePolicy
+    Default: 1
+
+  NodeAutoScalingGroupUpdatePolicyPauseTime:
+    Type: String
+    Description: PauseTime for the NodeGroup UpdatePolicy
+    Default: "PT0S"
```
nckturner: What is PT0S?  Do we need some documentation on what values are accepted here?

micahhausler: It is the default, zero value. Added a doc url.

amazon-eks-nodegroup.yaml
mogren: Maybe all 6 ASG settings in that case.

PR #108, opened 2018-11-27
Consolidate yaml style

timecode:
*Issue*
As a new user, I would appreciate concise clarity and consistency in the style used throughout a supplied template.

*Description of changes:*
Style changes ONLY

### DIFF
--- a/amazon-eks-nodegroup.yaml
+++ b/amazon-eks-nodegroup.yaml
@@ -18,86 +18,86 @@ Parameters:
     Default: t3.medium
     ConstraintDescription: Must be a valid EC2 instance type
     AllowedValues:
-    - t2.small
-    - t2.medium
-    - t2.large
-    - t2.xlarge
-    - t2.2xlarge
-    - t3.nano
-    - t3.micro
-    - t3.small
-    - t3.medium
-    - t3.large
-    - t3.xlarge
-    - t3.2xlarge
-    - m3.medium
-    - m3.large
-    - m3.xlarge
-    - m3.2xlarge
-    - m4.large
-    - m4.xlarge
-    - m4.2xlarge
-    - m4.4xlarge
-    - m4.10xlarge
-    - m5.large
-    - m5.xlarge
-    - m5.2xlarge
-    - m5.4xlarge
-    - m5.12xlarge
-    - m5.24xlarge
-    - c4.large
-    - c4.xlarge
-    - c4.2xlarge
-    - c4.4xlarge
-    - c4.8xlarge
-    - c5.large
-    - c5.xlarge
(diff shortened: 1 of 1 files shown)


amazon-eks-nodegroup.yaml
```diff
@@ -217,7 +210,7 @@ Resources:
       Description: Allow node to communicate with each other
       GroupId: !Ref NodeSecurityGroup
       SourceSecurityGroupId: !Ref NodeSecurityGroup
-      IpProtocol: '-1'
+      IpProtocol: -1
```
mogren: Is it really ok to remove the quotes here? `IpProtocol` takes a string argument, and `-` should be quoted since it's an indicator character.

amazon-eks-nodegroup.yaml
```diff
@@ -178,23 +176,19 @@ Resources:
...
-        Version: '2012-10-17'
+        Version: 2012-10-17
         Statement:
-        - Effect: Allow
-          Principal:
-            Service:
-            - ec2.amazonaws.com
-          Action:
-          - sts:AssumeRole
-      Path: "/"
```
micahhausler: Please add the `Path` back. While this is the default, changing this value will replace the role and could cause a disruptive change

amazon-eks-nodegroup.yaml
micahhausler: This should be enclosed in quotes

amazon-eks-nodegroup.yaml
```diff
@@ -283,31 +276,30 @@ Resources:
...
       Tags:
-      - Key: Name
-        Value: !Sub "${ClusterName}-${NodeGroupName}-Node"
-        PropagateAtLaunch: 'true'
-      - Key: !Sub 'kubernetes.io/cluster/${ClusterName}'
-        Value: 'owned'
-        PropagateAtLaunch: 'true'
+        - Key: Name
+          Value: !Sub ${ClusterName}-${NodeGroupName}-Node
+          PropagateAtLaunch: true
```
micahhausler: I'll have to check if this unquoted boolean is accepted by CloudFormation

timecode: Please verify, but from my observation it gets parsed, correctly, by CloudFormation's yaml parser. YAML itself considers Booleans to be expressed with true and false (no quotes).

amazon-eks-nodegroup.yaml
timecode: Path: "/" re-included

timecode:
@micahhausler in answer to your question, these are hand curated change

micahhausler:
Thanks!

PR #109, opened 2018-11-30
disable docker0 bridge as it is not used on EKS

mbartsch:
This will reduce the number of possible conflicts to deploy as docker0 is disabled.
*Issue #, if available:* None

*Description of changes:*
As many users may already have 172.17.0.0/16 in their network configuration, this cause some issues with that setup.
A solution is to disable the docker0 bridge , as it is not used should not create any impact.

### DIFF
--- a/files/docker-daemon.json
+++ b/files/docker-daemon.json
@@ -1,5 +1,5 @@
 {
-  "bip": "203.0.113.1/30",
+  "bridge": "none",
   "log-driver": "json-file",
   "log-opts": {
     "max-size": "10m",
--- a/files/docker-daemon.json
+++ b/files/docker-daemon.json
@@ -1,4 +1,5 @@
 {
+  "bip": "203.0.113.1/30",
   "log-driver": "json-file",
   "log-opts": {
     "max-size": "10m",


mgresko:
change bridge to none made our image unusable with EKS.  None of the containers could resolve dns, get to default gateway etc...  Not understanding the exact problem this is attempting to solve?  It seems that without docker0 nothing works.  Probably missing something fundamental.

micahhausler:
@mgresko are you running containers on the host outside of K8s? What CNI are you using?

mgresko:
We are using the aws-cni.  The issue we saw is that coredns would not start and it was because it could not get to 172.20.0.1 (cluster ip).  Once we added the bridge back all was well.

micahhausler:
@mgresko can you open a new issue for that? Can you also add your VPC CIDR range?

johnjeffers:
Wanted to also note that this is a breaking change if you use docker-in-docker. For example, we run Kubernetes build slaves in Jenkins. Those build slaves sometimes run jobs that have `docker build` commands in them, so it runs a container inside the container. This breaks networking in the docker-in-docker container. To fix it, you have to add the `--network host` param to the `docker build` command.

rprashad:
Yup same here, had to gut the `/etc/docker/daemon.json` config on every worker node and reboot to make it work.  We need this to do docker-in-docker image builds using Jenkins w/ Kubernetes plugin.

Not a good assumption.

left4taco:
Had to say, just because it's not used by EKS doesn't mean you guys have to disable it...We need to do some hack to bring the default bridge back.

PR #110, opened 2018-12-04
Always restart kubelet

devkid:
We observed multiple instances of the kubelet just exiting without stating any reason, exit code 0.
It makes sense to restart it on all occasions, not only when it failed.

### DIFF
--- a/files/kubelet.service
+++ b/files/kubelet.service
@@ -13,7 +13,7 @@ ExecStart=/usr/bin/kubelet --cloud-provider aws \
     --container-runtime docker \
     --network-plugin cni $KUBELET_ARGS $KUBELET_EXTRA_ARGS
 
-Restart=on-failure
+Restart=always
 RestartSec=5
 KillMode=process
 


micahhausler:
Thanks for this PR. Do you have any other information about the state of the kubelet exiting? What version of the AMI are you using? Any custom kubelet configuration?

devkid:
What we observed was just that the kubelet was stopped (`inactive` in `systemctl status kubelet`, exit code 0), and `journalctl -u kubelet` ended with normal log messages, nothing indicating a crash or any other reason. At the same time, we had another pod on the instance go OOM, so that might or might not be related. Anyways, is there a valid reason for *not* restarting the kubelet, independent of exit code?

travcunn:
@devkid Could you provide the output of `systemctl status kubelet` during a time when kubelet is stopped? We fixed an issue in #137 that makes systemd restarts kubelet in case the process is terminated with `signal=PIPE` and I want to check if this also fixes your use case.

devkid:
I will record the exact output if this happens again. I assume you might be right with `SIGPIPE`. Is it possible that journald was OOM killed in this low-memory situation?

micahhausler:
Closing with no response. Feel free to reopen if you find anything

Issue #113, opened 2018-12-10
Add support `p3dn.24xlarge` instance type


JackWindows:
**What happened**: Get the following error when trying to spin up a `p3dn.24xlarge` EC2 instance with the GPU AMI.
```
scale-up failed (UnsupportedOperation): The instance configuration for this AWS Marketplace product is not supported. Please see the AWS Marketplace site for more information about supported instance types, regions, and operating systems.
```

**What you expected to happen**: Successfully launch `p3dn.24xlrage` EC2 instance with the GPU AMI.

**How to reproduce it (as minimally and precisely as possible)**: In `us-east-1`, launch a `p3dn.24xlarge` EC2 instance with AMI `ami-0c974dde3f6d691a1`.

**Anything else we need to know?**: The `bootstrap.sh` probably also needs to be adjusted to support `p3dn.24xlarge`

**Environment**:
- AWS Region: `us-east-1`
- Instance Type(s): `p3dn.24xlarge`
- EKS Platform version (use `aws eks describe-cluster --name eks-mhausler-11-16-2018 --query cluster.platformVersion`): `eks.3`
- Kubernetes version (use `aws eks describe-cluster --name eks-mhausler-11-16-2018 --query cluster.version`): `1.10`
- AMI Version: `ami-0c974dde3f6d691a1`
- Kernel (e.g. `uname -a`):
- Release information (run `cat /tmp/release` on a node):

```
```

micahhausler:
As stated on #114, we're waiting on a CNI release with support for this

PR #114, opened 2018-12-10
Add new instance p3dn.24xlarge support

Jeffwan:
*Issue #, if available:*
Address #113 

*Description of changes:*
This change adds new instance support for p3dn.24xlarge and it will unblock customers to use this flavor for EKS cluster.

### DIFF
--- a/files/eni-max-pods.txt
+++ b/files/eni-max-pods.txt
@@ -101,6 +101,7 @@ p2.16xlarge 234
 p3.2xlarge 58
 p3.8xlarge 234
 p3.16xlarge 234
+p3dn.24xlarge 737
 r3.large 29
 r3.xlarge 58
 r3.2xlarge 58
--- a/amazon-eks-nodegroup.yaml
+++ b/amazon-eks-nodegroup.yaml
@@ -79,6 +79,7 @@ Parameters:
     - p3.2xlarge
     - p3.8xlarge
     - p3.16xlarge
+    - p3dn.24xlarge
     - r5.large
     - r5.xlarge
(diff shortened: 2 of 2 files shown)


files/eni-max-pods.txt
```diff
@@ -101,6 +101,7 @@ p2.16xlarge 234
 p3.2xlarge 58
 p3.8xlarge 234
 p3.16xlarge 234
+p3dn.24xlarge 737
```
gautamkmr: whats the number you have used for below expression
# of ENI * (# of IPv4 per ENI - 1)  + 2

Jeffwan: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-eni.html#AvailableIpPerENI
Based on the documentation, it could have up to 15 interfaces and each with 50 IpV4 address. 
15 * (50 - 1) + 2 = 737

Jeffwan:
Not sure the process to add new instance. EKS team would like to test new instance first before rolling out. Let's hold this PR and wait for testing result.

micahhausler:
Thanks for opening this PR. KubeCon is this week and it will take us a few extra days to review this.

hmizuma:
Instance type needs to be supported by CNI https://github.com/aws/amazon-vpc-cni-k8s/pull/275 before it can be added to the node group template.

micahhausler:
As stated above, we need a CNI release with this instance type before we can merge it into the AMI

micahhausler:
Added support in #171

PR #132, opened 2018-12-24
Increase ulimits for docker containers.

agcooke:
*Issue #, if available:*

*Description of changes:*
Increase the default ulimits for containers.

We see that on some of our (monolith) apps we run out of file-descriptors. This increases the limit from 1024 to 2048 for soft-limits. The values for the limits are just chosen, we did not research what they should be. We have not had issues with file-descriptors since making the change.

### DIFF
--- a/files/docker-daemon.json
+++ b/files/docker-daemon.json
@@ -3,5 +3,12 @@
   "log-opts": {
     "max-size": "10m",
     "max-file": "10"
+  },
+  "default-ulimits": {
+    "nofile": {
+      "Name": "nofile",
+      "Soft": 2048,
+      "Hard": 8192
+    }
   }
 }
--- a/install-worker.sh
+++ b/install-worker.sh
@@ -54,6 +54,9 @@ sudo amazon-linux-extras enable docker
 sudo yum install -y docker-17.06*
 sudo usermod -aG docker $USER
 
+# Remove all options from sysconfig docker.
+sudo sed -i '/OPTIONS/d' /etc/sysconfig/docker
+
 sudo mkdir -p /etc/docker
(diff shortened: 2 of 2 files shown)


micahhausler:
default-ulimits is not available in Docker 17.03: https://docs.docker.com/v17.03/engine/reference/commandline/cli/

agcooke:
In the install_workers docker 17.06 is used. I can verify if it works on a 17.06 AMI built with these setting, doing this locally does make me think it does work:
```
$ ./dockerd -v     
Docker version 17.06.0-ce, build 02c1d87
$ ./dockerd -h | grep ulimit
Flag shorthand -h has been deprecated, please use --help
      --default-ulimit ulimit                   Default ulimits for containers (default [])

```
@micahhausler Let me know if I should do that or are you against increasing the ulimits?

agcooke:
Here is the link to the [17.06 dockerd docs](https://docs.docker.com/v17.06/edge/engine/reference/commandline/dockerd/)

and the [17.03 dockerd docs](https://docs.docker.com/v17.03/edge/engine/reference/commandline/dockerd/). 

They both support the default-ulimits flag. I have not verified if it works with 17.03/06 but have done in on 18.06. I am willing to do that verification if it helps.

micahhausler:
Thanks for the doc links. Have you been able to get 17.06 to work with the ulimits set in the config file? Every time I try to set that config file option I get:

```
unable to configure the Docker daemon with file /etc/docker/daemon.json: \
    the following directives don't match any configuration option: default-ulimits
```

agcooke:
OK, I did not verify it, so if you did it must not be possible. Sorry for the inconvenience! 

The pull request cannot be merged until the default docker version is updated for everybody.

PR #133, opened 2018-12-24
Feature/specify docker version cni version

agcooke:
*Issue #, if available:*

*Description of changes:*
Made it possilbe to set the docker and CNI version in the packer file.

I did upgrade docker to 18.06 and CNI_PLUGIN_VERSION to v0.7.4. I can downgrade that if it is too big of a change!

### DIFF
--- a/eks-worker-al2.json
+++ b/eks-worker-al2.json
@@ -5,11 +5,14 @@
     "binary_bucket_name": "amazon-eks",
     "binary_bucket_region": "us-west-2",
     "binary_bucket_path": "1.11.5/2018-12-06/bin/linux/amd64",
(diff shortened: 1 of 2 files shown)


micahhausler:
Hey thanks for this PR, but these changes aren't quite the route we want to go.

In regard to the docker version, we still use `17.03` because it is what is tested with the versions of Kubernetes we use. We'd rather not make 18 the default.

In regard to the CNI plugin versions, we actually want to move those out of this AMI. They should be installed by the CNI provider, and there is an [open issue on the AWS CNI plugin for this](https://github.com/aws/amazon-vpc-cni-k8s/issues/158).

PR #134, opened 2018-12-24
    Add a loopback cni device to fix:

agcooke:
*Issue #, if available:*

*Description of changes:*
Added a loopback CNI device. This removed error messages when the kubelet starts-up.

I have not dived deep into if it is needed or not. With the latest AMI (built by ourselves) we did see that the Node often went quickly into NonReady state after boot. After a reboot that did not happen.

With this change the Node booted correctly and does not go into NonReady state.

### DIFF
--- a/install-worker.sh
+++ b/install-worker.sh
@@ -80,8 +80,17 @@ sudo mkdir -p /etc/kubernetes/manifests
 sudo mkdir -p /var/lib/kubernetes
 sudo mkdir -p /var/lib/kubelet
 sudo mkdir -p /opt/cni/bin
+sudo mkdir -p /etc/cni/net.d
 
 CNI_VERSION=${CNI_VERSION:-"v0.6.0"}
+# Add the CNI loopback device
+cat <<EOF | sudo tee -a /etc/cni/net.d/99-loopback.conf
+{
+	"cniVersion": "${CNI_VERSION}",
+	"name": "lo",
+	"type": "loopback"
+}
+EOF
 wget https://github.com/containernetworking/cni/releases/download/${CNI_VERSION}/cni-amd64-${CNI_VERSION}.tgz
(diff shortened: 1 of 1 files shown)


install-worker.sh
```diff
@@ -80,8 +80,17 @@ sudo mkdir -p /etc/kubernetes/manifests
 sudo mkdir -p /var/lib/kubernetes
 sudo mkdir -p /var/lib/kubelet
 sudo mkdir -p /opt/cni/bin
+sudo mkdir -p /etc/cni/net.d
 
 CNI_VERSION=${CNI_VERSION:-"v0.6.0"}
+# Add the CNI loopback device
```
vsiddharth: Its unclear how this change is addressing the issue. 
Could you please provide additional details?

agcooke: This fixed this issue on kubelet startup:
```
    dec 21 10:21:07 ip-10-0-1-189.eu-west-1.compute.internal kubelet[12670]: W1221 10:21:07.776122   12670 cni.go:172] Unable to update cni config: No networks found in /etc/cni/net.d
    dec 21 10:21:07 ip-10-0-1-189.eu-west-1.compute.internal kubelet[12670]: W1221 10:21:07.786670   12670 cni.go:172] Unable to update cni config: No networks found in /etc/cni/net.d
```
Those error lines disappeared.

I do not know if the loopback CNI device is necessary, I copied it from @albertrdixon changes to his fork. I went through his code and pulled changes back into our fork.

micahhausler:
While this does set the node status to `Ready`, if any pods try to launch with the loopback plugin they won't really work. It is better to have an accurate `NotReady`, so we're not going to merge this change.

agcooke:
Great, that makes sense, I will remove it from our fork as well.

PR #135, opened 2018-12-24
Make it possible to use docker-compose packer for building.

agcooke:
*Issue #, if available:*

*Description of changes:*
This is a group of mostly cosmetic changes:
- make it possible to use docker-compose run packer to run packer instead of having to install packer locally.
- Use t3.small spot instances to run packer to save money.
- Make it possible to override the AWS_DEFAULT_REGION from the command line.
- Added echo at the end of the install workers to make sure you know everything ran when debugging (not necessary but nice to have)


micahhausler:
Thanks for the PR, but we'd like to keep this simple just stick to packer binaries. Using docker-compose is not a bad thing, but in order to use it you need docker running and a compatible docker-compose installed. 

In regards to the t3/spot pricing, we'll keep it at m4.large. In major US regions m4.large is $0.10/hour, and the instance is only up for a few minutes. At 3 minutes, that comes to around USD $0.005 to build an AMI. You can certainly override it yourself by using `packer build -var instance_type=t3.small`

Thanks!

agcooke:
Completely understandable, this is a nice to have and not that important.

PR #137, opened 2018-12-26
Update kubelet.service to be resilient to SIGPIPE exits

### LINKED ISSUE #122: Kubelet shutdown when Journald restarts
**What happened**:
For some reason the Kubelet process on one of my EKS nodes running `1.10.3` terminated from a SIGPIPE. After some research it seems this is a common problem that occurs when Journald restarts. Both Kubelet and Docker was affected until the [Docker project decided to ignore SIGPIPE](https://github.com/moby/moby/issues/7087) signals, [Kubelet still is](https://github.com/kubernetes/kubernetes/issues/69161).

I didn't manage to figure out why Journald might have failed because as soon as I restarted Kubelet to see what had caused the issue, the Cluster Autoscaler terminated it (amateur mistake).

[`ignoreSIGPIPE`](https://www.freedesktop.org/software/systemd/man/systemd.exec.html#IgnoreSIGPIPE=) is set for the kubelet service, but it doesn't seem to make a difference:
```
$ systemctl show kubelet.service | grep PIPE
IgnoreSIGPIPE=yes
```

Ultimately the Kubelet should of course deal with this, but there's less pressure on the Kubernetes project to address it as it's been [fixed in later versions of systemd](https://github.com/kubernetes/kubernetes/issues/38028#issuecomment-265012422)

There's also a few relatively simple ways of fixing this issue by configuring the kubelet's systemd service unit. There's two good options raised [here](https://github.com/influxdata/influxdb/issues/7040#issuecomment-234355479) that could solve this issue. We could either:
* Add a `Restart=always` to the Kubelet unit file which should cause the Kubelet to always restart. This of course restricts you to using `systemctl` only. We'd have to make sure we don't rely on any `SIGTERM`'s from scripts anywhere. Or
* Add a `RestartForceExitStatus=SIGPIPE` to the unit file which causes the service to interpret `SIGPIPE` as failure and restart the Kubelet process.

The second option looks a lot cleaner and that feature was added in systemd 215, the `1.10.3` EKS AMI is running systemd 219, so there shouldn't be anything stopping us.

* SSH in to a node you're okay sacrificing for the greater good
* execute `sudo systemctl kill systemd-journald.service`.
* execute `systemctl status kubelet`

Journald will immediately restart, but the kubelet will remain terminated by a SIGPIPE.
```
● kubelet.service - Kubernetes Kubelet
   Loaded: loaded (/etc/systemd/system/kubelet.service; enabled; vendor preset: disabled)
   Active: inactive (dead) since Sun 2018-12-16 22:09:58 UTC; 10s ago
     Docs: https://github.com/kubernetes/kubernetes
  Process: 4284 ExecStart=/usr/bin/kubelet --address=0.0.0.0 --authentication-token-webhook --authorization-mode=Webhook --allow-privileged=true --cloud-provider=aws --cluster-dns=10.100.0.10 --cluster-domain=cluster.local --cni-bin-dir=/opt/cni/bin --cni-conf-dir=/etc/cni/net.d --container-runtime=docker --max-pods=20 --node-ip=172.22.27.120 --network-plugin=cni --pod-infra-container-image=602401143452.dkr.ecr.us-west-2.amazonaws.com/eks/pause-amd64:3.1 --cgroup-driver=cgroupfs --register-node=true --kubeconfig=/var/lib/kubelet/kubeconfig --feature-gates=RotateKubeletServerCertificate=true --anonymous-auth=false --client-ca-file=/etc/kubernetes/pki/ca.crt (code=killed, signal=PIPE)
 Main PID: 4284 (code=killed, signal=PIPE)
```

As pointed out above, the Kubernetes project is unlikely to give this much attention since it's a solved problem going forward and there's a bunch of easy fixes in the meantime.

**Environment**:
- AWS Region: us-west-2
- Instance Type(s): not important, but the above verification test was carried out on an `m4.large`
- EKS Platform version: `eks.3`
- Kubernetes version: `1.10`
- AMI Version: `eks-worker-v22 (ami-02415125ccd555295)`
- Kernel (e.g. `uname -a`): `Linux kdev-node-0f0248ba8ca1fa485 4.14.62-70.117.amzn2.x86_64 #1 SMP Fri Aug 10 20:14:53 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux`
- Release information (run `cat /tmp/release` on a node): No such file, might be that my AMI is too old?

travcunn:
Closes #122

https://github.com/awslabs/amazon-eks-ami/pull/110 should be closed in favor of this change.

*Description of changes:*

This makes Kubelet more resilient to crashing. Systemd will now restart kubelet in case it exits with a SIGPIPE as mentioned in #122

### DIFF
--- a/files/kubelet.service
+++ b/files/kubelet.service
@@ -13,7 +13,8 @@ ExecStart=/usr/bin/kubelet --cloud-provider aws \
     --container-runtime docker \
     --network-plugin cni $KUBELET_ARGS $KUBELET_EXTRA_ARGS
 
-Restart=on-failure
+Restart=Always
+RestartForceExitStatus=SIGPIPE
 RestartSec=5
 KillMode=process
 
--- a/files/kubelet.service
+++ b/files/kubelet.service
@@ -13,7 +13,7 @@ ExecStart=/usr/bin/kubelet --cloud-provider aws \
     --container-runtime docker \
     --network-plugin cni $KUBELET_ARGS $KUBELET_EXTRA_ARGS
 
-Restart=Always
+Restart=on-failure
 RestartForceExitStatus=SIGPIPE
 RestartSec=5
 KillMode=process


whereisaaron:
Thanks for this fix @travcunn! You suggest not also setting the kubelet to always be restarted as in #110. Wouldn’t that be a good change too, in case there are other kubelet exits that don’t look like failures? I wondered what the use case is for letting the kubelet on node exit without being restarted? When would that ever be a good thing? 😀

PR #130, opened 2018-12-21
Use chrony for time and make sure it is enabled on startup.

agcooke:
*Issue #, if available:*

*Description of changes:*
Used chrony to set time on the Node.

We noticed with the default AMI, NTP does not actually start.

We had an alarm:

```
ALERT ClockSyncBroken
  IF          node_timex_sync_status != 1
  FOR         5m
  LABELS      { severity="warning" }
  ANNOTATIONS {
    summary = "The clock is not being synced.",
    impact = "Random things are about to break for our users",
    detail = "Node: {{$labels.node}}",
  }
```
firing all the time.

Now with the changes made, chronyd is started on boot:
```
[ec2-user@ip-10-0-0-134 ~]$ sudo systemctl status chronyd
● chronyd.service - NTP client/server
   Loaded: loaded (/usr/lib/systemd/system/chronyd.service; enabled; vendor preset: enabled)
   Active: active (running) since do 2018-12-20 17:44:30 UTC; 13h ago
     Docs: man:chronyd(8)
           man:chrony.conf(5)
 Main PID: 3112 (chronyd)
    Tasks: 1
   Memory: 1.9M
   CGroup: /system.slice/chronyd.service
           └─3112 /usr/sbin/chronyd

dec 20 21:14:23 ip-10-0-0-134.eu-west-1.compute.internal chronyd[3112]: Source 2a00:1288:110:f600::10...6
dec 20 21:44:42 ip-10-0-0-134.eu-west-1.compute.internal chronyd[3112]: Source 193.1.219.116 replaced...6
Warning: Journal has been rotated since unit was started. Log output is incomplete or unavailable.
Hint: Some lines were ellipsized, use -l to show in full.
```

Used https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/set-time.html to setup chrony.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #142, opened 2019-01-09
Enable RTC sync in chronyd.

agcooke:
*Issue #, if available:*
The chronyd service was running but not syncing the RTC clock.

This alarm was still firing:
```
ALERT ClockSyncBroken
  IF          node_timex_sync_status != 1
  FOR         5m
  LABELS      { severity="warning" }
  ANNOTATIONS {
    summary = "The clock is not being synced.",
    impact = "Random things are about to break for our users",
    detail = "Node: {{$labels.node}}",
  }

```
as described in https://github.com/awslabs/amazon-eks-ami/pull/130

***Description of changes:***
After comparing the configuration between an Ubuntu AMI and the AL2 AMI I saw that the chrony.conf differerd.

AL2 was configured to not sync the RTC clock.

After making the changes it does sync and the alarm disappeared:

```
[ec2-user@ip-10-0-19-151 ~]$ sudo systemctl start chronyd
[ec2-user@ip-10-0-19-151 ~]$ ./adjtimex 
    mode:         0
-o  offset:       0 us
-f  freq.adjust:  2155554 (65536 = 1ppm)
    maxerror:     1098
    esterror:     564
    status:       0 ()
-p  timeconstant: 7
    precision:    1 us
    tolerance:    32768000
-t  tick:         9994 us
    time.tv_sec:  1547032282
    time.tv_usec: 176261
    return value: 0 (clock synchronized)
```
Links:
- [Chrony Configuration](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/system_administrators_guide/sect-understanding_chrony_and-its_configuration)
- [Prometheus Node Exporter timex metric](https://github.com/prometheus/node_exporter/blob/master/collector/timex.go)

### DIFF
--- a/install-worker.sh
+++ b/install-worker.sh
@@ -29,6 +29,15 @@ sudo yum install -y \
 # Make sure Amazon Time Sync Service starts on boot.
 sudo chkconfig chronyd on
 
+# Make sure that chronyd syncs RTC clock to the kernel.
+cat <<EOF | sudo tee -a /etc/chrony.conf
+
+# This directive enables kernel synchronisation (every 11 minutes) of the
+# real-time clock. Note that it can’t be used along with the 'rtcfile' directive.
+rtcsync
+
+EOF
+
 curl "https://bootstrap.pypa.io/get-pip.py" -o "get-pip.py"
 sudo python get-pip.py
 rm get-pip.py
--- a/install-worker.sh
+++ b/install-worker.sh
@@ -31,11 +31,9 @@ sudo chkconfig chronyd on
 
 # Make sure that chronyd syncs RTC clock to the kernel.
 cat <<EOF | sudo tee -a /etc/chrony.conf
-
 # This directive enables kernel synchronisation (every 11 minutes) of the
 # real-time clock. Note that it can’t be used along with the 'rtcfile' directive.
 rtcsync
-
 EOF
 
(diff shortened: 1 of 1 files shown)


install-worker.sh
```diff
@@ -29,6 +29,15 @@ sudo yum install -y \
 # Make sure Amazon Time Sync Service starts on boot.
 sudo chkconfig chronyd on
 
+# Make sure that chronyd syncs RTC clock to the kernel.
+cat <<EOF | sudo tee -a /etc/chrony.conf
+
```
micahhausler: nit: can you remove the trailing and leading empty newlines?

agcooke: Updated

errordeveloper:
Thanks for digging into this, we've noticed the problem a while ago, but it wasn't clear to me whether it's how node exporter implements the probe is inappropriate for all NTP daemons.

agcooke:
No problem. We have had lot's of head aches with our nodes. We got our Nodes stable now. But I am not an expert on getting Nodes healthy, thus some of the hand wavy pull requests I opened here, I just wanted them to work. 

Now you guys can take a look into https://github.com/weaveworks/flux/issues/1639 ;)

We love Flux by the way!

errordeveloper:
Thanks, I will pass it on!

Issue #79, opened 2018-10-31
Node becomes NotReady


agcooke:
We are running EKS in Ireland and our nodes are going unhealthy regularly.

It is not possible to SSH to the host, pods are not reachable. We have experienced this with t2.xlarge, t2.small and t3.medium instances.

We could ssh to another node in the cluster and ping the NotReady node, but are not able to ssh it either.

Graphs show the memory goes high at about the same time that the journalctl logs below. The EBS IO also goes high. The exact time is hard to pinpoint. I added logs with interesting 'failures' around the time that we think the node disappeared.

We are using the cluster for running tests, so pods are getting created and destroyed often.

We have not done anything described in https://github.com/awslabs/amazon-eks-ami/issues/51 for log rotation.

**Cluster Information:**
CNI: Latest daemonset with image: 602401143452.dkr.ecr.us-west-2.amazonaws.com/amazon-k8s-cni:1.2.1
Region: eu-west-1

**LOGS**

** Node AMI**
```
AMI ID amazon-eks-node-v24 (ami-0c7a4976cb6fafd3a)
```
** File system **
```
$ df -h
Filesystem      Size  Used Avail Use% Mounted on
devtmpfs        1,9G     0  1,9G   0% /dev
tmpfs           1,9G     0  1,9G   0% /dev/shm
tmpfs           1,9G  2,3M  1,9G   1% /run
tmpfs           1,9G     0  1,9G   0% /sys/fs/cgroup
/dev/nvme0n1p1   64G   40G   25G  62% /
tmpfs           389M     0  389M   0% /run/user/1000
```

** kubectl describe node**
```
Name: ip-<secret>.eu-west-1.compute.internal
Roles: <none>
Labels: beta.kubernetes.io/arch=amd64
beta.kubernetes.io/instance-type=t3.medium
beta.kubernetes.io/os=linux
failure-domain.beta.kubernetes.io/region=eu-west-1
failure-domain.beta.kubernetes.io/zone=eu-west-1b
kubernetes.io/hostname=ip-<secret>.eu-west-1.compute.internal
Annotations: node.alpha.kubernetes.io/ttl: 0
volumes.kubernetes.io/controller-managed-attach-detach: true
CreationTimestamp: Tue, 30 Oct 2018 11:25:48 +0100
Taints: <none>
Unschedulable: false
Conditions:
Type Status LastHeartbeatTime LastTransitionTime Reason Message
---- ------ ----------------- ------------------ ------ -------
[...]

PR #143, opened 2019-01-10
Added a separate 'ext4' based partion for /var/lib/docker.

agcooke:
*Issue #, if available:*
https://github.com/awslabs/amazon-eks-ami/issues/79

*Description of changes:*
We have large images in our monolithic application. This commit attaches a separate EBS volume for docker artefacts to accomdate them.

It is possible to specify that the image gets formatted as XFS or ext4. We have been running with ext4 for some weeks now and have not had any issues. We did have issues with XFS, but did not pinpoint as the root cause because there where many other things that we fixed.

The work is mostly copied from @albertrdixon work in the [LumosLab fork](https://github.com/lumoslabs/amazon-eks-ami)

### DIFF
--- a/eks-worker-al2.json
+++ b/eks-worker-al2.json
@@ -18,7 +18,8 @@
     "cni_plugin_version": "v0.7.1",
     "aws_access_key_id": "{{env `AWS_ACCESS_KEY_ID`}}",
     "aws_secret_access_key": "{{env `AWS_SECRET_ACCESS_KEY`}}",
-    "aws_session_token": "{{env `AWS_SESSION_TOKEN`}}"
+    "aws_session_token": "{{env `AWS_SESSION_TOKEN`}}",
+    "docker_ebs_name": "/dev/sdf"
   },
 
   "builders": [
@@ -52,6 +53,12 @@
           "volume_type": "gp2",
           "volume_size": 20,
           "delete_on_termination": true
+        },
+        {
+          "device_name": "{{user `docker_ebs_name`}}",
+          "volume_type": "gp2",
+          "volume_size": 64,
+          "delete_on_termination": true
         }
       ],
       "ssh_username": "ec2-user",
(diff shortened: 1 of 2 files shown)


agcooke:
This is not a backwards compatible, if we want to go ahead with it I can update the CloudFormation templates.

We had to update our terraform files to add the extra volume.

kinghajj:
Is the purpose of this change primarily for more storage? If so, couldn't one simply increase the root volume size? You wrote in the comments for #79, "this would take strain off the OS network IO, since all EBS volumes are actually connected over the network." But since the root volume is also EBS, shouldn't that isolation hold for it, too?

dawidmalina:
@micahhausler can you please take a look on this pull request?

dawidmalina:
@agcooke can you rebase your pull request?

I have also few questions:
- script allow us to configure FSTYPE but there is no way to configure it with packer variable
- script allow us to configure FSTAB_OPTIONS but there is no way to configure it with packer variable
- volume size is currently hard-coded but it would be nice to configure it with packer variable

@micahhausler do you think such change will be accepted by you? We are looking for this to be aligned with CIS benchmark
> 1.1 – Ensure a separate partition for containers has been created

agcooke:
Looks like this has been added to master in another commit: https://github.com/awslabs/amazon-eks-ami/commit/f8c117f90d055b46ff09b5e9a9c081c612316478

agcooke:
Oh, my mistake, that is something else...

agcooke:
@dawidmalina  I rebased to master and removed the variables. I have moved on to another project so difficult for me to test right now, can you take a look at it? Then we poke maintainers to get it reviewed...

dawidmalina:
@agcooke recently added ability to add custom source ami during build https://github.com/awslabs/amazon-eks-ami/pull/203 gives me no reasons to continue this change. I think that there are too many corner cases related to operation system configuration (like hardening, different partition layout, etc) so IMHO we should keep here only changes related to k8s node setup. If someone is not satisfied with default source ami - no problem it can be easily changed now.

Anyway final decision must be done by @micahhausler, @mogren

PR #144, opened 2019-01-10
Feature/specify docker version cni version

agcooke:
*Issue #, if available:*
https://github.com/awslabs/amazon-eks-ami/issues/79
https://github.com/awslabs/amazon-eks-ami/pull/133

*Description of changes:*
Made it possilbe to set the docker and CNI version in the packer file.

Docker version fixed to 17.06 and CNI_PLUGIN_VERSION  upgraded to v0.7.4. I can downgrade that if it is too big of a change!

@micahhausler I think this is really useful. We have upgraded to 18.6 and have had good results. Having the docker version out of the packer spec makes it easy for other people to do it as well. If these other changes to be made then I can do it.

### DIFF
--- a/eks-worker-al2.json
+++ b/eks-worker-al2.json
@@ -5,11 +5,14 @@
     "binary_bucket_name": "amazon-eks",
     "binary_bucket_region": "us-west-2",
     "binary_bucket_path": "1.11.5/2018-12-06/bin/linux/amd64",
+    "docker_version": "18",
     "creator": "{{env `USER`}}",
     "instance_type": "m4.large",
     "source_ami_id": "",
     "encrypted": "false",
-    "kms_key_id": ""
+    "kms_key_id": "",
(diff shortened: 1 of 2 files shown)


dawidmalina:
It would be also nice to bump docker version to 18.06 as default version.

dawidmalina:
@micahhausler as `cni_plugin_version` was reverted to default version `v0.7.1` is there any other blocker which blocks you from merge?

eks-worker-al2.json
```diff
@@ -45,10 +48,14 @@
           "creator": "{{user `creator`}}"
       },
       "tags": {
-          "created": "{{timestamp}}"
+          "created": "{{timestamp}}",
```
micahhausler: I'm fine adding these tags, but do be aware that tags are not visible on public AMIs.

Issue #157, opened 2019-01-21
Enable live-restore for docker daemon


someword:
**What would you like to be added**:
The provided docker configuration does not enable live-restore for docker (https://docs.docker.com/config/containers/live-restore/).  

**Why is this needed**:
 Live restore comes in handy when the docker daemon crashes/restarts.    The running containers keep running.   With the current configuration when docker restarts/dies all of the containers die as well.

micahhausler:
Closed with #158

PR #158, opened 2019-01-21
Enable live-restore capability in docker daemon

someword:
This change addresses issue #157

### DIFF
--- a/files/docker-daemon.json
+++ b/files/docker-daemon.json
@@ -3,5 +3,6 @@
   "log-opts": {
     "max-size": "10m",
     "max-file": "10"
-  }
+  },
+  "live-restore": true
 }


files/docker-daemon.json
```diff
@@ -3,5 +3,6 @@
   "log-opts": {
     "max-size": "10m",
     "max-file": "10"
-  }
+  },
+  "no-new-privileges": true
```
mogren: Sure this is the right change?  I would have expected `"live-restore": true`

`no_new_privs` might be a good idea anyway, but maybe a different issue?

someword: ugh, my bad.  Let me fix

someword: @mogren i've put in the correct change and rebased.
Sorry for the confusion :(

dawidmalina:
LGTM. @micahhausler when this can be merged? Btw how often pull requests are reviewed and accepted (merged) or rejected (which is fine in some cases)? Most of the existing pull requests are not even reviewed :( those are not comfortable conditions encouraging to contributions (maybe that's why there are so many independent forks)

This process is not transparent or at least I don't see it :(

Issue #162, opened 2019-01-26
Allow (optional) smaller EBS root volumes with container optimized AMI


talawahtech:
**What would you like to be added**:
The default  root volume size for EKS is 20GB. That is a good default, but I would like the option to use smaller volume sizes as well. Overriding the BlockDeviceMappings of the launch template allows us to increase the size of the root device, but we cannot *decrease* it beyond the size of the *snapshot* associated with the AMI.

**Why is this needed**:
If I want to use the EKS AMI to launch a large number of small instances with minimal containers, I am forced to use 20GB root volumes. In the most extreme  example if I were launching a bunch of t3.nano instances, the EBS cost would be almost as much as the instance cost

**Additional context**
It is possible to maintain the 20GB *default* for the AMI while still allowing instances to be launched with a smaller root volume size as long as **the snapshot** associated with the AMI is smaller. The size of the snapshot is determined by the size of the block device on the instance used to build the AMI.

The current [Packer configuration](https://github.com/awslabs/amazon-eks-ami/blob/master/eks-worker-al2.json) uses the `launch_block_device_mappings` property to set the volume size to 20GB and in doing so forces the snapshot to be 20GB as well.

If instead the volume size specified under [launch_block_device_mappings](https://www.packer.io/docs/builders/amazon-ebs.html#launch_block_device_mappings) was 4GB **and** the [ami_block_device_mappings](https://www.packer.io/docs/builders/amazon-ebs.html#ami_block_device_mappings) property was added with a volume size of 20GB then we would have the best of both worlds. The **default would still be 20GB** but the AMI user could override BlockDeviceMappings to resize the volume as low as 4GB or as high as EBS supports.

The Amazon Linux 2 minimal AMI has a snapshot size of 2GB. After the EKS AMI is built disk usage is about 1.5GB, so 4GB seems like a reasonable *minimum* value to me.

PR #163

micahhausler:
Sounds good to me.

micahhausler:
Fixed in #163

PR #163, opened 2019-01-26
Allow (optional) smaller EBS root volumes with container optimized AMI

talawahtech:
This PR addresses #162 

*Description of changes:*
It is possible to maintain the 20GB *default* root volume size for the AMI while still allowing instances to be launched with a smaller root volume size as long as **the snapshot** associated with the AMI is smaller. The size of the snapshot is determined by the size of the block device on the instance used to build the AMI.

The current [Packer configuration](https://github.com/awslabs/amazon-eks-ami/blob/master/eks-worker-al2.json) uses the `launch_block_device_mappings` property to set the volume size to 20GB and in doing so forces the snapshot to be 20GB as well.

This PR reduces the volume size specified under [launch_block_device_mappings](https://www.packer.io/docs/builders/amazon-ebs.html#launch_block_device_mappings) to 4GB **and** adds the [ami_block_device_mappings](https://www.packer.io/docs/builders/amazon-ebs.html#ami_block_device_mappings) property with a volume size of 20GB. 

This gives us the best of both worlds. The **default is still 20GB** but the AMI user can override BlockDeviceMappings to resize the volume as low as 4GB or as high as EBS supports.

The Amazon Linux 2 minimal AMI has a snapshot size of 2GB. After the EKS AMI is built disk usage is about 1.5GB, so 4GB seems like a reasonable *minimum* value to me as it provides room for additional base packages.

### DIFF
--- a/eks-worker-al2.json
+++ b/eks-worker-al2.json
@@ -33,6 +33,14 @@
         {
           "device_name": "/dev/xvda",
           "volume_type": "gp2",
+          "volume_size": 4,
+          "delete_on_termination": true
+        }
+      ],
+      "ami_block_device_mappings": [
+        {
+          "device_name": "/dev/xvda",
+          "volume_type": "gp2",
           "volume_size": 20,
           "delete_on_termination": true
         }


dawidmalina:
I agree that main partition should be relatively small. In most cases the place where you would need more space is`/var/lib/docker` but this should be dedicated volume anyway: https://github.com/awslabs/amazon-eks-ami/pull/143

PR #175, opened 2019-02-09
Allow pulling node binaries from private bucket

shyamjvs:
This is now possible by using `aws cp` with pre-configured credentials.

/cc @micahhausler @nckturner

### DIFF
--- a/install-worker.sh
+++ b/install-worker.sh
@@ -113,15 +113,22 @@ if [ "$BINARY_BUCKET_REGION" = "us-east-1" ]; then
     S3_DOMAIN="s3"
 fi
 S3_URL_BASE="https://$S3_DOMAIN.amazonaws.com/$BINARY_BUCKET_NAME/$BINARY_BUCKET_PATH"
-
+S3_PATH="s3://$BINARY_BUCKET_NAME/$BINARY_BUCKET_PATH"
 BINARIES=(
     kubelet
     kubectl
     aws-iam-authenticator
 )
 for binary in ${BINARIES[*]} ; do
-    sudo wget $S3_URL_BASE/$binary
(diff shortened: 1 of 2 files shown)


install-worker.sh
```diff
@@ -110,15 +110,20 @@ if [ "$BINARY_BUCKET_REGION" = "us-east-1" ]; then
...
+S3_PATH="s3://$BINARY_BUCKET_NAME/$BINARY_BUCKET_PATH"
 BINARIES=(
     kubelet
     kubectl
     aws-iam-authenticator
 )
 for binary in ${BINARIES[*]} ; do
-    sudo wget $S3_URL_BASE/$binary
-    sudo wget $S3_URL_BASE/$binary.sha256
+    if which aws >/dev/null; then
```
nckturner: Would this be confusing in the case where someone wants to use a private bucket, but does not have the awscli installed?  Maybe print a message that its defaulting to a public URL -- or add an argument to the script to select what it attempts to do, to make it more clear.

shyamjvs: Good point. Added print statements explaining the cases, PTAL.

nckturner:
LGTM!

xinghusc:
@shyamjvs @nckturner why do we need this change which allows us to pull `kubectl`, `kubelet`, etc from private bucket instead of the public bucket? And why do we make this option as default option?

shyamjvs:
@xinghusc Private bucket is not the default option. Using `aws s3` is the default option (and that can be used both with private and public buckets).

> why do we need this change

Because someone might want to build custom AMI using binaries from their own bucket and we shouldn't need to impose that it's public.

xinghusc:
@shyamjvs because aws cli was installed in previous steps so `which aws >/dev/null` will always be true which effectively makes it a default option.

PR #182, opened 2019-02-13
Clean resolv.conf before snapshotting

micahhausler:
*Issue #, if available:*
N/A

*Description of changes:*

`/etc/resolv.conf` was getting the search domain appended with the region the AMI was built in. In the EKS case, this resulted in the following contents for an instance in us-east-2:

```
$ cat /etc/resolv.conf 
; generated by /usr/sbin/dhclient-script
search us-east-2.compute.internal us-west-2.compute.internal
options timeout:2 attempts:5
nameserver 192.168.0.2
```

New instances will now just get the search domain for the region they are launched in:

```
$ cat /etc/resolv.conf 
; generated by /usr/sbin/dhclient-script
search us-east-2.compute.internal 
options timeout:2 attempts:5
nameserver 192.168.0.2
```

### DIFF
--- a/install-worker.sh
+++ b/install-worker.sh
@@ -175,17 +175,22 @@ sudo rm -rf \
 
 # Clean up files to reduce confusion during debug
 sudo rm -rf \
+    /etc/hostname \
     /etc/machine-id \
+    /etc/resolv.conf \
     /etc/ssh/ssh_host* \
-    /root/.ssh/authorized_keys \
     /home/ec2-user/.ssh/authorized_keys \
-    /var/log/secure \
-    /var/log/wtmp \
-    /var/lib/cloud/sem \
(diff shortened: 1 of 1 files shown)


install-worker.sh
```diff
@@ -173,19 +173,30 @@ sudo rm -rf \
     $TEMPLATE_DIR  \
     /var/cache/yum
 
+unset IFS
+# Kill dhclient to not cache search domains
+for DHPID in $(/usr/sbin/pidof dhclient) ; do
```
brycecarman: dhclient is started by systemd so it will get restarted if you just kill it. Maybe `sudo sysctl stop network` instead?

micahhausler: I double checked, and this isn't actually required.

PR #186, opened 2019-02-15
Increase ulimits for docker containers.

micahhausler:
*Issue #, if available:*

Rebase of #132 for docker 18.06. The configuration failed for docker 17.06, but works on 18.06

*Description of changes:*

Increase the default ulimits for containers

### DIFF
--- a/files/docker-daemon.json
+++ b/files/docker-daemon.json
@@ -6,5 +6,12 @@
     "max-file": "10"
   },
   "live-restore": true,
-  "max-concurrent-downloads": 10
+  "max-concurrent-downloads": 10,
+  "default-ulimits": {
+    "nofile": {
+      "Name": "nofile",
+      "Soft": 2048,
+      "Hard": 8192
+    }
+  }
 }
--- a/install-worker.sh
+++ b/install-worker.sh
@@ -63,6 +63,9 @@ DOCKER_VERSION=${DOCKER_VERSION:-"18.06"}
 sudo yum install -y docker-${DOCKER_VERSION}*
 sudo usermod -aG docker $USER
 
+# Remove all options from sysconfig docker.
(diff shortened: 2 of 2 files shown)


victortrac:
This change seemed to have broken elasticsearch on EKS:
```
ERROR: [1] bootstrap checks failed
[1]: max file descriptors [8192] for elasticsearch process is too low, increase to at least [65536]
```
ES pods on the new AMI released today are failing with that error. Pods on nodes created a few weeks ago are working as expected.

nitin302:
Is there any way some other default was being effected before? I am using amazon-eks-node-1.10-v20181210 and containers already have 65536 as the ulimit setting. The host level setting is 1024 though!

geerlingguy:
Another 'me too'—the latest AMI broke our Elasticsearch pods and we had to manually revert the change.

athilaelhuno:
Which version has this merge, because the amazon-eks-node-1.11-v20190220 (ami-0eeeef929db40543c) has the unlimit set.

cpanato:
same issue for me :/

When the new AMI will be released?

agnewp:
just updated to the latest version of AMI, broke all my elastic search nodes with the file descriptor error. This kind of thing should be adjustable!

whereisaaron:
@agnewp  @cpanato I think watch [this issue](https://github.com/aws/containers-roadmap/issues/188), we're expecting 1.11.8 images for EKS any day, which hopefully will include this fix.

https://github.com/aws/containers-roadmap/issues/188

MMartyn:
New AMI broke my elasticsearch as well. I opened it up to see what was going on and it appears that /etc/sysconfig/docker still had the OPTION that has the ulimit. It is almost like this isn’t having the intended affect https://github.com/awslabs/amazon-eks-ami/blob/master/install-worker.sh#L66

PR #187, opened 2019-02-15
Add enable-docker-bridge bootstrap argument

micahhausler:
*Issue #, if available:*

Fixes #183

*Description of changes:*

* Updated changelog git command
* Added enable-docker-bridge bootstrap argument

The default will be to leave the bridge network disabled, but users can now add `--enable-docker-bridge true` as a bootstrap argument in order to preserve the old behavior.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #192, opened 2019-02-19
Fixed --enable-docker-bridge bootstrap arg

micahhausler:
*Issue #, if available:*

Fixes bug introduced in #187 

*Description of changes:*

* Correctly set the docker bridge env var
* Safely write the docker config file

### DIFF
--- a/files/bootstrap.sh
+++ b/files/bootstrap.sh
@@ -53,7 +53,7 @@ while [[ $# -gt 0 ]]; do
             shift
             ;;
         --enable-docker-bridge)
-            ENABLE_DOCKER_BIDGE=$2
+            ENABLE_DOCKER_BRIDGE=$2
             shift
             shift
             ;;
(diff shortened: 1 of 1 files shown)


micahhausler:
FYI @0xAhmed @msvechla

0verc1ocker:
This is the dumbest bootstrapping process of kubelet I have seen yet. Congratulations.

micahhausler:
> This is the dumbest bootstrapping process of kubelet I have seen yet. Congratulations.

@0verc1ocker is there something you’d like to see done differently?

Micah Hausler
Systems Development Engineer – EKS
Amazon Web Services

0verc1ocker:
> This is the dumbest bootstrapping process of kubelet I have seen yet. Congratulations.
> @0verc1ocker is there something you’d like to see done differently?

I'm sorry @micahhausler . It has been a frustrating day. :cry:  I just find it really confusing when arguments get passed from CloudFormation parameter into another script and then trying to figure out where things broke. In this sense, bootstrap.sh is acting as a single point of failure and it has multiple flags of where it can break.

My opnion is that allowing users to just append a startup script as a service, similar to `kops` hooks would be better experience overall if using CloudFormation to spin up the instances. Eventually they will have to use UserData for their own custom operations anyways.

moshid:
Thank you for this fix! I'm new to eks, how do I apply the flag --enable-docker-bridge on an already running stack? I've used `jx create cluster eks` to create the cluster

PR #201, opened 2019-02-22
Add PublicIpAddress parameter

Luizm:
*Description of changes:*

I added the public IP address as a parameter, but keep `true` as default.

I think that some times the nodes will never be EIP associated with it and in most cases, prefer to use Load Balancer to publish the k8s service.


micahhausler:
Thanks for this PR, but we want to keep the possible options in the CloudFormation template to a minimum. Since adding the public IP has no cost associated with it, we'd like to keep the default on.

Issue #172, opened 2019-02-05
Ability to provide custom ami during build.


dawidmalina:
**What would you like to be added**:
It would be nice to have ability to provide custom ami during build. So we can override:
- `name` (should points by default to `amzn2-ami-minimal-hvm-*`)
- `owners` (should points by default to `137112412989`)

Currently it's looks like this:
```
      "source_ami": "{{user `source_ami_id`}}",
      "source_ami_filter": {
        "filters": {
          "architecture": "x86_64",
          "name": "amzn2-ami-minimal-hvm-*",
          "root-device-type": "ebs",
          "state": "available",
          "virtualization-type": "hvm"
      },
      "owners": [ "137112412989" ],
```
But this `source_ami_id` only allows to extend filters not fully override.

> If you set this in conjunction with source_ami, the source_ami will be added to the filter. 

**Why is this needed**:
To allow end user to use custom ami (for example hardened one) without need to keep own fork of this repository.

mogren:
Fix merged

PR #203, opened 2019-02-25
Ability to provide custom ami during build

dawidmalina:
*Issue #, if available:*
https://github.com/awslabs/amazon-eks-ami/issues/172

*Description of changes:*
This change will allow us to provide custom ami name prefix within source_ami_filter section and different image owner.

### DIFF
--- a/eks-worker-al2.json
+++ b/eks-worker-al2.json
@@ -9,6 +9,8 @@
     "creator": "{{env `USER`}}",
     "instance_type": "m4.large",
     "source_ami_id": "",
+    "source_ami_owners": "137112412989",
+    "source_ami_filter_name": "amzn2-ami-minimal-hvm-*",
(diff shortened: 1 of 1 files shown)


eks-worker-al2.json
```diff
@@ -58,6 +60,8 @@
       "tags": {
           "created": "{{timestamp}}",
           "docker_version": "{{ user `docker_version`}}",
+          "source_ami_id": "{{ user `source_ami_id`}}",
+          "source_ami_filter_name": "{{ user `source_ami_filter_name`}}",
```
micahhausler: Can you remove this tag? It is just a query filter, and `source_ami_id` should be enough to debug any issues

dawidmalina: Done

PR #206, opened 2019-02-26
Removed ulimit reduction

### LINKED ISSUE #193: Latest AMI version lowers ulimit and breaks Elasticsearch
**What happened**:

Upgrading to the latest AMI _reduced_ the `ulimit` to 8192 which broke Elasticsearch deployments.  Elasticsearch requires a `ulimit` of 65536.

Upgrading AMIs shouldn't reduce the `ulimit` (which is the opposite effect of what the PR #186 implies).

Use the latest AMI `amazon-eks-node-1.11-v20190220` - containers will have a max ulimit of 8192, not unlimited or 65536.

This was broken by #186 - that PR _reduced_ the default limit, rather than increasing it.

**Environment**:
- AWS Region: `us-east-1`
- Instance Type(s): `m5.xlarge`
- AMI Version: `amazon-eks-node-1.11-v20190220`

micahhausler:
*Issue #, if available:*

Fixes #193
Reverts #186

*Description of changes:*

Revert ulimit reduction

### DIFF
--- a/files/docker-daemon.json
+++ b/files/docker-daemon.json
@@ -6,12 +6,5 @@
     "max-file": "10"
   },
   "live-restore": true,
-  "max-concurrent-downloads": 10,
-  "default-ulimits": {
-    "nofile": {
-      "Name": "nofile",
-      "Soft": 2048,
-      "Hard": 8192
-    }
-  }
+  "max-concurrent-downloads": 10
 }


cpanato:
hello this change is applied in the current AMI for eks version 1.11 ? (ami-0eeeef929db40543c)

lekspyl:
Hello,

looks like `amazon-eks-node-1.11-v20190220` is the latest AMI and it does not contain this patch.
Could you tell when is it expected for a new AMI to be released? Asking due to https://github.com/awslabs/amazon-eks-ami/issues/193 which I can't make to work on the current AMI version. If there is any working way to override max file descriptors value I will be more than thankful if you'll share it.

dmarkey:
Hey, is there an ETA for a release with this fix in it?

annawinkler:
I would also be interested to find out when this fix will be released - thank you! ✨

sherifkayad:
@micahhausler ,

Is it possible to know when that fix will be available/when a new AMI is going to be available?

tabern:
This will be in the marketplace-available optimized AMI with K8s version 1.11.8. See https://github.com/aws/containers-roadmap/issues/188

PR #208, opened 2019-03-01
Added 1.12 build in Makefile

shyamjvs:
/cc @micahhausler

### DIFF
--- a/eks-worker-al2.json
+++ b/eks-worker-al2.json
@@ -1,11 +1,11 @@
 {
   "variables": {
-    "kubernetes_version": "1.11",
+    "kubernetes_version": "1.12",
     "aws_region": "us-west-2",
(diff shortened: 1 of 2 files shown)


eks-worker-al2.json
```diff
@@ -4,7 +4,7 @@
     "ami_name": "amazon-eks-node-{{timestamp}}",
     "binary_bucket_name": "amazon-eks",
     "binary_bucket_region": "us-west-2",
-    "binary_bucket_path": "1.11.5/2018-12-06/bin/linux/amd64",
+    "binary_bucket_path": "1.12.6/<fill>/bin/linux/amd64",
```
mcrute: It doesn't seem like there's actually a reasonable default here. Why not just remove the default and force the Makefile to set it.

shyamjvs: Good point. Fixed it.

PR #220, opened 2019-03-15
Use local-ipv4 for determining $TEN_RANGE

drewhemm:
*Description of changes:*

* Simpler method of determining whether the instance is deployed into a
subnet with a 10. CIDR
* Eliminates redundant calls to metadata service and parsing of their
responses
* Better handles scenarios where an instance spans multiple subnets (CNI
custom networking)
* Potential fix for incorrect setting of $TEN_RANGE, which causes DNS resolution to fail for all pods on the node.

### DIFF
--- a/files/bootstrap.sh
+++ b/files/bootstrap.sh
@@ -111,9 +111,8 @@ kubectl config \
 
 ### kubelet.service configuration
 
-MAC=$(curl -s http://169.254.169.254/latest/meta-data/network/interfaces/macs/ -s | head -n 1 | sed 's/\/$//')
-CIDRS=$(curl -s http://169.254.169.254/latest/meta-data/network/interfaces/macs/$MAC/vpc-ipv4-cidr-blocks)
-TEN_RANGE=$(echo $CIDRS | grep -c '^10\..*' || true )
+INTERNAL_IP=$(curl -s http://169.254.169.254/latest/meta-data/local-ipv4)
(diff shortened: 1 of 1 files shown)


micahhausler:
This won't work, because the local IPv4 could be in a secondary CIDR subnet (`100.64.0.0/10`), while the primary VPC range could be `10/8`. In that case, the DNS service IP would be `172.20.0.10`, but the instances in the `100.64.0.0/10` subnet would incorrectly choose `10.100.0.10`. The service CIDR is set by the control plane using corresponding logic

drewhemm:
Hi @micahhausler, the existing code also does not cover all the bases, since the anchored regex only checks the first of the CIDRs returned. For example:

```
$ MAC=$(curl -s http://169.254.169.254/latest/meta-data/network/interfaces/macs/ -s | head -n 1 | sed 's/\/$//')
$ CIDRS=$(curl -s http://169.254.169.254/latest/meta-data/network/interfaces/macs/$MAC/vpc-ipv4-cidr-blocks)
$ echo $CIDRS
10.50.112.0/24 10.9.0.0/16
$ echo $CIDRS | grep -c '^10\..*'
1
# This response should be 2
```

Both of these CIDRs are in the 10.x range, but only the first is checked. If the first was `100.64.0.0/10` and the second was `10.9.0.0/16` the regex would fail to match.

drewhemm:
`grep -cE [^\d]?10\.[\.\d\/]*` matches 10.x irrespective of where in the string it appears...

PR #221, opened 2019-03-18
Fixed regex to match 10.x in all cases

drewhemm:
*Description of changes:*

* Replaced ^ anchor
* regex needs to match irrespective of the position in the string

Related to, but not necessarily a solution for https://github.com/awslabs/amazon-eks-ami/issues/222

### DIFF
--- a/files/bootstrap.sh
+++ b/files/bootstrap.sh
@@ -113,7 +113,7 @@ kubectl config \
 
 MAC=$(curl -s http://169.254.169.254/latest/meta-data/network/interfaces/macs/ -s | head -n 1 | sed 's/\/$//')
 CIDRS=$(curl -s http://169.254.169.254/latest/meta-data/network/interfaces/macs/$MAC/vpc-ipv4-cidr-blocks)
-TEN_RANGE=$(echo $CIDRS | grep -c '^10\..*' || true )
+TEN_RANGE=$(echo $CIDRS | grep -cE '[^\d]?10\.[\.\d\/]*' || true )
 DNS_CLUSTER_IP=10.100.0.10
 if [[ "$TEN_RANGE" != "0" ]] ; then
     DNS_CLUSTER_IP=172.20.0.10;


files/bootstrap.sh
```diff
@@ -113,7 +113,7 @@ kubectl config \
 
 MAC=$(curl -s http://169.254.169.254/latest/meta-data/network/interfaces/macs/ -s | head -n 1 | sed 's/\/$//')
 CIDRS=$(curl -s http://169.254.169.254/latest/meta-data/network/interfaces/macs/$MAC/vpc-ipv4-cidr-blocks)
-TEN_RANGE=$(echo $CIDRS | grep -c '^10\..*' || true )
+TEN_RANGE=$(echo $CIDRS | grep -cE '[^\d]?10\.[\.\d\/]*' || true )
```
mogren: This would match `210.50.112.0/24` as well, because of the `\d`.

mogren: Oh, matching non-digits...

drewhemm: I did test it first ;)

[regexr.com](https://regexr.com) is quite useful for such tasks.

A nice improvement would be to use a 'look behind' so that the space in front of the second 10.x CIDR isn't included in the match. For the purpose of this use-case, it's unnecessary (and I'm not sure Grep extended re supports it), but could be considered the cherry on the cake.

mogren: Yeah, the non-digit doesn't really work since you have the ? (0 or 1). Examples that should not match:
```
> echo "210.50.112.0/24 99.20.112.0/24" | grep -cE '[^\d]?10\.[\.\d\/]*'     
1
> echo "220.50.112.0/24 99.10.112.0/24" | grep -cE '[^\d]?10\.[\.\d\/]*'   
1
```

drewhemm: Good catch!

PR #223, opened 2019-03-18
Updated describe-cluster retry/backoff exponential with jitter

micahhausler:
*Issue #, if available:*
N/A

*Description of changes:*

Updated retry logic to be exponential with jitter

### DIFF
--- a/files/bootstrap.sh
+++ b/files/bootstrap.sh
@@ -101,7 +101,7 @@ if [[ -z "${B64_CLUSTER_CA}" ]] && [[ -z "${APISERVER_ENDPOINT}" ]]; then
     # Retry the DescribleCluster API for API_RETRY_ATTEMPTS
     for attempt in `seq 0 $API_RETRY_ATTEMPTS`; do
         if [[ $attempt -gt 0 ]]; then
-        echo "Attempt $attempt of $API_RETRY_ATTEMPTS"
+            echo "Attempt $attempt of $API_RETRY_ATTEMPTS"
         fi
         aws eks describe-cluster \
             --region=${AWS_DEFAULT_REGION} \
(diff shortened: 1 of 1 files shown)


files/bootstrap.sh
```diff
@@ -114,7 +114,8 @@ if [[ -z "${B64_CLUSTER_CA}" ]] && [[ -z "${APISERVER_ENDPOINT}" ]]; then
         if [[ $attempt -eq $API_RETRY_ATTEMPTS ]]; then
             exit $rc
         fi
-        sleep_sec="$(( $((1+$attempt)) * $API_RETRY_ATTEMPTS * 5))"
+        jitter=$((1 + RANDOM % 10))
+        sleep_sec="$(( $(( 5 << $((1+$attempt)) )) + $jitter))"
```
mogren: Neat.

whereisaaron: bit shifts and jitter? now you’re just showing off 😀👍

Is there a cap on ((1+$attempt)) somehow? I guess the bit shift could overflow eventually?
Once sleep_sec reaches ~600 seconds, probably that is enough back-up for retries after that?

micahhausler: >  Is there a cap on ((1+$attempt)) somehow? I guess the bit shift could overflow eventually?

Yea, this script is in no way checking user input for correctness. There is only so much that you can sensibly do in bash :)

whereisaaron: How about:
```
jitter=$(( RANDOM % 10 ))
if (( attempt > 6 )); then
  sleep_sec=$(( 600 + $jitter ))
else
  sleep_sec="$(( (5 << $attempt) + $jitter))"
fi
```
Which yields sequences like: 6,10,23,43,80,160,329,602,608,600,601,609,604,604,602,608,609,607,600,605,605

Issue #222, opened 2019-03-18
Kubelet cluster-dns parameter is set incorrectly


drewhemm:
**What happened**: Potentially related to https://github.com/awslabs/amazon-eks-ami/issues/78 and https://github.com/awslabs/amazon-eks-ami/issues/197. 

`/etc/kubernetes/kubelet/kubelet-config.json` has the wrong value for clusterDNS, which is set by bootstrap.sh. This results in pods being unable to resolve DNS names. 

**What you expected to happen**: I would expect to see the correct nameserver in `/etc/kubernetes/kubelet/kubelet-config.json` and in `/etc/resolv.conf` on all pods running in the cluster.

**How to reproduce it (as minimally and precisely as possible)**: Create EKS service. My VPC happens to have two subnets (10.50.112.0/26 and 10.9.0.0/18, not sure that is relevant). Create worker instances. Most have the correct nameserver values in `/etc/kubernetes/kubelet/kubelet-config.json`, but not all. For the worker nodes that have the wrong value, pods  cannot resolve DNS queries. 

**Anything else we need to know?**:

**Environment**:
- AWS Region: eu-west-1
- Instance Type(s): c5.2xlarge, r5.2xlarge, other instance types too
- EKS Platform version: eks.2
- Kubernetes version: 1.11
- AMI Version: https://github.com/awslabs/amazon-eks-ami/commit/6090f200669ba1f76ce68f23e6496b3df9bc588a
- Kernel (e.g. `uname -a`): Linux 4.14.104-95.84.amzn2.x86_64
- Release information (run `cat /tmp/release` on a node):

```
cat: /tmp/release: No such file or directory
```

This issue still exists with the latest AMI code. An instance I created last week from an AMI built from the master branch (https://github.com/awslabs/amazon-eks-ami/commit/6090f200669ba1f76ce68f23e6496b3df9bc588a) has the wrong nameserver address in `/etc/kubernetes/kubelet/kubelet-config.json`:

```
...
"clusterDNS": [
  "10.100.0.10"
],
...
```
The instance is deployed into a VPC with two subnet CIDRs: 10.50.112.0/26 and 10.9.0.0/18.

It does not happen all the time. Other instances created around the same time and from the same AMI have the correct value. It is an intermittent bug that I have encountered on numerous occasions but am yet to find the exact cause.

whereisaaron:
Ref #221

micahhausler:
Can you provide a few more details about your VPC setup? 

[...]

PR #224, opened 2019-03-18
Fixed search for 10.x CIDR range

micahhausler:
*Issue #, if available:*

Simplified search for 10.x CIDR ranges.

*Description of changes:*

Bash replaces the newlines from the instance metadata service with a space, making the `grep` regex more complicated. By removing the intermediate CIDRS variable, we can get a newline separated string and properly grep it. The logic introduced in #221 did work, but the regex `[^\d]?` (optional non-digit) is easily confused with `^[\d]?` (optional leading digit).

```
$ curl -s http://169.254.169.254/latest/meta-data/network/interfaces/macs/$MAC/vpc-ipv4-cidr-blocks 
192.168.0.0/16
100.64.0.0/16
```

### DIFF
--- a/files/bootstrap.sh
+++ b/files/bootstrap.sh
@@ -135,8 +135,7 @@ kubectl config \
 ### kubelet.service configuration
 
 MAC=$(curl -s http://169.254.169.254/latest/meta-data/network/interfaces/macs/ -s | head -n 1 | sed 's/\/$//')
-CIDRS=$(curl -s http://169.254.169.254/latest/meta-data/network/interfaces/macs/$MAC/vpc-ipv4-cidr-blocks)
-TEN_RANGE=$(echo $CIDRS | grep -cE '[^\d]?10\.[\.\d\/]*' || true )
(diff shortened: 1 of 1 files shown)


drewhemm:
Ref https://github.com/awslabs/amazon-eks-ami/issues/222

The downside of this method is that it makes it impossible to test if the curl was successful, thereby guaranteeing the wrong `clusterDNS` in the event of a failed call to the metadata service and no way to catch it. This would be a better approach and is what I am currently working on:

```
MAC=$(curl -s http://169.254.169.254/latest/meta-data/network/interfaces/macs/ | head -n 1 | sed 's/\/$//')
if [ -z "$MAC" ]; then
    echo "MAC is empty"
    exit  1
fi

CIDRS=$(curl -s http://169.254.169.254/latest/meta-data/network/interfaces/macs/$MAC/vpc-ipv4-cidr-blocks)
if [ -z "$CIDRS" ]; then
    echo "CIDRS is empty"
    exit  1
fi

TEN_RANGE=$(echo $CIDRS | tr ' ' '\n' | grep -c '^10\..*' || true ) # translate the space-separated string into lines, then test with grep
```

I am also contemplating using `--retry 5`, since curl has exponential backoff built-in and `-S` instead of `-s` so that any errors will be output rather than muted.

PR #225, opened 2019-03-20
Fixed retrieval of eth0 MAC address during bootstrap.sh

drewhemm:
*Description of changes:*

* Replaced call to metadata service with cat of
`/sys/class/net/eth0/address`, since the data is available on-instance
* Restored separate setting of `CIDRS` and `TEN_RANGE` variables for easier
debugging
* Use `tr` to translate any spaces to new lines in `CIDRS` variable

*Issue #, if available:*

https://github.com/awslabs/amazon-eks-ami/issues/222

### DIFF
--- a/files/bootstrap.sh
+++ b/files/bootstrap.sh
@@ -134,8 +134,9 @@ kubectl config \
 
 ### kubelet.service configuration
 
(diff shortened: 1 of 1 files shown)


drewhemm:
Hmm, is it possible that an instance's default network interface might not be `eth0` ?

PR #226, opened 2019-03-20
Fixed setting of DNS_CLUSTER_IP in bootstrap.sh

drewhemm:
*Description of changes:*

* Added route table lookup to determine interface handling default route
* Replaced call to metadata service with cat of
`/sys/class/net/$DEFAULT_ROUTE_INTERFACE/address`, since the data is
available on-instance
* Restored separate setting of `CIDRS` and `TEN_RANGE` variables for easier
debugging
* Use `tr` to translate any spaces to new lines in `CIDRS` variable

*Issue #, if available:*

https://github.com/awslabs/amazon-eks-ami/issues/222


drewhemm:
I just checked on my laptop and while connected to both ethernet and wifi, I have two `default` interfaces:

```
$ route
Kernel IP routing table
Destination     Gateway         Genmask         Flags Metric Ref    Use Iface
default         _gateway        0.0.0.0         UG    100    0        0 enp0s31f6
default         _gateway        0.0.0.0         UG    600    0        0 wlp3s0

$ route | grep '^default' | grep -o '[^ ]*$'
enp0s31f6
wlp3s0
```

Could this ever happen on an EC2 instance? So many gotchas to consider!

Adding `head -n 1` as was originally in the code would resolve that:

```
$ route | grep '^default' | head -n 1 | grep -o '[^ ]*$'
enp0s31f6
```

As would setting the `--max-count` flag for grep:

```
$ route | grep -m 1 '^default' | grep -o '[^ ]*$'       
enp0s31f6
```

whereisaaron:
@drewhemm with two default routes you what the one with the lowest metric.

One approach to to ask where there server would route to, for a non-local address, and then take the interface it would use.

```
$ ip route get 0.0.0.0
none 0.0.0.0 via 192.168.20.1 dev eth0 proto unspec src 192.168.20.34
```

So the interface is:
```
$ ip route get 0.0.0.0 | grep -oP 'dev \K\S+'
eth0
```

drewhemm:
@whereisaaron I'm afraid that doesn't work; on both my laptop and an EC2 instance, the default route has a gateway other than `0.0.0.0`, so that command returns the loopback device:

```
$ ip route get 0.0.0.0
local 0.0.0.0 dev lo src 127.0.0.1 uid 1000 
    cache <local>
```

My laptop:
```
$ route
Kernel IP routing table
Destination     Gateway         Genmask         Flags Metric Ref    Use Iface
default         _gateway        0.0.0.0         UG    100    0        0 enp0s31f6
default         _gateway        0.0.0.0         UG    600    0        0 wlp3s0
```

EC2 instance:
```
$ route
Kernel IP routing table
Destination     Gateway         Genmask         Flags Metric Ref    Use Iface
default         gateway         0.0.0.0         UG    0      0        0 eth0
```

But, looking at the route table, my two default interfaces have different metrics and the one with the lowest metric is on top. I need to do some research to see if that is coincidental or predictable.

drewhemm:
Confirmed table ordering is by metric by changing the metric:

Compare
```
$ route                                           
Kernel IP routing table
Destination     Gateway         Genmask         Flags Metric Ref    Use Iface
default         dsldevice.lan   0.0.0.0         UG    600    0        0 wlp3s0
default         0.0.0.0         0.0.0.0         U     1000   0        0 enp0s31f6
```
with
```
route
Kernel IP routing table
Destination     Gateway         Genmask         Flags Metric Ref    Use Iface
default         0.0.0.0         0.0.0.0         U     100    0        0 enp0s31f6
default         dsldevice.lan   0.0.0.0         UG    600    0        0 wlp3s0
```
So we can take the first default entry in the route table. @mogren Are you happy for me to update this PR accordingly?

whereisaaron:
Oh thanks @drewhemm, my mistake for using the special case '0.0.0.0' address as the target/test address 🤦‍♂️. Just use any normal test address. Here is a test on an EC2 instance:
```
$ ip route get 0.0.0.0
local 0.0.0.0 dev lo  src 127.0.0.1 
    cache <local> 
$ ip route get 1.2.3.4
1.2.3.4 via 172.32.4.1 dev eth0  src 172.32.4.140 
    cache 
$ ip route get 8.8.8.8
1.2.3.4 via 172.32.4.1 dev eth0  src 172.32.4.140 
    cache 
```
This can't really not work, since you are asking the kernel, 'hey were would route default traffic, given your route table`.

I'd say asking the kernel is _technically_ more reliable than scraping the route table output. However, I image scraping the route table will also be 100% effective. 😄

drewhemm:
I quite like the idea of using `ip route get`, as long as we are happy to hard code an external IP address like `8.8.8.8`; does that add any kind of dependency on that IP address being up and reachable?

whereisaaron:
@drewhemm no the target IP doesn't even have to exist or be accessible, it is just a routing decision, no packets are sent. The gotcha would be if there was a separate, more specific, routing table entry that matched the test address, e.g. if there was a route table entry to send all 8.8.0.0/16 addresses via `eth1`. 

Whatever you decide, it might be better to base it around the [iproute2](https://en.wikipedia.org/wiki/Iproute2) `ip route` command rather that the deprecated iproute `route` command. Apart from not being deprecated, the `ip` output is generally more script friendly.

Also, it is probably not ideal to rely on the sort order of the `route` human-friendly output, as I don't see anything in the man page that guarantees the sort order. Better to sort yourself and be sure.

If you want to use your suggested method to grep for lowest-metric route. You can use `ip route show default` and sort/sed/head/grep that output. Something like:
```bash
ip route show default | sort -nk 9 | head -1 | grep -oP 'dev \K\S+'
```
That does still assume the number and order of columns, but you can remove that assumption and pick out the name+value pairs you want (in any order in the output), to get something that should be quite robust.
```bash
ip route show default | sed -nre 's/.*(^|\s)((dev|metric) \w+).*\s((dev|metric) \w+).*/\2 \4/p' | sort -nk 3 | head -1 | grep -oP 'dev \K\S+'
```

drewhemm:
Hi @whereisaaron, I tested those two options on an EC2 instance: the first works but the second returns no output...

whereisaaron:
@drewhemm looks like Amazon Linux has a real old version of the `ip` utlity. Although you set the metric when adding route, there appears no way to display the metric when listing routes!! It also doesn't support 'default', you have to use '0.0.0.0/0'.

```
$ ip -V
ip utility, iproute2-ss160111
$ ip --details route
unicast default via 172.25.0.1 dev eth0  proto boot  scope global 
unicast 169.254.169.254 dev eth0  proto boot  scope global 
unicast 172.25.0.0/20 dev eth1  proto kernel  scope link  src 172.25.0.10 
unicast 172.25.0.0/16 via 172.25.0.1 dev eth1  proto boot  scope global
```

Compare with:
```
$ ip -V
ip utility, iproute2-ss180129
$ ip --details route
none default via 192.168.2.1 dev eth0 proto unspec scope global metric 0
none 192.168.2.0/24 dev eth0 proto unspec scope global metric 256
none 192.168.2.40 dev eth0 proto unspec scope global metric 256
```

So with that old `ip` utility is looks like the best you can do is get the first listed default device. This means you aren't sorting by metric and counting on the output order 😞 
```
ip route show 0.0.0.0/0 | head -1 | grep -oP 'dev \K\S+'
```

drewhemm:
@whereisaaron That works on both the EC2 instance and my machine, so I am happy with that. 

Although it looks like the metric isn't displayed with the older version of `ip route`, it merely doesn't show the metric if it is _zero_. I confirmed this by adding two more routes:

```
$ route -n
Kernel IP routing table
Destination     Gateway         Genmask         Flags Metric Ref    Use Iface
0.0.0.0         10.50.112.1     0.0.0.0         UG    0      0        0 eth0
0.0.0.0         0.0.0.0         0.0.0.0         U     5000   0        0 enid4de56097eb
0.0.0.0         0.0.0.0         0.0.0.0         U     6000   0        0 eni00c29516e4a

$ ip route show 0.0.0.0/0
default via 10.50.112.1 dev eth0 
default dev enid4de56097eb scope link metric 5000 
default dev eni00c29516e4a scope link metric 6000
```

The route ordering is still metric-based, again confirmed by altering the metrics of the interfaces, so taking the first one is a solid approach.

I've updated my PR accordingly.

drewhemm:
Is something holding this up from being merged?

micahhausler:
Thanks for the detailed writeup and so sorry for the delayed response.

While this does side-step a possible 404 on the `http://169.254.169.254/latest/meta-data/network/interfaces/macs/` endpoint, there is no guarantee that the `http://169.254.169.254/latest/meta-data/network/interfaces/macs/$MAC/vpc-ipv4-cidr-blocks` endpoint will not 404 either. What do you think of adding a bootstrap argument where you can set the CIDR blocks? That way you could bypass the metadata calls for this case.

drewhemm:
Yeah, I think you're right as I am still occasionally seeing nodes with the wrong cluster DNS. A bootstrap argument should be bulletproof. I will update the PR.

drewhemm:
> What do you think of adding a bootstrap argument where you can set the CIDR blocks? That way you could bypass the metadata calls for this case.

I've amended the PR to use an argument instead, with the default value of `10.100.0.10` unchanged. However, this introduces a breaking change for anyone who needs `DNS_CLUSTER_IP` to be `172.20.0.10` unless they modify their code to add the necessary `--dns-cluster-ip` argument.

drewhemm:
Perhaps a better approach would be to keep the existing logic, but allow the argument to override it. It's a bit messy, but avoids the breaking change.

drewhemm:
> Perhaps a better approach would be to keep the existing logic, but allow the argument to override it. It's a bit messy, but avoids the breaking change.

@micahhausler Any thoughts?

drewhemm:
Hi @micahhausler @mogren, I have updated my PR to fall back to the existing logic if the `--dns-cluster-ip` positional argument is absent. 

I would appreciate your review so this can be merged as it will eliminate the need for us to build this AMI in our pipeline and we want to upgrade to the newer versions of EKS.

files/bootstrap.sh
```diff
@@ -262,11 +268,15 @@ sed -i s,CLUSTER_NAME,$CLUSTER_NAME,g /var/lib/kubelet/kubeconfig
 sed -i s,MASTER_ENDPOINT,$APISERVER_ENDPOINT,g /var/lib/kubelet/kubeconfig
 ### kubelet.service configuration
 
-MAC=$(curl -s http://169.254.169.254/latest/meta-data/network/interfaces/macs/ -s | head -n 1 | sed 's/\/$//')
-TEN_RANGE=$(curl -s http://169.254.169.254/latest/meta-data/network/interfaces/macs/$MAC/vpc-ipv4-cidr-blocks | grep -c '^10\..*' || true )
-DNS_CLUSTER_IP=10.100.0.10
-if [[ "$TEN_RANGE" != "0" ]] ; then
-    DNS_CLUSTER_IP=172.20.0.10;
+if [ -z ${DNS_CLUSTER_IP+x} ]; then
```
mogren: @drewhemm Wait a minute, should the `+x` really be there or was that just for testing?

drewhemm: @mogren It's something to do with how the `if` condition handles the variable if it is not defined:

https://stackoverflow.com/questions/48926946/what-does-variablex-mean-in-bash

So it's for testing in the context of testing the condition, not development/test code

mogren: Ah... I should have checked the docs properly. I wasn't sure about the `+`. 

https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_02

Makes sense!

PR #230, opened 2019-03-23
Fixed an explanation for the usage of bootstrap.sh with option

ffjlabo:
*Issue #, if available:*

*Description of changes:*
- fix the phrase for how to use bootstrap.sh with an option 
from `"usage: $0 [options] <cluster-name>"` to `"usage: $0 <cluster-name> [options]"`.

This is because that <cluster-name> is `$1` and [options] is `$2` in the script.


mogren:
Wait a minute, the documentation is correct, since we shift the arguments. The $2 is the argument to the option.

PR #248, opened 2019-04-19
Allow custom config for docker daemon

shyamjvs:
This will allow users to customize their docker daemon, by passing `--docker-config-json=<json contents of docker config file>` through the BootstrapArguments parameter in the worker node cloudformation template - https://amazon-eks.s3-us-west-2.amazonaws.com/cloudformation/2019-02-11/amazon-eks-nodegroup.yaml.

/cc @micahhausler

### DIFF
--- a/files/bootstrap.sh
+++ b/files/bootstrap.sh
@@ -22,6 +22,7 @@ function print_help {
     echo "--kubelet-extra-args Extra arguments to add to the kubelet. Useful for adding labels or taints."
     echo "--enable-docker-bridge Restores the docker default bridge network. (default: false)"
     echo "--aws-api-retry-attempts Number of retry attempts for AWS API call (DescribeCluster) (default: 3)"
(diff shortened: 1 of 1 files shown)


deiwin:
Another option to perhaps consider would be to merge the user-provided JSON with the default. This would have the benefit of automatically getting all the updates of the default JSON from subsequent AMI versions.

We currently use the following snippet to merge these.

```bash
# Apply our Docker conf, merging with existing
{
cat /etc/docker/daemon.json
cat << EOF
  {
    "default-ulimits": {
      "nofile": {
        "Name": "nofile",
        "Soft": 1048576,
        "Hard": 1048576
      }
    },
    "mtu": 8951,
    "log-opts": {
      "max-size": "50m",
      "max-file": "3"
    }
  }
EOF
} | jq -s '.[0] * .[1]' > /tmp/docker-daemon.json
mv /tmp/docker-daemon.json /etc/docker/daemon.json
chown root:root /etc/docker/daemon.json

# Restart Docker for conf to take effect
systemctl daemon-reload
systemctl restart docker
```

PR #236, opened 2019-03-29
Add new instance types

mogren:
*Description of changes:*

* Added a1, c5n, m5ad, r5ad

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #250, opened 2019-04-23
Add ARM support

### LINKED ISSUE #125: Add Support for A1 Instance type
EKS Optimized AMI for A1 Instance 
**Why is this needed**:

- Explore possibilities to optimize cost/performance

mcrute:
*Issue #:* Fixes #125

*Description of changes:*
Adds support for the ARM a1 instance class to the AMI builder.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,48 +1,72 @@
-AWS_REGION ?= us-west-2
+DOCKER_VERSION ?= 18.06
+KUBERNETES_BUILD_DATE ?= 2019-03-27
+CNI_VERSION ?= v0.6.0
+CNI_PLUGIN_VERSION ?= v0.7.5
+ARCH ?= x86_64
+BINARY_BUCKET_NAME ?= amazon-eks
+SOURCE_AMI_OWNERS ?= 137112412989
 
-# Defaults to Amazon Linux 2 LTS AMI
-# * use the us-west-2 minimal hvm image
-# https://aws.amazon.com/amazon-linux-2/release-notes/
-SOURCE_AMI_ID ?= $(shell aws \
-	--region $(AWS_REGION) \
-	ec2 describe-images \
-	--output text \
-	--filters \
-		Name=owner-id,Values=137112412989 \
-		Name=virtualization-type,Values=hvm \
-		Name=root-device-type,Values=ebs \
-		Name=name,Values=amzn2-ami-minimal-hvm-* \
-		Name=architecture,Values=x86_64 \
-		Name=state,Values=available \
-	--query 'max_by(Images[], &CreationDate).ImageId')
+ifeq ($(ARCH), arm64)
+INSTANCE_TYPE ?= a1.large
+else
+INSTANCE_TYPE ?= m4.large
+endif
 
-.PHONY: all validate ami 1.12 1.11 1.10
+DATE ?= $(shell date +%Y-%m-%d)
 
-all: 1.12
+AWS_DEFAULT_REGION ?= us-west-2
 
+T_RED := \e[0;31m
+T_GREEN := \e[0;32m
+T_YELLOW := \e[0;33m
+T_RESET := \e[0m
+
+.PHONY: all
+all: 1.10 1.11 1.12
+
+.PHONY: validate
 validate:
-	packer validate eks-worker-al2.json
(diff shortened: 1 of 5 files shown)


files/eni-max-pods.txt
```diff
@@ -11,6 +11,11 @@
 # access the instance metadata, VPC DNS, and Time Sync services from the 32nd IP
 # address onwards. If access to these services is needed from all IP addresses
 # on the interface, we recommend using a maximum of 31 IP addresses per interface.
+a1.medium 8
+a1.large 14
+a1.xlarge 20
+a1.2xlarge 20
+a1.4xlarge 39
```
mogren: Hmm, I think this should be:
```
a1.medium 8
a1.large 29
a1.xlarge 58
a1.2xlarge 58
a1.4xlarge 234
```
See https://github.com/awslabs/amazon-eks-ami/pull/236/files#diff-6ec08afbedd68be4a24ff9c1f9ca668fR14

Makefile
```diff
@@ -1,48 +1,71 @@
...
 validate:
-	packer validate eks-worker-al2.json
+	packer validate \
+		-var instance_type=$(INSTANCE_TYPE) \
+		eks-worker-al2.json
 
-1.10: validate
+.PHONY: k8s
+k8s: validate
+	@echo "$(T_GREEN)Building AMI for version $(T_YELLOW)$(VERSION)$(T_GREEN) on $(T_YELLOW)$(ARCH)$(T_RESET)"
```
mogren: Fancy with all the colors. 👍

Makefile
```diff
@@ -1,48 +1,71 @@
...
+	$(eval SOURCE_AMI_ID := $(shell aws ec2 describe-images \
+		--output text \
+		--filters \
+			Name=owner-id,Values=137112412989 \
```
mogren: `SOURCE_AMI_OWNERS=137112412989` at the top, then `$(SOURCE_AMI_OWNERS)` here?

left4taco:
@mcrute Hi there. I'm trying to use your script to create Arm AMI. But it looks like aws-cli isn't able to find the base image. 
```bash
$ aws ec2 describe-images \
                --output text \
                --filters \
                        Name=owner-id,Values=137112412989 \
                        Name=virtualization-type,Values=hvm \
                        Name=root-device-type,Values=ebs \
                        Name=name,Values=amzn2-ami-minimal-hvm-* \
                        Name=architecture,Values=arm64 \
                        Name=state,Values=available \
                --query 'max_by(Images[], &CreationDate).ImageId'
None
```

This is just the first step of building arm image. Am I using it wrong? Thanks.

mcrute:
@left4taco what region are you trying to use for AMI building? A1 instances are not [available in all regions](https://aws.amazon.com/about-aws/global-infrastructure/regional-product-services/) yet and you won't be able to build the AMI in an unsupported region.

left4taco:
@mcrute I'm using us-west-2 Oregon. A1 instances are available in this region. 

It turned out that `aws ec2 describe-images` command in `Makefile` is using the default one of the system, instead of the one specified in `eks-worker-al2.json`. Problem solved. Thanks.

Makefile
```diff
@@ -1,48 +1,72 @@
...
-	packer validate eks-worker-al2.json
+	packer validate \
+		-var instance_type=$(INSTANCE_TYPE) \
+		eks-worker-al2.json
 
-1.10: validate
+.PHONY: k8s
+k8s: validate
+	@echo "$(T_GREEN)Building AMI for version $(T_YELLOW)$(VERSION)$(T_GREEN) on $(T_YELLOW)$(ARCH)$(T_RESET)"
+	$(eval SOURCE_AMI_ID := $(shell aws ec2 describe-images \
```
sinneduy: @mcrute I don't follow why this is necessary.  Shouldn't packer's `source_ami_filter` be able to choose based on the architecture passed in [here](https://github.com/awslabs/amazon-eks-ami/blob/master/eks-worker-al2.json#L34)?

Doesn't doing this force the `Makefile` to always force using the `source_ami` instead?  https://github.com/awslabs/amazon-eks-ami/blob/master/eks-worker-al2.json#L30

Could you provide some additional insight here?

PR #255, opened 2019-04-30
add support for t3a.* instance types

### LINKED ISSUE #254: add support for t3a instance types
Support for t3a instance types.

**Why is this needed**:
As `t3a.*` instance types are generally preferable to `t3.*`.

jhoblitt:
Note that the limits are copies of the t3 values and may need to be adjusted correct.

Resolves #254


micahhausler:
Closing in favor of #236

PR #264, opened 2019-05-08
Use kubernetes minor version to choose kubelet config

Jeffwan:
*Issue #, if available:*
#268 (Thanks @apr-1985)

*Description of changes:*
`KUBERNETES_VERSION ` expects a revision version number but this is used to compare with minor version and get right kubelet config files. It will always use `kubelet-config-with-secret-polling.json` for 1.10 and 1.11. 

Get minor version var from original version helps in this case.

### DIFF
--- a/eks-worker-al2.json
+++ b/eks-worker-al2.json
@@ -91,9 +91,8 @@
       "script": "install-worker.sh",
       "environment_vars": [
         "KUBERNETES_VERSION={{user `kubernetes_version`}}",
-        "BINARY_BUCKET_NAME={{user `binary_bucket_name`}}",
-        "KUBERNETES_VERSION={{user `kubernetes_version`}}",
         "KUBERNETES_BUILD_DATE={{user `kubernetes_build_date`}}",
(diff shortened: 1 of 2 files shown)


eks-worker-al2.json
```diff
@@ -90,10 +90,9 @@
       "type": "shell",
       "script": "install-worker.sh",
       "environment_vars": [
-        "KUBERNETES_VERSION={{user `kubernetes_version`}}",
-        "BINARY_BUCKET_NAME={{user `binary_bucket_name`}}",
         "KUBERNETES_VERSION={{user `kubernetes_version`}}",
```
Jeffwan: This is a duplicate record.

apr-1985:
I raised this issue https://github.com/awslabs/amazon-eks-ami/issues/268, before I saw this PR was open. Can use this issue number if you like

Jeffwan:
@apr-1985 Thanks. Attach the issue in the description.  I tested the packer scripts and they looks good to me.

Issue #249, opened 2019-04-19
Make TSC the default system clock source


cwhatley:
**What would you like to be added**:

Make the default system clock source "tsc" instead of "xen".

**Why is this needed**:

Docker containers that do frequent system time lookups can be badly affected by the overhead around the xen system clock source.

AWS's own tuning recommendations suggest this switch.

Is there any reason not to switch to tsc as the default for the EKS images?

JasonSwindle:
I have a soon pending PR for this.  I need to test it first.


https://github.com/JasonSwindle/amazon-eks-ami/commit/9d57d76229bd21ab78f827c5668e03b7d6a60621

JasonSwindle:
This issue can be closed.

micahhausler:
Closed in #272

cwhatley:
This doesn't appear to be working anymore. See comments in #321

mattbmidigator:
seeing the same behavior in v1.15.11-eks-065dce

cartermckinnon:
Has anyone continued to experience this issue on currently-supported AMI's?

bryantbiggs:
I don't know if its related, but I usually do not have to open NTP/123 on my NACLs/security groups but with the EKS AMI I am seeing a lot of REJECTs in flow logs 

Per this https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/set-time.html
> The Amazon Time Sync Service is available through NTP at the 169.254.169.123 IPv4 address or the fd00:ec2::123 IPv6 address for any instance running in a VPC. The IPv6 address is only accessible on [Instances built on the Nitro System](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-types.html#ec2-nitro-instances). Your instance does not require access to the internet, and you do not have to configure your security group rules or your network ACL rules to allow access. The latest versions of Amazon Linux 2 and Amazon Linux AMIs synchronize with the Amazon Time Sync Service by default.

Is there any way to use this service at 169.254.169.123 instead of having to open up NTP/123 to a public route?

cartermckinnon:
Looking into this. I think we should remove the `chrony` config, as the base AL2 AMI should already do what's necessary for TSS.

[...]

PR #272, opened 2019-05-21
Making tsc the clock source

JasonSwindle:
*Issue #, if available:*
#249

*Description of changes:*
Make tsc the clock default.

### DIFF
--- a/install-worker.sh
+++ b/install-worker.sh
@@ -79,9 +79,9 @@ EOF
 
 # Make tsc the clock source
 if grep --quiet tsc /sys/devices/system/clocksource/clocksource0/available_clocksource; then
-    sudo bash -c "echo tsc > /sys/devices/system/clocksource/clocksource0/current_clocksource"
+    echo "tsc" | sudo tee /sys/devices/system/clocksource/clocksource0/current_clocksource
 else
-    sudo logger "tsc as a clock source is not applicable, skipping."
+    echo "tsc as a clock source is not applicable, skipping."
 fi
 
 ################################################################################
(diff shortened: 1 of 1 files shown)


install-worker.sh
```diff
@@ -73,6 +77,13 @@ cat <<EOF | sudo tee -a /etc/chrony.conf
 rtcsync
 EOF
 
+# Make tsc the clock source
+if grep --quiet tsc /sys/devices/system/clocksource/clocksource0/available_clocksource; then
+    sudo bash -c "echo tsc > /sys/devices/system/clocksource/clocksource0/current_clocksource"
+else
+    sudo logger "tsc as a clock source is not applicable, skipping."
```
micahhausler: Can you make this just `echo "tsc..."`, no sudo required

install-worker.sh
```diff
@@ -73,6 +77,13 @@ cat <<EOF | sudo tee -a /etc/chrony.conf
 rtcsync
 EOF
 
+# Make tsc the clock source
+if grep --quiet tsc /sys/devices/system/clocksource/clocksource0/available_clocksource; then
+    sudo bash -c "echo tsc > /sys/devices/system/clocksource/clocksource0/current_clocksource"
```
micahhausler: You can probably just do
```bash
echo "tsc" | sudo tee /sys/devices/system/clocksource/clocksource0/current_clocksource
```

instead of echoing in a shell

JasonSwindle:
Changes made.  Thank you for the review. :)

jacknagel:
Nitro-based instances use kvm-clock instead of xen, is it really advisable to switch to tsc in that case?

mogren:
Hmm, good point. We should check if `/sys/devices/system/clocksource/clocksource0/current_clocksource` is `xen` and only set it to `tsc` in that case.

Issue #271, opened 2019-05-21
Bootstrap configure wrong server name on kubeconfig


jaksonwkr:
**What happened**: I raised up a new cluster with Terraform, the workers don't get registered into the cluster, when I checked out into logs details, the kubelet was not working, the kubeconfig was misconfigured, the ca `/etc/kubernetes/pki/ca.crt` was wrong as well.

The file `/tmp/describe_cluster_result.txt` created by bootstrap.sh was filled with the content below:

```
None	None
``` 

**What you expected to happen**: The bootstrap got the API Endpoint and CA correctly from AWS API.

**How to reproduce it (as minimally and precisely as possible)**:
It's not something easy to reproduce because it depends on the time and on AWS API that has eventual consistency.

**Anything else we need to know?**:

**Environment**:
- AWS Region: us-east-1
- Instance Type(s): c5.xlarge
- EKS Platform version (use `aws eks describe-cluster --name <name> --query cluster.platformVersion`): eks.2
- Kubernetes version (use `aws eks describe-cluster --name <name> --query cluster.version`): 1.12
- AMI Version: amazon-eks-node-1.12-v20190329
- Kernel (e.g. `uname -a`): 4.14.106-97.85.amzn2.x86_64
- Release information (run `cat /etc/eks/release` on a node): 
```
BASE_AMI_ID="ami-027c5e2ccf2970def"
BUILD_TIME="Fri Mar 29 20:07:35 UTC 2019"
BUILD_KERNEL="4.14.104-95.84.amzn2.x86_64"
ARCH="x86_64"
```

whereisaaron:
@jaksonwkr did this Terraform build used to work? If so, do you use custom domain names for nodes? If so, this [recent change/fix for certificates](https://github.com/awslabs/amazon-eks-ami/issues/244#issuecomment-494139338) might be relevant.

> If you are using a custom non-EC2 assigned hostname, you'll need to configure the kubelet using the --hostname-override flag to an EC2-identifiable hostname (Ex: the instance's private DNS name) so that our certificate signer can verify that the kubelet certificate is for a legitimate target.

jaksonwkr:
@whereisaaron I didn't build the AMI with Terraform, I used the official AMI and I'm not using a custom domain. 
I think if the command `aws eks describe-cluster` used to get that information returns "None" it should retry instead pass to the next step and could resolve this problem.

jaksonwkr:
Problem solved. Closing issue.

PR #274, opened 2019-05-21
Retry if AWS API returns None

jaksonwkr:
*Issue #, if available:*

#271 

*Description of changes:*

Retry command `aws eks describe-cluster` if API returns None.


jaksonwkr:
Hi @micahhausler, thanks for your feedback. 
I did a few more tests and checked that is because the cluster isn't active yet. I'll change to wai until the cluster is ready.

PR #276, opened 2019-05-30
Restrict kubelet to only use TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256 for tls cipher suites

leakingtapan:


whereisaaron:
Perhaps this change is related to problem in #281 with `kubectl logs`?

PR #279, opened 2019-06-04
files/kubelet: specify "TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256"

gyuho:
~~ref. https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-52r1.pdf~~

Edit: Changing to `TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256`, root cause is still pending for investigation.

### DIFF
--- a/files/kubelet-config-with-secret-polling.json
+++ b/files/kubelet-config-with-secret-polling.json
@@ -31,5 +31,5 @@
   "serializeImagePulls": false,
   "serverTLSBootstrap": true,
   "configMapAndSecretChangeDetectionStrategy": "Cache",
-  "tlsCipherSuites": ["TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256"]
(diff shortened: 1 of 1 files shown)


whereisaaron:
That’s a five year-old 2014 recommendation, and even then it was the ‘allowed’ list, not the best. Should we still be using encryption without forward secrecy in 2019? And SHA hashes? We just got done disabling SHA in all our systems in 2016 😄 

We can certainly expand the list, but I personal think any list these days should have TLS 1.2, SHA256, and forward secrecy as a minimum requirement. Is there a reason to compromise?

leakingtapan:
@whereisaaron we are changing to TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256 since the generated key on kubelet is EC key.

whereisaaron:
Perhaps this change is related to problem in #281 with `kubectl logs`?

PR #295, opened 2019-07-01
remove kubectl dependency

M00nF1sh:
*Issue #, if available:*
N/A

*Description of changes:*
Remove kubectl dependency

### DIFF
--- a/files/bootstrap.sh
+++ b/files/bootstrap.sh
@@ -159,13 +159,7 @@ fi
 echo $B64_CLUSTER_CA | base64 -d > $CA_CERTIFICATE_FILE_PATH
 
 sed -i s,CLUSTER_NAME,$CLUSTER_NAME,g /var/lib/kubelet/kubeconfig
-kubectl config \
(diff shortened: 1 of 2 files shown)


naraink:
@M00nF1sh Could you please point me to place where this is discussed and decided to remove kubectl dependency?

When we pulled latest AMI today, our bootstrap scripts started failing.

M00nF1sh:
@naraink Sorry for broken your bootstrap scripts...
We planned to update the `container roadmap` for a deprecation period but missed to do so.
For migration, you can install the kubectl binary from `amazon-eks` S3 bucket, we'll continue vendor it.
BTW, The purpose of removing kubectl binary from ami is because it's an unnecessary parts of AMI(in terms of functionality)

naraink:
@M00nF1sh thanks for the response. OK got it. I see in EKS AMI makescript you guys were taking kubectl version and build date as input. What would be the suggested way to programmatically retrieve the right kubectl version, as we bake our own AMi using AWS AMi as base AMi. We use packer too.

PR #296, opened 2019-07-02
Handle unset AWS_ACCESS_KEY_ID

willthames:
*Description of changes:*

Use `aws sts get-caller-identity` to check whether AWS CLI is installed, rather than the presence of an unnecessary AWS_ACCESS_KEY_ID variable (instance profiles, AWS CLI profiles, and others are reasons it's not necessary)

### DIFF
--- a/scripts/install-worker.sh
+++ b/scripts/install-worker.sh
@@ -216,7 +216,7 @@ BINARIES=(
     aws-iam-authenticator
 )
 for binary in ${BINARIES[*]} ; do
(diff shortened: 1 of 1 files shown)


danielkubat:
thanks for this PR. related issue: https://github.com/awslabs/amazon-eks-ami/issues/527

willthames:
I've updated this PR to use a better check now

cartermckinnon:
We need to overhaul the way creds are handled with Packer in general -- we shouldn't be passing keys via env vars like this. I'm going to close this, because this approach isn't getting to the underlying issue.

PR #340, opened 2019-09-26
Added max-pods-override flat to bootstrap.sh

jwenz723:
*Issue #, if available:*
n/a

*Description of changes:*
I would like to propose adding a flag that would allow the ability to override the number of max pods per node that is stored in the max pods file. This change would assist in being able to set the max pods value to a lower value when using the AWS VPC CNI Custom Networking (results in 1 less ENI available, thus less pods).

### DIFF
--- a/files/bootstrap.sh
+++ b/files/bootstrap.sh
@@ -17,6 +17,7 @@ function print_help {
     echo ""
     echo "-h,--help print this help"
     echo "--use-max-pods Sets --max-pods for the kubelet when true. (default: true)"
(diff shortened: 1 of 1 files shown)


jwenz723:
An alternative approach would be to add the max number of pods capable for each instance type while making use of VPC CNI Custom Networking into the [eni-max-pods.txt](https://github.com/awslabs/amazon-eks-ami/blob/master/files/eni-max-pods.txt) file, for instance:

```
c5.large 29
c5.large.customnetworking 20
```

jwenz723:
Figured out I can just use the kubelet args like this:

```
--use-max-pods false --kubelet-extra-args --max-pods=123
```

Issue #207, opened 2019-02-27
Worker node fails join cluster with --rotate-certificates=true


Constantin07:
**What happened**:

I've passed `--kubelet-extra-args --feature-gates=RotateKubeletServerCertificate=true --rotate-certificates=true` to the `bootstrap.sh` script but the nodes fail to join the cluster.

**What you expected to happen**:
Worker nodes to join cluster as per https://kubernetes.io/docs/tasks/tls/certificate-rotation

**How to reproduce it (as minimally and precisely as possible)**:
Spin up worker nodes with the above args.

**Environment**:
- AWS Region: eu-west-1
- Instance Type(s): t3.large
- EKS Platform version: eks.1
- Kubernetes version: 1.11
- AMI Version: ami-098fb7e9b507904e7
- Kernel (e.g. `uname -a`): ```Linux ip-10-0-2-116.eu-west-1.compute.internal 4.14.97-90.72.amzn2.x86_64 #1 SMP Tue Feb 5 20:46:19 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux```
- Release information (run `cat /etc/os-release` on a node):
```
NAME="Amazon Linux"
VERSION="2"
ID="amzn"
ID_LIKE="centos rhel fedora"
VERSION_ID="2"
PRETTY_NAME="Amazon Linux 2"
ANSI_COLOR="0;33"
CPE_NAME="cpe:2.3:o:amazon:amazon_linux:2"
HOME_URL="https://amazonlinux.com/"
```

micahhausler:
We don't support certificate-based auth for EKS nodes, but we do have server certificates that auto rotate, see [kubelet-config.json](https://github.com/awslabs/amazon-eks-ami/blob/master/files/kubelet-config.json#L29).

Constantin07:
Thanks @micahhausler Am I correct that the `--rotate-certificates=true` is used for certificate-based auth ?

micahhausler:
Yep. [`{"featureGates": {"RotateKubeletServerCertificate": true}}`](https://github.com/awslabs/amazon-eks-ami/blob/master/files/kubelet-config.json#L29) in the kubelet config file turns on auto-rotation the kubelet's HTTPS serving certificate.

Can we close this out?

Constantin07:
@micahhausler sure. Many thanks for explanation.

Issue #99, opened 2018-11-15
Consider applying kube-bench/CIS k8s benchmark to node


tcotav:
I've attached a run of [kube-bench](https://github.com/aquasecurity/kube-bench) that applies the [k8s CIS benchmark](https://www.cisecurity.org/benchmark/kubernetes/) against the nodes.  The remediations are included in the output of the tool.

Should this be the defaults for the created AMI (or should it be applied in another fashion)?  Obviously I think it should be the default.
 
[kube-bench_node_output.txt](https://github.com/awslabs/amazon-eks-ami/files/2586281/kube-bench_node_output.txt)

brycecarman:
I took a look at kube-bench and completely agree that these tests should be passing. It looks like there are some issues with the tool that invalidate a lot of the results.

I didn't look into all the checks yet, but this is what I have found so far:
   * The tool looks in the wrong locations for config files. Rather then telling you it can't find a file, it just marks the check as failed.
   * The tool expects all kubelet arguments to be passed on the commandline. We just switched over to using a kubelet config file so the tool doesn't see a lot of the actual configuration.

tcotav:
I'll create a custom config and see if I can get more valid results (and pass them along).

hobbsh:
So I am not using k8s CIS benchmarks (just ubuntu/al2 benchmarks) here: https://github.com/hobbsh/hardened-eks-ami. Would be great to get k8s benchmarks in anywhere!

brandond:
We've got this on our internal backlog here at work, I'd be glad to put some time into a PR if no-one else has started work on this yet.

johndavies91:
> I took a look at kube-bench and completely agree that these tests should be passing. It looks like there are some issues with the tool that invalidate a lot of the results.
> 
> I didn't look into all the checks yet, but this is what I have found so far:
> 
> * The tool looks in the wrong locations for config files. Rather then telling you it can't find a file, it just marks the check as failed.
> * The tool expects all kubelet arguments to be passed on the commandline. We just switched over to using a kubelet config file so the tool doesn't see a lot of the actual configuration.

I believe this is now solved here - https://github.com/aquasecurity/kube-bench/blob/master/job-eks.yaml

I have been able to get 20 checks passing and 2 failing by populating the kubelet-config.json file with the appropriate values. on initial setup my kubelet-config file looks as follows:
```json
{
    \"kind\": \"KubeletConfiguration\",
[...]

PR #347, opened 2019-10-09
Make 'kube-bench' happy.

bmcstdio:
*Issue #, if available:*

#99

*Description of changes:*

This PR updates the kubelet configuration file as recommended by https://github.com/aquasecurity/kube-bench (except, notably, for `--rotate-certificates=true` which according to #207 is not supported).

```
[INFO] 2 Worker Node Security Configuration
[INFO] 2.1 Kubelet
[PASS] 2.1.1 Ensure that the --anonymous-auth argument is set to false (Scored)
[PASS] 2.1.2 Ensure that the --authorization-mode argument is not set to AlwaysAllow (Scored)
[PASS] 2.1.3 Ensure that the --client-ca-file argument is set as appropriate (Scored)
[PASS] 2.1.4 Ensure that the --read-only-port argument is set to 0 (Scored)
[PASS] 2.1.5 Ensure that the --streaming-connection-idle-timeout argument is not set to 0 (Scored)
[PASS] 2.1.6 Ensure that the --protect-kernel-defaults argument is set to true (Scored)
[PASS] 2.1.7 Ensure that the --make-iptables-util-chains argument is set to true (Scored)
[PASS] 2.1.8 Ensure that the --hostname-override argument is not set (Scored)
[PASS] 2.1.9 Ensure that the --event-qps argument is set to 0 (Scored)
[PASS] 2.1.10 Ensure that the --tls-cert-file and --tls-private-key-file arguments are set as appropriate (Scored)
[INFO] 2.1.11 [DEPRECATED] Ensure that the --cadvisor-port argument is set to 0
[FAIL] 2.1.12 Ensure that the --rotate-certificates argument is not set to false (Scored)
[PASS] 2.1.13 Ensure that the RotateKubeletServerCertificate argument is set to true (Scored)
[PASS] 2.1.14 Ensure that the Kubelet only makes use of Strong Cryptographic Ciphers (Not Scored)
[INFO] 2.2 Configuration Files
[PASS] 2.2.1 Ensure that the kubelet.conf file permissions are set to 644 or more restrictive (Scored)
[PASS] 2.2.2 Ensure that the kubelet.conf file ownership is set to root:root (Scored)
[PASS] 2.2.3 Ensure that the kubelet service file permissions are set to 644 or more restrictive (Scored)
[PASS] 2.2.4 Ensure that the kubelet service file ownership is set to root:root (Scored)
[PASS] 2.2.5 Ensure that the proxy kubeconfig file permissions are set to 644 or more restrictive (Scored)
[PASS] 2.2.6 Ensure that the proxy kubeconfig file ownership is set to root:root (Scored)
[WARN] 2.2.7 Ensure that the certificate authorities file permissions are set to 644 or more restrictive (Scored)
[PASS] 2.2.8 Ensure that the client certificate authorities file ownership is set to root:root (Scored)
[PASS] 2.2.9 Ensure that the kubelet configuration file ownership is set to root:root (Scored)
[PASS] 2.2.10 Ensure that the kubelet configuration file has permissions set to 644 or more restrictive (Scored)

== Remediations ==
2.1.12 If using a Kubelet config file, edit the file to add the line rotateCertificates: true.
If using command line arguments, edit the kubelet service file /etc/systemd/system/kubelet.service
on each worker node and add --rotate-certificates=true argument to the KUBELET_CERTIFICATE_ARGS variable.
Based on your system, restart the kubelet service. For example:
systemctl daemon-reload
systemctl restart kubelet.service

2.2.7 Run the following command to modify the file permissions of the --client-ca-file
chmod 644 <filename>


== Summary ==
21 checks PASS
1 checks FAIL
1 checks WARN
1 checks INFO
```


Pluies:
Oooh, exactly what I was looking for! This would be great to be merged in, especially for silly things like `readOnlyPort: 0` making kube-bench complain that the EKS AMI is not compliant because the value isn't set... Despite the default value being 0.

I guess the only "controversial" change here is `protectKernelDefaults`, as it would mean that pods can't then go ahead and modify `sysctl.conf`? I vaguely remember ElasticSearch in particular needing tweaks there ([docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/vm-max-map-count.html))... But in general, I'd say it's _probably_ fine to err on the side of safety in this official image, and let people who need special-cases fork it and allow this instead.

marcincuber:
@bmcstdio Would you mind sharing the command you are running to execute kube-bench on your EKS cluster and to receive those results?

Thanks

Pluies:
@marcincuber (not who you requested, but I have a vested interest in this issue so I'll contribute 😄  ) here's how we run kube-bench here:
https://gist.github.com/Pluies/8836b5353c2e15b1de168a33b0368772
Run this script with a kubectl that has access to EKS, configured to be in the `default` namespace, and it will run the kube-bench job & output the results above.

mogren:
@bmcstdio Sorry this has been stuck without getting reviewed for a while. Would you mind rebasing and resolving the conflict? The change looks good.

bmcstdio:
@mogren no worries, glad it looks good! I've just rebased and fixed the conflicts.

files/kubelet-config.json
```diff
@@ -25,9 +25,12 @@
   "hairpinMode": "hairpin-veth",
   "cgroupDriver": "cgroupfs",
   "cgroupRoot": "/",
+  "eventRecordQPS": 0,
   "featureGates": {
     "RotateKubeletServerCertificate": true
   },
+  "protectKernelDefaults": true,
+  "readOnlyPort": 0,
```
mogren: Thanks!

stefansedich: Is setting eventRecordQPS to 0 a good idea? from what I understand this will disable any rate limiting of event creation, are there any negatives to this?

mogren: @stefansedich You are right that `eventRecordQPS` (`--event-qps`) defaults to 5, and that setting it to 0 will disable rate limiting. The article you linked to says that the default would be good for around 30 pods on a node.  

https://github.com/kubernetes/kubelet/blob/b81ee8851698006169372c9dce21feabc4bc3c99/config/v1beta1/types.go#L232-L239

That said, I have seen [several](https://www.stackrox.com/post/2019/09/12-kubernetes-configuration-best-practices/) [articles](https://rancher.com/docs/rancher/v2.x/en/security/hardening-2.2/) saying 0 is a good default. Do you think having 50 be a better default limit?

stefansedich: @mogren I think the issue here is that the CIS benchmarks state to use `0` and I have never really seen the rationale behind this, to me it appears as though this param should be tuned on a per-cluster basis, I would love input from someone with more knowledge on the subject.

files/bootstrap.sh
```diff
@@ -331,3 +331,13 @@ fi
 systemctl daemon-reload
 systemctl enable kubelet
 systemctl start kubelet
+
+# Wait for the TLS certificate and private key to be present and add these to the config file so they are picked up the next time the kubelet starts.
+# These cannot be added upfront because they are not present when the kubelet first starts, and adding them would cause the kubelet to crash.
+TLS_CERT_FILE="/var/lib/kubelet/pki/kubelet-server-current.pem"
```
M00nF1sh: TBH, i'm against this change, it's risky and ties to undefined behavior.
It works for current kubelet code, but if upstream kubelet changes behavior to be like "if tlsCertFile is specified, stop automatically rotate certificates", then we'll be break.

Additionally, tlsCertFile/tlsPrivateKeyFile flags are not actually used under current code[https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/client-go/util/certificate/certificate_store.go#L125], which means this is just an trick to make kube-bench happy. Instead, i'd prefer to change kube-bench to consider `serverTLSBootstrap` as secure as well

stefansedich: A few of these changes made scare me, which is why I am never in favor of just making kube-bench happy, I think you are right we need to go and make some changes on that end.

Pluies:
Update for anybody looking at this thread, these changes were reverted in #381

files/kubelet-config.json
joebowbeer: @stefansedich Look at the source code.

`eventRecordQPS=0` is unlimited, which the CIS Benchmark warns can result in DOS.

Note that a value of 0 for `--event-qps` flag (deprecated) resulted in a value of 5, which was the default when this flag was omitted.

The default for `eventRecordQPS` (field) is also 5, but an explicit 0 is unlimited.

In other words: Do not add an explicit 0. Just omit it and it will default to 5.

PR #362, opened 2019-10-22
#361 - custom pause container image support

### LINKED ISSUE #361: Allow to specify pause container to be sourced from a custom repo

Override the pause container image to be from a custom repo. Currently this is semi hard coded.

**Why is this needed**:

This is not supported and all images must be sourced from internal repositories.

ajohnstone:
*Issue #, if available:*

fixes #361 

*Description of changes:*

Support passing in custom pause container.


ajohnstone:
@mogren thanks, any idea when a new ami would be baked with this in?

PR #367, opened 2019-11-01
Set kubeReserved and evictionHard in the kubelet-config

natherz97:
### **Overview:**
This change to bootstrap.sh sets kubeReserved and evictionHard in /etc/kubernetes/kubelet/kubelet-config.json for worker nodes. The values used for kubeReserved are dynamically calculated based on the available CPU and memory of the instance. The amount of CPU and memory allocated for a given EC2 instance is calculated using this formula from GKE, https://cloud.google.com/kubernetes-engine/docs/concepts/cluster-architecture#node_allocatable. Note that the values used for evictionHard are hard-coded. 

Here is a related issue: Kube/System Reserved resources should be enabled by default (https://github.com/awslabs/amazon-eks-ami/issues/318)

### **Testing:**
We are testing to see if the memory reserved for the kubelet in kubeReserved is enforced on the worker node. Specifically, we are trying to observe if pods that are requesting a large percentage of available CPU and memory resources on a worker node are continuously evicted and replaced with new pods before they can use memory resources reserved for the kubelet when kubeReserved is set. In the case that kubeReserved is not set, these pods which are causing memory starvation of the kubelet, should continue to run without being recycled. We expect this to cause the worker node to transition a not ready status.

**Procedure:**
1. Launch a cluster with one worker node with instance type t3.medium (2 vCPU, 4 GiB)
2. Create a deployment with the given specification and scale the number of pods to equal the number of gibibytes available to the instance. In this case, scale the deployment to four replicas. 
	$ kubectl create -f mem-deployment.yaml # mem-deployment.yaml is from [mem-deployment.txt](https://github.com/awslabs/amazon-eks-ami/files/3796273/mem-deployment.txt)
	$ kubectl scale deployment cpu-deployment --replicas=4
3. Wait five minutes and then view the cluster events. Check to see if any of the pods were evicted. Since kubeReserved is not set on the worker node, the pods shouldn’t be getting evicted. We expect the worker node to become become unresponsive and enter status “Not Ready.”
	$ kubectl get events --sort-by=.metadata.creationTimestamp
	$ kubectl get nodes # check to see if the node has status “Not Ready”
4. Clean up the deployment.
	$ kubectl delete deployment mem-deployment
5. Now, we want to compare this behavior against having kubeReserved set on your worker node. SSH into your worker node and replace /etc/eks/bootstrap.sh with the version from this PR, which calculates and sets kubeReserved and evictionHard in /etc/kubernetes/kubelet/kubelet-config.json.
6. Rerun the bootstrap.sh script and restart kubelet by running, $ systemctl restart kubectl. See the instructions below for more details about how to rerun bootstrap.sh. 
7. Now repeat steps 2-4. When viewing cluster events, check to see if any of the pods were evicted and if any new pods were created. Since kubeReserved is now set, we expect the pods to be recycling since they’re trying to use memory set aside to keep the kubelet healthy. 
8. Repeat steps 1-7 with other worker node instance types.

Instructions for how to rerun bootstrap.sh on a worker node and restart the kubelet:
1. Modify /etc/eks/bootstrap.sh to be the version included in this PR. You need to run $ sudo vi /etc/eks/bootstrap.sh to modify the file.
2. On the worker node, run: $ curl http://169.254.169.254/latest/user-data
3. Copy and paste variables for B64_CLUSTER_CA, APISERVER_ENDPOINT
4. Run: $ sudo /etc/eks/bootstrap.sh ClusterName  --b64-cluster-ca $B64_CLUSTER_CA --apiserver-endpoint $APISERVER_ENDPOINT
5. Run $ cat /etc/kubernetes/kubelet/kubelet-config.json to confirm that kubeReserved has been set 
6. To restart kubelet to use the modified kubelet-config.json run $ sudo systemctl restart kubelet

### **Results Using a t3.medium Worker Node (2 vCPU, 4 GiB):**

**When kubeReserved and evictionHard are not set in the kubelet-config:**
```
$ kubectl get pods
NAME                              READY   STATUS    RESTARTS   AGE
mem-deployment-5d85bd8c74-fw65j   1/1     Running   0          2m44s
mem-deployment-5d85bd8c74-mgdz5   1/1     Running   0          2m44s
mem-deployment-5d85bd8c74-pgnx5   1/1     Running   0          2m44s
mem-deployment-5d85bd8c74-rlr6c   1/1     Running   0          2m57s
```
```
$ kubectl get nodes
NAME                                            STATUS     ROLES    AGE    VERSION
ip-192-168-180-252.us-west-2.compute.internal   NotReady   <none>   104m   v1.13.11-eks-5876d6
```
Note that not setting kubeReserved causes the worker node to become unavailable. The four pods continue to run and starve the kubelet of resources.

**When kubeReserved and evictionHard are set in the kubelet-config:**

Values for evictionHard and kubeReserved in /etc/kubernetes/kubelet/kubelet-config.json:
```
  "evictionHard": {
    "memory.available": "100Mi",
    "nodefs.available": "10%",
    "nodefs.inodesFree": "5%"
  },
  "kubeReserved": {
    "cpu": "70m",
    "ephemeral-storage": "1Gi",
    "memory": "768Mi"
  }
```
```
$ kubectl get pods
NAME                              READY   STATUS      RESTARTS   AGE
mem-deployment-5d85bd8c74-4q94g   0/1     Evicted     0          54s
mem-deployment-5d85bd8c74-4ts4s   0/1     Pending     0          48s
mem-deployment-5d85bd8c74-6hv7n   1/1     Running     2          89s
mem-deployment-5d85bd8c74-6njp4   0/1     Evicted     0          51s
mem-deployment-5d85bd8c74-7fspp   0/1     Evicted     0          52s
mem-deployment-5d85bd8c74-7qjdl   0/1     Evicted     0          89s
mem-deployment-5d85bd8c74-87kt8   0/1     Evicted     0          53s
mem-deployment-5d85bd8c74-959mc   0/1     OOMKilled   2          98s
mem-deployment-5d85bd8c74-b4rvq   0/1     Evicted     0          49s
mem-deployment-5d85bd8c74-gbfhr   1/1     Running     2          89s
mem-deployment-5d85bd8c74-gzdqt   0/1     Evicted     0          52s
mem-deployment-5d85bd8c74-pbnz4   0/1     Evicted     0          49s
mem-deployment-5d85bd8c74-rqhx4   0/1     Evicted     0          53s
mem-deployment-5d85bd8c74-rs59v   0/1     Evicted     0          49s
mem-deployment-5d85bd8c74-slqts   0/1     Evicted     0          54s
mem-deployment-5d85bd8c74-srdnl   0/1     Evicted     0          51s
mem-deployment-5d85bd8c74-sxkxs   0/1     Evicted     0          54s
mem-deployment-5d85bd8c74-v9px4   0/1     Evicted     0          54s
mem-deployment-5d85bd8c74-x55fb   0/1     Evicted     0          52s
mem-deployment-5d85bd8c74-x8lr4   0/1     Evicted     0          54s
mem-deployment-5d85bd8c74-z79hr   0/1     Evicted     0          50s
```
```
$ kubectl get nodes
NAME                                            STATUS   ROLES    AGE    VERSION
ip-192-168-180-252.us-west-2.compute.internal   Ready    <none>   119m   v1.13.11-eks-5876d6
```
Note that the worker node and the kubelet stays responsive when kubeReserved and evictionHard are set. The pods are evicted and replaced with new pods instead of being allowed to continue running.

### **Conclusion**
Setting kubeReserved and evictionHard by default is necessary because a majority of customers would prefer to have their worker nodes stay alive and responsive over having their pods persist on the node, while potentially using CPU and memory resources that are needed by the kubelet.

### DIFF
--- a/files/bootstrap.sh
+++ b/files/bootstrap.sh
@@ -112,6 +112,91 @@ function get_pause_container_account_for_region () {
     esac
 }
 
+# Helper function which calculates the amount of the given resource (either CPU or memory)
+# to reserve in a given resource range, specified by a start and end of the range and a percentage
+# of the resource to reserve. Note that we return zero if the start of the resource range is
+# greater than the total resource capacity on the node. Additionally, if the end range exceeds the total
+# resource capacity of the node, we use the total resource capacity as the end of the range.
+# Args:
+#   $1 total available resource on the worker node in input unit (either millicores for CPU or Mi for memory)
+#   $2 start of the resource range in input unit
+#   $3 end of the resource range in input unit
+#   $4 percentage of range to reserve in percent*100 (to allow for two decimal digits)
+# Return:
+#   amount of resource to reserve in input unit
+get_resource_to_reserve_in_range() {
+  local total_resource_on_instance=$1
+  local start_range=$2
+  local end_range=$3
+  local percentage=$4
+  resources_to_reserve="0"
+  if (( $total_resource_on_instance > $start_range )); then
+    resources_to_reserve=$(((($total_resource_on_instance < $end_range ? \
+        $total_resource_on_instance : $end_range) - $start_range) * $percentage / 100 / 100))
+  fi
+  echo $resources_to_reserve
+}
+
+# Calculates the amount of memory to reserve for the kubelet in mebibytes from the total memory available on the instance.
+# From the total memory capacity of this worker node, we calculate the memory resources to reserve
+# by reserving a percentage of the memory in each range up to the total memory available on the instance.
+# We are using these memory ranges from GKE (https://cloud.google.com/kubernetes-engine/docs/concepts/cluster-architecture#node_allocatable):
+# 255 Mi of memory for machines with less than 1024Mi of memory
+# 25% of the first 4096Mi of memory
+# 20% of the next 4096Mi of memory (up to 8192Mi)
+# 10% of the next 8192Mi of memory (up to 16384Mi)
+# 6% of the next 114688Mi of memory (up to 131072Mi)
+# 2% of any memory above 131072Mi
+# Args:
+#   $1 total available memory on the machine in Mi
+# Return:
+#   memory to reserve in Mi for the kubelet
+get_memory_mebibytes_to_reserve() {
+  local total_memory_on_instance=$1
+  local memory_ranges=(0 4096 8192 16384 131072 $total_memory_on_instance)
+  local memory_percentage_reserved_for_ranges=(2500 2000 1000 600 200)
+  if (( $total_memory_on_instance <= 1024 )); then
+    memory_to_reserve="255"
+  else
+    memory_to_reserve="0"
+    for i in ${!memory_percentage_reserved_for_ranges[@]}; do
+      local start_range=${memory_ranges[$i]}
+      local end_range=${memory_ranges[(($i+1))]}
+      local percentage_to_reserve_for_range=${memory_percentage_reserved_for_ranges[$i]}
+      memory_to_reserve=$(($memory_to_reserve + \
+          $(get_resource_to_reserve_in_range $total_memory_on_instance $start_range $end_range $percentage_to_reserve_for_range)))
+    done
+  fi
+  echo $memory_to_reserve
+}
+
+# Calculates the amount of CPU to reserve for the kubelet in millicores from the total number of vCPUs available on the instance.
+# From the total core capacity of this worker node, we calculate the CPU resources to reserve by reserving a percentage
+# of the available cores in each range up to the total number of cores available on the instance.
+# We are using these CPU ranges from GKE (https://cloud.google.com/kubernetes-engine/docs/concepts/cluster-architecture#node_allocatable):
+# 6% of the first core
+# 1% of the next core (up to 2 cores)
+# 0.5% of the next 2 cores (up to 4 cores)
+# 0.25% of any cores above 4 cores
+# Args:
+#   $1 total number of millicores on the instance (number of vCPUs * 1000)
+# Return:
+#   CPU resources to reserve in millicores (m)
+get_cpu_millicores_to_reserve() {
+  local total_cpu_on_instance=$1
+  local cpu_ranges=(0 1000 2000 4000 $total_cpu_on_instance)
+  local cpu_percentage_reserved_for_ranges=(600 100 50 25)
+  cpu_to_reserve="0"
+  for i in ${!cpu_percentage_reserved_for_ranges[@]}; do
+    local start_range=${cpu_ranges[$i]}
+    local end_range=${cpu_ranges[(($i+1))]}
+    local percentage_to_reserve_for_range=${cpu_percentage_reserved_for_ranges[$i]}
+    cpu_to_reserve=$(($cpu_to_reserve + \
+        $(get_resource_to_reserve_in_range $total_cpu_on_instance $start_range $end_range $percentage_to_reserve_for_range)))
+  done
+  echo $cpu_to_reserve
+}
+
 if [ -z "$CLUSTER_NAME" ]; then
     echo "CLUSTER_NAME is not defined"
     exit  1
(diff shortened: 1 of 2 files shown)


files/bootstrap.sh
```diff
@@ -112,6 +112,77 @@ function get_pause_container_account_for_region () {
     esac
 }
 
+# Gets resource to reserve in a given region
```
nckturner: I'm not understanding what the "in a given region" refers to

nckturner: Ok, I understand now the link in the PR description.  Can we add the following 3 pieces of text as comments:

How Kubernetes calculates `Allocatable`:
```
Allocatable = Capacity - Reserved - Eviction Threshold
```

Reservations for memory resources:
```
255 MiB of memory for machines with less than 1 GB of memory
25% of the first 4GB of memory
20% of the next 4GB of memory (up to 8GB)
10% of the next 8GB of memory (up to 16GB)
6% of the next 112GB of memory (up to 128GB)
2% of any memory above 128GB
```

Reservations for CPU resources:
```
6% of the first core
1% of the next core (up to 2 cores)
0.5% of the next 2 cores (up to 4 cores)
0.25% of any cores above 4 cores
```

files/bootstrap.sh
```diff
@@ -187,6 +258,15 @@ fi
 KUBELET_CONFIG=/etc/kubernetes/kubelet/kubelet-config.json
 echo "$(jq ".clusterDNS=[\"$DNS_CLUSTER_IP\"]" $KUBELET_CONFIG)" > $KUBELET_CONFIG
 
+# set kubeReserved in the kubelet-config.json
+MEMORY_GI=$(free --si -g | grep Mem | awk '{print $2}')
```
M00nF1sh: 1. free **--si** gives you Gib instead of Gib.  It's '65' vs '62' difference on a system with 62Kib memory.
2. I'd prefer to get ride of the multiplier arg in the function. we can normalize  available_memory to Mib and available_cpu to millicpu before calling it. 
3. I'd prefer to use Mib instead Gib to calculate it preciously.
4. i prefer to use `cat /proc/meminfo  | grep MemTotal | awk '{print $2}` to denote it's memory**Total** (but use free is fine to me :D).

files/bootstrap.sh
```diff
@@ -112,6 +112,77 @@ function get_pause_container_account_for_region () {
...
+#   $1 total available memory in Gi
+# Output: memory to reserve in Mi
+get_memory_mebibytes_to_reserve() {
+  local total=$1
+  local multiplier=1024
+  # 0   4   8   16  128  $total
+  #  25% 20% 10%  6%   2%
+  local regions=(0 4 8 16 128 $total)
+  local percentages=(2500 2000 1000 600 200)
+  if (( $total <= 1 )); then
```
nckturner: This won't work for machines with less than 1 GB of memory, because bash doesn't support math for non-integers AFAIK.  Out of curiosity, why did you pick bash and not python, since theres a decent amount of calculation here?  I'd prefer the latter, and I think the math would be easier to follow and easier to test.  

(There *are* machines with less than 1GB of memory, for example the [t3.nano)](https://aws.amazon.com/ec2/instance-types/).  Whether or not it's realistic for someone to use one, we shouldn't break if they find a way.

wongma7: python is not installed.

We could install it then run the script then uninstall it all during bootstrap.

I think that would take an unacceptable amount of time though

files/bootstrap.sh
wongma7: Good catch on backwards --si arg.

The only reason we had the multiplier arg is because the `regions` (`start` and `end`) args were easier to read. i.e., 64 instead of 64*1024. Not a very good reason in hindsight, better to have all values in the same unit (Mi or milicpu)!

files/bootstrap.sh
wongma7: Also i think it still "works' for machines less than 1GB since if $total is 0, 256 MiB is returned. Anyway, switching from GB to MiB will make this a non-issue.

wongma7: Actually python2 is installed 🤦‍♂ I never even checked.

Let's leave it as a TODO to convert to a python script. Long-term, this much bash will be hard to maintain.

wongma7:
lgtm but the commits look weird

natherz97:
For different instance types here are the values calculated for kubeReserved:

t3.nano (2 vCPU, 0.5 GiB):
"kubeReserved": {
    "cpu": "70m",
    "ephemeral-storage": "1Gi",
    "memory": "255Mi"
}

m5.large (2 vCPU, 8 GiB):
"kubeReserved": {
    "cpu": "70m",
    "ephemeral-storage": "1Gi",
    "memory": "1741Mi"
}

c5.24xlarge (96 vCPU, 192 GiB):
"kubeReserved": {
    "cpu": "310m",
    "ephemeral-storage": "1Gi",
    "memory": "10704Mi"
}

r5.24xlarge (96 vCPU, 768 GiB):
"kubeReserved": {
    "cpu": "310m",
    "ephemeral-storage": "1Gi",
    "memory": "22236Mi"
}

leakingtapan:
> Needs a bash math fix and better documentation of the formulas being used. Additionally, I'd consider doing at least the math portion in python, as I think it would be easier to follow and test.

Agree. As our bootstrap script grow, we definitely should replace it with a go program or python script.

leakingtapan:
@natherz97 and I'm a bit confused by the cpu/mem calculation algorithm. Could you clarify more one how are they calculated?

cshivashankar:
Does it help to set ephermal storage as well ?

PR #368, opened 2019-11-04
Remove the ec2-net-utils package

Kausheel:
*Description of changes:*

I've encountered an issue when using Packer with the following AMI filter: amzn2-ami-hvm-*

The above filter allows users to use the "full" Amazon Linux 2 AMI (not the minimal AMI, or EKS optimized AMI).

The problem is that the full AM2 AMI contains the ec2-net-utils package. This package contains scripts to automatically create route table entries, which interferes with Pod networking. In my tests, the ec2-net-utils package would create route entries in the "Local" Route Table to direct Pod traffic to an incorrect interface. We want the CNI to take care of setting up the route entries, and this package conflicts with that.

The default AMI filter "amzn2-ami-minimal-hvm-*" and also the EKS optimized AMIs don't contain the ec2-net-utils package, so it's clearly not required for the Worker Nodes to function correctly.

This Pull Request will check to see if the ec2-net-utils package is installed, and if yes, it removes it. I've tested that Packer completes the build successfully with the change included. I've launched some Worker Nodes to verify that the package is successfully removed by checking /var/log/yum.log, and also by running `yum list installed`.


install-worker.sh
```diff
@@ -64,6 +64,8 @@ sudo yum install -y \
     unzip \
     wget
 
+if yum list installed | grep ec2-net-utils; then sudo yum remove ec2-net-utils -y -q; fi
+
```
mogren: Thanks, it is true that this package causes problems with the AWS CNI. Could you please add a comment before this line to explain why we remove it.

Kausheel:
I've added a 1 line code commit now.

If you'd like me to be more descriptive, let me know and I'll be happy to do so. If you want, I can include the full paragraph explanation above if you think that level of detail would be important in the commit. Otherwise, for now, I've just added the 1 line explanation.

bhavin192:
@Kausheel thanks for adding this :) This package wasted a lot of time while debugging issue with our image few months ago. It never came to my mind that we can check if it's installed and then remove it like this.

jyotimahapatra:
> I've added a 1 line code commit now.
> 
> If you'd like me to be more descriptive, let me know and I'll be happy to do so. If you want, I can include the full paragraph explanation above if you think that level of detail would be important in the commit. Otherwise, for now, I've just added the 1 line explanation.

@Kausheel We are currently debating over an effort to bring back the ec2-net-utils into the ami. Could you recollect the issues that you observed back then? We would like to make sure we are covering those issues in the tests.

Kausheel:
I've reached out to you offline with some more information.

The summary of the original problem is that ec2-net-utils would try to apply entries in the Linux routing table which override/interfere with the routes created by the CNI. Pod networking would then break.

You can test this by launching a Node without the package, deploy a Pod and verify network to/from the Pod is working. Check the state of the routing table entries.

The install ec2-net-utils, and wait for it to execute (/var/log/messages will have an entry when ec2-net-utils is triggered, which might happen on a DHCP lease refresh). Once the program executes, check the state of Pod networking again, and check for changes in the routing tables. We should observe that the routes managing traffic to/from the Pod have been modified by ec2-net-utils.

PR #392, opened 2020-01-08
Set protectKernelDefaults to true

Pluies:
*Issue

#99 

*Description of changes:*

See CIS Benchmark 4.2.6


install-worker.sh
```diff
@@ -227,6 +227,16 @@ EOF
 sudo mv /tmp/release /etc/eks/release
 sudo chown root:root /etc/eks/*
 
+################################################################################
+### Stuff required by "protectKernelDefaults=true" #############################
+################################################################################
+
+cat <<EOF | sudo tee -a /etc/sysctl.d/99-amazon.conf
+vm.overcommit_memory=1
+kernel.panic=10
```
abeer91, line 236: Any reason to pick 10 here? I was trying to learn more about these config params and ran into these docs

```
The value in this file represents the number of seconds the kernel
waits before rebooting on a panic. When you use the software watchdog,
the recommended setting is 60.
```
https://www.kernel.org/doc/Documentation/sysctl/kernel.txt

mkarklinsScout: Looks like kubernetes expects it to be 10 seconds - https://github.com/kubernetes/kubernetes/blob/d8695d06b7191db56ebbbc0340da263833c9bb6f/pkg/util/sysctl/sysctl.go#L56

@abeer91 Given that I'd need to have this merged. Would you want me to find out why k8s wants exactly 10 second timeout?

abeer91:
Apologies for the delay on this one, I verified the current configuration on the AMI, from my understanding without these params we would be defaulting to the values configured there. So this change isn't going to override any of those params.

```
[ec2-user@ip-192-168-189-37 vm]$ cat overcommit_memory
1
[ec2-user@ip-192-168-189-37 kernel]$ cat panic
10
[ec2-user@ip-192-168-189-37 kernel]$ cat panic_on_oops
1
```

Can you rebase this commit again?

Pluies:
Thanks for the update @abeer91 , it's rebased! :)

> Also do we need to add all the properties here https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/cm/container_manager_linux.go#L391-L399 or will this just fallback to whatever is configured as default?

I have to admit, I'm not 100% sure – @bmcustodio was the original author of these changes in #347 in October 2019, and only set these three parameters. The Kubelet code at the time already did set these parameters, so I assume it's all good?

PR #398, opened 2020-01-21
adding support for china regions

M00nF1sh:
Adding support for China regions.

### DIFF
--- a/install-worker.sh
+++ b/install-worker.sh
@@ -157,11 +157,11 @@ sudo tar -xvf cni-plugins-${ARCH}-${CNI_PLUGIN_VERSION}.tgz -C /opt/cni/bin
 rm cni-plugins-${ARCH}-${CNI_PLUGIN_VERSION}.tgz cni-plugins-${ARCH}-${CNI_PLUGIN_VERSION}.tgz.sha512
 
 echo "Downloading binaries from: s3://$BINARY_BUCKET_NAME"
-S3_DOMAIN="s3-$BINARY_BUCKET_REGION"
-if [ "$BINARY_BUCKET_REGION" = "us-east-1" ]; then
-    S3_DOMAIN="s3"
+S3_DOMAIN="amazonaws.com"
+if [ "$BINARY_BUCKET_REGION" = "cn-north-1" ] || [ "$BINARY_BUCKET_REGION" = "cn-northwest-1" ]; then
+    S3_DOMAIN="amazonaws.com.cn"
 fi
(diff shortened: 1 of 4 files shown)


josselin-c:
You'll need to change the arn for the managed policies in amazon-eks-nodegroup.yaml. They should look like:
```
        -  !Sub "arn:${AWS::Partition}:iam::aws:policy/AmazonEKSWorkerNodePolicy"
```

install-worker.sh
```diff
@@ -157,11 +157,11 @@ sudo tar -xvf cni-plugins-${ARCH}-${CNI_PLUGIN_VERSION}.tgz -C /opt/cni/bin
...
 echo "Downloading binaries from: s3://$BINARY_BUCKET_NAME"
-S3_DOMAIN="s3-$BINARY_BUCKET_REGION"
-if [ "$BINARY_BUCKET_REGION" = "us-east-1" ]; then
-    S3_DOMAIN="s3"
+S3_DOMAIN="amazonaws.com"
+if [ "$BINARY_BUCKET_REGION" = "cn-north-1" ] || [ "$BINARY_BUCKET_REGION" = "cn-northwest-1" ]; then
+    S3_DOMAIN="amazonaws.com.cn"
 fi
-S3_URL_BASE="https://$BINARY_BUCKET_NAME.$S3_DOMAIN.amazonaws.com/$KUBERNETES_VERSION/$KUBERNETES_BUILD_DATE/bin/linux/$ARCH"
+S3_URL_BASE="https://$BINARY_BUCKET_NAME.s3.$BINARY_BUCKET_REGION.$S3_DOMAIN/$KUBERNETES_VERSION/$KUBERNETES_BUILD_DATE/bin/linux/$ARCH"
```
josselin-c: What is $BINARY_BUCKET_NAME for China?

M00nF1sh:
> You'll need to change the arn for the managed policies in amazon-eks-nodegroup.yaml. They should look like:
> 
> ```
>         -  !Sub "arn:${AWS::Partition}:iam::aws:policy/AmazonEKSWorkerNodePolicy"
> ```
Made the change, but CFN/IAM actually is fine to work with `arn:aws:iam::aws:policy/AmazonEKSWorkerNodePolicy` in China partiton, it transforms it to aws-cn on-the-fly

install-worker.sh
M00nF1sh: an private bucket for our internal build pipeline :D

M00nF1sh:
WARNING: our current released binaries won't work with this change, we'll release new set of binaries which contains new authenticator

davidygs:
Does this mean that the managed node group has been hitting the global sts endpoint the whole time? If so, anytime line as to when a new ami will be available to fix this?

M00nF1sh:
@davidygs yes, it depends on aws-sdk-go's default setting. some region hits regional endpoint.
This change will force all region to use it's regional endpoint

PR #402, opened 2020-01-22
Update install-worker.sh and eks-worker-al2.json

zadowsmash:
*Issue #, if available:*
NA

*Description of changes:*

Updated chown command within install-worker.sh. This resolves the  "cannot access ‘/etc/eks/*’: No such file or directory" error.

Added ability to specify remote_folder on shell providers.  When /tmp is mounted with noexec flag, script is unable to execute from default remote_folder.

Added ability to specify volume size on "launch_block_device_mappings". This allows AMIs with a larger root device size to be used.

Added "should wait for iptables lock" on files/1.14/kubelet.service

Added ability to share ami's across accounts


eks-worker-al2.json
```diff
@@ -94,6 +96,7 @@
   "provisioners": [
     {
       "type": "shell",
+      "remote_folder": "{{ user `remote_folder`}}",
       "inline": ["mkdir -p /tmp/worker/"]
```
micahhausler: Won't `/tmp` here need to be substituted out with ```{{ user `remote_folder`}}```?

eks-worker-al2.json
```diff
@@ -29,7 +29,9 @@
     "ssh_username": "ec2-user",
     "temporary_security_group_source_cidrs": "",
     "associate_public_ip_address": "",
-    "subnet_id": ""
+    "subnet_id": "",
+    "remote_folder": "",
```
micahhausler: Should you set this to `/tmp` as a default, so it gets propagated to the `install-worker.sh` and elsewhere?

zadowsmash: Hi @micahhausler ,
Thanks for the review. 

As per the packer doc, the remote_folder is just where the "install-worker.sh" script gets uploaded to and then executed from:
https://www.packer.io/docs/provisioners/shell.html#remote_folder

In this case, /tmp/worker is just used as a staging folder for the other config files (iptables etc) and nothing is executed out of it.

Also, if nothing is specified in the variables for remote_folder, it just defaults to /tmp so its ok to leave it blank.

I have tested my changes on a couple of image builds and it works fine.

Cheers.

micahhausler: You're right that `/tmp` is the packer default, but the `install-worker.sh` script won't know where the default is. It will be easiest to follow if the default here is `/tmp`, and then you can plumb that value down to the bash script

zadowsmash: `remote_folder` in this scenario is where packer will upload the script that it generates. So for example when this runs, packer creates a file that will have the contents of what's in the inline block on the host:
`/tmp/random-name.sh`
```
#!/bin/bash
mkdir -p /tmp/worker
```
So in this case the `install-worker.sh` script doesn't need to know where packer is putting it's executables as all the references in `install-worker.sh` are absolute and aren't affected by where the script is bring ran.

treacher: Any updates on this one? Our team could really use this change as we're using a CIS hardened Amazon Linux AMI which doesn't allow exec from `/tmp`.

micahhausler: Ah got it

files/1.14/kubelet.service
```diff
@@ -5,7 +5,7 @@ After=docker.service
 Requires=docker.service
 
 [Service]
-ExecStartPre=/sbin/iptables -P FORWARD ACCEPT
+ExecStartPre=/sbin/iptables -P FORWARD ACCEPT -w 5
```
micahhausler: Can you remove this change? I don't see any documented reason for this

zadowsmash: Hey @micahhausler 

This was actually already implemented in:

https://github.com/awslabs/amazon-eks-ami/commit/128a7149b6b50f0392d9a6951e24bd30f19f8e2a

I added it to my fork for the time being as it helped an issue we had

zadowsmash:
Hello @micahhausler,
Any chance we can get this one merged?
Is there any other issues you have with it?

Cheers

jitran:
I've reviewed the changes in this pull request and with packer defaulting remote_folder to /tmp, the changes here won't affect the existing consumers of this packer file.

files/1.14/kubelet.service
mogren: We should wait for it to complete, so 👍for this change.

zadowsmash: thanks @mogren  , made sense to me to do the same for 1.14
cheers :)

micahhausler:
Can you rebase to resolve conflicts?

zadowsmash:
> Can you rebase to resolve conflicts?

@micahhausler all done mate.
Cheers :)

mogren:
> > Can you rebase to resolve conflicts?
> 
> @micahhausler all done mate.
> Cheers :)

What did you rebase on? Not `awslabs:master` it seems, GitHub still shows a conflict.

zadowsmash:
Hi @mogren ,
I can't see any conflicts?

image

mogren:
I kept reloading the page, but GitHub wouldn't let me merge it. Today it worked... 🤷‍♂️

Thanks for the fix, and sorry it took such a long time.

zadowsmash:
> I kept reloading the page, but GitHub wouldn't let me merge it. Today it worked... 
> 
> Thanks for the fix, and sorry it took such a long time.

no worries,
Thanks for merging it in :)

PR #376, opened 2019-11-28
Regression: TLS Ciphersuite: restrict to TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256

Pluies:
*Description of changes:*

See section 2.1.14 of the CIS benchmark:

> [2.1.14] Ensure that the Kubelet only makes use of Strong Cryptographic Ciphers
> If using a Kubelet config file, edit the file to set TLSCipherSuites: to TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305,TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305,TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,TLS_RSA_WITH_AES_256_GCM_SHA384,TLS_RSA_WITH_AES_128_GCM_SHA256
> If using executable arguments, edit the kubelet service file /etc/systemd/system/kubelet.service on each worker node and set the below parameter.
> --tls-cipher-suites=TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305,TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305,TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,TLS_RSA_WITH_AES_256_GCM_SHA384,TLS_RSA_WITH_AES_128_GCM_SHA256

Note that this is a regression, this had been set previously in PR #276
but got lost in #352.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #403, opened 2020-01-22
fix master->node communication by change tls suit to recommended by CIS bench

M00nF1sh:
Fix master -> node communication broken by 
1. https://github.com/awslabs/amazon-eks-ami/pull/376
2. https://github.com/awslabs/amazon-eks-ami/pull/276

We generated cluster ca key with ECDSA algorithm instead of RSA, so `TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256` won't work with EKS and will result in broken kubelet.  At least `TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256` should be added. 

However, I added all ciphers suggested by CIS bench mark for two reasons:
1. more choices for TLS negotiation:  e.g. ECDHE-ECDSA-AES256-GCM-SHA384 can be used as well.
2. even these RSA ciphers cannot be used, i added here just in case we have old EKS clusters with RSA ca.


rtripat:
ship it!

PR #408, opened 2020-01-25
update binaries to use latest ones

M00nF1sh:
Update binaries to use latest ones

## Kubernetes versions
* 1.14.9
* 1.13.12
* 1.12.10

## Authenticator version
v0.5.0

NOTE: we haven't released new AMIs using these binaries yet, it's already staged in our pipeline and will be released soon


id-jim-bach:
need to remove `1.11` from all target as well
https://github.com/M00nF1sh/amazon-eks-ami/blob/8b089cb15359e5e41ba6c57def05b896874caf1c/Makefile#L27

PR #424, opened 2020-03-05
Fix regression introduced for images built in us-east-1

ericlarssen-wf:
When https://github.com/awslabs/amazon-eks-ami/commit/7695621fce7e2400eec946bafb3fca7ce521f4fd#diff-4de22a3515c1c580a6b6a45c86b7289c was added, us-east-1 no longer was being stripped from the S3_BASE_URL variable. Specifying the region on us-east-1 will result in an invalid url. In this pr I want to add the logic back in.

### DIFF
--- a/scripts/install-worker.sh
+++ b/scripts/install-worker.sh
@@ -157,11 +157,17 @@ sudo tar -xvf cni-plugins-${ARCH}-${CNI_PLUGIN_VERSION}.tgz -C /opt/cni/bin
 rm cni-plugins-${ARCH}-${CNI_PLUGIN_VERSION}.tgz cni-plugins-${ARCH}-${CNI_PLUGIN_VERSION}.tgz.sha512
 
 echo "Downloading binaries from: s3://$BINARY_BUCKET_NAME"
-S3_DOMAIN="amazonaws.com"
-if [ "$BINARY_BUCKET_REGION" = "cn-north-1" ] || [ "$BINARY_BUCKET_REGION" = "cn-northwest-1" ]; then
-    S3_DOMAIN="amazonaws.com.cn"
+
+S3_DOMAIN="s3.$BINARY_BUCKET_REGION"
+if [ "$BINARY_BUCKET_REGION" = "us-east-1"]; then
+    S3_DOMAIN="s3"
 fi
-S3_URL_BASE="https://$BINARY_BUCKET_NAME.s3.$BINARY_BUCKET_REGION.$S3_DOMAIN/$KUBERNETES_VERSION/$KUBERNETES_BUILD_DATE/bin/linux/$ARCH"
+
+AMZ_DOMAIN="amazonaws.com"
+if [ "$BINARY_BUCKET_REGION" = "cn-north-1" ] || [ "$BINARY_BUCKET_REGION" = "cn-northwest-1" ]; then
+    AMZ_DOMAIN="amazonaws.com.cn"
+fi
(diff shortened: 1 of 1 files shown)


scripts/install-worker.sh
```diff
@@ -157,11 +157,17 @@ sudo tar -xvf cni-plugins-${ARCH}-${CNI_PLUGIN_VERSION}.tgz -C /opt/cni/bin
 rm cni-plugins-${ARCH}-${CNI_PLUGIN_VERSION}.tgz cni-plugins-${ARCH}-${CNI_PLUGIN_VERSION}.tgz.sha512
 
 echo "Downloading binaries from: s3://$BINARY_BUCKET_NAME"
-S3_DOMAIN="amazonaws.com"
+
+S3_DOMAIN="s3-$BINARY_BUCKET_REGION"
```
M00nF1sh: we should use `s3.$region` instead of `s3-region` from the docs.

scripts/install-worker.sh
```diff
@@ -157,11 +157,17 @@ sudo tar -xvf cni-plugins-${ARCH}-${CNI_PLUGIN_VERSION}.tgz -C /opt/cni/bin
...
+    S3_DOMAIN="s3"
+fi
+
+AMZ_DOMAIN="amazonaws.com"
 if [ "$BINARY_BUCKET_REGION" = "cn-north-1" ] || [ "$BINARY_BUCKET_REGION" = "cn-northwest-1" ]; then
-    S3_DOMAIN="amazonaws.com.cn"
+    AMZ_DOMAIN="amazonaws.com.cn"
 fi
-S3_URL_BASE="https://$BINARY_BUCKET_NAME.s3.$BINARY_BUCKET_REGION.$S3_DOMAIN/$KUBERNETES_VERSION/$KUBERNETES_BUILD_DATE/bin/linux/$ARCH"
+S3_URL_BASE="https://$BINARY_BUCKET_NAME.$S3_DOMAIN.$AMZ_DOMAIN/$KUBERNETES_VERSION/$KUBERNETES_BUILD_DATE/bin/linux/$ARCH"
```
M00nF1sh: i don't think this works for legacy bucket(if you mean the DNS name when u say by legacy), which cannot be access via the virtual host style.
Have u tested whether this works? Also, i think `https://s3.$region.amazonaws.com/$bucket-name/$key-name`  should work for legacy bucket as well

ericlarssen-wf: I have to apologize, the link to the documentation is not necessary... In this PR I simply wanted to fix the regression made in this commit https://github.com/awslabs/amazon-eks-ami/commit/7695621fce7e2400eec946bafb3fca7ce521f4fd#diff-4de22a3515c1c580a6b6a45c86b7289c. After trying to build our images with this script in us-east-1 it looks like the S3_URL_BASE equates to https://amazon-eks.s3.us-east-1.amazonaws.com/1.14.9/2020-01-22/bin/linux/amd64 which is invalid. With this update I simply wanted to revert the change to the us-east-1 region case. So it will be https://amazon-eks.s3.amazonaws.com/1.14.9/2020-01-22/bin/linux/amd64 or https://amazon-eks.s3-us-west-2.amazonaws.com/1.14.9/2020-01-22/bin/linux/amd64 depending on the region.

M00nF1sh: no worries.
If it's amazon-eks, then we don't need to change this PR. that bucket is actually in us-west-2. so you should access it via `https://amazon-eks.s3.us-west-2.amazonaws.com/1.14.9/2020-01-22/bin/linux/amd64/kube-proxy.md5`

M00nF1sh: you should be able to build with `make k8s.... binary_bucket_region=us-west-2` or `make 1.14 binary_bucket_region=us-west-2`

ericlarssen-wf: thanks will close!

PR #430, opened 2020-03-12
Consistent Docker GID version in Image

JanisOrlovs:
Docker install across versions change GID for docker, this causes problems for consistency. This commit solves it by adding same GID docker group

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 manifest.json
 *.swp
+.idea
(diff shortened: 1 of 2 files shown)


Jeffwan:
@JanisOrlovs @mogren  Seems 1950 gid is not available and this command will fail. 
Any reason to choose 1950? 

```
INSTALL_DOCKER="${INSTALL_DOCKER:-true}"
if [[ "$INSTALL_DOCKER" == "true" ]]; then
    sudo useradd --gid 1950 docker

 amazon-ebs: useradd: group '1950' does not exist
```

mogren:
Ugh, I assume we need to do: 
```
sudo groupadd -f -g 1950 docker && sudo useradd --gid 1950 docker
```

mmerkes:
@mogren @JanisOrlovs Setting this to a hardcoded GID causes problems for GPU AMIs, and [we're looking at removing it.](https://github.com/awslabs/amazon-eks-ami/pull/622) Do you have any context on why it's desirable to hardcode it and if we stop hard-coding it, is it going to break things that customers might depend on?

PR #436, opened 2020-03-24
Move compressed file to /var/log

mogren:
*Description of changes:*
* Put archived log files in `/var/log/` instead of the bin directory
* Also collect `kube-system_cni-metrics-helper` logs
* Ignore missing logs (Not all nodes run CoreDNS for example)
* Only copy archives matching `/var/log/eks_i*` when uploading to S3 in the SSM doc

@nithu0115, please take a look.


nithu0115:
LGTM

PR #437, opened 2020-03-25
Force create the group id

mogren:
*Description of changes:*
* Make sure the exists a group called `docker` with GID `1950`


Jeffwan:
Great! Thanks for the quick fix. It looks good to me.

dkhandeshi:
this still has an issue - needs `sudo` right before `useradd`
So instead of:
`sudo groupadd -fog 1950 docker && useradd --gid 1950 docker`
it needs to be
`sudo groupadd -fog 1950 docker && sudo useradd --gid 1950 docker`

PR #440, opened 2020-03-26
Remove authenticator binary as a dependency

abeer91:
*Description of changes:*

Moving to use ```aws eks get-token``` instead of using the authenticator binary, so we can reduce the things that have to be bundled in the AMI.

*Testing*

Created an AMI using these changes and it was able to join my cluster.


files/kubelet-kubeconfig
```diff
@@ -16,10 +16,11 @@ users:
...
-      command: /usr/bin/aws-iam-authenticator
+      command: aws
       args:
-        - "token"
-        - "-i"
-        - "CLUSTER_NAME"
         - --region
-        - "AWS_REGION"
\ No newline at end of file
+        - "AWS_REGION"
```
M00nF1sh: the "AWS_REGION" arg should be here if i'm correct..

files/kubelet-kubeconfig
```diff
@@ -16,10 +16,11 @@ users:
...
+        - eks
+        - get-token
```
M00nF1sh: will aws eks get-token use regional sts for us-west-2?
if not, consider add an env `AWS_STS_REGIONAL_ENDPOINTS=regional ` and test it. you should find the sts endpoint in the token itself.

abeer91: Looks like it uses the IAD Endpoint. Will set that environment variable in the Kubeconfig to redirect to regional endpoint.

PR #462, opened 2020-05-04
Cluster CA and API server endpoint in params

shabin-slr:
Accept cluster CA and API server endpoint in parameters and embed them in userdata

*Issue #, if available:*

*Description of changes:*


abeer91:
Hi, can you describe your use case a little more? Since adding these changes means that users of this template will now need to add more properties each time joining nodegroups.

Users today can directly pass these arguments directly into the bootstrap args section.

Issue #466, opened 2020-05-07
boostrap.sh does not work for Local Zones and Wavelength Zones


Jeffwan:
**What happened**:

[AWS Local Zones](https://aws.amazon.com/about-aws/global-infrastructure/localzones/faqs/) and [AWS Wavelength](https://aws.amazon.com/wavelength/faqs/) have different AZ settings. 

The AZ value for the LAX61 Local Zone is `us-west-2-lax-1a`. The AWS_DEFAULT_REGION will be set to `us-west-2-lax-1` which is incorrect, this will be used to call `us-west-2-lax-1.amazonaws.com`

**What you expected to happen**:
call `us-west-2` instead.

**How to reproduce it (as minimally and precisely as possible)**:
Create an instance in Local Zones and Wavelength Zones and this can be easily reproduced by running 
https://github.com/awslabs/amazon-eks-ami/blob/master/files/bootstrap.sh#L199

**Anything else we need to know?**:

**Environment**:
- AWS Region: 
- Instance Type(s):
- EKS Platform version (use `aws eks describe-cluster --name <name> --query cluster.platformVersion`):
- Kubernetes version (use `aws eks describe-cluster --name <name> --query cluster.version`):
- AMI Version:
- Kernel (e.g. `uname -a`):
- Release information (run `cat /etc/eks/release` on a node):

```
```

PR #467, opened 2020-05-07
Change to use instance-identity/document to fetch region directly

Jeffwan:
*Issue #, if available:*
#466 

We are using ZONE to infer the REGION which is not the best way. There're edge cases like #466 describes that this method can not cover. The most elegant way is to read REGION directly from ec2 metadata service. 
https://github.com/awslabs/amazon-eks-ami/blob/master/files/bootstrap.sh#L198-L199

`http://169.254.169.254/latest/dynamic/instance-identity/document` provides the information and here's an output example

```
{
  "accountId" : "xxxx",
  "architecture" : "x86_64",
  "availabilityZone" : "us-west-2d",
  "billingProducts" : null,
  "devpayProductCodes" : null,
  "marketplaceProductCodes" : null,
  "imageId" : "ami-0c13bb9cbfd007e56",
  "instanceId" : "i-040xxxxx5b",
  "instanceType" : "m5.large",
  "kernelId" : null,
  "pendingTime" : "2020-04-02T01:03:19Z",
  "privateIp" : "192.168.x.169",
  "ramdiskId" : null,
  "region" : "us-west-2",
  "version" : "2017-09-30"
}

```

*Description of changes:*
Change to use instance-identity/document to fetch region directly

### DIFF
--- a/files/bootstrap.sh
+++ b/files/bootstrap.sh
@@ -195,9 +195,10 @@ if [ -z "$CLUSTER_NAME" ]; then
     exit  1
 fi
 
-ZONE=$(curl -s http://169.254.169.254/latest/meta-data/placement/availability-zone)
-AWS_DEFAULT_REGION=$(echo $ZONE | awk '{print substr($0, 1, length($0)-1)}')
-AWS_SERVICES_DOMAIN=$(curl -s http://169.254.169.254/2018-09-24/meta-data/services/domain)
+
+TOKEN=$(curl -X PUT -H "X-aws-ec2-metadata-token-ttl-seconds: 600" "http://169.254.169.254/latest/api/token")
+AWS_DEFAULT_REGION=$(curl -s --retry 5 -H "X-aws-ec2-metadata-token: $TOKEN" http://169.254.169.254/latest/dynamic/instance-identity/document | jq .region -r)
+AWS_SERVICES_DOMAIN=$(curl -s --retry 5 -H "X-aws-ec2-metadata-token: $TOKEN" http://169.254.169.254/2018-09-24/meta-data/services/domain)
 
 MACHINE=$(uname -m)
 if [ "$MACHINE" == "x86_64" ]; then


files/bootstrap.sh
```diff
@@ -195,8 +195,7 @@ if [ -z "$CLUSTER_NAME" ]; then
     exit  1
 fi
 
-ZONE=$(curl -s http://169.254.169.254/latest/meta-data/placement/availability-zone)
-AWS_DEFAULT_REGION=$(echo $ZONE | awk '{print substr($0, 1, length($0)-1)}')
+AWS_DEFAULT_REGION=$(curl -s http://169.254.169.254/latest/dynamic/instance-identity/document | jq .region -r)
```
mogren, line 198: To make @nckturner's life a bit easier, can we use IMDSv2 for this? Something like this:
```bash
TOKEN=$(curl -X PUT -H "X-aws-ec2-metadata-token-ttl-seconds: 600" "http://169.254.169.254/latest/api/token")
IMDS_CURL_ARGS="-s -H \"X-aws-ec2-metadata-token: $TOKEN\""
AWS_DEFAULT_REGION=$(curl $IMDS_CURL_ARGS http://169.254.169.254/latest/dynamic/instance-identity/document | jq .region -r)
AWS_SERVICES_DOMAIN=$(curl $IMDS_CURL_ARGS http://169.254.169.254/2018-09-24/meta-data/services/domain)
```

Also, can you confirm that `jq` is always available?

Jeffwan: Haha, for sure. I will upgrade to IMDSv2 directly in this PR to save time on potential migration. 

We install jq here and have some references in bootstrap.sh.
https://github.com/awslabs/amazon-eks-ami/blob/06cb77298984d84fc0edd0a876fb571b576d6ae8/scripts/install-worker.sh#L61

Jeffwan:
@mogren address the comment. Thanks for the suggested scripts. :D

PR #471, opened 2020-05-12
Add a new manifest containing the AMI name

saurav-agarwalla:
*Issue #, if available:*

*Description of changes:* This commit adds a new manifest which contains AMI name in the manifest filename so that parallel builds can be triggered. Even though the new manifest is now generated along with the current one for backwards compatibility, eventually the old manifest (manifest.json) will be deprecated.

### DIFF
--- a/eks-worker-al2.json
+++ b/eks-worker-al2.json
@@ -162,6 +162,11 @@
       "type": "manifest",
       "output": "manifest.json",
       "strip_path": true
+    },
+    {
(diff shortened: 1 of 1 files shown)


saurav-agarwalla:
> this may break customers, are we ok to do so
> if so, we need to call it out in release note or produce two manifest (manifest.json) and (ami-name-manifest.json)

Good point. I have added a new manifest for now for backwards compatibility.

PR #474, opened 2020-05-19
Moved the file copy after the execution of upgrade_kernel.sh

jitran:
Upon running a recent packer build on AL2 today, it failed with the following error while executing the install-worker.sh script:
```
19-May-2020 14:52:03 | amazon-ebs: mv: cannot stat ‘/tmp/worker/iptables-restore.service’: No such file or directory
```

*Description of changes:*
Moved the file copy after the execution of upgrade_kernel.sh. upgrade_kernel.sh performs a system reboot which will remove the contents of /tmp.

I was able to successfully build a new AMI on an x86_64 arch.

### DIFF
--- a/eks-worker-al2.json
+++ b/eks-worker-al2.json
@@ -5,7 +5,7 @@
     "creator": "{{env `USER`}}",
     "encrypted": "false",
     "kms_key_id": "",
-    
+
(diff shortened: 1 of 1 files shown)


jitran:
Hi everyone, thanks for the approval. Can we get the third and last green tick too? We need to build a new AMI next week.

Cheers

PR #517, opened 2020-07-28
update pause container to use multi-arch repo

nprab428:
*Description of changes:*
update pause container to use multi-arch repo


nprab428:
Ran some basic tests to validate cluster will deploy pods/services as expected. It's good to merge now.

PR #525, opened 2020-08-20
AMI release to get Cluster DNS IP from SCIDR for Self Managed Nodegroups

ljosyula:
Issue:
Getting Cluster DNS IP from SCIDR for Self Managed Nodegroups, when DNS IP is not provided by user. This is being done as a part of SCIDR Feature Launch

Description of changes:
Added a code to find SCIDR from Describe Cluster and assign DNS IP using that. This is done as a part of new SCIDR feature, where we allow user to give SCIDR value as well. For Mnaged Nodegroups this is handles in ISO. For unmanaged nodegroups we are handling it by creating a new AMI version. If the SCIDR is given then the Cluster DNS IP is set using that. To make it backward compatible, if SCIDR is null DNS IP will be set 10... or 172... depending on the vpc-ipv4-cidr-blocks as it was happening earlier.

Testing:
Since the feature is not launched yet. I added the service model in the nodegroup created during testing by : aws configure add-model
Created a file in the nodegroup that holds the value of SCIDR fetched from Describe Cluster API and DNS Cluster IP.
Created a cluster with SCIDR, then created a Self managed nodegroup with new Service Model loaded in it and one with current API.
Created a cluster and Self managed nodegroup in external-client-account as well to test backward compatibility

### DIFF
--- a/files/bootstrap.sh
+++ b/files/bootstrap.sh
@@ -100,6 +100,7 @@ set -u
 USE_MAX_PODS="${USE_MAX_PODS:-true}"
 B64_CLUSTER_CA="${B64_CLUSTER_CA:-}"
 APISERVER_ENDPOINT="${APISERVER_ENDPOINT:-}"
+DNS_CLUSTER_IP="${DNS_CLUSTER_IP:-}"
 KUBELET_EXTRA_ARGS="${KUBELET_EXTRA_ARGS:-}"
 ENABLE_DOCKER_BRIDGE="${ENABLE_DOCKER_BRIDGE:-false}"
 API_RETRY_ATTEMPTS="${API_RETRY_ATTEMPTS:-3}"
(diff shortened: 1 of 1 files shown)


files/bootstrap.sh
```diff
@@ -218,7 +219,7 @@ PAUSE_CONTAINER="$PAUSE_CONTAINER_IMAGE:$PAUSE_CONTAINER_VERSION"
 CA_CERTIFICATE_DIRECTORY=/etc/kubernetes/pki
 CA_CERTIFICATE_FILE_PATH=$CA_CERTIFICATE_DIRECTORY/ca.crt
 mkdir -p $CA_CERTIFICATE_DIRECTORY
-if [[ -z "${B64_CLUSTER_CA}" ]] && [[ -z "${APISERVER_ENDPOINT}" ]]; then
+if [[ -z "${DNS_CLUSTER_IP}" ]] || [[ -z "${B64_CLUSTER_CA}" ]] || [[ -z "${APISERVER_ENDPOINT}" ]]; then
```
blxdaniel, line 222: ``[ -z "${B64_CLUSTER_CA}" ]] && [[ -z "${APISERVER_ENDPOINT}" ]`` isn't the ``&&`` important because we need the cluster CA to access to API server?

blxdaniel: it's just a check to get all the info if it's missing

ljosyula: We are trying to get all the info from describe-cluster if any of them does not exists. Therefore, I replaced && with ||

dgr237:
This change has broken any users which are not using managed node groups and are running in a private vpc i.e. we pass the ca bundle and the endpoint but not the DNS IP as we want the bootstrap.sh to calculate this.

PR #552, opened 2020-10-20
Adding t4g instance types + some missing instances types

vishalkg:
*Issue #, if available:*

*Description of changes:*
* Added T4g instance types
* Added instances types which are missing as against [eni-max-pods.txt](https://github.com/awslabs/amazon-eks-ami/blob/master/files/eni-max-pods.txt)

### DIFF
--- a/amazon-eks-nodegroup.yaml
+++ b/amazon-eks-nodegroup.yaml
@@ -88,321 +88,356 @@ Parameters:
     Type: String
     Default: t3.medium
     AllowedValues:
-      - a1.medium
-      - a1.large
-      - a1.xlarge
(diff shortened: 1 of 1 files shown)


vishalkg (commented):
Copied over instance list directly from [max-pods-eni.txt](https://github.com/awslabs/amazon-eks-ami/blob/master/files/eni-max-pods.txt). This ensures, none of the instances are left behind and instances appear in the same order in both places and makes it easy to track and maintain future changes.

abeer91 (approved):

abeer91 (approved):

PR #561, opened 2020-11-15
Missing sha256 check sum for 'else' path

mtilson:
*Issue #, if available:*
N/A

*Description of changes:*
When using 'wget' to fetch cni binaries from s3, checksum file is downloaded but not used


mmerkes (approved):
Thanks for the contribution. Looks reasonable to me.

Issue #567, opened 2020-11-18
Binaries missing/incorrect release notes for v20201117 release


janbeerden:
From the release notes:
```
Binaries used to build these AMIs are published :
s3://amazon-eks/1.18.9/2020-11-07/
s3://amazon-eks/1.17.12/2020-11-07/
s3://amazon-eks/1.16.15/2020-11-07/
s3://amazon-eks/1.15.12/2020-11-07/
```

Neither one of the K8s version directories contains a `2020-11-07` subdirectory, only a `2020-11-02` subdirectory is present.

Are the binaries missing or are the references in the release notes incorrect?

janbeerden:
@vishalkg, I see you updated the changelog for v20201117, perhaps you can identify what the problem is here?

vishalkg:
@janbeerden We have fixed the bucket name.

janbeerden:
@vishalkg, awesome, thanks!

PR #570, opened 2020-11-19
Updated changedlog with correct s3 bucket name

sehayoun:
*Issue #, if available:*
https://github.com/awslabs/amazon-eks-ami/issues/567
*Description of changes:*
For v20201117 AMI release, update s3 binary bucket name from 11-07 to 11-02

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,10 +15,10 @@
 * amazon-eks-node-1.15-v20201117
 
 Binaries used to build these AMIs are published :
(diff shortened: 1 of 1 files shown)


abeer91 (approved):

abeer91:
Actually we need to update the dates here as well - https://github.com/awslabs/amazon-eks-ami/blob/master/Makefile#L47

sehayoun:
Yes, I'm sending one more PR to update the Makefile -  https://github.com/awslabs/amazon-eks-ami/blob/master/Makefile#L47

PR #573, opened 2020-11-23
GPU Boost clock setup for performance improvement

sehayoun:
*Issue #, if available:*

*Description of changes:*
Setup gpu boost clock to optimize gpu performance. Setting the gpu boost clock plays a role in NVswitch performance which is very important in p4d instance type. [reference](https://github.com/aws-samples/aws-efa-nccl-baseami-pipeline/blob/master/nvidia-efa-ami_base/aws-gpu-boost-clock.sh)

### DIFF
--- a/files/bootstrap.sh
+++ b/files/bootstrap.sh
@@ -342,3 +342,31 @@ fi
 systemctl daemon-reload
 systemctl enable kubelet
 systemctl start kubelet
+
+# gpu boost clock
+if  command -v nvidia-smi &>/dev/null ; then
+   echo "nvidia-smi found"
+
+   sudo nvidia-smi -pm 1 # set persistence mode
+   sudo nvidia-smi --auto-boost-default=0
+
+   GPUNAME=$(nvidia-smi -L | head -n1)
+   echo $GPUNAME
+
+   # set application clock to maximum
+   if [[ $GPUNAME == *"A100"* ]]; then
+      nvidia-smi -ac 1215,1410
+   elif [[ $GPUNAME == *"V100"* ]]; then
+      nvidia-smi -ac 877,1530
+   elif [[ $GPUNAME == *"K80"* ]]; then
+      nvidia-smi -ac 2505,875
+   elif [[ $GPUNAME == *"T4"* ]]; then
+      nvidia-smi -ac 5001,1590
+   elif [[ $GPUNAME == *"M60"* ]]; then
+      nvidia-smi -ac 2505,1177
+   else
+      echo "unsupported gpu"
+   fi
+else
+    echo "nvidia-smi not found"
+fi


files/bootstrap.sh
```diff
@@ -342,3 +342,31 @@ fi
 systemctl daemon-reload
 systemctl enable kubelet
 systemctl start kubelet
+
+# gpu boost clock
+if  command -v nvidia-smi &>/dev/null ; then
+   echo "nvidia-smi found"
+
+   sudo nvidia-smi -pm 1 # set persistence mode
```
mmerkes, line 350: Can you add comments on each of these options explaining what they do and why you're setting them to the values you are?

files/bootstrap.sh
```diff
@@ -342,3 +342,31 @@ fi
...
+   sudo nvidia-smi --auto-boost-default=0
+
+   GPUNAME=$(nvidia-smi -L | head -n1)
+   echo $GPUNAME
+
+   # set application clock to maximum
+   if [[ $GPUNAME == *"A100"* ]]; then
+      nvidia-smi -ac 1215,1410
```
mmerkes, line 358: What's the source of these numbers other than the other references script?

mmerkes (commented):
You mention that you're doing this for p4 instances. Does it have any implications on non-p4 instances that people might be running the nvidia driver?

files/bootstrap.sh
sehayoun: - nvidia-smi -pm 1 : this sets persistence mode as enabled, which is a necessary change to be able to change application clock as we want. more info from https://developer.nvidia.com/blog/increase-performance-gpu-boost-k80-autoboost/
- nvidia-smi --auto-boost-default=0 : we disable auto boost mode in favor of application clock. more info from https://developer.download.nvidia.com/compute/DCGM/docs/nvidia-smi-367.38.pdf

files/bootstrap.sh
sehayoun: Those values are from running [nvidia-smi -i 0 -q -d SUPPORTED_CLOCKS | grep -F 'Memory' -A1 | head -n2], which means maximum supported memory and core clock respectively, that we want to set our application clock as. It depends on GPU type we have in our gpu instances, I ran the command for each gpu type to confirm those values.

abeer91 (approved):

files/bootstrap.sh
```diff
@@ -342,3 +342,31 @@ fi
...
+      nvidia-smi -ac 2505,875
+   elif [[ $GPUNAME == *"T4"* ]]; then
+      nvidia-smi -ac 5001,1590
+   elif [[ $GPUNAME == *"M60"* ]]; then
+      nvidia-smi -ac 2505,1177
+   else
+      echo "unsupported gpu"
+   fi
+else
+    echo "nvidia-smi not found"
```
abeer91, line 371: Nit: This log line could be made more verbose like:

nvidia-smi not found, this is expected if using the non-gpu AMI.

mmerkes (approved):

heybronson (approved):

files/bootstrap.sh
zerog2k: was there an answer to:
> Does it have any implications on non-p4 instances that people might be running the nvidia driver?
we use g3/g4 instance families, and also curious if there are implications outside of p4

sehayoun: Yes, impact for G4 and G3 is also included, we're setting them to achieve better performance. You can find more details from https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/optimize_gpu.html

PR #575, opened 2020-12-01
Adding containerd_version as a packer variable

heybronson:
*Issue #, if available:*

*Description of changes:*
- Adds containerd version to the AMI description.
- Adds containerd version as a packer variable. 
- Fixes Makefile to point to correct targets.

Validation Done:
- Created an AMI successfully via `make 1.15`
- Validated that the correct containerd version was installed:
```
[ec2-user@ip-192-168-34-16 ~]$ rpm -q containerd
containerd-1.4.1-2.amzn2.x86_64
```
- Created a customAMI Managed Nodegroup with this AMI. Validated that the AMI joined the cluster and that I was able to run containers on the AMI.

### DIFF
--- a/scripts/install-worker.sh
+++ b/scripts/install-worker.sh
@@ -24,6 +24,7 @@ validate_env_set() {
 validate_env_set BINARY_BUCKET_NAME
 validate_env_set BINARY_BUCKET_REGION
 validate_env_set DOCKER_VERSION
+validate_env_set CONTAINERD_VERSION
 validate_env_set CNI_PLUGIN_VERSION
 validate_env_set KUBERNETES_VERSION
 validate_env_set KUBERNETES_BUILD_DATE
@@ -125,7 +126,7 @@ if [[ "$INSTALL_DOCKER" == "true" ]]; then
     # Due to an issue with the latest containerd, customers are seeing
     # pods getting stuck in terminating, so we need to downgrade containerd
     # until the issue is fixed upstream or we have another workaround.
-    sudo yum downgrade -y containerd-1.3.2-1.amzn2
+    sudo yum downgrade -y containerd-${CONTAINERD_VERSION}
 
     # Enable docker daemon to start on boot.
     sudo systemctl daemon-reload
--- a/scripts/install-worker.sh
+++ b/scripts/install-worker.sh
@@ -122,10 +122,6 @@ if [[ "$INSTALL_DOCKER" == "true" ]]; then
     sudo mv $TEMPLATE_DIR/docker-daemon.json /etc/docker/daemon.json
     sudo chown root:root /etc/docker/daemon.json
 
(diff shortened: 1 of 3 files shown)


scripts/install-worker.sh
```diff
@@ -125,7 +126,7 @@ if [[ "$INSTALL_DOCKER" == "true" ]]; then
     # Due to an issue with the latest containerd, customers are seeing
     # pods getting stuck in terminating, so we need to downgrade containerd
     # until the issue is fixed upstream or we have another workaround.
-    sudo yum downgrade -y containerd-1.3.2-1.amzn2
+    sudo yum downgrade -y containerd-${CONTAINERD_VERSION}
```
ravisinha0506, line 129: Should this be removed or changed to use `yum install`?

scripts/install-worker.sh
```diff
@@ -125,7 +126,7 @@ if [[ "$INSTALL_DOCKER" == "true" ]]; then
     # Due to an issue with the latest containerd, customers are seeing
```
ravisinha0506, line 126: Should this comment be removed?

scripts/install-worker.sh
heybronson: `yum install` will fail if we're trying to install a version that older than the current version installed (containerd is installed as a dependency on docker).

`yum downgrade` won't fail if we're a) downgrading to an older version or b) downgrading to the current version that's already installed when we installed docker.

ravisinha0506:
PR looks good to me!!

scripts/install-worker.sh
```diff
@@ -121,11 +122,7 @@ if [[ "$INSTALL_DOCKER" == "true" ]]; then
     sudo mv $TEMPLATE_DIR/docker-daemon.json /etc/docker/daemon.json
     sudo chown root:root /etc/docker/daemon.json
 
-    # https://github.com/awslabs/amazon-eks-ami/issues/563
-    # Due to an issue with the latest containerd, customers are seeing
-    # pods getting stuck in terminating, so we need to downgrade containerd
-    # until the issue is fixed upstream or we have another workaround.
-    sudo yum downgrade -y containerd-1.3.2-1.amzn2
+    sudo yum downgrade -y containerd-${CONTAINERD_VERSION}
```
mmerkes, line 125: Will this work if you pass in either a higher or the same version of containerd?

mmerkes (commented):
Overall, looks good. One comment. Also, please update the PR description with the testing that you've done to verify that this all works.

scripts/install-worker.sh
heybronson: This won't work if we pass in a higher version than what's already installed.

This is a no-op if we pass the same version as what's already installed.

By default, yum installs the latest version of containerd that's compatible with our pinned dockerd version.

mmerkes (approved):

PR #578, opened 2020-12-10
Update systemd to always restart kubelet to support dynamic kubelet configuration

chotiwat:
*Description of changes:*
We were trying to enable [dynamic kubelet configuration](https://kubernetes.io/docs/tasks/administer-cluster/reconfigure-kubelet/) on our cluster but the new nodes failed to become ready because `systemd` didn't restart `kubelet.service` after it cleanly terminated to reload the new configuration.

`kubelet` reloads dynamic configuration by exiting the process with code 0 and assume that a babysitter process will restart the service. We should configure `systemd` to restart `kubelet` on exit code 0 as well (or use `Restart=always`).

See:
- https://kubernetes.io/docs/tasks/administer-cluster/reconfigure-kubelet/#understanding-how-the-kubelet-checkpoints-config
- https://github.com/kubernetes/kubernetes/blob/4c00c3c459261e8ff3381c1070ddf798f0131956/pkg/kubelet/kubeletconfig/configsync.go#L191-L207

*Workaround*
Since we are already using a userdata script to add `--dynamic-config-dir` as an extra argument, we add the below script to append the exit code to `kubelet.service` configuration.
```bash
cat << EOF > /etc/systemd/system/kubelet.service.d/90-restart-on-exit-code-0.conf
[Service]
RestartForceExitStatus=0
EOF
```

### DIFF
--- a/files/kubelet.service
+++ b/files/kubelet.service
@@ -13,7 +13,7 @@ ExecStart=/usr/bin/kubelet --cloud-provider aws \
     --network-plugin cni $KUBELET_ARGS $KUBELET_EXTRA_ARGS
 
 Restart=on-failure
-RestartForceExitStatus=SIGPIPE SIGKILL
+RestartForceExitStatus=0 SIGPIPE SIGKILL
 RestartSec=5
 KillMode=process
 
--- a/files/kubelet.service
+++ b/files/kubelet.service
@@ -12,8 +12,7 @@ ExecStart=/usr/bin/kubelet --cloud-provider aws \
     --container-runtime docker \
     --network-plugin cni $KUBELET_ARGS $KUBELET_EXTRA_ARGS
 
-Restart=on-failure
-RestartForceExitStatus=0 SIGPIPE SIGKILL
+Restart=always
 RestartSec=5
 KillMode=process
 


heybronson (dismissed):
Thanks for the PR, LGTM.

files/kubelet.service
```diff
@@ -13,7 +13,7 @@ ExecStart=/usr/bin/kubelet --cloud-provider aws \
     --network-plugin cni $KUBELET_ARGS $KUBELET_EXTRA_ARGS
 
 Restart=on-failure
```
heybronson, line 15: Let's instead change this to `Restart=always`.

chotiwat: Sounds good

chotiwat: I have also updated the PR title accordingly

heybronson (approved):

mmerkes (approved):
LGTM

mmerkes:
If I remember correctly, Dynamic Kubelet config was stuck in beta for a long time, and SIG-Node has discussed removing it due to a lack of understanding usage and not having an advocate in the SIG. If this is a feature that you depend on, that's something you should be aware of. I would share your use case with them, and I'm happy to help advocate if I understand your use case.

chotiwat:
@mmerkes Thanks for the heads up! I didn't know that.
Our team is in a process of migrating from Kops to EKS and we are trying to achieve feature parity with what we have on Kops. We use Dynamic Kubelet Config mainly to configure system resource reservation and eviction thresholds dynamically without the need to rolling update the whole cluster (similar to https://github.com/kubernetes/enhancements/issues/281#issuecomment-695751859). This has been especially useful when we run into out-of-resource issues on the nodes and need to quickly tune the cluster outside of the maintenance window. While it would not break our clusters if the feature ends up getting deprecated, it would slow down our development and operational speed unless an alternative is offered.

PR #521, opened 2020-08-10
Change cgroup driver to systemd

reegnz:
Fixes #490

*Issue #, if available:*
#490

*Description of changes:*

Kubernetes documentation indicates that for stability reasons
one should run kubernetes with the systemd cgroup driver if the
init system itself is systemd.

https://kubernetes.io/docs/setup/production-environment/container-runtimes/#cgroup-drivers

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #587, opened 2020-12-17
Revert "Change cgroup driver to systemd (#521)"

abeer91:
This reverts commit f9d2c39535a8b53a951c1a896590aef5254dc896.

*Description of changes:*

EksCtl uses a nodebootsrap process where it configures the kubelet config, but not the docker config

Before we revert this commit, we need to finalize on how we can safely update eksctl to also set these values.

https://github.com/weaveworks/eksctl/blob/1555c35b1b315ff31c062a789c5f756a6122d057/docs/current-001-node_bootstrap.md 

### DIFF
--- a/files/docker-daemon.json
+++ b/files/docker-daemon.json
@@ -1,8 +1,5 @@
 {
   "bridge": "none",
-  "exec-opts": [
-    "native.cgroupdriver=systemd"
-  ],
   "log-driver": "json-file",
   "log-opts": {
     "max-size": "10m",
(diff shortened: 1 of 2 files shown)


mmerkes (approved):

reegnz:
I don't necessarily understand why a revert. A tool downstream is breaking, why should that influence what happens in the upstream?

reegnz:
Also could you please link to the ticket in eksctl that tracks this issue, so we can re-apply the change once the downstream manages to fix their incompatibility?

Issue #490, opened 2020-06-09
Use "systemd" cgroup driver as default instead of dockers' "cgroupfs"


jjjms:
**What would you like to be added**:
EKS AMI by default to use "systemd" cgroups driver for both kubelet and docker.

**Why is this needed**:
Since AL2 is using systemd and used systemd driver for cgroups managing, kubelet and docker using cgroupfs would result in systemd unaware of the resource allocation by cgroupfs and could result in system crash in certain cases.

https://kubernetes.io/docs/setup/production-environment/container-runtimes/#cgroup-drivers

I have tested this by performing the following change in config files and adding the node back to master. In my testing the node was marked as Ready and I was able to create pods in this node. 

```
### Cordoned a workernode:
k drain ip-192-168-0-171.us-west-2.compute.internal --ignore-daemonsets
node/ip-192-168-0-171.us-west-2.compute.internal already cordoned
WARNING: ignoring DaemonSet-managed Pods: kube-system/aws-node-jfnzw, kube-system/kube-proxy-sdx2q
node/ip-192-168-0-171.us-west-2.compute.internal drained


### Remove node entry from EKS so that node will be joined as a new entity altogether
k delete no ip-192-168-0-171.us-west-2.compute.internal


### Stopped kubelet:
[ec2-user@ip-192-168-0-171 ~]$ sudo systemctl stop kubelet docker

### Edited kubelet and docker config files to add systemd as cgroups manager:
[ec2-user@ip-192-168-0-171 ~]$ cat /etc/docker/daemon.json
{
  "bridge": "none",
  "exec-opts": ["native.cgroupdriver=systemd"],
  "log-driver": "json-file",
  "log-opts": {
    "max-size": "10m",
    "max-file": "10"
  },
  "live-restore": true,
  "max-concurrent-downloads": 10
}

### Since I'm using EKSCTl to create my cluster and nodegroup, I have modified the following file:
[root@ip-192-168-66-171 ec2-user]# cat /etc/eksctl/kubelet.yaml
address: 0.0.0.0
apiVersion: kubelet.config.k8s.io/v1beta1
authentication:
  anonymous:
    enabled: false
  webhook:
    cacheTTL: 2m0s
    enabled: true
  x509:
    clientCAFile: /etc/eksctl/ca.crt
authorization:
  mode: Webhook
  webhook:
    cacheAuthorizedTTL: 5m0s
    cacheUnauthorizedTTL: 30s
cgroupDriver: systemd
[...]

PR #593, opened 2021-01-06
Change cgroup driver to systemd

reegnz:
Kubernetes documentation indicates that for stability reasons
one should run kubernetes with the systemd cgroup driver if the
init system itself is systemd.

https://kubernetes.io/docs/setup/production-environment/container-runtimes/#cgroup-drivers

### DIFF
--- a/files/docker-daemon.json
+++ b/files/docker-daemon.json
@@ -1,5 +1,8 @@
 {
   "bridge": "none",
+  "exec-opts": [
+    "native.cgroupdriver=systemd"
+  ],
   "log-driver": "json-file",
   "log-opts": {
     "max-size": "10m",
--- a/files/kubelet-config.json
+++ b/files/kubelet-config.json
@@ -24,7 +24,7 @@
   "clusterDomain": "cluster.local",
   "hairpinMode": "hairpin-veth",
   "readOnlyPort": 0,
-  "cgroupDriver": "cgroupfs",
+  "cgroupDriver": "systemd",
   "cgroupRoot": "/",
   "featureGates": {
     "RotateKubeletServerCertificate": true


reegnz:
https://github.com/awslabs/amazon-eks-ami/pull/587 reverted this change earlier, now that eksctl was made compatible with this change in https://github.com/weaveworks/eksctl/pull/2962 it should be OK to merge again.

reegnz:
This is another run at https://github.com/awslabs/amazon-eks-ami/issues/490

abeer91:
Hi, now that eksctl has released a change, we're looking into how to make this release possible. There's still a condition where if we release an AMI with this change and a user is using an older version of eksctl, this could still lead to failures. We will be update this issue with more details, once we have a plan to pick this change.

reegnz:
@abeer91 I do understand the concern of breaking downstream tooling, so I might look into how that could be handled, although I can't promise anything (thinking about maybe using a [path unit](https://www.freedesktop.org/software/systemd/man/systemd.path.html) to keep the kubelet config in sync with the `daemon.json` but that is very hacky).

How long do you think we should wait for the newer eksctl to be adopted?
Does it warrant a changelog note that this might break some tooling downstream (that assumes only kubelet config needs to be changed)?

cyplenochek (approved):

jagadeeshi2i:
Hi, upon cluster creation with GPU instances i could not fetch nodes with `kubectl get nodes`. 
I see the below error on kubelet logs

`failed to run Kubelet: misconfiguration: kubelet cgroup driver: "systemd" is different from docker cgroup driver: "cgroupfs"
`
eksctl version : 0.51.0
eks version: 1.20
kubectl version: v1.21.0
AMI: ami-0b450512a684db582

Will the above fix solves this issue. @abeer91 when can we expect this fix.

stevehipwell:
@reegnz you might want to add [ systemd cgroup support to containerd](https://kubernetes.io/docs/setup/production-environment/container-runtimes/#containerd-systemd) in this PR now it's been released.

@abeer91 if this is still being blocked due to concerns about `eksctl` how about a containerd only PR (see https://github.com/aws/containers-roadmap/issues/1210#issuecomment-888176554)?

josephprem:
note:  [ systemd cgroup support to containerd](https://kubernetes.io/docs/setup/production-environment/container-runtimes/#containerd-systemd) is not correct. We are in runc v2 not in v1

> @reegnz you might want to add [ systemd cgroup support to containerd](https://kubernetes.io/docs/setup/production-environment/container-runtimes/#containerd-systemd) in this PR now it's been released.
> 
> @abeer91 if this is still being blocked due to concerns about `eksctl` how about a containerd only PR (see [aws/containers-roadmap#1210 (comment)](https://github.com/aws/containers-roadmap/issues/1210#issuecomment-888176554))?

stevehipwell:
@josephprem I can see that runc cgroup v2 is supported but is it actually enabled in the AMI? Also I can't find any official documentation with the best being reading https://github.com/containerd/containerd/issues/4203.

josephprem:
> @josephprem ~I can see that runc cgroup v2 is supported but is it actually enabled in the AMI?~
> 
> I can't find any official documentation with the best being reading [containerd/containerd#4203](https://github.com/containerd/containerd/issues/4203) (specifically [containerd/containerd#4203 (comment)](https://github.com/containerd/containerd/issues/4203#issuecomment-651532765))

yes it is 
```bash
cat /etc/containerd/config.toml
version = 2
root = "/var/lib/containerd"
state = "/run/containerd"

[grpc]
address = "/run/dockershim.sock"

default_runtime_name = "runc"

bin_dir = "/opt/cni/bin"
conf_dir = "/etc/cni/net.d"
```

stevehipwell:
@josephprem I re-read the config and that's why I struck the comment through about v2 being enabled. I'm still interested if you have any better documentation than the comment I found and linked above.

josephprem:
> @josephprem I re-read the config and that's why I struck the comment through about v2 being enabled. I'm still interested if you have any better documentation than the comment I found and linked above.

https://github.com/containerd/cri/blob/master/docs/config.md

josephprem:
I have added this in my user-data
```bash
conf=/etc/eks/containerd/containerd-config.toml 
# EKS bootstrap script moves $conf to /etc/eks/containerd/config.toml 
section="\[plugins.\"io.containerd.grpc.v1.cri\".containerd.runtimes.runc.options\]"
key="SystemdCgroup"
grep -q $section $conf || sed -i "/^runtime_type.*/a $section" $conf
grep -q $key $conf || sed -i "/$section.*/a $key = true" $conf
```

stevehipwell:
@josephprem I assume that you could modify _/etc/kubernetes/kubelet/kubelet-config.json_ like the changes in this PR instead of setting `--cgroup-driver=systemd`?

josephprem:
> /etc/kubernetes/kubelet/kubelet-config.json

I guess so , but my preference for pushing variables is through systemd Drop-Ins

Callisto13:
Hey folks 👋  `eksctl` here. A couple of things to note on our side:

- We have changed the way that `eksctl` bootstraps: we no longer roll our own, and simply call the `/etc/eks/bootstrap.sh` to delegate to whatever is going on in the AMIs. Therefore `eksctl` will not need to do anything with cgroup stuff. (See details of this decision here https://github.com/weaveworks/eksctl/pull/3564)
- HOWEVER: The above switch comes with a breaking change for those using **Unmanaged Nodegroups** with **Custom AMIs**, therefore the legacy codepath is still in place for a little while longer. (See notice here https://github.com/weaveworks/eksctl/issues/3563.) On the legacy path we have **removed previously added changes which set the cgroup driver**. This was because we kept experiencing breakages every time the AMI folks moved something around (this pain was actually one of the reasons why we removed our own bootstrapper). This could mean that if the custom AMI is based on one which sets the _docker_ driver to be `systemd`, the kubelet will fail because `eksctl`'s legacy bootstrapper will override the _kubelet_ config, setting the driver to `cgroupfs`. 
  - BUT, users can get by this by using the `overrideBootstrapCommand` to set their own `userdata`.

Please note that we are no longer making changes to the legacy path.

To address this point:
> There's still a condition where if we release an AMI with this change and a user is using an older version of eksctl, this could still lead to failures.

We emphatically **do not support** older versions of `eksctl`. If users are running any version of `eksctl` which is not latest, they are doing it wrong 🙃 .

reegnz:
> We emphatically do not support older versions of eksctl. If users are running any version of eksctl which is not latest, they are doing it wrong 🙃 .

@Callisto13 thanks for emphasizing that!

@abeer91 so now that being said, **what is still blocking the merging of this PR?**

stevehipwell:
@reegnz I think what @Callisto13 was saying is that the latest version of `eksctl` still supports a legacy flow for compatibility and that is what's currently blocking this? This is just an observation; personally I don't think a higher level project should dictate the direction of a dependency, this is why strong APIs/contracts are essential.

reegnz:
@stevehipwell Legacy flow seems to be supported for Custom AMIs, not the official lEKS AMI, there they use the /etc/eks/bootstrap.sh, no legacy code-path there.

stevehipwell:
@reegnz I think the concern was that someone using legacy flow with a custom AMI built from this AMI would be broken after this PR was merged?

mmerkes:
I want to verify that there are no outstanding concerns/issues, and I'll try to get this merged this week.

cartermckinnon:
I'm going to close this; we've made `systemd` the cgroup driver for `containerd`, and we don't plan to touch `docker` until we remove it in 1.25.

PR #595, opened 2021-01-12
Updating limits for memlock ulimit

abeer91:
*Description of changes:*

Updating the ulimit for memlock.

This should be backward compatible and only affects those customers that use memlock. 

More about K8s memory management - https://github.com/kubernetes/kubernetes/issues/53533#issuecomment-427233471

Tested by building a nodegroup with this AMI and confirming that it joins with the cluster.


anguslees (approved):
lgtm, fwiw.

PR #596, opened 2021-01-12
fix for bootstrap.sh 401 issue: https://github.com/awslabs/amazon-eks…

westdart:
…-ami/issues/594

*Issue #, if available:* 594

*Description of changes:*
Add retries to getting an aws metadata service token and for obtaining the metadata.

### DIFF
--- a/files/bootstrap.sh
+++ b/files/bootstrap.sh
@@ -134,6 +134,72 @@ function get_pause_container_account_for_region () {
     esac
 }
 
+function _get_token() {
+  local token_result=
+  local http_result=
+
+  token_result=$(curl -s -w "\n%{http_code}" -X PUT -H "X-aws-ec2-metadata-token-ttl-seconds: 600" "http://169.254.169.254/latest/api/token")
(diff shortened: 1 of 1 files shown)


heybronson:
Thanks for the PR! What tests did you run to confirm your changes?

westdart:
> Thanks for the PR! What tests did you run to confirm your changes?

I built EKS clusters using the modified version by passing the modified version as part of the userdata, placing it in /etc/eks/bootstrap-v2.sh and then had my main userdata script invoke that instead of the original /etc/eks/bootstrap.sh.

The only conditional part in the script that was modified is around whether 'DNS_CLUSTER_IP' is supplied or not, and this was also tested.

Also the lines:
```
      local lines=$(echo "$metadata_result" | wc -l)
      echo "$metadata_result" | head -n $(( lines - 1 ))
```
could actually be changed to 
```
      echo "$metadata_result" | head -n -1    
```
but I opted for the former as head may not support negative indexes in all the main distros and I am not sure where this script ends up.

heybronson (approved):

PR #601, opened 2021-01-20
build ARM AMIs with m6g.large instance type

heybronson:
*Description of changes:*
`m6g.large` instance types are available in all AWS partitions while `a1.large` instance type is not. Use `m6g.large` instance type as the default when building ARM AMIs.


abeer91 (approved):

mmerkes:
LGTM

Issue #606, opened 2021-02-03
Patching nodes failing for non GPU instance


AshokMishra:
**What happened**: Using the latest aws eks ami cause the node patching to fail with below error

`Created symlink from /etc/systemd/system/multi-user.target.wants/kubelet.service to /etc/systemd/system/kubelet.service.
nvidia-smi found
NVIDIA-SMI has failed because it couldn't communicate with the NVIDIA driver. Make sure that the latest NVIDIA driver is installed and running.
Exited with error on line 354
Feb 02 21:37:14 cloud-init[5087]: util.py[WARNING]: Failed running /var/lib/cloud/instance/scripts/part-001 [9]
Feb 02 21:37:14 cloud-init[5087]: cc_scripts_user.py[WARNING]: Failed to run module scripts-user (scripts in /var/lib/cloud/instance/scripts)
Feb 02 21:37:14 cloud-init[5087]: util.py[WARNING]: Running module scripts-user (<module 'cloudinit.config.cc_scripts_user' from '/usr/lib/python2.7/site-packages/cloudinit/config/cc_scripts_user.pyc'>) failed
Feb 02 21:37:14 cloud-init[5087]: util.py[WARNING]: Running module ssh-authkey-fingerprints (<module 'cloudinit.config.cc_ssh_authkey_fingerprints' from '/usr/lib/python2.7/site-packages/cloudinit/config/cc_ssh_authkey_fingerprints.pyc'>) failed
Cloud-init v. 19.3-5.amzn2 finished at Tue, 02 Feb 2021 21:37:14 +0000. Datasource DataSourceEc2.  Up 71.89 seconds`

**What you expected to happen**: Expected output is mentioned below
`Created symlink from /etc/systemd/system/multi-user.target.wants/kubelet.service to /etc/systemd/system/kubelet.service.
+ /opt/aws/bin/cfn-signal -e 0 --stack INFRA-***-EKS-****-NODEGROUP --resource NodeGroup --region <REGION>
Feb 02 21:37:14 cloud-init[5113]: util.py[WARNING]: Running module ssh-authkey-fingerprints (<module 'cloudinit.config.cc_ssh_authkey_fingerprints' from '/usr/lib/python2.7/site-packages/cloudinit/config/cc_ssh_authkey_fingerprints.pyc'>) failed
[...]

PR #613, opened 2021-02-16
update gpu clocks on gpu instance types only

heybronson:
*Fixes issue* https://github.com/awslabs/amazon-eks-ami/issues/606
*Description of changes:*
- Updated GPU boost clock logic to check that `nvidia-smi -q` returns a valid response (confirming that nvidia drivers are correctly configured on the AMI/instance type) before attempting to change clock speeds.

*Tests*
Created a GPU AMI that included these changes. Created the nodegroups via customAMI managed nodes with the following instance types:
- non-GPU instance type
- GPU instance type

Confirmed that non-GPU instance type nodes ran successfully. Confirmed that the GPU instance type had the correct GPU clock configuration by running `nvidia-smi -q -d CLOCK` and comparing to the values we set in the script.

### DIFF
--- a/files/bootstrap.sh
+++ b/files/bootstrap.sh
@@ -417,25 +417,30 @@ systemctl start kubelet
 if  command -v nvidia-smi &>/dev/null ; then
    echo "nvidia-smi found"
 
-   sudo nvidia-smi -pm 1 # set persistence mode
-   sudo nvidia-smi --auto-boost-default=0
+   nvidia-smi -q > /tmp/nvidia-smi-check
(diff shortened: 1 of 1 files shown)


sehayoun (approved):

mmerkes:
Looks like you did some code linting as part of this change. Maybe it would be better to move that into a separate PR and then make the actual change you want to make it easier to track and review?

abeer91 (approved):

mmerkes:
LGTM

Issue #548, opened 2020-10-10
AWS image builder


yanivpaz:
**What would you like to be added**:
Support for https://aws.amazon.com/image-builder/

**Why is this needed**:
- Simplify the build process of AMI
- Have better integration to AWS services 
- Avoid maintaining separate EC2 with packer

yanivpaz:
closed by mistake

yanivpaz:
@micahhausler can we leverage AWS AMI  builder and the new [refresh ASG](https://aws.amazon.com/blogs/compute/introducing-instance-refresh-for-ec2-auto-scaling/) feature to simplify the upgrade of ec2 worker nodes?

grgizem:
@yanivpaz something that we have been working on, EKS AMI prep using Image builder service: https://github.com/sufleio/amazon-eks-ami based on this repository. hope this helps

yanivpaz:
Thanks @grgizem  !

cartermckinnon:
We don't have any current plans to rewrite this AMI template using Image Builder. We'll reopen this issue if that changes.

PR #617, opened 2021-02-18
CloudFormation template to deploy CodeBuild project

jplock:
*Issue #, if available:* https://github.com/awslabs/amazon-eks-ami/issues/548

*Description of changes:* This PR includes a CloudFormation template that can be used to deploy a CodeBuild project that will execute packer and build the AMI (and then copy the AMI to the region the project is running in if not us-west-2).

### DIFF
--- a/amazon-eks-codebuild.yaml
+++ b/amazon-eks-codebuild.yaml
@@ -132,8 +132,6 @@ Resources:
         EnvironmentVariables:
           - Name: AWS_DEFAULT_REGION
             Value: "us-west-2" # amazon-eks S3 bucket is in us-west-2
-          - Name: PACKER_VERSION
-            Value: "1.7.0"
           - Name: PACKER_LOG
             Value: !If
               - IsPackerLog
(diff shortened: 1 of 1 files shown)


amazon-eks-codebuild.yaml
```diff
@@ -0,0 +1,189 @@
...
+      Name: amazon-eks-ami
+      Source:
+        BuildSpec: |-
+          version: 0.2
+
+          phases:
+            install:
+              commands:
+                - echo "Installing Packer..."
+                - curl -qL -o packer.zip https://releases.hashicorp.com/packer/${PACKER_VERSION}/packer_${PACKER_VERSION}_linux_amd64.zip && unzip packer.zip
```
awsvikrant, line 162: Can we replace this line with following and we don't need to pass packer version as environment variable. 
      - curl -fsSL https://apt.releases.hashicorp.com/gpg | apt-key add -
      - apt-add-repository "deb [arch=amd64] https://apt.releases.hashicorp.com $(lsb_release -cs) main"
      - apt-get update && apt-get install packer build-essential

jplock: @awsvikrant I updated the PR, thanks for reviewing

cartermckinnon:
This is neat; though some of it is usurped by Packer (like `ami_regions`). Not sure it will get merged, but I think providing some scaffolding for build and distributing the AMI is a good idea.

cartermckinnon:
Until and unless we use something like this for our official builds, I don't think we should add this to our repository. It's still a handy resource, and I think you should commit it to your own repository. Thank you for contributing!

PR #620, opened 2021-02-22
Fix 'Unknown options:' when multiple /var/log/eks_i* files exists

ajoux-accor:
*Description of changes:*

The command s3 fails if _'/var/log/eksi*'_ results in multiple files.

Fix : 
- Upload only the last file
- If upload succeed : remove old files

### DIFF
--- a/log-collector-script/linux/eks-ssm-content.json
+++ b/log-collector-script/linux/eks-ssm-content.json
@@ -20,6 +20,8 @@
          "inputs": {
             "runCommand": [
                "curl -O https://raw.githubusercontent.com/awslabs/amazon-eks-ami/master/log-collector-script/linux/eks-log-collector.sh",
+               "echo \"Cleaning old eks-log-collector files\"",
+               "rm /var/log/eks_i*",
                "bash ./eks-log-collector.sh  >/dev/null 2>&1",
                "echo \"EKS logs collected\"",
                "if [ -f /usr/local/bin/aws ]; then",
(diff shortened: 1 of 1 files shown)


log-collector-script/linux/eks-ssm-content.json
```diff
@@ -34,7 +34,7 @@
                "/usr/local/bin/aws --version",
                "fi",
                "echo \"Pushing to S3\"",
-               "/usr/local/bin/aws s3 cp /var/log/eks_i* s3://{{bucketName}}",
+               "/usr/local/bin/aws s3 cp $(ls -1 /var/log/eks_i* | tail -1) s3://{{bucketName}} && rm /var/log/eks_i*",
```
mmerkes, line 37: Thanks for contributing. Maybe we should separate these statements? More specifically, I think we should move `rm /var/log/eks_i*` to run before we run the log collector so that we clean any previous runs first. This will remove the need to do any `ls` and `tail` massaging. I'm trying to think if there's a reason we'd want to keep the bundle on the instance after the run for troubleshooting or something, but probably not? If not, we could just run it before (for any missed ones) and after, if we want to reclaim the disk space.

ajoux-accor: As suggested, I placed the cleanup on top.

mmerkes:
Thanks for this change. It looks good. I just need to do some manual verification.

mmerkes (approved):
LGTM. I tested out the old and new SSM docs. Old one worked as described, and I was able to run multiple times with this change. Tx!

PR #622, opened 2021-02-22
remove hardcoded gid for docker

heybronson:
*Issue #, if available:*

*Description of changes:*
Removes explicit gid for docker. This commit fixes builds for GPU AMIs which can have conflicting docker gids.

*Testing*
Built a GPU/nonGPU AMI. Both AMIs were able to join my EKS cluster.

### DIFF
--- a/scripts/install-worker.sh
+++ b/scripts/install-worker.sh
@@ -111,7 +111,8 @@ sudo yum install -y yum-utils device-mapper-persistent-data lvm2
 INSTALL_DOCKER="${INSTALL_DOCKER:-true}"
 if [[ "$INSTALL_DOCKER" == "true" ]]; then
     sudo amazon-linux-extras enable docker
-    sudo groupadd -f docker
+    sudo groupadd -fog 1950 docker
+    sudo useradd --gid $(getent group docker | cut -d: -f3) docker
     sudo yum install -y docker-${DOCKER_VERSION}*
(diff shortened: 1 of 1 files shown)


scripts/install-worker.sh
```diff
@@ -111,7 +111,7 @@ sudo yum install -y yum-utils device-mapper-persistent-data lvm2
 INSTALL_DOCKER="${INSTALL_DOCKER:-true}"
 if [[ "$INSTALL_DOCKER" == "true" ]]; then
     sudo amazon-linux-extras enable docker
-    sudo groupadd -fog 1950 docker && sudo useradd --gid 1950 docker
+    sudo groupadd -f docker
```
mmerkes, line 114: To avoid backwards incompatibility issues in the event that someone depends on this, could you just update it to something like this?

```
sudo groupadd -fog 1950 docker && sudo useradd --gid $(id -g docker) docker
```

mmerkes: Looks like `groupadd` is a no-op if the group already exists. Also, that `id` command work for me, you'll probably need to do something else.

```
$ sudo groupadd -fog 1999 docker
$ sudo groupadd -fog 1950 docker
$ cut -d: -f3 < <(getent group homey)
> 1999
```

mmerkes (approved):
LGTM. Tx for updating to be backwards compatible.

PR #640, opened 2021-03-24
Add retry in case 'aws eks wait cluster-active' returns error due to rate limiting

prateekgogia:
…rate limiting

*Issue #, if available:*

*Description of changes:*
Adds a check if the command 'aws eks wait cluster-active' fails it should retry. 


cartermckinnon:
Closing as stale. If retries need to be used for any of the AWS CLI calls, we should use the built-int mechanisms. https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-retries.html

PR #90, opened 2018-11-07
Modifying kubelet to use config files instead of kubelet flags which are about to deprecate.

nithu0115:
*Description of changes:*

Changing Kubelet to use config files instead of using kubelet flags as they are deprecated. 

```
Oct 18 00:55:14 ip-172-31-25-244.test-sd1.stagingus-west-2.compute.internal systemd[1]: Starting Kubernetes Kubelet...
Oct 18 00:55:20 ip-172-31-25-244.test-sd1.stagingus-west-2.compute.internal kubelet[3718]: Flag --address has been deprecated, This parameter should be set via the config file specified by the Kubelet's --config flag. See https://kubernetes.i
Oct 18 00:55:20 ip-172-31-25-244.test-sd1.stagingus-west-2.compute.internal kubelet[3718]: Flag --authentication-token-webhook has been deprecated, This parameter should be set via the config file specified by the Kubelet's --config flag. See
Oct 18 00:55:20 ip-172-31-25-244.test-sd1.stagingus-west-2.compute.internal kubelet[3718]: Flag --authorization-mode has been deprecated, This parameter should be set via the config file specified by the Kubelet's --config flag. See https://k
Oct 18 00:55:20 ip-172-31-25-244.test-sd1.stagingus-west-2.compute.internal kubelet[3718]: Flag --allow-privileged has been deprecated, will be removed in a future version
Oct 18 00:55:20 ip-172-31-25-244.test-sd1.stagingus-west-2.compute.internal kubelet[3718]: Flag --cluster-domain has been deprecated, This parameter should be set via the config file specified by the Kubelet's --config flag. See https://kuber
Oct 18 00:55:20 ip-172-31-25-244.test-sd1.stagingus-west-2.compute.internal kubelet[3718]: Flag --cgroup-driver has been deprecated, This parameter should be set via the config file specified by the Kubelet's --config flag. See https://kubern
Oct 18 00:55:20 ip-172-31-25-244.test-sd1.stagingus-west-2.compute.internal kubelet[3718]: Flag --feature-gates has been deprecated, This parameter should be set via the config file specified by the Kubelet's --config flag. See https://kubern
Oct 18 00:55:20 ip-172-31-25-244.test-sd1.stagingus-west-2.compute.internal kubelet[3718]: Flag --anonymous-auth has been deprecated, This parameter should be set via the config file specified by the Kubelet's --config flag. See https://kuber
[...]

PR #642, opened 2021-03-25
Add --max-pods param for bootstrap script

alesavin:
Maximum number of pods per node specified at /etc/eks/eni-max-pods.txt
and it is the limit of network interfaces for the instance type. This
number form an amount of memory in Mi kubernetes reserved with formula
11 * $max_num_pods + 255.

Nodes could have a setup when maximum number of pods defined and
it is less than value at the file. Here I add an ability to use
command line value to cover such cases.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/files/bootstrap.sh
+++ b/files/bootstrap.sh
@@ -17,6 +17,7 @@ function print_help {
     echo ""
     echo "-h,--help print this help"
     echo "--use-max-pods Sets --max-pods for the kubelet when true. (default: true)"
+    echo "--max-pods Max number of pods for the node. (default: value per instance type from /etc/eks/eni-max-pods.txt)"
     echo "--b64-cluster-ca The base64 encoded cluster CA content. Only valid when used with --apiserver-endpoint. Bypasses calling \"aws eks describe-cluster\""
     echo "--apiserver-endpoint The EKS cluster API Server endpoint. Only valid when used with --b64-cluster-ca. Bypasses calling \"aws eks describe-cluster\""
     echo "--kubelet-extra-args Extra arguments to add to the kubelet. Useful for adding labels or taints."
@@ -39,6 +40,11 @@ while [[ $# -gt 0 ]]; do
             ;;
         --use-max-pods)
             USE_MAX_PODS="$2"
+            shift
+            shift
+            ;;
+        --max-pods)
+            MAX_PODS="$2"
             shift
             shift
             ;;
@@ -100,6 +106,7 @@ CLUSTER_NAME="$1"
 set -u
 
 USE_MAX_PODS="${USE_MAX_PODS:-true}"
+MAX_PODS="${MAX_PODS:-}"
 B64_CLUSTER_CA="${B64_CLUSTER_CA:-}"
 APISERVER_ENDPOINT="${APISERVER_ENDPOINT:-}"
 SERVICE_IPV4_CIDR="${SERVICE_IPV4_CIDR:-}"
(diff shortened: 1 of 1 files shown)


alesavin:
@mogren @natherz97 PTAL

mmerkes:
@alesavin Is there a reason you can't use the `--kubelet-extra-args` parameter to specify max pods? Looks like that should have the same effect. We try to avoid adding new params to the bootstrap script

alesavin:
@mmerkes there are reasons why I see extra args for kubelet not the best choice:

1. --max-pods, --kube-reserved are deprecated (... parameter should be set via the config file specified by the Kubelet's `--config` flag. at kubelet params info), config is generated by bootstrap script here
2. with --kube-reserved I have to specify not only memory but cpu, ephemeral-storage limits; but only reserved memory is the thing I want to touch

just to be clear, I want to change not only maxPods but kube reserved memory as well (as last one correlate to maxPods as 11 * maxPods + 255). I would be glad if you provide any other workaround to tweak these values, I probably missing some parts.

alesavin:
> @alesavin you probably want to limit the max-pods value passed as parameter to the value specified in the eni-max-pods.txt file. I believe the number contained there is the max number of network interfaces for the given instance type. You probably do not want to allow users asking for a max-pods value beyond that number

done, add upper boundary for cmd line param

mhulscher:
Any progress on this? I'd like to see this functionality added

alesavin:
@mmerkes any feedback for my PR vs "--kubelet-extra-args"?

mikesir87:
I, too, would like to see this flag added. The method to use `--kubelet-extra-args` should be avoided as the kubelet docs indicate the `max-pods` flag is [deprecated](https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/) (can't deep-link on that page). While it _could_ work, I feel it would be best not to encourage support through a deprecated option.

mikesir87:
But, I'd also like to see a value that can go above what the ENI-based max allows, as not everyone that uses the EKS-optimized AMI is using the VPC CNI (like myself).

yanilov:
Hi @alesavin 
We encountered (other) deprecated kubelet flags too.
Looking at the bootstrap script and PR #90, it appears that a flexible solution would be to transform kubelet-config.json **before** calling bootstrap.sh (via cloudinit or other means).

For example, a (terraform-templated) cloud-init.yaml file could look like:

```yaml
runcmd:
  - echo "$(jq "${kubelet_config_overrides_json} /etc/kubernetes/kubelet/kubelet-config.json)" > /etc/kubernetes/kubelet/kubelet-config.json
  - [
      /etc/eks/bootstrap.sh,
      --apiserver-endpoint,
      "${cluster_endpoint}",
      --b64-cluster-ca,
      "${cluster_ca}",
      "${cluster_id}",
    ]
```

A module which creates the EKS cluster will also export the node userdata, which could look like this:
```terraform
output "node_userdata" {
  value = templatefile("${path.module}/cloud-init.template.yaml", {
    cluster_endpoint = aws_eks_cluster.cluster.endpoint
    cluster_ca            = aws_eks_cluster.cluster.certificate_authority.0.data
    cluster_id             = aws_eks_cluster.cluster.id
    kubelet_config_overrides_json = {
      maxPods = 42
  })
}
```

alesavin:
@yanilov files/bootstrap.sh changes /etc/kubernetes/kubelet/kubelet-config.json directly like you mentioned, not passing kubelet args. If you transform it before the file data (like memory) will be over-written https://github.com/awslabs/amazon-eks-ami/blob/master/files/bootstrap.sh#L440. Am I missing something?

cartermckinnon:
We need to offer a way to specify kubelet config parameters, as we have with `--kubelet-extra-args`. Modifying the default kubelet config file before or after the bootstrap script executes is fragile, I wouldn't recommend that approach. I'm going to close this, we'll continue tracking in #873.

PR #652, opened 2021-04-16
Add PRIVATE_BIN_REPO support for pulling binaries

jnewblanc:
As a EKS ami builder, I want to be able to pull binaries from my own repo of cached artifacts so that I can ensure availability, reproducibility, and security.

In my particular case, the binaries are not currently available in s3://amazon-eks.s3.us-gov-west-1.amazonaws.com and egress rules are tightly controlled.

How to use:

1. User caches/stages the binaries in their own artifact repo.  At present, these are:
     - kubelet
     - aws-iam-authenticator
     - cni-plugins-linux

2. User adds the PRIVATE_BIN_REPO variable to their custom eks-worker config

Given that this is an edge uncommon use case, the PRIVATE_BIN_REPO user variable has not been added to the eks-worker config.  Users will continue to use the existing behavior by default.

### DIFF
--- a/scripts/install-worker.sh
+++ b/scripts/install-worker.sh
@@ -172,7 +172,11 @@ BINARIES=(
     aws-iam-authenticator
 )
 for binary in ${BINARIES[*]} ; do
-    if [[ -n "$AWS_ACCESS_KEY_ID" ]]; then
+    if [ "${PRIVATE_BIN_REPO}" != "" ]; then
+        echo "Using PRIVATE_BIN_REPO ${PRIVATE_BIN_REPO} as source for coping binaries."
+        sudo wget ${PRIVATE_BIN_REPO}/$binary
+        sudo wget ${PRIVATE_BIN_REPO}/$binary.sha256
+    elif [[ -n "$AWS_ACCESS_KEY_ID" ]]; then
         echo "AWS cli present - using it to copy binaries from s3."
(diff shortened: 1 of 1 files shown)


mmerkes:
You can already [configured a private S3 bucket](https://github.com/awslabs/amazon-eks-ami/blob/master/scripts/install-worker.sh#L24-L25). Is there a reason that's insufficient for your use case?

jnewblanc:
I'm trying to add additional flexibility.  I'm in a tightly controlled environment where all external artifacts must enter through a series of artifact repositories and promotion pipelines.  I've added this change to pull from the repository rather than me needing to have extra process/automation to manage redundant artifacts in a s3 bucket.

cartermckinnon:
> must enter through a series of artifact repositories

IIUC, `PRIVATE_BIN_REPO`/"artifact repositories" are just file servers?

> the binaries are not currently available in s3://amazon-eks.s3.us-gov-west-1.amazonaws.com 

This is something we're planning to address! Because the `aws-us-gov` regions aren't airgapped, we haven't replicated binaries into that partition in the past; but we've heard from several customers that can't do cross-partition access. If the binaries were available in `s3://amazon-eks` in this partition, would that solve your problem?

cartermckinnon:
We still intend to replicate these binaries into the `aws-us-gov` partition; but I don't think this PR makes sense in its current form.

PR #653, opened 2021-04-22
Revert "Revert "Removing dependency on Authenticator binary" (#446)"

toabctl:
This reintroduces the switch to use "aws eks get-token" instead of the
aws-iam-authenticator. The reason (see [0]) why that switch got
reverted was, that eksctl wasn't able to handle the situation where
aws-iam-authenticator was not there. But that changed (see [1] and
[2]) so switching to aws-cli for getting a token should be good now.

This reverts commit d6e021b87edb9861fd478dece0e09cfe4bead8c4.

[0] https://github.com/awslabs/amazon-eks-ami/pull/446
[1] https://github.com/weaveworks/eksctl/issues/788
[2] https://github.com/weaveworks/eksctl/pull/993

### DIFF
--- a/files/kubelet-kubeconfig
+++ b/files/kubelet-kubeconfig
@@ -16,10 +16,14 @@ users:
   user:
     exec:
       apiVersion: client.authentication.k8s.io/v1alpha1
-      command: /usr/bin/aws-iam-authenticator
+      command: aws
+      env:
+        - name: AWS_STS_REGIONAL_ENDPOINTS
+          value: regional
       args:
-        - "token"
-        - "-i"
-        - "CLUSTER_NAME"
+        - eks
         - --region
         - "AWS_REGION"
+        - get-token
+        - --cluster-name
+        - "CLUSTER_NAME"
--- a/scripts/install-worker.sh
+++ b/scripts/install-worker.sh
@@ -169,7 +169,6 @@ S3_PATH="s3://$BINARY_BUCKET_NAME/$KUBERNETES_VERSION/$KUBERNETES_BUILD_DATE/bin
 
 BINARIES=(
     kubelet
-    aws-iam-authenticator
 )
 for binary in ${BINARIES[*]} ; do
     if [[ -n "$AWS_ACCESS_KEY_ID" ]]; then


toabctl:
@abeer91 @M00nF1sh could you have a look please given that you did the commit and review for the revert?

toabctl:
Friendly ping, @abeer91 @M00nF1sh

toabctl:
anybody else who could have a look here?

toabctl:
friendly ping ... what's the problem with looking into this?

toabctl:
@M00nF1sh another friendly ping after some more month....

toabctl:
@M00nF1sh friendly ping. it's now half a year. can you please review this?

M00nF1sh:
@toabctl hmm, just pinged peers in nodes team to review this PR.
seems this is a breaking change on for certain eksctl versions, i'll let ppl from nodes teams to comment on this

mmerkes:
Sorry this PR has been languishing. Overall, I like the change. However, there is one blocker. For K8s version 1.24, `client.authentication.k8s.io/v1alpha1` is no longer supported, so we now have a dependency on [this AWS CLI change](https://github.com/aws/aws-cli/commit/99d0fa1c96f7e76a570fe671f099163ec859cde7). The AmazonLinux yum repos are using an old AWS CLI version, which doesn't include that change. They're working on upgrading the CLI soon, and when it's available, our AMIs will automatically pull it in and we can go forward with this change.

If you can update this PR with `client.authentication.k8s.io/v1beta1`, we'll revisit this as soon as the AWS CLI is updated and try to merge this.

@cartermckinnon We on the same page here?

mmerkes:
Actually, I'll update the API version as part of [this issue](https://github.com/awslabs/amazon-eks-ami/issues/1020), so you'll just need to rebase after that.

toabctl:
> Actually, I'll update the API version as part of [this issue](https://github.com/awslabs/amazon-eks-ami/issues/1020), so you'll just need to rebase after that.

ack

mmerkes:
We've been discussing this internally. The blocker has been that the `awscli` version in the `yum` repos is fairly out of date, and while we're working on installing the `awscli` via another method in [this PR](https://github.com/awslabs/amazon-eks-ami/pull/1066), it's not an option in all regions. We have limited control over the availability of the `awscli`, but we do have control over the `authenticator` availability as we publish those binaries. For that reason, I don't think we're going to go this direction. This issue could be resolved soon, but we'd be in the same boat if the API updated again. We like the simplification and reduced dependencies here, but we may still need to provide the authenticator here if other customers possibly depend on that.

I'm going to close this for now. If there's advantages for doing this other than the simplification, let me know and we can reconsider.

PR #655, opened 2021-04-27
Validate that SONOBUOY_E2E_REGISTRY is set.

gdamberg:
*Issue #, if available:*

*Description of changes:*

Add a test that SONOBUOY_E2E_REGISTRY variable is set. The script will fail later if is unset due to -o nounset.

### DIFF
--- a/scripts/install-worker.sh
+++ b/scripts/install-worker.sh
@@ -246,6 +246,7 @@ sudo mv $TEMPLATE_DIR/eni-max-pods.txt /etc/eks/eni-max-pods.txt
 sudo mv $TEMPLATE_DIR/bootstrap.sh /etc/eks/bootstrap.sh
 sudo chmod +x /etc/eks/bootstrap.sh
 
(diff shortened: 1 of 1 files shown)


scripts/install-worker.sh
```diff
@@ -30,6 +30,7 @@ validate_env_set CNI_PLUGIN_VERSION
 validate_env_set KUBERNETES_VERSION
 validate_env_set KUBERNETES_BUILD_DATE
 validate_env_set PULL_CNI_FROM_GITHUB
+validate_env_set SONOBUOY_E2E_REGISTRY
```
mmerkes, line 33: To keep the expected behavior (skip the sonobuoy block below if unset), we can initialize the variable rather than assert that it's set.

```
SONOBUOY_E2E_REGISTRY="${SONOBUOY_E2E_REGISTRY:-}"
```

IMO, that's probably how we should handle that. Good catch and thanks for the contribution!

mmerkes (approved):
LGTM

heybronson (approved):

PR #656, opened 2021-04-27
Container runtime bootstrap

rothgar:
*Issue #, if available:* aws/containers-roadmap#313 and weaveworks/eksctl#3572

*Description of changes:*

Add a `--container-runtime` flag to the bootstrap.sh script to help make container runtime switching consistent. Provides a default of `dockerd` which performs no actions with the current AL2 AMI defaults and eventually can be switched to default to containerd with an optional switch to dockerd.

### DIFF
--- a/files/bootstrap.sh
+++ b/files/bootstrap.sh
@@ -26,6 +26,7 @@ function print_help {
     echo "--dns-cluster-ip Overrides the IP address to use for DNS queries within the cluster. Defaults to 10.100.0.10 or 172.20.0.10 based on the IP address of the primary interface"
     echo "--pause-container-account The AWS account (number) to pull the pause container from"
     echo "--pause-container-version The tag of the pause container"
+    echo "--container-runtime Specify a container runtime (default: dockerd)"
 }
 
 POSITIONAL=()
@@ -87,6 +88,11 @@ while [[ $# -gt 0 ]]; do
             shift
             shift
             ;;
+        --container-runtime)
+            CONTAINER_RUNTIME=$2
+            shift
+            shift
+            ;;
         *)    # unknown option
             POSITIONAL+=("$1") # save it in an array for later
             shift # past argument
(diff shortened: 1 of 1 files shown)


pavan-tri:
With K8S 1.20 in EKS, this is a important fix which needs attention,  +1

heybronson (approved):

files/bootstrap.sh
```diff
@@ -392,6 +399,22 @@ Environment='KUBELET_EXTRA_ARGS=$KUBELET_EXTRA_ARGS'
 EOF
 fi
 
+# Replace dockerd with containerd
+if [[ "$CONTAINER_RUNTIME" -eq "containerd" ]]; then
```
shivamdixit, line 403: @rothgar I believe this is incorrect because `-eq` operator is used for numeric comparison, not string comparison. [1] This should be`=`.

[1] https://tldp.org/LDP/abs/html/comparison-ops.html#ICOMPARISON1

files/bootstrap.sh
```diff
@@ -392,6 +399,22 @@ Environment='KUBELET_EXTRA_ARGS=$KUBELET_EXTRA_ARGS'
...
+[Service]
+ExecStart=
+ExecStart=/usr/bin/kubelet --cloud-provider aws \
+    --config /etc/kubernetes/kubelet/kubelet-config.json \
+    --kubeconfig /var/lib/kubelet/kubeconfig \
+    --container-runtime remote \
+    --container-runtime-endpoint "unix:///run/containerd/containerd.sock" \
+    --network-plugin cni $KUBELET_ARGS $KUBELET_EXTRA_ARGS
+EOF
+elif [[ "$CONTAINER_RUNTIME" -ne "dockerd" ]]; then
```
shivamdixit, line 414: Same as before, we should be using `=` not `-ne` because this is for numeric comparison and will fail.

files/bootstrap.sh
```diff
@@ -392,6 +399,22 @@ Environment='KUBELET_EXTRA_ARGS=$KUBELET_EXTRA_ARGS'
 EOF
 fi
 
+# Replace dockerd with containerd
+if [[ "$CONTAINER_RUNTIME" -eq "containerd" ]]; then
+    cat <<EOF > /etc/systemd/system/kubelet.service.d/50-kubelet-container-runtime.conf
```
shivamdixit, line 404: We need to quote the `EOF` like `cat <<'EOF'` else it will try to expand the variable `KUBELET_ARGS` and fail because the variable is not defined. See man bash [1]:

> If any characters in word are quoted, the delimiter is the result of quote removal on word, and the lines in the here-document are not expanded. If word is unquoted, all lines of the here-document are subjected to parameter expansion, command substitution, and arithmetic expansion. 

[1] https://linux.die.net/man/1/bash

shivamdixit (changes requested):

8enmann:
What's the status here? Would love to get this in

heybronson:
https://github.com/awslabs/amazon-eks-ami/pull/698 has addressed changing container runtime on node bootstrap.

ravisinha0506:
Going to close this PR in favor of #698.

PR #677, opened 2021-06-14
preload docker images

maksim-paskal:
*Description of changes:*
To reduce image pull time in pod creation - it's will be great to have some preloaded docker images in AMI. For cluster that have spot instances there are many docker pull operation on kubernetes node creation (daemonsets). Preloaded docker images make cluster more reliable

Default volume size for `amazon-eks-ami` is 4Gb - it will be not enough for preloaded images - this change also add parameter `launch_block_device_mappings_volume_size` for custom user values

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 PACKER_BINARY ?= packer
-PACKER_VARIABLES := aws_region ami_name binary_bucket_name binary_bucket_region kubernetes_version kubernetes_build_date kernel_version docker_version containerd_version runc_version cni_plugin_version source_ami_id source_ami_owners source_ami_filter_name arch instance_type security_group_id additional_yum_repos pull_cni_from_github sonobuoy_e2e_registry
+PACKER_VARIABLES := aws_region ami_name binary_bucket_name binary_bucket_region kubernetes_version kubernetes_build_date kernel_version docker_version containerd_version runc_version cni_plugin_version source_ami_id source_ami_owners source_ami_filter_name arch instance_type security_group_id additional_yum_repos pull_cni_from_github sonobuoy_e2e_registry launch_block_device_mappings_volume_size
 
 K8S_VERSION_PARTS := $(subst ., ,$(kubernetes_version))
 K8S_VERSION_MINOR := $(word 1,${K8S_VERSION_PARTS}).$(word 2,${K8S_VERSION_PARTS})
--- a/scripts/install-worker.sh
+++ b/scripts/install-worker.sh
@@ -295,6 +295,27 @@ echo fs.inotify.max_user_watches=524288 | sudo tee -a /etc/sysctl.conf
 echo fs.inotify.max_user_instances=8192 | sudo tee -a /etc/sysctl.conf
 echo vm.max_map_count=524288 | sudo tee -a /etc/sysctl.conf
 
+################################################################################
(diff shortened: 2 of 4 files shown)


scripts/install-worker.sh
```diff
@@ -289,6 +289,27 @@ echo fs.inotify.max_user_watches=524288 | sudo tee -a /etc/sysctl.conf
...
 
+################################################################################
+### Preload docker images ######################################################
+################################################################################
+
+# start docker for image pulling
+sudo systemctl start docker
+
+# (optional) login to private docker
+cat $TEMPLATE_DIR/docker-preload-images-login | bash
```
sylr, line 300:
```suggestion
bash $TEMPLATE_DIR/docker-preload-images-login
```

Makefile
```diff
@@ -1,5 +1,5 @@
 PACKER_BINARY ?= packer
-PACKER_VARIABLES := aws_region ami_name binary_bucket_name binary_bucket_region kubernetes_version kubernetes_build_date kernel_version docker_version containerd_version runc_version cni_plugin_version source_ami_id source_ami_owners source_ami_filter_name arch instance_type security_group_id additional_yum_repos pull_cni_from_github sonobuoy_e2e_registry
+PACKER_VARIABLES := aws_region ami_name binary_bucket_name binary_bucket_region kubernetes_version kubernetes_build_date kernel_version docker_version containerd_version runc_version cni_plugin_version source_ami_id source_ami_owners source_ami_filter_name arch instance_type security_group_id additional_yum_repos pull_cni_from_github sonobuoy_e2e_registry launch_block_device_mappings_volume_size
```
sylr, line 2: Did you really mean to add this ?

scripts/install-worker.sh
```diff
@@ -289,6 +289,27 @@ echo fs.inotify.max_user_watches=524288 | sudo tee -a /etc/sysctl.conf
...
+
+# remove comments from docker images list
+sed -i '/^#/d' $TEMPLATE_DIR/docker-preload-images
+
+# preload docker images, skip empty lines
+for line in `sed '/^$/d' $TEMPLATE_DIR/docker-preload-images`; do
```
sylr, line 306:
```suggestion
# remove comments and empty lines from docker images list
sed -i -e '/^#/d' -e '/^$/d' $TEMPLATE_DIR/docker-preload-images

# preload docker images
for line in $(cat $TEMPLATE_DIR/docker-preload-images); do
```

maksim-paskal:
@sylr thank for your suggession, looks good, added to PR

cartermckinnon:
There's work being done [here](https://github.com/awslabs/amazon-eks-ami/pull/938) to cache images used by all EKS nodes (not user images). That PR has a much more obvious benefit for EKS users, IMO.

If a user bakes their workload images into their AMI, they'll need to rebuild their AMI every time their `DaemonSet`/`Deployment`/etc image changes to continue benefiting from the cache. While that tight coupling may be worth it for some use-cases, I'm not sure it makes sense to add support for this to the base AMI (because users will need to build a custom AMI anyway).

Additionally, we shouldn't add any new usage of the `docker` CLI, because it's slated for removal in upcoming EKS releases.

maksim-paskal:
Ok, thanks for feedback

PR #692, opened 2021-07-09
Add time to AMI name, making it unique

masrur:
*Issue #, if available:*

*Description of changes:*


cartermckinnon:
We use the `YYYYMMDD` format for our official AMI builds, that's the motivation for using it here. You can override the AMI name however you'd like by setting the `ami_name` argument, like:
```
make 1.22 ami_name=my-ami-name-1.22-$(date +%s)
```

PR #694, opened 2021-07-13
add packer variable files option

GwonsooLee:
*Issue #, if available:*

*Description of changes:*
Add packer variable file environment variables for users to specify the custom variable files for packer. Users now can only set  variables dynamically by setting `PACKER_VARIABLES`. If users want to modify few more variables which are not listed int `PACKER_VARIABLES`, then it would be helpful for them to define variables in json file and specify the file name to `PACKER_VAR_FILE` 

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,12 +33,12 @@ all: 1.15 1.16 1.17 1.18 1.19 1.20
 
 .PHONY: validate
 validate:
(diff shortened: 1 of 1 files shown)


cartermckinnon:
Looks fine to me; if you resolve the conflict and rename this `PACKER_VARIABLE_FILE`, I'm on board :+1:

cartermckinnon:
This will be resolved by #1079 .

PR #695, opened 2021-07-13
change default volume type from gp2 to gp3

GwonsooLee:
*Description of changes:*
This is really small PR. I think it would be better to use gp3 for default volume type which is cheaper than gp2.
I think changing default type of instance(which is m4.large) would also be helpful.

### DIFF
--- a/eks-worker-al2.json
+++ b/eks-worker-al2.json
@@ -61,7 +61,7 @@
       "launch_block_device_mappings": [
         {
           "device_name": "/dev/xvda",
-          "volume_type": "gp2",
+          "volume_type": "gp3",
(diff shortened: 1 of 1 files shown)


mmerkes:
Per #1045, we're making volume type configurable in make arguments, which will cause this PR to not work. If you want to update it based on that change (hopefully merged soon), we can use this PR for when we make `gp3` the default volume type sometime in the near future. Exact date TBD.

cartermckinnon:
I'm going to close this. We're aware of the feature request, but we have to make the change in a different way due to the lack of gp3 support in some environments (like local zones).

PR #698, opened 2021-07-13
Containerd runtime support

ravisinha0506:
*Issue #, if available:*

*Description of changes:*
changes done to add support for containerd runtime in eks worker ami. This is done by:
* Adding a new optional parameter to specify container-runtime in the bootstrap script.
* The default value of this parameter is dockerd to make it backward compatible to existing behavior.
* docker process wouldn't be running on the host when containerd runtime is chosen on the node.


heybronson (approved):

files/bootstrap.sh
```diff
@@ -392,24 +399,40 @@ Environment='KUBELET_EXTRA_ARGS=$KUBELET_EXTRA_ARGS'
...
+        echo "$DOCKER_CONFIG_JSON" > /etc/docker/daemon.json
+    fi
+    if [[ "$ENABLE_DOCKER_BRIDGE" = "true" ]]; then
+          # Enabling the docker bridge network. We have to disable live-restore as it
+          # prevents docker from recreating the default bridge network on restart
+          echo "$(jq '.bridge="docker0" | ."live-restore"=false' /etc/docker/daemon.json)" > /etc/docker/daemon.json
+    fi
+    systemctl daemon-reload
+    systemctl enable docker
+    systemctl start docker
```
take-five, line 429: Hi
In the previous version of this file the Docker daemon was always restarted after its config file was modified.  But after this commit nothing happens, the changes don't have an effect, probably because now the script runs `systemctl start docker` instead of `systemctl restart docker`

gclair: This change is impacting our automated testing too. Was wondering what the rationale between switching from `restart` to `start` was.

PR #706, opened 2021-07-20
Adding utility to calculate max-pods for kubelet

suket22:
*Issue #, if available:* N/A

*Description of changes:*
This adds a script that can be used to determine the Amazon EKS-recommended max-pods per node value, which depends upon several factors like instanceType, AWS VPC CNI version and the env variables set on the CNI DaemonSet like `ENABLE_PREFIX_DELEGATION`.

This script is written to be used in conjunction with the AWS VPC CNI plugin - https://github.com/aws/amazon-vpc-cni-k8s

### DIFF
--- a/files/max-pods-calculator.sh
+++ b/files/max-pods-calculator.sh
@@ -1,0 +1,151 @@
+#!/bin/bash
+
+set -o pipefail
+set -o nounset
+set -o errexit
+
+err_report() {
+    echo "Exited with error on line $1"
+}
+trap 'err_report $LINENO' ERR
+
+function print_help {
+    echo "usage: $0 <instance(s)> [options]"
+    echo "Calculates maxPods value to be used when starting up the kubelet."
+    echo "-h,--help print this help."
+    echo "--instance-type Specify the instance type to calculate max pods value."
+    echo "--instance-type-from-imds Use this flag if the instance type should be fetched from IMDS."
+    echo "--cni-version Specify the version of the CNI (example - 1.7.5)."
+    echo "--cni-custom-networking-enabled Use this flag to indicate if CNI custom networking mode has been enabled."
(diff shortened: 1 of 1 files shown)


MyannaHarris (approved):
lgtm

adammw:
@suket22 how is this script supposed to be called? will there be another PR to replace the --max-pods calculation in bootstrap.sh to use this logic?

adammw:
Also, would be handy if the `+2` is configurable for clusters where we have >2 hostNetwork daemonsets configured, and would help explain what the +2 represents (I saw this explained in elsewhere but not here).

snolan-amount:
This script sets a ceiling for max pods to 110 for "low CPU" instance types and 250 for "high CPU" instance types - what is the AWS reference for those numbers?

mikestef9:
This is not an exact science, but is based on our own internal testing. While the Kubernetes upstream project recommends no more than 110 pods per node, our testing has shown you can go higher on larger EC2 instance types. We’d recommend you run your own testing based on your applications, and you are free to override the kubelet settings and set a max pods value that meets your requirements. The important point is that with prefix assignment, IP addresses is no longer an artificial limitation to the number of pods you can run per node.

suket22:
@adammw 
We won't be updating the --max-pods calculation in bootstrap.sh. The bootstrap script is unable to determine which AWS CNI version is running on the cluster, and therefore cannot execute the max-pods-calculator script. 

The way to think about this is if you're using self-managed nodegroups, you can pre-compute what the EKS-recommended maximum value for max-pods should be based on the CNI version you're running, your need for custom networking and prefix delegation by executing this script.

You can then update the UserData of your instances (via your EC2 LaunchTemplate if you use one) and set the bootstrap script as - 
```
/etc/eks/bootstrap.sh <your-cluster> --use-max-pods false --kubelet-extra-args '--max-pods=<your_value>'
```

PR #711, opened 2021-07-23
Modify the linux SSM runCommand to install AWS CLI version 2

a2ush:
## Description of changes:

Change to install AWS CLI v2 so that it does not depend on other Python packages.

## Background:
I've confirmed that we cannot sometimes install AWS CLI v1 with the current sample JSON file because of the Python version[1][2].
For example, when we launch EC2 instance with EKS optimized AMI[3], we fail to do the sample runCommand because the default Python version is 2.7 in this instance.

So, I think it's better to install AWS CLI v2 which has no dependencies on other Python packages.

[1] https://docs.aws.amazon.com/cli/latest/userguide/welcome-versions.html
```
AWS CLI version	                 Supported Python version
Versions starting 7/15/2021	        Python 3.6+
```
[2] The result of each commands in this EC2 instance
```
# python -V
Python 2.7.18
# which python
/usr/bin/python
# ls -l /usr/bin/python
lrwxrwxrwx 1 root root 7 Jul  6 17:24 /usr/bin/python -> python2
# ./awscli-bundle/install -i /usr/local/aws -b /usr/local/bin/aws
Unsupported Python version detected: Python 2.7
To continue using this installer you must use Python 3.6 or later.
For more information see the following blog post: https://aws.amazon.com/blogs/developer/announcing-end-of-support-for-python-2-7-in-aws-sdk-for-python-and-aws-cli-v1/
```
[3] https://docs.aws.amazon.com/eks/latest/userguide/eks-optimized-ami.html

### Note
* When we do the command `unzip /usr/bin/awscliv2.zip`, we get the "aws" directory. Therefore, if the directory or file named as "aws" already exits before doing this command, it will be failed.
So, I prepared the "awscli-v2-install" directory to avoid this problem.
* I've confirmed that the updated JSON file(eks-ssm-content.json) works fine with EKS Optimized AMI.


log-collector-script/linux/eks-ssm-content.json
```diff
@@ -21,17 +21,17 @@
...
+               "rm -f /var/log/eks_i*",
                "bash ./eks-log-collector.sh  >/dev/null 2>&1",
                "echo \"EKS logs collected\"",
                "if [ -f /usr/local/bin/aws ]; then",
                "echo \"AWS_already_installed\"",
                "else",
-               "echo \"Installing AWSCLI\"",
-               "curl \"https://s3.amazonaws.com/aws-cli/awscli-bundle.zip\" -o \"awscli-bundle.zip\"  >/dev/null 2>&1",
+               "echo \"Installing AWSCLI v2\"",
+               "curl \"https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip\" -o \"awscliv2.zip\"  >/dev/null 2>&1",
```
cartermckinnon, line 31: The `ARCH` may not always be `x86_64`, can you add handling for that?

cartermckinnon:
We're installing v2 in the AMI itself (when possible), so I think it makes sense to do so here.

log-collector-script/linux/eks-ssm-content.json
a2ush: Modified

a2ush: @cartermckinnon 
Could you please re-check it?

cartermckinnon:
We made this change in #1135 👍

PR #721, opened 2021-08-04
Include Multus CNI logs in EKS Logs Collector script

Kausheel:
*Description of changes:*

As of today, the Multus CNI is supported by EKS. The EKS Logs Collector script does not currently include logs for this component. Since CNI troubleshooting is a key use case for this script, we should ensure Multus logs are included here.

I've added a new function to collect the Multus CNI Pod logs. Let me know if you have any feedback.


cgchinmay:
> Hi cgchinmay@
> 
> The test for the PR is as follows:
> 
> * Connect to a Worker Node via SSH.
> * Download the script via the normal steps:
> 
> ```
> curl -O https://raw.githubusercontent.com/awslabs/amazon-eks-ami/master/log-collector-script/linux/eks-log-collector.sh
> sudo bash eks-log-collector.sh
> ```
> 
> * Observe the output includes "Trying to collect Multus logs..."
> * Extract the log bundle, and verify the var_log dir contains the kube-system_kube-multus-ds log directory. This will contain the Multus logs (if you have Multus installed).

Thanks, Will validate it and then close the PR

Kausheel:
Hi @cgchinmay let me know if there's any other information you need from me to help out with this.

log-collector-script/linux/eks-log-collector.sh
```diff
@@ -446,6 +447,13 @@ get_ipamd_info() {
   ok
 }
 
+get_multus_info() {
+  try "collect Multus logs"
```
cgchinmay, line 451: I think we should try to collect Multus logs only if its installed, similar to what we do for SELinux , is it possible to add that check here ?

PR #723, opened 2021-08-05
containerd restart instead of start

cavdhut:
*Issue #, if available:*  

*Description of changes:*

If containerd service is already started it doesn't pickup the config changes.  Similar to what this script already does for docker, replacing `systemctl start containerd` with `systemctl restart containerd`.

### DIFF
--- a/files/bootstrap.sh
+++ b/files/bootstrap.sh
@@ -411,7 +411,7 @@ if [[ "$CONTAINER_RUNTIME" = "containerd" ]]; then
     sudo chown root:root /etc/systemd/system/kubelet.service
     systemctl daemon-reload
     systemctl enable containerd
-    systemctl start containerd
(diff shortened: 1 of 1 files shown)


files/bootstrap.sh
```diff
@@ -411,7 +411,7 @@ if [[ "$CONTAINER_RUNTIME" = "containerd" ]]; then
     sudo chown root:root /etc/systemd/system/kubelet.service
     systemctl daemon-reload
     systemctl enable containerd
-    systemctl start containerd
+    systemctl restart containerd
```
ravisinha0506, line 414: This is already merged as a part of [this commit](https://github.com/awslabs/amazon-eks-ami/commit/5d109a125ef3876209850d0d4308d1a77ddf601c)

Issue #733, opened 2021-08-16
Make Loglevel configurable in bootstrap.sh


steveizzle:
**What would you like to be added**:

The Verbosity of kubelet logs is hard coded in bootstrap.sh 
https://github.com/awslabs/amazon-eks-ami/blob/5d109a125ef3876209850d0d4308d1a77ddf601c/files/bootstrap.sh#L396

I would like to configure the verbosity of the logs (kubelet --v flag) dynamically without having to rewrite the 10-kubelet-args.conf file and reload the daemon.

Is there a reason why this is hard coded? Could it not stay in the default loglevel (dont know what that is..) and when needed change with kubelet-extra-args?

An easy solution would also be to simply add a new parameter with default=2. I would be glad to add this.

**Why is this needed**:

It seems like a parameter which should be configurable. Also i opened a support case and i was asked to set the loglevel more verbose, which seems a little complicated. I always try to change the least possible and stay in the best supported state.

steveizzle:
I added the loglevel  as extra-arg and now have two --v flags in the running kubelet process. It seems that the latter one is used, which means it currently works. But it seems a little hacky to have a process running with the same flag twice with two differnet values

dschunack:
Make absolutely sense for me to implement a parameter to configure it directly in the bootstrap and not as double parameter.

PR #748, opened 2021-08-31
Added Log level Option

steveizzle:
*Issue #, if available:*

#733 

*Description of changes:*

Added Option to bootstrap.sh for setting the loglevel of kubelets. Defaults to 2 which was hard encoded before.

### DIFF
--- a/files/bootstrap.sh
+++ b/files/bootstrap.sh
@@ -58,6 +58,11 @@ while [[ $# -gt 0 ]]; do
             shift
             shift
             ;;
+        --kubelet-log-level)
+            KUBELET_LOG_LEVEL=$2
+            shift
+            shift
+            ;;
         --enable-docker-bridge)
             ENABLE_DOCKER_BRIDGE=$2
             shift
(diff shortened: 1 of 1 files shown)


abeer91:
You can update log level currently by passing the log level as additional kubelet args, which is already supported in the bootstrap script.

Something like:
```
      #!/bin/bash
      /etc/eks/bootstrap.sh <cluster_name> --kubelet-extra-args '--v=8'
```

steveizzle:
@abeer91 i did that and it works, but i find it a little weird and hacky for such a professional product like the official eks workers to have a process running with two same flags. https://github.com/awslabs/amazon-eks-ami/issues/733#issuecomment-900119187

cartermckinnon:
I don't disagree that this looks a bit odd; but it's just the behavior of https://github.com/spf13/pflag. This applies for any flag on kubelet or any other k8s binary.

I don't think adding a new bootstrap flag for this specific kubelet flag makes sense, because we already have the general-purpose `--kubelet-extra-args`. As it stands, if the user defines a verbosity in `--kubelet-extra-args`, that's what takes effect.

PR #752, opened 2021-09-07
Support Packer AMI Builder's ami_region feature

suharshan:
*Issue #, if available:*

*Description of changes:*

Packer has an [ami_regions](https://www.packer.io/docs/builders/amazon/ebs#ami_regions) feature which copies the AMI across multiple regions, so that the user doesn't have to copy AMIs across regions manually.


cartermckinnon:
This seems handy; can you add `ami_regions` to `PACKER_VARIABLES` in the Makefile?

cartermckinnon:
In my tests, this doesn't work properly with the manifest post-processor. For example, with `ami_regions: us-west-2,us-east-1`, the resulting `manifest.json` only contains an entry for the `us-east-1` artifact.

suharshan:
> This seems handy; can you add `ami_regions` to `PACKER_VARIABLES` in the Makefile?

Apologies I missed your reply. Added to `PACKER_VARIABLES`.

cartermckinnon (approved):

PR #759, opened 2021-09-15
Expose flags for max-pods-calculator to use a configurable max pods ceiling

a7i:
*Issue #, if available:*

*Description of changes:*
Provide the ability to pass in the ceiling for max-pods instead of relying on the default values for low/high cpu count.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #766, opened 2021-09-29
Add show-max-allowed flag to max pods calculator script

overdrive3000:
*Description of changes:*

Add the `--show-max-allowed` flag to the `max-pods-calculator.sh` helper script in order to show the actual max number of IP addresses allowed in given EC2 instance type.

Useful when using the `--cni-prefix-delegation-enable` flag and the user needs to set a higher number than the value currently recommended  by the script.

### DIFF
--- a/files/max-pods-calculator.sh
+++ b/files/max-pods-calculator.sh
@@ -19,6 +19,7 @@ function print_help {
     echo "--cni-custom-networking-enabled Use this flag to indicate if CNI custom networking mode has been enabled."
     echo "--cni-prefix-delegation-enabled Use this flag to indicate if CNI prefix delegation has been enabled."
     echo "--cni-max-eni specify how many ENIs should be used for prefix delegation. Defaults to using all ENIs per instance."
+    echo "--show-max-allowed Use this flag to show max number of Pods allowed to run in Worker Node. Otherwise the script will show the recommended value"
 }
 
 POSITIONAL=()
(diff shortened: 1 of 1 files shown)


files/max-pods-calculator.sh
```diff
@@ -144,8 +150,15 @@ fi
 MAX_POD_CEILING_FOR_LOW_CPU=110
 MAX_POD_CEILING_FOR_HIGH_CPU=250
 CPU_COUNT=$(echo $DESCRIBE_INSTANCES_RESULT | jq -r '.CpuCount' )
+
+if [ "$SHOW_MAX_ALLOWED" = true ] ; then
+  echo $max_pods
```
suket22, line 155: Some kind of ceiling is better than nothing at all. With IP limitations out of the way, the theoretical number (which is what `$max_pods` is here) could be really high. A c5.xlarge could give you > 500 pods so I'm not sure how useful this might be.

I think https://github.com/awslabs/amazon-eks-ami/pull/759/files is probably a better approach where you can override the ceiling value to be used in case you want to go higher *or* lower than the 110/250 recommendation this script currently gives out. That way the caller of the script can explicitly guide what number they're comfortable with.

overdrive3000: Hi @suket22,

I agree with you on setting some ceiling. But sometimes I only need to calculate the exact number of pods that I could run in an specific EC2 instance size, how to set the right ceiling depends on multiple factors like the kind of workload, max number of pods allowed, etc.

In my opinion both #759 and this feature are useful.

suket22: I'm revisiting some of these changes on the EKS-optimized AMI and I think this is valuable. I also realized that even outside of CNI Prefix Delegation, this script is currently spitting out 250 if you're using saying an m5.16xlarge.

suket22 (approved):

files/max-pods-calculator.sh
o6uoq: Any update?

PR #790, opened 2021-10-20
Implement --containerd-extra-config-toml in bootstrap.sh

codablock:
This flag allows to pass extra configuration for containerd, which is
appended to /etc/containerd/config.toml. This allows to specify things
like mirros/pull-through-caches without interfering with the EKS provided
config.

### DIFF
--- a/files/bootstrap.sh
+++ b/files/bootstrap.sh
@@ -23,6 +23,7 @@ function print_help {
     echo "--enable-docker-bridge Restores the docker default bridge network. (default: false)"
     echo "--aws-api-retry-attempts Number of retry attempts for AWS API call (DescribeCluster) (default: 3)"
     echo "--docker-config-json The contents of the /etc/docker/daemon.json file. Useful if you want a custom config differing from the default one in the AMI"
+    echo "--containerd-extra-config-toml Extra configuration to append to the containerd configuration. Useful if you want a custom config differing from the default one in the AMI"
     echo "--dns-cluster-ip Overrides the IP address to use for DNS queries within the cluster. Defaults to 10.100.0.10 or 172.20.0.10 based on the IP address of the primary interface"
     echo "--pause-container-account The AWS account (number) to pull the pause container from"
     echo "--pause-container-version The tag of the pause container"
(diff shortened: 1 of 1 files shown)


vavdoshka:
yes please I need it as well

anusha-ragunathan:
LGTM!

Thanks for taking care of this.

vavdoshka (approved):

files/bootstrap.sh
```diff
@@ -461,6 +468,9 @@ if [[ "$CONTAINER_RUNTIME" = "containerd" ]]; then
     sudo mkdir -p /etc/containerd
     sudo mkdir -p /etc/cni/net.d
     sudo sed -i s,SANDBOX_IMAGE,$PAUSE_CONTAINER,g /etc/eks/containerd/containerd-config.toml  
+    if [[ -n "$CONTAINERD_EXTRA_CONFIG_TOML" ]]; then
+        echo "$CONTAINERD_EXTRA_CONFIG_TOML" >> /etc/eks/containerd/containerd-config.toml
+    fi
```
anusha-ragunathan, line 473: This block should be moved above L470, so that we have the ability to specify SANDBOX image in custom containerd config and get is replaced by the right pause container in the bootstrap.sh script.

codablock: Makes sense, I'm going to change this and force-push.

anusha-ragunathan (approved):

files/bootstrap.sh
```diff
@@ -75,6 +76,11 @@ while [[ $# -gt 0 ]]; do
             shift
             shift
             ;;
+        --containerd-extra-config-toml)
```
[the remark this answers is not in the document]
codablock: I would prefer it as you described it, but decided to stick with the way --docker-config-json is implemented to avoid unexpected surprises. If enough EKS folks tell me that the file-pointer way is the preferred one, I'll go with it.

suket22:
Sounds good, thanks @codablock. I'll get on it and should have a PR soon.

bryantbiggs:
@suket22 / @codablock - is it possible to make whatever solution comes out of this PR compatible with managed node groups? Right now, with the way managed node groups inject the bootstrap script at the base of the user data file, users do not have access to all of the bootstrap script flags which leads to [hacky workarounds](https://github.com/awslabs/amazon-eks-ami/issues/844). I know it might be beyond the scope of what you were envisioning, but I think if there is a path to cater to both self-managed and EKS managed node groups, that would be amazing. Something to consider, thank you!

suket22:
Alright we've got this merged as part of #929 and this should be make it as part of our next AMI release.

PR #792, opened 2021-10-27
Include max-pods-calculator.sh in image

### LINKED ISSUE #750: include max-pods-calculator.sh into AMI

Please include the script https://github.com/awslabs/amazon-eks-ami/blob/master/files/max-pods-calculator.sh into the AMI image.

**Why is this needed**:

Downloading this script from github during bootstrapping is error prone and will likely cause a situation of a dependency unavailability due to rate limiting or something else, which will cause a downtime for EKS nodes.

andyspiers:
resolves #750

**Include max-pods-calculator.sh**

Currently we have a hard-coded map between instance-type and max pods. Having this very useful script present in the AMI would enable us to automate this calculation in user-data.

This PR simply makes the script _available_ for use in custom user-data scripts. It's not an attempt to automate usage of it, though that may be an interesting idea later.

As noted in #750, it's not really acceptable for us to download it from Github using curl at provision time when it could simply be baked into the AMI.

### DIFF
--- a/scripts/install-worker.sh
+++ b/scripts/install-worker.sh
@@ -272,6 +272,8 @@ sudo mkdir -p /etc/eks
 sudo mv $TEMPLATE_DIR/eni-max-pods.txt /etc/eks/eni-max-pods.txt
 sudo mv $TEMPLATE_DIR/bootstrap.sh /etc/eks/bootstrap.sh
 sudo chmod +x /etc/eks/bootstrap.sh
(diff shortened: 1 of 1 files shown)


suket22 (approved):

mmerkes (approved):

andyspiers:
Thanks for reviewing @mmerkes & @suket22 🙂

Issue #529, opened 2020-08-27
Improve README to help users create custom AMIs


mmerkes:
**What would you like to be added**:

The README should include more complete instructions on creating custom AMIs. [This article](https://github.com/awslabs/amazon-eks-ami/issues/524) provides a decent overview of some options, but some of the instructions are incomplete or not correct. The README for this repo should include up-to-date instructions for using this tool, including covering most of the scenarios covered in that article.

**Why is this needed**:

This tool should own its own documentation. Users are relying on [other articles](https://github.com/awslabs/amazon-eks-ami/issues/524), which don't work as is, and it would be best to have a more authoritative source.

PR #794, opened 2021-10-29
Update Readme file to have clear instructions about how to build usin…

icarthick:
…g different versions of Kubernetes binaries

*Issue #, if available:*
https://github.com/awslabs/amazon-eks-ami/issues/529

*Description of changes:*
Updated Readme file to include instructions about how to build AMI using different versions of Kubernetes including custom binaries


ravisinha0506 (approved):

README.md
```diff
@@ -21,13 +21,72 @@ when building this AMI.
...
-A Makefile is provided to build the AMI, but it is just a small wrapper around
+A Makefile is provided to build the Amazon EKS Worker AMI, but it is just a small wrapper around
 invoking Packer directly. You can initiate the build process by running the
 following command in the root of this repository:
 
 ```bash
 make
 ```
+By default, Makefile will build Amazon EKS Worker AMI with Kubernetes binaries from these K8s versions
+1.16.15, 1.17.17, 1.18.20, 1.19.14, 1.20.10, 1.21.4
```
suket22, line 32: This is no longer accurate since we're not generating 1.16 or 1.17 AMIs and are using diffferent kubelet versions for the others. 

Since this will keep happening in the future as we bring in new kubelet binaries, I'd recommend we change this line to something like "The Makefile chooses a particular kubelet binary to use per kubernetes version which you can [view here](https://github.com/awslabs/amazon-eks-ami/blob/master/Makefile)."

README.md
```diff
@@ -21,13 +21,72 @@ when building this AMI.
...
+To build an Amazon EKS Worker AMI for a particular Kubernetes version run the following command
+```bash
+make 1.21.4 ## Build a Amazon EKS Worker AMI for k8s 1.21.4
```
suket22, line 35: I think this should be `make 1.21` ?

In the rest of the doc changes as well, we've been referencing the patch version of the K8s version to build, but we tend to recommend customers only specify the minor version and let the Makefile pick up the latest available patch.

cartermckinnon:
I agree with @suket22's comments, resetting approval until they are addressed.

README.md
icarthick: Made the change

README.md
icarthick: Made the change

icarthick:
@suket22 @cartermckinnon  please verify the changes

cartermckinnon (approved):

PR #826, opened 2021-12-06
Adds maxPods for g5g, m5n.metal and dl1.24xlarge

suket22:
*Issue #, if available:*
N/A

*Description of changes:*
Adding new instanceTypes to the max-pods file. 

### DIFF
--- a/files/eni-max-pods.txt
+++ b/files/eni-max-pods.txt
@@ -11,7 +11,7 @@
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
 #
-# This file was generated at 2021-11-30T17:52:24+01:00
+# This file was generated at 2021-12-06T09:52:52-08:00
 #
 # Mapping is calculated from AWS EC2 API using the following formula:
 # * First IP on each ENI is not used for pods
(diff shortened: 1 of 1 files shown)


files/eni-max-pods.txt
```diff
@@ -134,6 +134,7 @@ d3en.4xlarge 38
 d3en.6xlarge 58
 d3en.8xlarge 78
 d3en.xlarge 10
+dl1.24xlarge 2942
```
suket22, line 137: @mikestef9 pointed out that this isn't right. With the VPC CNI, only the first network card is used and the rest are reserved for use by the EFA. 

So rather than `2942`, which was `60 * (50 - 1) + 2`, I'll change this value to `737` which is `15 * (50 - 1) + 2`. 

We'll need to also embed this logic within these two scripts - 

https://github.com/aws/amazon-vpc-cni-k8s/blob/master/scripts/gen_vpc_ip_limits.go#L146
https://github.com/awslabs/amazon-eks-ami/blob/master/files/max-pods-calculator.sh

abeer91 (approved):

PR #834, opened 2021-12-24
fix bug for the permission of  /etc/logrotate.conf 

yibaoren:
permission of  /etc/logrotate.conf  is not right, which result in fail for all logrotate

*Issue #, if available:*

*Description of changes:*


yibaoren:
I have tested and used in production environment

PR #845, opened 2022-01-18
changes to build kontain AMI for k8s 1.21

johnmuth81:
This is a look at the changes to build a Kontain AMI for EKS. This AMI is based on the Amazon AMI. There are 
a couple todo's:

1) For testing, kkm.run is being pulled from a s3 bucket I made because kontain_bin.tar in v.0.9.4 doesn't contain it. This will change on the next KM release.
2) Runtime manager config changes are TBD. Waiting for AWS to retire dockershim and support containerd. Expected thus quater.


johnmuth81:
mistake. Sorry

Issue #844, opened 2022-01-17
How to enable `containerd` when using EKS managed node group


bryantbiggs:
I know this may not be the best place to ask, but I wanted to give it a shot anyways. 

Is there a way to simply just set environment variables when using EKS managed node groups (*WITHOUT specifying a custom AMI ID and therefore relying on the default user data bootstrap script provided by the EKS managed node group) and still enable `contianerd`?

You can see more details [here](https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1770#issuecomment-1014644200), but it looks like in order to get `containerd` we have to go into some bash fu in the user data that we/users provide that gets pre-pended to the user data that the EKS managed node group service provides in order to get this to work:

```bash
#!/bin/bash
set -ex
cat <<-EOF > /etc/profile.d/bootstrap.sh
export CONTAINER_RUNTIME="containerd"
export USE_MAX_PODS=false
export KUBELET_EXTRA_ARGS="--max-pods=110"
EOF
# Source extra environment variables in bootstrap script
sed -i '/^set -o errexit/a\\nsource /etc/profile.d/bootstrap.sh' /etc/eks/bootstrap.sh
```

I just wanted to see if it were possible to do the following instead do (at least, thats what the "API" looks like when viewing the [`bootstrap.sh` script](https://github.com/awslabs/amazon-eks-ami/blob/master/files/bootstrap.sh#L120-L132):
```bash
export CONTAINER_RUNTIME="containerd"
export USE_MAX_PODS=false
export KUBELET_EXTRA_ARGS="--max-pods=110"
```

thank you!

cartermckinnon:
For this specific case, yes; this is the best way to accomplish it. The upcoming 1.22 release will make such workarounds unnecessary, as `containerd` will become the default runtime.

cartermckinnon:
Sorry, my info was outdated -- the change to `containerd` has been pushed to the 1.23 release; there is more context available [in the docs](https://docs.aws.amazon.com/eks/latest/userguide/dockershim-deprecation.html).

stevehipwell:
@bryantbiggs @cartermckinnon I'm planning on opening a PR in this repo to update _bootstrap.sh_ with a known file to source when present. This will be a no-op unless someone creates the file to set these variables (or other variables such as dynamic labels).

lobeck:
This does not seem to work anymore.

It only uses the args passed through the call to bootstrap.sh and the environment variable gets overwritten.

[...]

PR #855, opened 2022-01-27
Support bootstrap env from file

stevehipwell:
https://github.com/awslabs/amazon-eks-ami/issues/844

*Description of changes:*

Support providing environment variables to _bootstrap.sh_ via an optional file (_/etc/profile.d/bootstrap-env.sh_) that is sourced if it exists. This solves the problem of no variable persistence between cloud-init steps which currently stops MNGs providing bootstrap overrides. To support adding `KUBELET_EXTRA_ARGS` to MNGs if  `KUBELET_EXTRA_ARGS` is set before the args are processed it will be persisted to  `KUBELET_EXTRA_ARGS_ENV` which will be appended to any value pass in via the args.

From non-MNG nodes this functionality can be used to add dynamic node labels based on instance metadata amongst other things.

This pattern has been in use for while with no issues, but this official change will make it safer as it removes the need to modify _bootstrap.sh_.

### DIFF
--- a/files/bootstrap.sh
+++ b/files/bootstrap.sh
@@ -11,6 +11,19 @@ trap 'err_report $LINENO' ERR
 
 IFS=$'\n\t'
 
+# Source bootstrap env file if it exists, this file can be used to expose variable for bootstrapping wihtout modifying this script.
+if [[ -f /etc/profile.d/bootstrap-env.sh ]]
+then
+  source /etc/profile.d/bootstrap-env.sh
+fi
+
+# If KUBELET_EXTRA_ARGS have been exported they will be overridden so preserve them to merge together
+if [[ -n "${KUBELET_EXTRA_ARGS:-}" ]]
+then
+  KUBELET_EXTRA_ARGS_ENV="${KUBELET_EXTRA_ARGS}"
+  KUBELET_EXTRA_ARGS=""
+fi
+
 function print_help {
     echo "usage: $0 [options] <cluster-name>"
     echo "Bootstraps an instance into an EKS cluster"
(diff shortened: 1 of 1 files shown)


stevehipwell:
@cartermckinnon could you take a look at this please?

suket22:
The alternative I've been thinking of is that the bootstrap script supports accepting two filePaths : one that points to the kubelet configuration, and one to the container runtime configuration. The latter now exists for [containerd](https://github.com/awslabs/amazon-eks-ami/pull/929).

If we were to re-write the bootstrapping in a different language / as a binary, it might be easier to continue to honor these configuration files rather than honoring *all* bash environment variables as defined in the original bootstrap script. 

I'm making an assumption here that overriding kubelet configuration, and container runtime configuration are the only use cases. I don't think most folks are looking to override the API Server endpoint / Cluster certificate etc. Having said that, if someone's looking to do that, they should probably just bootstrap the whole thing themselves.

stevehipwell:
@suket22 I missed your second reply. Many people are making other changes such as adding custom CAs to the OS and adding OCI credentials to kubelet. The configuration options in Bottlerocket are a good starting point, but remember that it has init containers for the real edge cases; say dynamic node labels based on the subnet the node starts in.

suket22:
> That said, I think you'll have to offer a switch between the new binary and old shell bootstrapping; so I don't think it matters if more work is done to the current logic as long as there is a way to do the same with the new binary

Agreed, I just don't know how we can offer the switch via the Managed Nodegroups API. I don't think we're going to be adding a new EKS API parameter for this. I think the best CX would be for us to switch over to a newer binary without anyone with an existing nodegroup having to change anything. 

> Many people are making other changes such as adding custom CAs to the OS and adding OCI credentials to kubelet

Fair point. I'd assume this is very niche though right? I think as a managed service we need to draw the line somewhere. At this level of customization, it probably makes sense to take complete ownership of the bootstrapping (like how custom AMIs + MNG does it).

suket22:
Playing devil's advocate, there's no way enforce backwards compatibility regardless :(
If someone's using Managed Nodegroups and is editing the `/etc/eks/bootstrap.sh` file on disk directly in their MIME part, all their edits become useless if Managed Nodegroups starts calling a different binary. 
So maybe a switch is something we'd need regardless.

cartermckinnon:
Closing along with #875 .

PR #860, opened 2022-02-10
DNS_CLUSTER_IP cannot be specified in IPv6 Cluster

literalice:
*Description of changes:*

`DNS_CLUSTER_IP` isn't recognized in a IPv6 cluster. It is fixed to the IP from `SERVICE_IPV6_CIDR`.

The PR enables setting `DNS_CLUSTER_IP` or `--dns-cluster-ip` in IPv6 env.

### DIFF
--- a/files/bootstrap.sh
+++ b/files/bootstrap.sh
@@ -385,7 +385,7 @@ MAC=$(get_meta_data 'latest/meta-data/network/interfaces/macs/' | head -n 1 | se
 if [[ -z "${DNS_CLUSTER_IP}" ]]; then
   if [[ "${IP_FAMILY}" == "ipv6" ]]; then
     if [[ -z "${SERVICE_IPV6_CIDR}" ]]; then
-      echo "Service Ipv6 Cidr must be provided when ip-family is specified as IPV6"
+      echo "Either --service-ipv6-cidr or --cluster-dns-ip must be provided when --ip-family is set to ipv6"
       exit 1
     fi
     DNS_CLUSTER_IP=$(awk -F/ '{print $1}' <<< $SERVICE_IPV6_CIDR)a
--- a/files/bootstrap.sh
+++ b/files/bootstrap.sh
@@ -299,11 +299,6 @@ if [[ ! -z "${IP_FAMILY}" ]]; then
         echo "Invalid IpFamily. Only ipv4 or ipv6 are allowed"
         exit 1
   fi
-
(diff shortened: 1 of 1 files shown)


files/bootstrap.sh
```diff
@@ -385,13 +380,17 @@ sed -i s,MASTER_ENDPOINT,$APISERVER_ENDPOINT,g /var/lib/kubelet/kubeconfig
 sed -i s,AWS_REGION,$AWS_DEFAULT_REGION,g /var/lib/kubelet/kubeconfig
 ### kubelet.service configuration
 
-if [[ "${IP_FAMILY}" == "ipv6" ]]; then
-      DNS_CLUSTER_IP=$(awk -F/ '{print $1}' <<< $SERVICE_IPV6_CIDR)a
```
literalice, line 389: It's run only when `--dns-cluster-ip` isn't specified in this PR.

files/bootstrap.sh
```diff
@@ -385,13 +380,17 @@ sed -i s,MASTER_ENDPOINT,$APISERVER_ENDPOINT,g /var/lib/kubelet/kubeconfig
...
-fi
-
 MAC=$(get_meta_data 'latest/meta-data/network/interfaces/macs/' | head -n 1 | sed 's/\/$//')
 
 if [[ -z "${DNS_CLUSTER_IP}" ]]; then
+  if [[ "${IP_FAMILY}" == "ipv6" ]]; then
+    if [[ -z "${SERVICE_IPV6_CIDR}" ]]; then
+      echo "Service Ipv6 Cidr must be provided when ip-family is specified as IPV6"
```
ravisinha0506, line 388: Can we update this error message saying "either service ipv6 cidr or cluster-dns-ip need to be present"?

bwagner5, line 388:
```suggestion
      echo "Either --service-ipv6-cidr or --cluster-dns-ip must be provided when --ip-family is set to ipv6"
```

bwagner5 (approved):
If the new error message is good with @ravisinha0506 then, lgtm

suket22 (approved):
Changes LGTM. 

I suppose this could only break those that are currently setting **both**  `--dns-cluster-ip` && `--service-ipv6-cidr`, and have been setting an incorrect value for the former. Everything is working fine for them right now because the latter is overriding the former. I don't expect anybody's currently doing this though so let's move forward. 


For the future - I think we can further reduce the complexity in the bootstrap script by never accepting `--ip-family` at all.

Please wait for @ravisinha0506 for the final approval and merge.

Issue #857, opened 2022-02-02
Upgrade Linux Kernel to 5.10


DavidSpek:
**What would you like to be added**:
Upgrade the Linux kernel to 5.10

**Why is this needed**:
[Amazon Linux 2 AMI has been upgraded to Linux 5.10](https://aws.amazon.com/about-aws/whats-new/2021/11/amazon-linux-2-ami-kernel-5-10/), so I believe the EKS AMI should also be upgraded. This is also necessary as some features of Calico and Cilium, like WireGuard for encrypting traffic, require a newer version of the Linux kernel, as mentioned in [this](https://github.com/awslabs/amazon-eks-ami/issues/727) issue.

learhy:
Hi there! I'm wondering if there is a public timeline as to when the linked PR might be merged. Thanks!

khacminh:
Hi @cartermckinnon,

Because [GPU AMI doesn't come with Kernel 5.10](https://github.com/awslabs/amazon-eks-ami/releases/tag/v20230217), IMO, this issue should be re-open.

cartermckinnon:
@khacminh I've opened a new issue to track this, #1222 .

PR #862, opened 2022-02-11
Use kernel 5.10

noony:
*Issue #, if available:*
#857 

*Description of changes:*
Upgrade kernel version to 5.10 for kubernetes version above ~~1.19~~ 1.22.
It's useful for wireguard transparent encryption, it also includes performance improvements for Intel Ice Lake processors and AWS Graviton2 processors.

More details [here](https://aws.amazon.com/about-aws/whats-new/2021/11/amazon-linux-2-ami-kernel-5-10/?nc1=h_ls)

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,7 +78,7 @@ templates in the same bucket under the path `s3://amazon-eks/1.10.3/2018-06-05/`
 
 By default, the `amazon-eks-ami` uses a [source_ami_filter](https://github.com/awslabs/amazon-eks-ami/blob/e3f1b910f83ad1f27e68312e50474ea6059f052d/eks-worker-al2.json#L46) that selects the latest [hvm](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/virtualization_types.html) AL2 AMI for the given architecture as the base AMI. For more information on what kernel versions are running on published Amazon EKS optimized Linux AMIs, see [the public documentation](https://docs.aws.amazon.com/eks/latest/userguide/eks-linux-ami-versions.html).
 
-When building an AMI, you can set the `kernel_version` to `4.14` or `5.4` to customize the kernel version. The [upgrade_kernel.sh script](https://github.com/awslabs/amazon-eks-ami/blob/master/scripts/upgrade_kernel.sh#L26) contains the logic for updating and upgrading the kernel. For Kubernetes versions 1.18 and below, it uses the `4.14` kernel if not set, and it will install the latest patches. For Kubernetes version 1.19 and above, it uses the `5.4` kernel if not set.
+When building an AMI, you can set the `kernel_version` to `4.14`, `5.4` or `5.10` to customize the kernel version. The [upgrade_kernel.sh script](https://github.com/awslabs/amazon-eks-ami/blob/master/scripts/upgrade_kernel.sh#L26) contains the logic for updating and upgrading the kernel. For Kubernetes versions 1.18 and below, it uses the `4.14` kernel if not set, and it will install the latest patches. For Kubernetes version 1.19 and above, it uses the `5.4` kernel if not set. For Kubernetes version 1.22 and above, it uses the `5.10` kernel if not set.
 
 ## Security
 
--- a/scripts/upgrade_kernel.sh
+++ b/scripts/upgrade_kernel.sh
@@ -16,8 +16,10 @@ if [[ -z "$KERNEL_VERSION" ]]; then
 
     if (( ADDR[0] == 1 && ADDR[1] < 19 )); then
         KERNEL_VERSION=4.14
-    else
+    elif (( ADDR[0] == 1 && ADDR[1] < 22 )); then
         KERNEL_VERSION=5.4
+    else
+        KERNEL_VERSION=5.10
     fi
 
     echo "kernel_version is unset. Setting to $KERNEL_VERSION based on kubernetes_version $KUBERNETES_VERSION"


noony:
Hi, can I have input on this PR ?
Is it something possible ? Should we build additional ami with kernel-5.10 in the name ?
Thanks in advance for your response.

README.md
```diff
@@ -78,7 +78,7 @@ templates in the same bucket under the path `s3://amazon-eks/1.10.3/2018-06-05/`
...
 
-When building an AMI, you can set the `kernel_version` to `4.14` or `5.4` to customize the kernel version. The [upgrade_kernel.sh script](https://github.com/awslabs/amazon-eks-ami/blob/master/scripts/upgrade_kernel.sh#L26) contains the logic for updating and upgrading the kernel. For Kubernetes versions 1.18 and below, it uses the `4.14` kernel if not set, and it will install the latest patches. For Kubernetes version 1.19 and above, it uses the `5.4` kernel if not set.
+When building an AMI, you can set the `kernel_version` to `4.14`, `5.4` or `5.10` to customize the kernel version. The [upgrade_kernel.sh script](https://github.com/awslabs/amazon-eks-ami/blob/master/scripts/upgrade_kernel.sh#L26) contains the logic for updating and upgrading the kernel. For Kubernetes versions 1.18 and below, it uses the `4.14` kernel if not set, and it will install the latest patches. For Kubernetes version 1.19 and above, it uses the `5.10` kernel if not set.
```
cartermckinnon, line 81: We should probably remove `5.4` here.

noony: fixed thanks

noony:
Hi @cartermckinnon , thanks for your response,
I'm pretty confident that it's will not break things, we already use this kernel with kops clusters without any issues (I agree with you, it's not the EKS ami)

cartermckinnon:
I got feedback from the wider team today: we'll have to make this change in tandem with a Kubernetes release. While 5.10 has largely stabilized, it isn't as battle-tested as 5.4; and we shouldn't ask users to undertake this sort of upgrade in the middle of their k8s version's support cycle. I'll update the title to reflect this.

@rtripat @prasad0896 Do you think this is realistic for 1.22?

noony:
I modified the pull request to get the kernel 5.10 for kubernetes version >= 1.22 regarding your last comment. 👍

stefansedich:
With [Dirty Pipe](https://dirtypipe.cm4all.com/) I assume this upgrade may be more pressing?

mars64:
> With [Dirty Pipe](https://dirtypipe.cm4all.com/) I assume this upgrade may be more pressing?

My .02: probably not exactly. 

The sha of the fix for dirty pipe: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/lib/iov_iter.c?id=9d2231c5d74e13b2a0546fee6737ee4446017903

[cinlin](https://cinlin.io/) lists backported patches in: `5.16.11`, `5.15.25`, `5.10.102`, `5.4.181`, 4.19.231`, `4.14.268`, and `4.9.303`

At the time of this writing the [latest ami (1.21.5-20220303)](https://docs.aws.amazon.com/eks/latest/userguide/eks-linux-ami-versions.html) is still on `5.4.176-91.338.amzn2`. Its yet unclear to me whether this kernel is modified outside the upstream versioning -- it'd be great to get confirmation from this team whether these `amzn2` kernel versions are expected to precisely match upstream.

That is to say, the shortest path to patching dirty pipe seems to be `5.4.181`, which is a somewhat distinct issue from upgrading to `5.10` -- since they're talking about bundling with a `1.22` upgrade, etc.

matdehaast:
@cartermckinnon since EKS 1.22 has been released and this didn't land, is there any idea when we can get this kernel upgrade landed?

cartermckinnon:
We're in a bind here, because:
1. At present, 5.10 isn't/wasn't stable enough to use as the default for 1.22, and we doubt that will change in time for 1.23.
2. We can't add a bootstrap flag for customers to opt-in to 5.10, because choosing a kernel at runtime isn't acceptable (it requires a reboot).
3. We can't ship a AMI variant for things like this, because it blows out our matrix and would change our support contract by requiring a deprecation path.

My best guess is that 5.10 will be the default kernel for the 1.24 release. In the meantime, if your workload necessitates 5.10, you should use a custom AMI.

domderen:
For others that land here, my workaround for this issue, was to create an EKS node group based on the [Bottlerocket OS](https://github.com/bottlerocket-os/bottlerocket). The officially supported image for EKS versions 1.20 and above is using kernel 5.10. It is a completely different architecture of a host system, so it might not be suitable for usecases that required deep modifications of the host environment. But if you are just looking to use features of 5.10 kernel in your workloads, like utilizing Wireguard, then Bottlerocket OS 1.5.3 (aws-k8s-1.20) can help.

cartermckinnon:
@LCaparelli we're not aware of any active kernel CVE's in our current AMI's; it's possible the one you're referring to has already been addressed. Please follow the instructions [here](https://aws.amazon.com/security/vulnerability-reporting/) if otherwise.

stevehipwell:
@cartermckinnon could we get a status update on this for EKS v1.24 (November 2022 I think)?

stevehipwell:
@cartermckinnon @bwagner5 could we get a status update on this as the first EKS v1.24 AMI has been released and I don't see any mention of the v5.10 kernel? This is going to be a major issue if we can't use tools which require a modern kernel (e.g. eBPF).

cartermckinnon:
You're correct, apologies for not updating the status here. Timelines haven't align as planned with 1.24 GA and the 5.10 migration; but we're working on it, and I expect 1.24 AMI's to use 5.10 before the end of the year.

stevehipwell:
@cartermckinnon does AL 2022 come into this discussion at all? Based on the docs it looks to be using kernel v5.15 by default. Is there a plan to move this to AL 2022 or to add an additional image based on it?

cartermckinnon:
We do plan to use an AL2022 base when possible; my current forecast is 1.25 will ship atop AL2022, assuming the GA dates are reasonably aligned.

PR #863, opened 2022-02-14
without sudo command doesn't work

OneideLuizSchneider:
*Issue #, if available:*

*Description of changes:*
- Changed documentation, added `sudo` on the first command

Additional information:
- k8s v1.21
- error message = `warning: Failed to create the file eks-log-collector.sh: Permission denied`
- Tested today

### DIFF
--- a/log-collector-script/linux/README.md
+++ b/log-collector-script/linux/README.md
@@ -10,7 +10,7 @@ At a high level, you run this script on your Kubernetes node, and it will collec
 
 * Run this project as the root user:
 ```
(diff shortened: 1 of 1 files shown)


cartermckinnon:
Can you provide more information about to reproduce this? You should not need to escalate if you have write permissions for your `$PWD`.

cartermckinnon:
The step you're updating in this PR is preceded by:
>  Run this project ***as the root user***:

You will never need to `sudo` if you are already `root`. I'll open a PR to remove `sudo` from the second line, because that's not necessary and is confusing.

If you intend to execute these steps as a user other than `root`, just make sure you're in a writable directory for your user, such as `$HOME`.

PR #867, opened 2022-02-17
Downgrade and lock ec2-utils at 1.2-45 until 1.2-47 is available.

cartermckinnon:
*Description of changes:*

`ec2-utils-1.2.46` introduced a bug related to device symlinks, in which some nvme devices would not have standard symlinks properly maintained. Our latest AMI (`v20220210` at the time of writing) includes this version of `ec2-utils`.

This change downgrades this AMI to `ec2-utils-1.2-45` and locks the version. We'll need to remove this patch once `ec2-utils-1.2-47` is available.

More info available here: https://github.com/aws/amazon-ec2-utils/issues/22

### DIFF
--- a/scripts/install-worker.sh
+++ b/scripts/install-worker.sh
@@ -59,13 +59,18 @@ sudo yum install -y \
     chrony \
     conntrack \
     curl \
-    jq \
     ec2-instance-connect \
+    ipvsadm \
+    jq \
     nfs-utils \
     socat \
     unzip \
     wget \
-    ipvsadm
+    yum-plugin-versionlock
+
+# Downgrade and lock ec2-utils until 1.2-47 is available: https://github.com/aws/amazon-ec2-utils/issues/22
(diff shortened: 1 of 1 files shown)


scripts/install-worker.sh
```diff
@@ -59,13 +59,18 @@ sudo yum install -y \
...
     nfs-utils \
     socat \
     unzip \
     wget \
-    ipvsadm
+    yum-plugin-versionlock
+
+# Downgrade and lock ec2-utils until 1.2-47 is available: https://github.com/aws/amazon-ec2-utils/issues/22
+sudo yum downgrade -y ec2-utils-1.2-45.amzn2.noarch
+sudo yum versionlock ec2-utils-*
```
ravisinha0506, line 73: How did we validate this change?

cartermckinnon: I built a `1.21` AMI, created a `t2.micro` with it, and verified that `yum list | grep ec2-utils` showed `1.2-45`.

PR #871, opened 2022-02-25
Use only secure protocols for SSH

SorinBaiuta:
*Issue #, if available:* It is noticed that the SSH server supports the following weak KEX algorithm(s):

KEX algorithm                                        | Reason
diffie-hellman-group-exchange-sha1 | Using SHA-1
diffie-hellman-group1-sha1                  | Using Oakley Group 2 (a 1024-bit MODP group) and SHA-1

Impact: An attacker can quickly break individual connections.
A nation-state can break a 1024-bit prime.

Millions of HTTPS, SSH, and VPN servers all use the same prime numbers for Diffie-Hellman key
exchange. Practitioners believed this was safe as long as new key exchange messages were generated
for every connection. However, the first step in the number field sieve-the most efficient
algorithm for breaking a Diffie-Hellman connection-is dependent only on this prime.

*Description of changes:* Specify sha256 for KEX Algorithms

### DIFF
--- a/scripts/install-worker.sh
+++ b/scripts/install-worker.sh
@@ -99,6 +99,8 @@ fi
 
 # Disable weak ciphers
 echo -e "\nCiphers aes128-ctr,aes256-ctr,aes128-gcm@openssh.com,aes256-gcm@openssh.com" | sudo tee -a /etc/ssh/sshd_config
+# Use only secure protocols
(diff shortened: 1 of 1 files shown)


SorinBaiuta:
Hi @ravisinha0506 ,
which would be the amount of time interval for reviews?

KR,
Sorin

cartermckinnon:
Change seems reasonable, but I don't think we should configure this sort of thing separately from Amazon Linux. It's easy enough for users to do in their user data if it's necessary for their use case; but I'd probably recommend SSM over SSH anyway.

Issue #873, opened 2022-03-01
Let users pass in a kubelet config file which is merged with the original after bootstrap.sh is finished


Skarlso:
**What would you like to be added**:

Right now, bootstrap.sh accepts extra kubelet flags. That's not nice if the thing we try to configure is hard to express via flags, or doesn't exist altogether. I propose that we let users pass in a kubelet config file parsed as JSON and then do something like this:

```bash
# note, KUBELET_EXTRA_CONFIG is provided by the user such as: {"evictionHard":{"memory.available":"200Mi","nodefs.available":"10%"},"featureGates":{"RotateKubeletServerCertificate":false},"kubeReserved":{"cpu":"321m","ephemeral-storage":"1Gi","memory":"445Mi"},"kubeReservedCgroup":"/kube-reserved","systemReserved":{"cpu":"333m","ephemeral-storage":"1Gi","memory":"333Mi"}}


TMP_KUBE_CONF='/tmp/kubelet-conf-temp.json'
KUBELET_CONFIG_EXTRA='/tmp/kubelet-conf-extra.json'
echo '${KUBELET_EXTRA_CONFIG}' > ${KUBELET_CONFIG_EXTRA}


echo "merging user options into kubelet-config.json"
trap 'rm -f ${TMP_KUBE_CONF} ${KUBELET_CONFIG_EXTRA}' EXIT
jq -s '.[0] + .[1]' "${KUBELET_CONFIG}" "${KUBELET_CONFIG_EXTRA}" > "${TMP_KUBE_CONF}"
mv "${TMP_KUBE_CONF}" "${KUBELET_CONFIG}"
```

**Why is this needed**:

This is needed because bootstrap overwrites the kubeconfig after any initial scripts would have had the change to update its values to something else.

A good example on why this would be nice is to normalise the usage of parameter configuration in eksctl for example relateing to this issue: https://github.com/weaveworks/eksctl/issues/4459

Also it would make other issues on here about kubelet and configurations a lot easier, since people could just plain create a kubelet config and use that.

suket22:
We were trying to solve a similar problem with Karpenter too and an alternative approach we've adopted [is introducing a different flag in the bootstrap script](https://github.com/awslabs/amazon-eks-ami/pull/905) that doesn't enable the kubelet but does everything else.

[...]

PR #875, opened 2022-03-02
Adding JSON configuration options for overwrites to bootstrap script

Skarlso:
Closes https://github.com/awslabs/amazon-eks-ami/issues/873

*Description of changes:*

Changes `bootstrap.sh` to accept a JSON kubelet config which merges it with the existing config after all changes have been applied by the bootstrap script.


files/bootstrap.sh
```diff
@@ -55,6 +56,11 @@ while [[ $# -gt 0 ]]; do
             shift
             shift
             ;;
+        --kubelet-extra-json-config)
```
suket22, line 59: * You've defined the arg as `--kubelet-extra-json-args` in the helper on #L23. Can we instead have this as `--kubelet-extra-json-config-file` and take a pointer to a file instead of the file contents?
* You probably also want to default `KUBELET_EXTRA_JSON_CONFIG` like on #L126, otherwise it might fail with this variable not being bound in the `if -n` check later on.

files/bootstrap.sh
```diff
@@ -498,6 +504,16 @@ else
     exit 1
 fi
 
+if [[ -n "${KUBELET_EXTRA_JSON_CONFIG}" ]]; then
+    TMP_KUBE_CONF='/tmp/kubelet-conf-temp.json'
+    KUBELET_CONFIG_EXTRA='/tmp/kubelet-conf-extra.json'
+    echo '%s' > ${KUBELET_CONFIG_EXTRA}
```
suket22, line 510: If we take in a pointer to a file, could we change this entire section to something simpler like - 
```
if [[ -n "${KUBELET_EXTRA_JSON_CONFIG}" ]]; then
    echo "bootstrap.sh: merging user json options into kubelet-config.json"
    jq -s '.[0] + .[1]' "${KUBELET_CONFIG}" "${KUBELET_EXTRA_JSON_CONFIG}" > $KUBELET_CONFIG
fi
```

`jq` doesn't produce a partial result, so we shouldn't need to redirect to a temp file and then copy it over unless I'm missing something else :)

files/bootstrap.sh
```diff
@@ -20,6 +20,7 @@ function print_help {
     echo "--b64-cluster-ca The base64 encoded cluster CA content. Only valid when used with --apiserver-endpoint. Bypasses calling \"aws eks describe-cluster\""
     echo "--apiserver-endpoint The EKS cluster API Server endpoint. Only valid when used with --b64-cluster-ca. Bypasses calling \"aws eks describe-cluster\""
     echo "--kubelet-extra-args Extra arguments to add to the kubelet. Useful for adding labels or taints."
+    echo "--kubelet-extra-json-args Extra json configuration for kubelet. Userful for adding all kinds of extra configuration such as evictionHard or kubeReserved settings in JSON format."
```
suket22, line 23: Can we also add `Will override any conflicting setting that the script defaults to.`

suket22 (commented):
@cartermckinnon anything you want to add to this?

files/bootstrap.sh
Skarlso: Yeah if I remember correctly, this indirection was added because of a different thing. But it was a couple months ago now, so I'll try to remember why.

files/bootstrap.sh
Skarlso: Hm, I'm not sure we can. I remember something about this, but we pass it in as definitive data. I don't think we can make it be on the instance when this is being executed.

cartermckinnon: My only nit is I'd call this `--kubelet-extra-config`/`KUBELET_EXTRA_CONFIG`. We don't need to mention the format of the config file here, and this should have the same semantics as `--kubelet-extra-args`.

files/bootstrap.sh
```diff
@@ -517,6 +523,16 @@ else
...
+if [[ -n "${KUBELET_EXTRA_JSON_ARGS}" ]]; then
+    TMP_KUBE_CONF='/tmp/kubelet-conf-temp.json'
+    KUBELET_CONFIG_EXTRA='/tmp/kubelet-conf-extra.json'
+    echo '%s' > ${KUBELET_CONFIG_EXTRA}
+
+    echo "bootstrap.sh: merging user json options into kubelet-config.json"
+    trap 'rm -f ${TMP_KUBE_CONF} ${KUBELET_CONFIG_EXTRA}' EXIT
+    jq -s '.[0] + .[1]' "${KUBELET_CONFIG}" "${KUBELET_CONFIG_EXTRA}" > "${TMP_KUBE_CONF}"
+    mv "${TMP_KUBE_CONF}" "${KUBELET_CONFIG}"
+fi
```
cartermckinnon, line 535:
```suggestion
if [[ -n "${KUBELET_EXTRA_CONFIG_FILE}" ]]; then
    echo "bootstrap.sh: merging $KUBELET_EXTRA_CONFIG_FILE into ${KUBELET_CONFIG}"
    jq -s '.[0] + .[1]' "${KUBELET_CONFIG}" "${KUBELET_EXTRA_CONFIG_FILE}" > $KUBELET_CONFIG
fi
```

cartermckinnon (changes requested):
Just some clarifications on the flag name and doc.

Skarlso:
But that's a separate issue. This issue deals with the fact when there is extra input from the user to a node on which there are no extra files. Only incoming data that we would like to merge with kube config. I can add an extra flag that deals with that. I think making this value to be both needlessly complicates things including the parsing and detecting if it's a file or not. Not to mention overloading the naming and the purpose of a single  flag. SRP and all that?

files/bootstrap.sh
Skarlso: Create the file with what content? :)

suket22: > eksctl is no longer constructing userdata

Maybe I'm misunderstanding, but isn't eksctl creating userData with `/etc/eks/bootstrap.sh $args`? I don't think I quite understand why eksctl cannot take in the json input for `kubeletExtraConfig`, put it in a file and then provide it to the bootstrap script. 

If instead, it an eksctl customer that's fully authoring their UserData (including their bootstrap script), then I'd expect them to do - 
```
cat <<EOF > /etc/eks/kubelet-custom-config.json
{
  "kubeReserved": 240m
}
EOF
/etc/eks/bootstrap.sh --kubelet-extra-config /etc/eks/kubelet-custom-config.json $otherArgs
```

I understand that the eksctl API accepts additional kubelet config as JSON / YAML input if I'm [reading this right](https://github.com/weaveworks/eksctl/issues/4459), but I don't think we need to force the bootstrap script's interface to also match that semantic.

suket22:
Overloading a single arg to either be a file or actual content seems odd to me, and don't think we should go down the route.

cartermckinnon:
I suggested it, because:
1. When you only want to set a parameter or two, like in @suket22 's example, it feels a little silly to throw that in a file instead of just doing `--kubelet-extra-config '{"kubeReserved": "240m"}'`.
2. The implementation is pretty simple:
```
if [[ -n "${KUBELET_EXTRA_CONFIG}" ]]; then
    echo "bootstrap.sh: merging user json options into kubelet-config.json"
    if [ ! -f "${KUBELET_EXTRA_CONFIG" ]
    then
        # value isn't a file, so treat it as JSON
        TEMP_FILE=$(mktemp)
        echo "$KUBELET_EXTRA_CONFIG" > $TEMP_FILE
        KUBELET_EXTRA_CONFIG=$TEMP_FILE
    fi
    jq -s '.[0] + .[1]' "${KUBELET_CONFIG}" "${KUBELET_EXTRA_CONFIG}" > $KUBELET_CONFIG
fi
```
3. We should only have one flag related to extra kubelet JSON config, i.e. we can't have one for JSON inline and one for a JSON file. We'd need to do a 3-way merge 👎 .
4. Accepting either content inline or content from a file isn't really novel. One example of this is `curl`'s `--data`/`-d` flag. Another is the AWS CLI, which accepts a file for any parameter value (commonly used for flags like `--filters`; which also usually accept two different content formats inline -- JSON and some other "shorthand").

It's not a hill I need to die on, but it sounds like inline JSON was @Skarlso 's original goal here.

files/bootstrap.sh
Skarlso: It doesn't work, because anything we do just gets overwritten ultimately in this script. Whatever we do outside, the value just gets clobbered over with some defaults.

The reason we don't do the first part is that using `kubelet-extra-config` is cumbersome and overly complex for simple values and passing in some json data. :)

Skarlso:
@cartermckinnon Ah I might have misunderstood what you want. You would like to change an existing flag to either accept a file or a set of JSON data? That wouldn't be a problem actually.

How we would leverage it actually is by just providing a few things without them getting overwritten with a nice value which doesn't require us to pass in a convoluted thing where we have to pre-create files and edit everything.

We have existing code which translates and creates data based on custom YAML data which is then just directly forwarded to the bootstrap script. 

```yaml
kind: ClusterConfig

metadata:
  name: dev-cluster-1
  region: eu-north-1

nodeGroups:
  - name: ng-1
    instanceType: m5a.xlarge
    desiredCapacity: 1
    kubeletExtraConfig:
        kubeReserved:
            cpu: "300m"
            memory: "300Mi"
            ephemeral-storage: "1Gi"
        kubeReservedCgroup: "/kube-reserved"
        systemReserved:
            cpu: "300m"
            memory: "300Mi"
            ephemeral-storage: "1Gi"
        evictionHard:
            memory.available:  "200Mi"
            nodefs.available: "10%"
        featureGates:
            RotateKubeletServerCertificate: true # has to be enabled, otherwise it will be disabled
```

This is super easy and convenient for the users. We don't write this out into a file and try to merge them together with other data. We pass in our thing, and would pass this in which than uses the bash magic to create the extra kube config that then can be used to set things up. It's really slick and nice user flow IMHO.

Now, if I could pass this very same thing into kube-extra-config I'm okay with that too as long as everything gets applied properly. :)

Skarlso:
I know how to implement it, that wasn't my problem with it. :D

```bash
if [[ -n "${KUBELET_EXTRA_CONFIG}" ]]; then
```

So you're thinking about a scenario where there is a custom AMI and the file has been already backed into it, I guess? I can see that being a thing, however, I just would like to point out that just because others do something it doesn't make it a good thing. :D 

But I can get behind that I guess.

suket22:
> So you're thinking about a scenario where there is a custom AMI and the file has been already backed into it, I guess?

I was actually thinking of using this in the context of MIME multi-parts, where the first part (maybe user managed) populates the file and the second part uses this file. This could help folks override kubelet settings on Managed Nodes and Karpenter, without going the full custom AMI route. 

We'd establish a runtime contract stating "always put your kubeletConfig in this path `/etc/eks/config` in your UserData section". This way we don't need to change the EKS API shape to include kubeletParameters, but rather just rely on folks using UserData in their launch templates.

Skarlso:
>I was actually thinking of using this in the context of MIME multi-parts, where the first part (maybe user managed) populates the file and the second part uses this file.

If the data can be passed in, sure. I don't see the point though since you can merge the data immediately. But maybe I'm missing something there. :)

The main point is, that incoming data trying to do anything will be overwritten with data defaulted in the bootstrap script. That's what we are trying to avoid / fix here.

stevehipwell:
@suket22 doesn't #855 cover most of the use case here (I know it's not perfect)? Basically additional kubelet args can be set but you can't override values explicitly set in bootstrap.sh.

I'd think that the desired logic, however it is implemented, should be that values which bootstrap.sh sets explicitly need inputs to override if it's valid to do so and anything else should able to be appended. In addition to work with MNGs there needs to be a way to persist the user input after it's passed in as multi part data.

Skarlso:
@stevehipwell as written in https://github.com/weaveworks/eksctl/issues/4459#issuecomment-1055690876 we would like to avoid setting a gazillion specific arguments that are a maintenance nightmare and can change over time. Where as, passing in a stream of JSON data can be arbitrary in any regard and up to the user to define entirely.

stevehipwell:
@Skarlso the only specific arguments or env values are to override bootstrap.sh explicitly set kubelet args, which need to integrate with the bootstrap logic. In your desired design the full context of the bootstrapping is lost from your modifications, so there will be more settings off limits especially for MNGs.

Skarlso:
Sorry, is this an incomplete sentence? :D 
>the only specific arguments or env values are to override bootstrap.sh explicitly set kubelet args, which need to integrate with the bootstrap logic.

I can't really parse what you mean here.

> In your desired design the full context of the bootstrapping is lost from your modifications, so there will be more settings off limits especially for MNGs.

What? :D Which context are you referring to?

stevehipwell:
@Skarlso I don't think it's incomplete; if you're using bootstrap.sh you will need to provide variables to customise it's behaviour around the configuration it explicitly sets. For everything else you should be able to pass through your configuration, but you do need to know what can't be passed through. The context comes from the fact that kubelet args may, and often do, have corresponding changes required and from the fact that in the context of a MNG there are more constraints.

If done right bootstrapping should follow an open/closed design; you provide constrained input variables to configure the closed functionality and then extend on top of the result within the constraints allowed, which can be dynamically set based on the bootstrap inputs. E.g. disable default kube reserved values being calculated with a bootstrap variable and pass kube reserved values through (only allowed because of the bootstrap input).

Skarlso:
ah, thanks for the longer explanation, that makes more sense now. Yes, the values are dependent, sure. And some things need to be calculated based on those values. But where is that requirement coming from? What I mean is, can it be left to the user saying, listen, of you set X you need to be sure you set Y properly for things to work otherwise, they might break and we don't take responsibility on that. If you know what I'm saying?

Otherwise, I agree that we should let things lying around with which the user can shoot themselves in the foot I guess. In which case, the argument based approach would make more sense because it gives more control on other calculations?

stevehipwell:
@Skarlso longer term I think a binary to process a well defined configuration is the way forward (as mentioned in #855). My personal vote goes for the Bottlerocket configuration (or a subset) being processed by a Go binary specific to AL2 (potentially AL2022 too).

Skarlso:
@stevehipwell For sure, but that's a massive step up from this small little option to define custom Kube config, right?

stevehipwell:
@Skarlso it probably depends on the scope of the configuration supported, but it'd solve a number of hard problems.

CC @bwagner5

Skarlso:
Hmmm. I'm not completely against doing it, if it would be beneficial :)

bwagner5:
I'm convinced a new bootstrap binary is needed and probably isn't too hard to implement. We need better input options into a bootstrap script and we need better testing. The bash script has gotten us pretty far, but a bootstrap script that uses the same or similar format to Bottlerocket would be optimal IMO. The configuration scope in a new bootstrap binary could probably be scoped down quite a bit at first. We could start by supporting mostly what the current bootstrap script supports but fix the mechanisms for passing input and the order of overrides / merges.  We could also get rid of some older support that bootstrap script has, like the `dockerd` support.  

There are other advantages to a new bootstrap script, for example, unifying the AL2, eventual AL2022, and Windows bootstrap processes into one project.

Skarlso:
Heck, let's do it. I'm game. Do you have a ticket? Should I close this? What do you prefer?

stevehipwell:
@bwagner5 with a limited initial config scope and init container support all use cases would be catered for. As more of the Bottlerocket config is supported the fewer use cases would need init containers.

I'd suggest that this binary is created in a new repo and that a future aim is to make the logic available as a package to be imported where required (e.g. Karpenter to replicate reserved logic).

I'd be keen to contribute to this effort.

Skarlso:
So..... what happens to this now? :D

bennesp:
@stevehipwell @cartermckinnon do you know if something has being started in this context ? I would like to help but I don't know if something has changed in the last months

bryantbiggs:
for kubelet specifically, I would recommend waiting for 1.28 and utilizing the new drop-in configuration directory https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/#kubelet-conf-d

This would allow the EKS AMI to set the required values and allow users to augment by writing the values they wish to set into a config file in that drop in directory. This should also play quite nicely with managed nodegroups where the EKS service is actually calling the bootstrap command which makes it difficult for users to pass additional flags 

The one change we might need to make here is to add support for the `--config-dir` flag to make this process easier with managed nodegroups

Issue #848, opened 2022-01-18
Add ability to customize containerd-config.toml


anusha-ragunathan:
**What would you like to be added**:

Today, a default [containerd-config.toml](https://github.com/awslabs/amazon-eks-ami/blob/master/files/containerd-config.toml) is made available via [bootstrap.sh](https://github.com/awslabs/amazon-eks-ami/blob/master/files/bootstrap.sh#L465)
Add the ability to pass a custom container config to bootstrap.sh

A similar option is available when using docker as the CRI implementation via $DOCKER_CONFIG_JSON. Expose something like that for containerd as well.

**Why is this needed**:

Needed so custom containerd configurations can be passed to EKS bootstrapper.

bryantbiggs:
relates to #790

cartermckinnon:
We've added the ability to import a `containerd` config file from a known directory: https://awslabs.github.io/amazon-eks-ami/usage/al2/#customizing-containerd-config

We've further improved this with `NodeConfig` on our AL2023-based AMI's, which accept arbitrary `containerd` config as input and merge it as necessary with our defaults.

anish:
hey @cartermckinnon sorry to bump an old thread, but do we have to worry about this when using `NodeConfig` to override containerd config ? https://github.com/containerd/containerd/issues/5837#issuecomment-894840240

ndbaker1:
hey @anish, we actually sorted this exact issue out awhile back in https://github.com/awslabs/amazon-eks-ami/pull/1780 giving you a raw unstructured merge with the data you provide to the `NodeConfig`

anish:
that's awesome ! thank you @ndbaker1

PR #905, opened 2022-04-26
Add --kubelet-enabled to bootstrap script for experimental use

suket22:
*Issue #, if available:*
N/A

*Description of changes:*
I'm looking to add Custom UserData and AMI support within Karpenter for the AWS provider. 
We've decided to add support [as described here](https://github.com/aws/karpenter/blob/main/designs/aws/custom-user-data-and-amis.md) using a sandwich like model where UserData is executed in three distinct segments - 
`Karpenter Managed` -> `User Managed` -> `Karpenter Managed`. 

* In the first segment, we intend to complete bootstrapping but not start the kubelet. 
* In the second segment, custom user data is executed and at this point you can either perform any package installation, or even directly modify kubelet parameters if desired. 
* In the third and final segment, all Karpenter will do is start the kubelet. 

This change in the AMI enables Karpenter to use this model of UserData merging. 
We can potentially extend this ability to Managed Nodegroups. This should be backwards compatible with how Managed Nodegroups currently functions and if we move to a UserData sandwiching model, we can then let everyone modify kubelet or container-runtime params without needing the Custom AMI mode.

**How was this tested?**

1. I created an AMI in my AWS account by running `make 1.21` with these changes. 
2. I launched some worker nodes using this AMI to verify that
   1.  There was no regression in the script and it still behaves as before when this flag is not provided. 
   2.  When specifying the flag as `--kubelet-enabled false`, I verified that all the bootstrapping was complete and only the kubelet wasn't enabled. I used the ssm-agent running on the node to enable the kubelet and the nodes joined the cluster thereafter. 
   3. Used a MIME multi-part choice to simulate the UserData sandwiching and made sure that worked too.

### DIFF
--- a/files/bootstrap.sh
+++ b/files/bootstrap.sh
@@ -29,6 +29,7 @@ function print_help {
     echo "--container-runtime Specify a container runtime (default: dockerd)"
     echo "--ip-family Specify ip family of the cluster"
     echo "--service-ipv6-cidr ipv6 cidr range of the cluster"
+    echo "--kubelet-enabled [EXPERIMENTAL] enable the kubelet at the end of bootstrapping (default: true). May be removed in the future."
 }
 
 POSITIONAL=()
@@ -105,6 +106,11 @@ while [[ $# -gt 0 ]]; do
             shift
             shift
             ;;
+        --kubelet-enabled)
+            KUBELET_ENABLED=$2
+            shift
+            shift
+            ;;
         *)    # unknown option
             POSITIONAL+=("$1") # save it in an array for later
             shift # past argument
@@ -130,6 +136,7 @@ PAUSE_CONTAINER_VERSION="${PAUSE_CONTAINER_VERSION:-3.1-eksbuild.1}"
 CONTAINER_RUNTIME="${CONTAINER_RUNTIME:-dockerd}"
 IP_FAMILY="${IP_FAMILY:-}"
 SERVICE_IPV6_CIDR="${SERVICE_IPV6_CIDR:-}"
+KUBELET_ENABLED="${KUBELET_ENABLED:-true}"
 
 function get_pause_container_account_for_region () {
     local region="$1"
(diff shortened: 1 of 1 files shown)


suket22:
> It's also leaks the details of bootstrap.sh in the sense that users would depend on the location of config files. 

Yeah I think this is fair. I've gone back and forth on whether we want the bootstrap script to be the one merging configuration or rather relying on the UserData author to do it. I think the flexibility this provides is that you've got the ability to make any customization what so ever and you don't need to tailor the bootstrap script to each use case since that requires a new flag each time. For example, overriding containerd config in #848 which would be additional surface for the bootstrap script). I do think this scales best, but I agree that this is essentially pushing the complexity off to the UserData author.

> I'd prefer to go with --kubelet-extra-config and --containerd-extra-config if it would cover Karpenter's use case?

This doesn't entirely solve Karpenter's use case either since users of Karpenter would need some way to wire through this configuration, thereby expanding the API surface in the Karpenter's provisioner API. eksctl currently has as a field `kubeletExtraConfig` which is something we're looking to avoid in the Karpenter API. The Managed Nodes API doesn't accept something like `kubeletExtraConfig`, which is why this wouldn't help the use case that #855 was targeting either. I think a UserData only approach through some form of a runtime contract probably keeps everyone's APIs clean. 

Maybe an alternative for us is that these args are pointers to files rather than stringified JSON / TOML? We can then have Karpenter and Managed Nodes pass in a hardcoded file name - and have those two services inform their users that they need to always populate those files if they want add additional configuration. It should also unblock eksctl I think.

stevehipwell:
@suket22 what you're doing is similar to what I do with current MNGs; I opened #855 to make the process easier (e.g. not have to manually modify _bootstrap.sh_).

This pattern should also work for your use case if you make each user data section a separate `text/x-shellscript` mime part. In the first mime part you need to replicate #855 (until it's merged) and then in all subsequent sections you need to source _/etc/profile.d/bootstrap-env.sh_ at the beginning and save any variables you want persisted to subsequent sections into _/etc/profile.d/bootstrap-env.sh_ at the end.

PR #918, opened 2022-05-09
Fail build if docker GID isn't 1950

cartermckinnon:
*Issue #, if available:*

*Description of changes:*

Some users rely on the GID of the `docker` group, and the build should fail if that expectation isn't met. The `-f` flag causes `groupadd` to succeed when the `docker` group already exists:
```
       -f, --force
           This option causes the command to simply exit with success status if the specified group already exists.
           When used with -g, and the specified GID already exists, another (unique) GID is chosen (i.e.  -g is
           turned off).
```


abeer91 (approved):

cartermckinnon:
I build a 1.21 AMI from this branch (`make 1.21`); the build succeeded. I then launched a `t2.micro` with it, and verified that the `docker` GID was correct:
```
> cat /etc/passwd | grep docker
docker:x:1001:1950::/home/docker:/bin/bash
```

PR #925, opened 2022-05-21
include ipvs loadbalancing modules

equick:
If we update kube_proxy to use ipvs and scheduler "sed" (shortest expected delay) it fails with:
```
Can't use the IPVS proxier: IPVS proxier will not be used because the following required kernel modules are not loaded: [ip_vs_sed]
```
This PR will make all the IPVS load-balancing algorithm modules available.


cartermckinnon:
Seems fine to me. @equick can you take a look at our [updated PR testing doc](https://github.com/awslabs/amazon-eks-ami/blob/master/CONTRIBUTING.md#testing-changes) and add details on how you validated this change?

For reference, this k8s blog post outlines the feature and the IPVS algorithm modules being loaded in this PR: https://kubernetes.io/blog/2018/07/09/ipvs-based-in-cluster-load-balancing-deep-dive/

MariaSu:
any news?
with addon config for ipvs I get
```
E0412 10:15:48.739711       1 server_others.go:107] "Can't use the IPVS proxier" err="IPVS proxier will not be used because the following required kernel modules are not loaded: [ip_vs_lc]"
```
1.24 eks

cartermckinnon:
I think this is more appropriate in a custom AMI today, because we don't have any specific testing around it; and it's not going to get much attention as a result of kube-proxy not using it in any default config.

PR #927, opened 2022-05-25
Add support iSCSI persistent volumes

maksim-paskal:
*Description of changes:*
Support high availability block volumes that can be mounted by ISCSI protocol to kubernetes pods. Ceph or Longhorn support ISCSI protocol, this PR can help users to mount HA volumes to their EKS clusters by default.

Signed-off-by: Maksim Paskal <paskal.maksim@gmail.com>

### DIFF
--- a/scripts/install-worker.sh
+++ b/scripts/install-worker.sh
@@ -63,6 +63,7 @@ sudo yum install -y \
     ipvsadm \
     jq \
     nfs-utils \
+    iscsi-initiator-utils \
     socat \
     unzip \
     wget \


cartermckinnon:
@maksim-paskal Can you provide some more details about how a user would leverage this ISCSI package in Ceph (or otherwise)? We're reluctant to add such software to this AMI unless it's applicable to a majority of our users, and this looks a bit niche to me.

maksim-paskal:
@cartermckinnon, thanks revieving my PR.

Sure, this PR will help EKS users to use cloud native storages in their clusters, for example [Longhorn](https://longhorn.io/) (or otherwise), without any additional worker configuration on all nodes in cluster. This storages use ISCSI protocol to mount their volumes to pods, to mount volumes, worker needs additional OS packages was installed before mount.

Current scenario installing EKS worker, contains package [nfs-utils](https://github.com/awslabs/amazon-eks-ami/blob/master/scripts/install-worker.sh#L65) that needed to mount NFS volumes to pods, I decide if AMI now contains ability to connect to NFS storages, why users can't connect to ISCSI volumes by defaults.

This PR add support for mounting ISCSI volumes in EKS clusters

cartermckinnon:
Thanks for the details! There's a clear case for NFS, because it's needed by the EFS CSI driver; which many, many EKS users utilize. My understanding is that comparatively few of our users roll their own block storage, and these users also tend to run custom AMIs. We try to strike a balance between making this AMI "general purpose" and "universal", taking into account that every piece of software we ship in the AMI requires ongoing support and maintenance (read: security updates).

I'll let others chime in; I'm not opposed to the change if there's sufficient interest.

PR #929, opened 2022-05-25
Add support for containerd-config-file to bootstrap script

### LINKED ISSUE #830: Make containerd config configurable
Make containerd config configurable, similarly to `--docker-config-json` option.

**Why is this needed**:
By default `max_container_log_line_size` set to `16384` which means when a line is bigger than this size will be split into multiple log messages.
Our problem with this, that it's based on byte size and sometimes it splits a UTF-8 character containing multiple bytes, which makes it invalid and unable to parse by fluent-bit/fluentd.
E.g.:
```
...,"custom3":{"label":"🥱🤭😓🤑👥🖖🤲🤲🤠🤠🤠🤠🤠👊👨👨�
2021-12-10T09:59:12.166751315Z stderr F ���‍♂️🧝🧝🧞🧩🎗️🏐🎫🎰🎇🏡🏯1️⃣🆕🈹",
```

It's possible to modify the config in a custom userdata script, but would be nicer to do this through a parameter to the bootstrap script.

suket22:
*Issue #, if available:*
This resolves #830 and #848 
This PR continues the work from @codablock in #790

*Description of changes:*
Adds a new parameter to the bootstrap script `--containerd-config-file` that completely overrides the config to be used with containerd.
Note - 
* We will continue to do a sed replace on `SANDBOX_IMAGE`. This is useful in many contexts, including running in `cn-north-1`, `cn-northwest-1` where you might want to pull the pause container from ECR due to network connectivity.
* We will use the entire contents of the config file - in-line with how it works for `--docker-config-json`. 

### Testing
* I created an AMI with these changes by running `make 1.21`
* I launched a worker node using dockerd and containerd and made sure they join the cluster (no regressions)
* I tested the parameter and made sure the node joined and the config was applied. 

To replicate my results, create some custom userData first - 
```
> cat user-data-crconfig.txt
#!/bin/bash -xe

cat <<EOF > /etc/eks/custom-containerd-config.toml
version = 2
root = "/var/lib/containerd"
state = "/run/containerd"

[grpc]
address = "/run/containerd/containerd.sock"

default_runtime_name = "runc"

max_container_log_line_size = 10000

bin_dir = "/opt/cni/bin"
conf_dir = "/etc/cni/net.d"
EOF

/etc/eks/bootstrap.sh 'karp-cluster' \
--container-runtime containerd \
--containerd-config-file /etc/eks/custom-containerd-config.toml \
--kubelet-extra-args '--node-labels=test=foo'
```

and then launch an instance and see if it joins your cluster
```
> aws ec2 run-instances \
    --image-id ami-xxx \
    --instance-type t3.large \
    --count 1 \
    --subnet-id subnet-xxx \
    --security-group-ids sg-xxx \
    --iam-instance-profile Name=xx \
    --tag-specifications 'ResourceType=instance,Tags=[{Key=kubernetes.io/cluster/my-cluster,Value=owned}]' \
    --user-data file://user-data-crconfig.txt
```

### DIFF
--- a/files/bootstrap.sh
+++ b/files/bootstrap.sh
@@ -23,6 +23,7 @@ function print_help {
     echo "--enable-docker-bridge Restores the docker default bridge network. (default: false)"
     echo "--aws-api-retry-attempts Number of retry attempts for AWS API call (DescribeCluster) (default: 3)"
     echo "--docker-config-json The contents of the /etc/docker/daemon.json file. Useful if you want a custom config differing from the default one in the AMI"
+    echo "--containerd-config-file File containing the containerd configuration to be used in place of AMI defaults."
     echo "--dns-cluster-ip Overrides the IP address to use for DNS queries within the cluster. Defaults to 10.100.0.10 or 172.20.0.10 based on the IP address of the primary interface"
     echo "--pause-container-account The AWS account (number) to pull the pause container from"
     echo "--pause-container-version The tag of the pause container"
@@ -72,6 +73,11 @@ while [[ $# -gt 0 ]]; do
             ;;
         --docker-config-json)
             DOCKER_CONFIG_JSON=$2
+            shift
+            shift
+            ;;
+        --containerd-config-file)
+            CONTAINERD_CONFIG_FILE=$2
             shift
             shift
             ;;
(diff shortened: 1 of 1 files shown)


mmerkes (approved):
LGTM

cartermckinnon (approved):

PR #930, opened 2022-05-27
Update eks-log-collector.sh

dsalamancaMS:
Removed 10MB limit on messages log file collection

*Issue #, if available:*

*Description of changes:*
Removing 10MB limit on Messages file collection as it might not capture the required information due to only capturing the last 10MB lines


cartermckinnon:
Can you elaborate on the motivation here? Generally, 10MB's of logs should be plenty to diagnose an issue on the node. I'm fine with raising the limit if you've found that it's too low, but I don't think we should remove the limit altogether.

cartermckinnon:
To summarize internal conversations on this:

A 100MB limit is the most reasonable to pass the resulting archive between the support and service teams. In the vast majority of cases, it would cover the entirely of `/var/log/messages`.

cartermckinnon (approved):

PR #938, opened 2022-06-03
cache pause, vpc-cni, and kube-proxy images in the AMI

bwagner5:
*Issue #, if available:*
https://github.com/awslabs/amazon-eks-ami/issues/1034
https://github.com/awslabs/amazon-eks-ami/issues/990
https://github.com/awslabs/amazon-eks-ami/issues/1099

*Description of changes:*
 - Cache the following images to improve node bootstrapping time:
   - pause
   - aws-node images (init and aws-node)
   - kube-proxy images (minimal and normal)

kubelet logs with image pull timings before caching (v=4):

```
Oct  7 17:05:07 ip-192-168-101-146 pull-sandbox-image.sh: done: 45.698568ms
Oct  7 17:05:17 ip-192-168-101-146 kubelet: I1007 17:05:17.995376    4028 event.go:291] "Event occurred" object="kube-system/kube-proxy-z7s74" kind="Pod" apiVersion="v1" type="Normal" reason="Pulled" message="Successfully pulled image \"602401143452.dkr.ecr.us-west-2.amazonaws.com/eks/kube-proxy:v1.22.12-minimal-eksbuild.1\" in 1.712432387s"
Oct  7 17:05:20 ip-192-168-101-146 kubelet: I1007 17:05:20.322544    4028 event.go:291] "Event occurred" object="kube-system/aws-node-c24g9" kind="Pod" apiVersion="v1" type="Normal" reason="Pulled" message="Successfully pulled image \"602401143452.dkr.ecr.us-west-2.amazonaws.com/amazon-k8s-cni-init:v1.11.4-eksbuild.1\" in 3.960642579s"
Oct  7 17:05:26 ip-192-168-101-146 kubelet: I1007 17:05:26.913619    4028 event.go:291] "Event occurred" object="kube-system/aws-node-c24g9" kind="Pod" apiVersion="v1" type="Normal" reason="Pulled" message="Successfully pulled image \"60240
1143452.dkr.ecr.us-west-2.amazonaws.com/amazon-k8s-cni:v1.11.4-eksbuild.1\" in 1.390664746s"
```

The total latency in these serial image pulls is ~7 seconds.

After caching all are used from the local host:

``` 
Oct  7 17:05:09 ip-192-168-97-141 kubelet: I1007 17:05:09.666150    4912 event.go:291] "Event occurred" object="kube-system/aws-node-ldvfh" kind="Pod" apiVersion="v1" type="Normal" reason="Pulled" message="Container image \"602401143452.dkr.ecr.us-west-2.amazonaws.com/amazon-k8s-cni-init:v1.11.4-eksbuild.1\" already present on machine"
Oct  7 17:05:12 ip-192-168-97-141 kubelet: I1007 17:05:12.375142    4912 event.go:291] "Event occurred" object="kube-system/aws-node-ldvfh" kind="Pod" apiVersion="v1" type="Normal" reason="Pulled" message="Container image \"602401143452.dkr.ecr.us-west-2.amazonaws.com/amazon-k8s-cni:v1.11.4-eksbuild.1\" already present on machine"
Oct  7 17:05:12 ip-192-168-97-141 kubelet: I1007 17:05:12.493594    4912 event.go:291] "Event occurred" object="kube-system/kube-proxy-79tp5" kind="Pod" apiVersion="v1" type="Normal" reason="Pulled" message="Container image \"602401143452.dkr.ecr.us-west-2.amazonaws.com/eks/kube-proxy:v1.22.12-minimal-eksbuild.1\" already present on machine"
```

Since this PR caches the sandbox image (pause), this means we can also enable containerd to start on boot rather than in the bootstrap script which removes the containerd startup time which results in ~3 second latency reduction for when kubelet starts (P50).

Screen Shot 2022-10-07 at 14 23 42

This graph shows the combined reduction of the node getting to Ready, starting w/ kube-proxy & pause cached and then adding the vpc cni images to the cache (P50):

Screen Shot 2022-10-07 at 14 17 08

Size of the containerd image cache:

```
du -h /var/lib/containerd/io.containerd.snapshotter.v1.overlayfs
...
1.1G	/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/
```

Size of a running node without any cached images (so the minimal images pulled needed to start the node):

```
du -h /var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/
...
470M	/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,6 +91,104 @@ Provisioner](https://www.packer.io/docs/provisioners/shell.html) runs the
 necessary configuration tasks.  Then, Packer creates an AMI from the instance
 and terminates the instance after the AMI is created.
 
+### Container Image Caching
+
+Optionally, some container images can be cached during the AMI build process in order to reduce the latency of the node getting to a `Ready` state when launched. 
+
+To turn on container image caching:
+
+```
+cache_container_images=true make 1.23
+```
+
+When container image caching is enabled, the following images are cached:
+ - 602401143452.dkr.ecr.<AWS_REGION>.amazonaws.com/eks/kube-proxy:<default and latest>-eksbuild.<BUILD_VERSION>
+ - 602401143452.dkr.ecr.<AWS_REGION>.amazonaws.com/eks/kube-proxy:<default and latest>-minimal-eksbuild.<BUILD_VERSION>
+ - 602401143452.dkr.ecr.<AWS_REGION>.amazonaws.com/eks/pause:3.5
+ - 602401143452.dkr.ecr.<AWS_REGION>.amazonaws.com/amazon-k8s-cni-init:<default and latest>
+ - 602401143452.dkr.ecr.<AWS_REGION>.amazonaws.com/amazon-k8s-cni:<default and latest>
+
+The account ID can be different depending on the region and partition you are building the AMI in. See [here](https://docs.aws.amazon.com/eks/latest/userguide/add-ons-images.html) for more details.
+
+Since the VPC CNI is not versioned with K8s itself, the latest version of the VPC CNI and the default version, based on the response from the EKS DescribeAddonVersions at the time of the AMI build, will be cached. 
+
+The images listed above are also tagged with each region in the partition the AMI is built in, since images are often built in one region and copied to others within the same partition. Images that are available to pull from an ECR FIPS endpoint are also tagged as such (i.e. `602401143452.dkr.ecr-fips.us-east-1.amazonaws.com/eks/pause:3.5`).
+
+When listing images on a node, you'll notice a long list of images. However, most of these images are simply tagged in different ways with no storage overhead. Images cached in the AMI total around 1.0 GiB. In general, a node with no images cached using the VPC CNI will use around 500 MiB of images when in a `Ready` state with no other pods running on the node.
+
+### IAM Permissions
+
+To build the EKS Optimized AMI, you will need the following permissions:
+
+```
+{
+    "Version": "2012-10-17",
+    "Statement": [
+        {
+            "Effect": "Allow",
+            "Action": [
+                "ec2:AttachVolume",
+                "ec2:AuthorizeSecurityGroupIngress",
+                "ec2:CopyImage",
+                "ec2:CreateImage",
+                "ec2:CreateKeypair",
+                "ec2:CreateSecurityGroup",
+                "ec2:CreateSnapshot",
+                "ec2:CreateTags",
+                "ec2:CreateVolume",
+                "ec2:DeleteKeyPair",
+                "ec2:DeleteSecurityGroup",
+                "ec2:DeleteSnapshot",
+                "ec2:DeleteVolume",
+                "ec2:DeregisterImage",
+                "ec2:DescribeImageAttribute",
+                "ec2:DescribeImages",
+                "ec2:DescribeInstances",
+                "ec2:DescribeInstanceStatus",
+                "ec2:DescribeRegions",
+                "ec2:DescribeSecurityGroups",
+                "ec2:DescribeSnapshots",
+                "ec2:DescribeSubnets",
+                "ec2:DescribeTags",
+                "ec2:DescribeVolumes",
+                "ec2:DetachVolume",
+                "ec2:GetPasswordData",
+                "ec2:ModifyImageAttribute",
+                "ec2:ModifyInstanceAttribute",
+                "ec2:ModifySnapshotAttribute",
+                "ec2:RegisterImage",
+                "ec2:RunInstances",
+                "ec2:StopInstances",
+                "ec2:TerminateInstances",
+                "eks:DescribeAddonVersions",
+                "ecr:GetAuthorizationToken"
+            ],
+            "Resource": "*"
+        },
+        {
+            "Effect": "Allow",
+            "Action": [
+                "ecr:BatchGetImage",
+                "ecr:BatchCheckLayerAvailability",
+                "ecr:GetDownloadUrlForLayer"
+            ],
+            "Resource": "arn:aws:ecr:us-west-2:602401143452:repository/*"
+        },
+        {
+            "Effect": "Allow",
+            "Action": [
+                "s3:GetObject"
+            ],
+            "Resource": "arn:aws:s3:::amazon-eks/*"
+        }
+    ]
+}
+```
+
+You will need to use the region you are building the AMI in to specify the ECR repository resource in the second IAM statement. You may also need to change the account if you are building the AMI in a different partition or special region. You can see a mapping of regions to account ID [here](https://docs.aws.amazon.com/eks/latest/userguide/add-ons-images.html).
+If you're using a custom s3 bucket to vend different K8s binaries, you will need to change the resource in the third IAM statement above to reference your custom bucket.
+For more information about the permissions required by Packer with different configurations, see the [docs](https://www.packer.io/plugins/builders/amazon#iam-task-or-instance-role).
+
 ## Using the AMI
 
 If you are just getting started with Amazon EKS, we recommend that you follow
--- a/eks-worker-al2.json
+++ b/eks-worker-al2.json
@@ -13,6 +13,7 @@
     "aws_session_token": "{{env `AWS_SESSION_TOKEN`}}",
     "binary_bucket_name": "amazon-eks",
     "binary_bucket_region": "us-west-2",
+    "cache_container_images": "false",
     "cni_plugin_version": "v0.8.6",
     "containerd_version": "1.6.6-1.amzn2.0.2",
     "creator": "{{env `USER`}}",
@@ -23,7 +24,8 @@
     "kms_key_id": "",
     "kubernetes_build_date": null,
     "kubernetes_version": null,
-    "launch_block_device_mappings_volume_size": "4",
+    "launch_block_device_mappings_volume_size": "8",
+    "pause_container_version": "3.5",
     "pull_cni_from_github": "true",
     "remote_folder": "",
     "runc_version": "1.1.3-1.amzn2.0.2",
@@ -161,7 +163,10 @@
         "AWS_ACCESS_KEY_ID={{user `aws_access_key_id`}}",
         "AWS_SECRET_ACCESS_KEY={{user `aws_secret_access_key`}}",
         "AWS_SESSION_TOKEN={{user `aws_session_token`}}",
-        "SONOBUOY_E2E_REGISTRY={{user `sonobuoy_e2e_registry`}}"
+        "SONOBUOY_E2E_REGISTRY={{user `sonobuoy_e2e_registry`}}",
+        "PAUSE_CONTAINER_VERSION={{user `pause_container_version`}}",
+        "KUBE_PROXY_VERSION_SUFFIX={{user `kube_proxy_version_suffix`}}",
+        "CACHE_CONTAINER_IMAGES={{user `cache_container_images`}}"
       ]
     },
     {
--- a/files/bootstrap.sh
+++ b/files/bootstrap.sh
@@ -178,51 +178,6 @@ SERVICE_IPV6_CIDR="${SERVICE_IPV6_CIDR:-}"
 ENABLE_LOCAL_OUTPOST="${ENABLE_LOCAL_OUTPOST:-}"
 CLUSTER_ID="${CLUSTER_ID:-}"
 
-function get_pause_container_account_for_region() {
-  local region="$1"
-  case "${region}" in
-    ap-east-1)
-      echo "${PAUSE_CONTAINER_ACCOUNT:-800184023465}"
-      ;;
-    me-south-1)
-      echo "${PAUSE_CONTAINER_ACCOUNT:-558608220178}"
-      ;;
-    cn-north-1)
-      echo "${PAUSE_CONTAINER_ACCOUNT:-918309763551}"
-      ;;
-    cn-northwest-1)
-      echo "${PAUSE_CONTAINER_ACCOUNT:-961992271922}"
-      ;;
-    us-gov-west-1)
-      echo "${PAUSE_CONTAINER_ACCOUNT:-013241004608}"
-      ;;
-    us-gov-east-1)
-      echo "${PAUSE_CONTAINER_ACCOUNT:-151742754352}"
-      ;;
-    us-iso-east-1)
-      echo "${PAUSE_CONTAINER_ACCOUNT:-725322719131}"
-      ;;
-    us-isob-east-1)
-      echo "${PAUSE_CONTAINER_ACCOUNT:-187977181151}"
-      ;;
-    af-south-1)
-      echo "${PAUSE_CONTAINER_ACCOUNT:-877085696533}"
-      ;;
-    eu-south-1)
-      echo "${PAUSE_CONTAINER_ACCOUNT:-590381155156}"
-      ;;
-    ap-southeast-3)
-      echo "${PAUSE_CONTAINER_ACCOUNT:-296578399912}"
-      ;;
-    me-central-1)
-      echo "${PAUSE_CONTAINER_ACCOUNT:-759879836304}"
-      ;;
-    *)
-      echo "${PAUSE_CONTAINER_ACCOUNT:-602401143452}"
-      ;;
-  esac
-}
-
 # Helper function which calculates the amount of the given resource (either CPU or memory)
 # to reserve in a given resource range, specified by a start and end of the range and a percentage
 # of the resource to reserve. Note that we return zero if the start of the resource range is
@@ -314,8 +269,8 @@ if [[ "$MACHINE" != "x86_64" && "$MACHINE" != "aarch64" ]]; then
   exit 1
 fi
 
-PAUSE_CONTAINER_ACCOUNT=$(get_pause_container_account_for_region "${AWS_DEFAULT_REGION}")
-PAUSE_CONTAINER_IMAGE=${PAUSE_CONTAINER_IMAGE:-$PAUSE_CONTAINER_ACCOUNT.dkr.ecr.$AWS_DEFAULT_REGION.$AWS_SERVICES_DOMAIN/eks/pause}
+ECR_URI=$(/etc/eks/get-ecr-uri.sh "${AWS_DEFAULT_REGION}" "${AWS_SERVICES_DOMAIN}" "${PAUSE_CONTAINER_ACCOUNT:-}")
+PAUSE_CONTAINER_IMAGE=${PAUSE_CONTAINER_IMAGE:-$ECR_URI/eks/pause}
 PAUSE_CONTAINER="$PAUSE_CONTAINER_IMAGE:$PAUSE_CONTAINER_VERSION"
 
 ### kubelet kubeconfig
@@ -525,29 +480,26 @@ if [[ "$CONTAINER_RUNTIME" = "containerd" ]]; then
 
   sudo mkdir -p /etc/containerd
   sudo mkdir -p /etc/cni/net.d
-  mkdir -p /etc/systemd/system/containerd.service.d
-  cat << EOF > /etc/systemd/system/containerd.service.d/10-compat-symlink.conf
-[Service]
-ExecStartPre=/bin/ln -sf /run/containerd/containerd.sock /run/dockershim.sock
-EOF
   if [[ -n "$CONTAINERD_CONFIG_FILE" ]]; then
     sudo cp -v $CONTAINERD_CONFIG_FILE /etc/eks/containerd/containerd-config.toml
   fi
   echo "$(jq '.cgroupDriver="systemd"' $KUBELET_CONFIG)" > $KUBELET_CONFIG
   sudo sed -i s,SANDBOX_IMAGE,$PAUSE_CONTAINER,g /etc/eks/containerd/containerd-config.toml
-  sudo cp -v /etc/eks/containerd/containerd-config.toml /etc/containerd/config.toml
-  sudo cp -v /etc/eks/containerd/sandbox-image.service /etc/systemd/system/sandbox-image.service
+
+  # Check if the containerd config file is the same as the one used in the image build.
+  # If different, then restart containerd w/ proper config
+  if ! cmp -s /etc/eks/containerd/containerd-config.toml /etc/containerd/config.toml; then
+    sudo cp -v /etc/eks/containerd/containerd-config.toml /etc/containerd/config.toml
+    sudo cp -v /etc/eks/containerd/sandbox-image.service /etc/systemd/system/sandbox-image.service
+    sudo chown root:root /etc/systemd/system/sandbox-image.service
+    systemctl daemon-reload
+    systemctl enable containerd sandbox-image
+    systemctl restart sandbox-image containerd
+  fi
   sudo cp -v /etc/eks/containerd/kubelet-containerd.service /etc/systemd/system/kubelet.service
   sudo chown root:root /etc/systemd/system/kubelet.service
-  sudo chown root:root /etc/systemd/system/sandbox-image.service
   # Validate containerd config
   sudo containerd config dump > /dev/null
-  systemctl daemon-reload
-  systemctl enable containerd
-  systemctl restart containerd
-  systemctl enable sandbox-image
-  systemctl start sandbox-image
-
 elif [[ "$CONTAINER_RUNTIME" = "dockerd" ]]; then
   mkdir -p /etc/docker
   bash -c "/sbin/iptables-save > /etc/sysconfig/iptables"
--- a/files/get-ecr-uri.sh
+++ b/files/get-ecr-uri.sh
@@ -1,0 +1,54 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# More details about the mappings in this file can be found here https://docs.aws.amazon.com/eks/latest/userguide/add-ons-images.html
+
+region=$1
+aws_domain=$2
+if [[ $# -eq 3 ]] && [[ ! -z $3 ]]; then
+  acct=$3
+else
+  case "${region}" in
+    ap-east-1)
+      acct="800184023465"
+      ;;
+    me-south-1)
+      acct="558608220178"
+      ;;
+    cn-north-1)
+      acct="918309763551"
+      ;;
+    cn-northwest-1)
+      acct="961992271922"
+      ;;
+    us-gov-west-1)
+      acct="013241004608"
+      ;;
+    us-gov-east-1)
+      acct="151742754352"
+      ;;
+    us-iso-east-1)
+      acct="725322719131"
+      ;;
+    us-isob-east-1)
+      acct="187977181151"
+      ;;
+    af-south-1)
+      acct="877085696533"
+      ;;
+    eu-south-1)
+      acct="590381155156"
+      ;;
+    ap-southeast-3)
+      acct="296578399912"
+      ;;
+    me-central-1)
+      acct="759879836304"
+      ;;
+    *)
+      acct="602401143452"
+      ;;
+  esac
+fi
+
+echo "${acct}.dkr.ecr.${region}.${aws_domain}"
--- a/files/pull-image.sh
+++ b/files/pull-image.sh
@@ -1,0 +1,27 @@
+#!/usr/bin/env bash
+
+img=$1
+region=$(echo "${img}" | cut -f4 -d ".")
+MAX_RETRIES=3
+
+function retry() {
+  local rc=0
+  for attempt in $(seq 0 $MAX_RETRIES); do
+    rc=0
+    [[ $attempt -gt 0 ]] && echo "Attempt $attempt of $MAX_RETRIES" 1>&2
+    "$@"
+    rc=$?
+    [[ $rc -eq 0 ]] && break
+    [[ $attempt -eq $MAX_RETRIES ]] && exit $rc
+    local jitter=$((1 + RANDOM % 10))
+    local sleep_sec="$(($((5 << $((1 + $attempt)))) + $jitter))"
+    sleep $sleep_sec
+  done
+}
+
+ecr_password=$(retry aws ecr get-login-password --region $region)
+if [[ -z ${ecr_password} ]]; then
+  echo >&2 "Unable to retrieve the ECR password."
+  exit 1
+fi
+retry sudo ctr --namespace k8s.io image pull "${img}" --user AWS:${ecr_password}
--- a/files/pull-sandbox-image.sh
+++ b/files/pull-sandbox-image.sh
@@ -1,27 +1,5 @@
 #!/usr/bin/env bash
+set -euo pipefail
 
-### fetching sandbox image from /etc/containerd/config.toml
-sandbox_image=$(awk -F'[ ="]+' '$1 == "sandbox_image" { print $2 }' /etc/containerd/config.toml)
-region=$(echo "$sandbox_image" | cut -f4 -d ".")
-ecr_password=$(aws ecr get-login-password --region $region)
-API_RETRY_ATTEMPTS=5
-
-for attempt in $(seq 0 $API_RETRY_ATTEMPTS); do
-  rc=0
-  if [[ $attempt -gt 0 ]]; then
-    echo "Attempt $attempt of $API_RETRY_ATTEMPTS"
-  fi
-  ### pull sandbox image from ecr
-  ### username will always be constant i.e; AWS
-  sudo ctr --namespace k8s.io image pull $sandbox_image --user AWS:$ecr_password
-  rc=$?
-  if [[ $rc -eq 0 ]]; then
-    break
-  fi
-  if [[ $attempt -eq $API_RETRY_ATTEMPTS ]]; then
-    exit $rc
-  fi
-  jitter=$((1 + RANDOM % 10))
-  sleep_sec="$(($((5 << $((1 + $attempt)))) + $jitter))"
-  sleep $sleep_sec
-done
+sandbox_image="$(awk -F'[ ="]+' '$1 == "sandbox_image" { print $2 }' /etc/containerd/config.toml)"
+/etc/eks/containerd/pull-image.sh "${sandbox_image}"
--- a/scripts/install-worker.sh
+++ b/scripts/install-worker.sh
@@ -4,6 +4,7 @@ set -o pipefail
 set -o nounset
 set -o errexit
 IFS=$'\n\t'
+export AWS_DEFAULT_OUTPUT="json"
 
 TEMPLATE_DIR=${TEMPLATE_DIR:-/tmp/worker}
 
@@ -30,6 +31,8 @@ validate_env_set CNI_PLUGIN_VERSION
 validate_env_set KUBERNETES_VERSION
 validate_env_set KUBERNETES_BUILD_DATE
 validate_env_set PULL_CNI_FROM_GITHUB
+validate_env_set PAUSE_CONTAINER_VERSION
+validate_env_set CACHE_CONTAINER_IMAGES
 
 ################################################################################
 ### Machine Architecture #######################################################
@@ -138,45 +141,17 @@ else
   sudo yum install -y awscli
 fi
 
-################################################################################
-### Docker #####################################################################
-################################################################################
-
-sudo yum install -y device-mapper-persistent-data lvm2
-
-INSTALL_DOCKER="${INSTALL_DOCKER:-true}"
-if [[ "$INSTALL_DOCKER" == "true" ]]; then
-  sudo amazon-linux-extras enable docker
-  sudo groupadd -og 1950 docker
-  sudo useradd --gid $(getent group docker | cut -d: -f3) docker
-
-  # install runc and lock version
-  sudo yum install -y runc-${RUNC_VERSION}
-  sudo yum versionlock runc-*
-
-  # install containerd and lock version
-  sudo yum install -y containerd-${CONTAINERD_VERSION}
-  sudo yum versionlock containerd-*
-
-  # install docker and lock version
-  sudo yum install -y docker-${DOCKER_VERSION}*
-  sudo yum versionlock docker-*
-  sudo usermod -aG docker $USER
-
-  # Remove all options from sysconfig docker.
-  sudo sed -i '/OPTIONS/d' /etc/sysconfig/docker
-
-  sudo mkdir -p /etc/docker
-  sudo mv $TEMPLATE_DIR/docker-daemon.json /etc/docker/daemon.json
-  sudo chown root:root /etc/docker/daemon.json
-
-  # Enable docker daemon to start on boot.
-  sudo systemctl daemon-reload
-fi
-
 ###############################################################################
 ### Containerd setup ##########################################################
 ###############################################################################
+
+# install runc and lock version
+sudo yum install -y runc-${RUNC_VERSION}
+sudo yum versionlock runc-*
+
+# install containerd and lock version
+sudo yum install -y containerd-${CONTAINERD_VERSION}
+sudo yum versionlock containerd-*
 
 sudo mkdir -p /etc/eks/containerd
 if [ -f "/etc/eks/containerd/containerd-config.toml" ]; then
@@ -195,7 +170,15 @@ fi
 sudo mv $TEMPLATE_DIR/kubelet-containerd.service /etc/eks/containerd/kubelet-containerd.service
 sudo mv $TEMPLATE_DIR/sandbox-image.service /etc/eks/containerd/sandbox-image.service
 sudo mv $TEMPLATE_DIR/pull-sandbox-image.sh /etc/eks/containerd/pull-sandbox-image.sh
+sudo mv $TEMPLATE_DIR/pull-image.sh /etc/eks/containerd/pull-image.sh
 sudo chmod +x /etc/eks/containerd/pull-sandbox-image.sh
+sudo chmod +x /etc/eks/containerd/pull-image.sh
+
+sudo mkdir -p /etc/systemd/system/containerd.service.d
+cat << EOF | sudo tee /etc/systemd/system/containerd.service.d/10-compat-symlink.conf
+[Service]
+ExecStartPre=/bin/ln -sf /run/containerd/containerd.sock /run/dockershim.sock
+EOF
 
 cat << EOF | sudo tee -a /etc/modules-load.d/containerd.conf
 overlay
@@ -207,6 +190,34 @@ net.bridge.bridge-nf-call-ip6tables = 1
 net.bridge.bridge-nf-call-iptables = 1
 net.ipv4.ip_forward = 1
 EOF
+
+################################################################################
+### Docker #####################################################################
+################################################################################
+
+sudo yum install -y device-mapper-persistent-data lvm2
+
+INSTALL_DOCKER="${INSTALL_DOCKER:-true}"
+if [[ "$INSTALL_DOCKER" == "true" ]]; then
+  sudo amazon-linux-extras enable docker
+  sudo groupadd -og 1950 docker
+  sudo useradd --gid $(getent group docker | cut -d: -f3) docker
+
+  # install docker and lock version
+  sudo yum install -y docker-${DOCKER_VERSION}*
+  sudo yum versionlock docker-*
+  sudo usermod -aG docker $USER
+
+  # Remove all options from sysconfig docker.
+  sudo sed -i '/OPTIONS/d' /etc/sysconfig/docker
+
+  sudo mkdir -p /etc/docker
+  sudo mv $TEMPLATE_DIR/docker-daemon.json /etc/docker/daemon.json
+  sudo chown root:root /etc/docker/daemon.json
+
+  # Enable docker daemon to start on boot.
+  sudo systemctl daemon-reload
+fi
 
 ################################################################################
 ### Logrotate ##################################################################
@@ -331,6 +342,8 @@ sudo systemctl disable kubelet
 ################################################################################
 
 sudo mkdir -p /etc/eks
+sudo mv $TEMPLATE_DIR/get-ecr-uri.sh /etc/eks/get-ecr-uri.sh
+sudo chmod +x /etc/eks/get-ecr-uri.sh
 sudo mv $TEMPLATE_DIR/eni-max-pods.txt /etc/eks/eni-max-pods.txt
 sudo mv $TEMPLATE_DIR/bootstrap.sh /etc/eks/bootstrap.sh
 sudo chmod +x /etc/eks/bootstrap.sh
@@ -361,6 +374,92 @@ if vercmp "$KUBERNETES_VERSION" gteq "1.22.0"; then
 
   # copying credential provider config file to eks folder
   sudo mv $TEMPLATE_DIR/ecr-credential-provider-config /etc/eks/ecr-credential-provider/ecr-credential-provider-config
+fi
+
+################################################################################
+### Cache Images ###############################################################
+################################################################################
+if [[ "$CACHE_CONTAINER_IMAGES" == "true" && "$BINARY_BUCKET_REGION" != "us-iso-east-1" && "$BINARY_BUCKET_REGION" != "us-isob-east-1" ]]; then
+  AWS_DOMAIN=$(imds 'latest/meta-data/services/domain')
+  ECR_URI=$(/etc/eks/get-ecr-uri.sh "${BINARY_BUCKET_REGION}" "${AWS_DOMAIN}")
+
+  PAUSE_CONTAINER="${ECR_URI}/eks/pause:${PAUSE_CONTAINER_VERSION}"
+  cat /etc/eks/containerd/containerd-config.toml | sed s,SANDBOX_IMAGE,$PAUSE_CONTAINER,g | sudo tee /etc/eks/containerd/containerd-cached-pause-config.toml
+  sudo cp -v /etc/eks/containerd/containerd-cached-pause-config.toml /etc/containerd/config.toml
+  sudo cp -v /etc/eks/containerd/sandbox-image.service /etc/systemd/system/sandbox-image.service
+  sudo chown root:root /etc/systemd/system/sandbox-image.service
+  sudo systemctl daemon-reload
+  sudo systemctl start containerd
+  sudo systemctl enable containerd sandbox-image
+
+  K8S_MINOR_VERSION=$(echo "${KUBERNETES_VERSION}" | cut -d'.' -f1-2)
+  KUBE_PROXY_ADDON_VERSIONS=$(aws eks describe-addon-versions --addon-name kube-proxy --kubernetes-version=${K8S_MINOR_VERSION})
+
+  DEFAULT_KUBE_PROXY_FULL_VERSION=$(echo "${KUBE_PROXY_ADDON_VERSIONS}" | jq -r '.addons[] .addonVersions[] | select(.compatibilities[] .defaultVersion==true).addonVersion')
+  DEFAULT_KUBE_PROXY_VERSION=$(echo "${DEFAULT_KUBE_PROXY_FULL_VERSION}" | cut -d"-" -f1)
+  DEFAULT_KUBE_PROXY_PLATFORM_VERSION=$(echo "${DEFAULT_KUBE_PROXY_FULL_VERSION}" | cut -d"-" -f2)
+
+  LATEST_KUBE_PROXY_FULL_VERSION=$(echo "${KUBE_PROXY_ADDON_VERSIONS}" | jq -r '.addons[] .addonVersions[] .addonVersion' | sort -V | tail -n1)
+  LATEST_KUBE_PROXY_VERSION=$(echo "${LATEST_KUBE_PROXY_FULL_VERSION}" | cut -d"-" -f1)
+  LATEST_KUBE_PROXY_PLATFORM_VERSION=$(echo "${LATEST_KUBE_PROXY_FULL_VERSION}" | cut -d"-" -f2)
+
+  KUBE_PROXY_IMGS=(
+    ## Default kube-proxy images
+    "${ECR_URI}/eks/kube-proxy:${DEFAULT_KUBE_PROXY_VERSION}-${DEFAULT_KUBE_PROXY_PLATFORM_VERSION}"
+    "${ECR_URI}/eks/kube-proxy:${DEFAULT_KUBE_PROXY_VERSION}-minimal-${DEFAULT_KUBE_PROXY_PLATFORM_VERSION}"
+
+    ## Latest kube-proxy images
+    "${ECR_URI}/eks/kube-proxy:${LATEST_KUBE_PROXY_VERSION}-${LATEST_KUBE_PROXY_PLATFORM_VERSION}"
+    "${ECR_URI}/eks/kube-proxy:${LATEST_KUBE_PROXY_VERSION}-minimal-${LATEST_KUBE_PROXY_PLATFORM_VERSION}"
+  )
+
+  #### Cache VPC CNI images starting with the addon default version and the latest version
+  VPC_CNI_ADDON_VERSIONS=$(aws eks describe-addon-versions --addon-name vpc-cni --kubernetes-version=${K8S_MINOR_VERSION})
+  DEFAULT_VPC_CNI_VERSION=$(echo "${VPC_CNI_ADDON_VERSIONS}" | jq -r '.addons[] .addonVersions[] | select(.compatibilities[] .defaultVersion==true).addonVersion')
+  LATEST_VPC_CNI_VERSION=$(echo "${VPC_CNI_ADDON_VERSIONS}" | jq -r '.addons[] .addonVersions[] .addonVersion' | sort -V | tail -n1)
+  CNI_IMG="${ECR_URI}/amazon-k8s-cni"
+  CNI_INIT_IMG="${CNI_IMG}-init"
+  CNI_IMGS=(
+    ## Default VPC CNI Images
+    "${CNI_IMG}:${DEFAULT_VPC_CNI_VERSION}"
+    "${CNI_INIT_IMG}:${DEFAULT_VPC_CNI_VERSION}"
+
+    ## Latest VPC CNI Images
+    "${CNI_IMG}:${LATEST_VPC_CNI_VERSION}"
+    "${CNI_INIT_IMG}:${LATEST_VPC_CNI_VERSION}"
+  )
+
+  CACHED_IMGS=(
+    "${PAUSE_CONTAINER}"
+    ${KUBE_PROXY_IMGS[@]}
+    ${CNI_IMGS[@]}
+  )
+
+  for img in "${CACHED_IMGS[@]}"; do
+    ## only kube-proxy-minimal is vended for K8s 1.24+
+    if [[ "${img}" == *"kube-proxy:"* ]] && [[ "${img}" != *"-minimal-"* ]] && vercmp "${K8S_MINOR_VERSION}" gteq "1.24"; then
+      continue
+    fi
+    /etc/eks/containerd/pull-image.sh "${img}"
+  done
+
+  #### Tag the pulled down image for all other regions in the partition
+  for region in $(aws ec2 describe-regions --all-regions | jq -r '.Regions[] .RegionName'); do
+    for img in "${CACHED_IMGS[@]}"; do
+      regional_img="${img/$BINARY_BUCKET_REGION/$region}"
+      sudo ctr -n k8s.io image tag "${img}" "${regional_img}" || :
+      ## Tag ECR fips endpoint for supported regions
+      if [[ "${region}" =~ (us-east-1|us-east-2|us-west-1|us-west-2|us-gov-east-1|us-gov-east-2) ]]; then
+        regional_fips_img="${regional_img/.ecr./.ecr-fips.}"
+        sudo ctr -n k8s.io image tag "${img}" "${regional_fips_img}" || :
+        sudo ctr -n k8s.io image tag "${img}" "${regional_fips_img/-eksbuild.1/}" || :
+      fi
+      ## Cache the non-addon VPC CNI images since "v*.*.*-eksbuild.1" is equivalent to leaving off the eksbuild suffix
+      if [[ "${img}" == *"-cni"*"-eksbuild.1" ]]; then
+        sudo ctr -n k8s.io image tag "${img}" "${regional_img/-eksbuild.1/}" || :
+      fi
+    done
+  done
 fi
 
 ################################################################################
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 PACKER_BINARY ?= packer
-PACKER_VARIABLES := aws_region ami_name binary_bucket_name binary_bucket_region kubernetes_version kubernetes_build_date kernel_version docker_version containerd_version runc_version cni_plugin_version source_ami_id source_ami_owners source_ami_filter_name arch instance_type security_group_id additional_yum_repos pull_cni_from_github sonobuoy_e2e_registry ami_regions volume_type
+PACKER_VARIABLES := $(shell $(PACKER_BINARY) inspect -machine-readable eks-worker-al2.json | grep 'template-variable' | awk -F ',' '{print $$4}')
 
 K8S_VERSION_PARTS := $(subst ., ,$(kubernetes_version))
 K8S_VERSION_MINOR := $(word 1,${K8S_VERSION_PARTS}).$(word 2,${K8S_VERSION_PARTS})


cartermckinnon:
I'm interested in where this is headed, but what kind of speedup are we talking about? These images are fairly small, and they're coming from within the region. I'm seeing <5 seconds total for image pulls in us-west-2:
```
> sudo ctr --namespace k8s.io image pull 602401143452.dkr.ecr.us-west-2.amazonaws.com/eks/pause:3.1-eksbuild.1 --user AWS:$ECR_PASSWORD
602401143452.dkr.ecr.us-west-2.amazonaws.com/eks/pause:3.1-eksbuild.1:            resolved       |++++++++++++++++++++++++++++++++++++++| 
index-sha256:1cb4ab85a3480446f9243178395e6bee7350f0d71296daeb6a9fdd221e23aea6:    done           |++++++++++++++++++++++++++++++++++++++| 
manifest-sha256:234b8785dd78afc0fbb27edad009e7eb253e5685fb7387d4f0145f65c00873ac: done           |++++++++++++++++++++++++++++++++++++++| 
config-sha256:106a8e54d5eb3f70fcd1ed46255bdf232b3f169e89e68e13e4e67b25f59c1315:   done           |++++++++++++++++++++++++++++++++++++++| 
layer-sha256:41d8806bd3d23e1ffb7e9825fa56a0c2e851dfeeb405477ab1d6bc3a34bc0da2:    done           |++++++++++++++++++++++++++++++++++++++| 
elapsed: 0.4 s                                                                    total:  1.2 Ki (3.1 KiB/s)                                       
unpacking linux/amd64 sha256:1cb4ab85a3480446f9243178395e6bee7350f0d71296daeb6a9fdd221e23aea6...
done

> sudo ctr --namespace k8s.io image pull 602401143452.dkr.ecr.us-west-2.amazonaws.com/amazon-k8s-cni:v1.11.2 --user AWS:$ECR_PASSWORD
602401143452.dkr.ecr.us-west-2.amazonaws.com/amazon-k8s-cni:v1.11.2:              resolved       |++++++++++++++++++++++++++++++++++++++| 
index-sha256:9da3824d4b058462912d6e781714c4faa30be0a306cc48790cc042f51ca70651:    done           |++++++++++++++++++++++++++++++++++++++| 
manifest-sha256:5778a70db82f9ed9fb3ed3cb1d882f7211637c064589cd894c261759ea77265a: done           |++++++++++++++++++++++++++++++++++++++| 
layer-sha256:85f9b0d1a1fdbd43edbe8f98bbbf3b11d250fcfb05155e43fcb21d0d04822b0d:    done           |++++++++++++++++++++++++++++++++++++++| 
layer-sha256:8de5b65bd171294b1e04e0df439f4ea11ce923b642eddf3b3d76d297bfd2670c:    done           |++++++++++++++++++++++++++++++++++++++| 
layer-sha256:f18cfec559f5efc0bc336da4bb168e5a850082ad64e090db084e04f6cd65f7c8:    done           |++++++++++++++++++++++++++++++++++++++| 
config-sha256:4e9a8bf255bb85e99421a7108d8306ba0cacf43145478e5ab9d07e60ba9177ec:   done           |++++++++++++++++++++++++++++++++++++++| 
layer-sha256:1b862f5f9fa1a496dc2cb43edfeedd44e25a2c227d8b8796e223910853d3339b:    done           |++++++++++++++++++++++++++++++++++++++| 
elapsed: 2.8 s                                                                    total:  94.1 M (33.6 MiB/s)                                      
unpacking linux/amd64 sha256:9da3824d4b058462912d6e781714c4faa30be0a306cc48790cc042f51ca70651...
done

> sudo ctr --namespace k8s.io image pull 602401143452.dkr.ecr.us-west-2.amazonaws.com/amazon-k8s-cni-init:v1.11.2 --user AWS:$ECR_PASSWORD
602401143452.dkr.ecr.us-west-2.amazonaws.com/amazon-k8s-cni-init:v1.11.2:         resolved       |++++++++++++++++++++++++++++++++++++++| 
index-sha256:aa5bd2e6b21e7f4167e9db19c239f2288159ca171558ec3a60a7c5f9ce0650d3:    done           |++++++++++++++++++++++++++++++++++++++| 
manifest-sha256:c9fab251a20bf364315f54d4e43708e8a839d6e2d24156d70462beea0567e756: done           |++++++++++++++++++++++++++++++++++++++| 
layer-sha256:edc3ce4cb2326060f5e32631030f0ce65c688b745b40e51b0f18effb7fc2618a:    done           |++++++++++++++++++++++++++++++++++++++| 
config-sha256:5abf9331a3a190e0d8c5cde96d540e39b737e63e4d89d864694717c8ab9607eb:   done           |++++++++++++++++++++++++++++++++++++++| 
layer-sha256:8de5b65bd171294b1e04e0df439f4ea11ce923b642eddf3b3d76d297bfd2670c:    exists         |++++++++++++++++++++++++++++++++++++++| 
layer-sha256:e5481bb799cb068f202c75d6d276f186821d31b7783b5e3e8627362b1d4ced02:    done           |++++++++++++++++++++++++++++++++++++++| 
layer-sha256:85ab13e61b22b5dd72a416bcd42bf052d8a1d4565b3229b551043ba74343d7f5:    done           |++++++++++++++++++++++++++++++++++++++| 
elapsed: 0.8 s                                                                    total:  34.6 M (43.2 MiB/s)                                      
unpacking linux/amd64 sha256:aa5bd2e6b21e7f4167e9db19c239f2288159ca171558ec3a60a7c5f9ce0650d3...
done

> sudo ctr --namespace k8s.io image pull 602401143452.dkr.ecr.us-west-2.amazonaws.com/eks/kube-proxy:v1.22.6-eksbuild.1 --user AWS:$ECR_PASSWORD
602401143452.dkr.ecr.us-west-2.amazonaws.com/eks/kube-proxy:v1.22.6-eksbuild.1:   resolved       |++++++++++++++++++++++++++++++++++++++| 
index-sha256:c8abb4b8efc94090458f34e5f456791d9f7f57b5c99517b6b4e197305c1f10f6:    done           |++++++++++++++++++++++++++++++++++++++| 
manifest-sha256:0256f9a055e60fb40bccb5e2f743eb0eea13dcf38952058ceaba8e1fd1e6c096: done           |++++++++++++++++++++++++++++++++++++++| 
config-sha256:c8c9982c9d03789fe4d09993eaa54b11acd7b9bc6ebbfa60a223696172ca7507:   done           |++++++++++++++++++++++++++++++++++++++| 
layer-sha256:20b09fbd30377e1315a8bc9e15b5f8393a1090a7ec3f714ba5fce0c9b82a42f2:    done           |++++++++++++++++++++++++++++++++++++++| 
layer-sha256:8e20184c86519fabc2c1c075723074fc9c3fd780fb4c3fa2722b99db746b6db1:    done           |++++++++++++++++++++++++++++++++++++++| 
elapsed: 0.7 s                                                                    total:  30.6 M (43.6 MiB/s)                                      
unpacking linux/amd64 sha256:c8abb4b8efc94090458f34e5f456791d9f7f57b5c99517b6b4e197305c1f10f6...
done
```

Maybe this is more of a benefit for Outpost?

cartermckinnon:
I'm interested in where this is headed, but what kind of speedup are we talking about? These images are fairly small, and they're coming from within the region. I'm seeing <5 seconds total for image pulls in us-west-2:
```
> sudo ctr --namespace k8s.io image pull 602401143452.dkr.ecr.us-west-2.amazonaws.com/eks/pause:3.1-eksbuild.1 --user AWS:$ECR_PASSWORD
602401143452.dkr.ecr.us-west-2.amazonaws.com/eks/pause:3.1-eksbuild.1:            resolved       |++++++++++++++++++++++++++++++++++++++| 
index-sha256:1cb4ab85a3480446f9243178395e6bee7350f0d71296daeb6a9fdd221e23aea6:    done           |++++++++++++++++++++++++++++++++++++++| 
manifest-sha256:234b8785dd78afc0fbb27edad009e7eb253e5685fb7387d4f0145f65c00873ac: done           |++++++++++++++++++++++++++++++++++++++| 
config-sha256:106a8e54d5eb3f70fcd1ed46255bdf232b3f169e89e68e13e4e67b25f59c1315:   done           |++++++++++++++++++++++++++++++++++++++| 
layer-sha256:41d8806bd3d23e1ffb7e9825fa56a0c2e851dfeeb405477ab1d6bc3a34bc0da2:    done           |++++++++++++++++++++++++++++++++++++++| 
elapsed: 0.4 s                                                                    total:  1.2 Ki (3.1 KiB/s)                                       
unpacking linux/amd64 sha256:1cb4ab85a3480446f9243178395e6bee7350f0d71296daeb6a9fdd221e23aea6...
done

> sudo ctr --namespace k8s.io image pull 602401143452.dkr.ecr.us-west-2.amazonaws.com/amazon-k8s-cni:v1.11.2 --user AWS:$ECR_PASSWORD
602401143452.dkr.ecr.us-west-2.amazonaws.com/amazon-k8s-cni:v1.11.2:              resolved       |++++++++++++++++++++++++++++++++++++++| 
index-sha256:9da3824d4b058462912d6e781714c4faa30be0a306cc48790cc042f51ca70651:    done           |++++++++++++++++++++++++++++++++++++++| 
manifest-sha256:5778a70db82f9ed9fb3ed3cb1d882f7211637c064589cd894c261759ea77265a: done           |++++++++++++++++++++++++++++++++++++++| 
layer-sha256:85f9b0d1a1fdbd43edbe8f98bbbf3b11d250fcfb05155e43fcb21d0d04822b0d:    done           |++++++++++++++++++++++++++++++++++++++| 
layer-sha256:8de5b65bd171294b1e04e0df439f4ea11ce923b642eddf3b3d76d297bfd2670c:    done           |++++++++++++++++++++++++++++++++++++++| 
layer-sha256:f18cfec559f5efc0bc336da4bb168e5a850082ad64e090db084e04f6cd65f7c8:    done           |++++++++++++++++++++++++++++++++++++++| 
config-sha256:4e9a8bf255bb85e99421a7108d8306ba0cacf43145478e5ab9d07e60ba9177ec:   done           |++++++++++++++++++++++++++++++++++++++| 
layer-sha256:1b862f5f9fa1a496dc2cb43edfeedd44e25a2c227d8b8796e223910853d3339b:    done           |++++++++++++++++++++++++++++++++++++++| 
elapsed: 2.8 s                                                                    total:  94.1 M (33.6 MiB/s)                                      
unpacking linux/amd64 sha256:9da3824d4b058462912d6e781714c4faa30be0a306cc48790cc042f51ca70651...
done

> sudo ctr --namespace k8s.io image pull 602401143452.dkr.ecr.us-west-2.amazonaws.com/amazon-k8s-cni-init:v1.11.2 --user AWS:$ECR_PASSWORD
602401143452.dkr.ecr.us-west-2.amazonaws.com/amazon-k8s-cni-init:v1.11.2:         resolved       |++++++++++++++++++++++++++++++++++++++| 
index-sha256:aa5bd2e6b21e7f4167e9db19c239f2288159ca171558ec3a60a7c5f9ce0650d3:    done           |++++++++++++++++++++++++++++++++++++++| 
manifest-sha256:c9fab251a20bf364315f54d4e43708e8a839d6e2d24156d70462beea0567e756: done           |++++++++++++++++++++++++++++++++++++++| 
layer-sha256:edc3ce4cb2326060f5e32631030f0ce65c688b745b40e51b0f18effb7fc2618a:    done           |++++++++++++++++++++++++++++++++++++++| 
config-sha256:5abf9331a3a190e0d8c5cde96d540e39b737e63e4d89d864694717c8ab9607eb:   done           |++++++++++++++++++++++++++++++++++++++| 
layer-sha256:8de5b65bd171294b1e04e0df439f4ea11ce923b642eddf3b3d76d297bfd2670c:    exists         |++++++++++++++++++++++++++++++++++++++| 
layer-sha256:e5481bb799cb068f202c75d6d276f186821d31b7783b5e3e8627362b1d4ced02:    done           |++++++++++++++++++++++++++++++++++++++| 
layer-sha256:85ab13e61b22b5dd72a416bcd42bf052d8a1d4565b3229b551043ba74343d7f5:    done           |++++++++++++++++++++++++++++++++++++++| 
elapsed: 0.8 s                                                                    total:  34.6 M (43.2 MiB/s)                                      
unpacking linux/amd64 sha256:aa5bd2e6b21e7f4167e9db19c239f2288159ca171558ec3a60a7c5f9ce0650d3...
done

> sudo ctr --namespace k8s.io image pull 602401143452.dkr.ecr.us-west-2.amazonaws.com/eks/kube-proxy:v1.22.6-eksbuild.1 --user AWS:$ECR_PASSWORD
602401143452.dkr.ecr.us-west-2.amazonaws.com/eks/kube-proxy:v1.22.6-eksbuild.1:   resolved       |++++++++++++++++++++++++++++++++++++++| 
index-sha256:c8abb4b8efc94090458f34e5f456791d9f7f57b5c99517b6b4e197305c1f10f6:    done           |++++++++++++++++++++++++++++++++++++++| 
manifest-sha256:0256f9a055e60fb40bccb5e2f743eb0eea13dcf38952058ceaba8e1fd1e6c096: done           |++++++++++++++++++++++++++++++++++++++| 
config-sha256:c8c9982c9d03789fe4d09993eaa54b11acd7b9bc6ebbfa60a223696172ca7507:   done           |++++++++++++++++++++++++++++++++++++++| 
layer-sha256:20b09fbd30377e1315a8bc9e15b5f8393a1090a7ec3f714ba5fce0c9b82a42f2:    done           |++++++++++++++++++++++++++++++++++++++| 
layer-sha256:8e20184c86519fabc2c1c075723074fc9c3fd780fb4c3fa2722b99db746b6db1:    done           |++++++++++++++++++++++++++++++++++++++| 
elapsed: 0.7 s                                                                    total:  30.6 M (43.6 MiB/s)                                      
unpacking linux/amd64 sha256:c8abb4b8efc94090458f34e5f456791d9f7f57b5c99517b6b4e197305c1f10f6...
done
```

Maybe this is more of a benefit for Outpost?

cartermckinnon:
> with ECR's [3,000 TPS GetDownloadUrlForLayer limit](https://docs.aws.amazon.com/AmazonECR/latest/userguide/service-quotas.html#:~:text=Rate%20of-,GetDownloadUrlForLayer,-requests), a 300 node scale-up would result in throttling and likely a smaller scale-up since the actual images that need to be pulled would also be taking up that TPS

This limit is *per second*, though -- it seems unlikely that 300 nodes would attempt to download all 10 of these layers within the same second, right? `containerd` also only downloads 3 layers in parallel [by default](https://github.com/containerd/containerd/blob/5bc2d2e42989404a3ecdfdfe95120fb0d25f6305/pkg/cri/config/config_unix.go#L101). Users might configure a higher concurrency, but they can also request an increased quota from ECR, too.

Anyway, not against the general idea; we just need to make sure we're caching the "right" image(s) for this to benefit most users.

scripts/install-worker.sh
```diff
@@ -69,6 +71,13 @@ sudo yum install -y \
     yum-plugin-versionlock \
     yum-utils
 
+# Install AWS CLI v2
+AWSCLIv2_DIR=$(mktemp -d)
+curl -L "https://awscli.amazonaws.com/awscli-exe-linux-${MACHINE}.zip" -o "${AWSCLIv2_DIR}/awscliv2.zip"
+unzip "${AWSCLIv2_DIR}/awscliv2.zip" -d ${AWSCLIv2_DIR}
+sudo "${AWSCLIv2_DIR}/aws/install"
+sudo mv /usr/local/bin/aws /usr/local/bin/awsv2
```
cartermckinnon, line 79: Unfortunately, we can't do this; because the image has to be buildable in airgapped environments. (We're working on getting the CLI updated in the AL yum repos)

bwagner5:
> > with ECR's [3,000 TPS GetDownloadUrlForLayer limit](https://docs.aws.amazon.com/AmazonECR/latest/userguide/service-quotas.html#:~:text=Rate%20of-,GetDownloadUrlForLayer,-requests), a 300 node scale-up would result in throttling and likely a smaller scale-up since the actual images that need to be pulled would also be taking up that TPS
> 
> This limit is _per second_, though -- it seems unlikely that 300 nodes would attempt to download all 10 of these layers within the same second, right? `containerd` also only downloads 3 layers in parallel [by default](https://github.com/containerd/containerd/blob/5bc2d2e42989404a3ecdfdfe95120fb0d25f6305/pkg/cri/config/config_unix.go#L101). Users might configure a higher concurrency, but they can also request an increased quota from ECR, too.
> 
> Anyway, not against the general idea; we just need to make sure we're caching the "right" image(s) for this to benefit most users.

It might be unlikely to download them all in the same second but not impossible. As the scale outs get larger and faster (which is what we're trying to do with Karpenter), things will start to bottleneck.

scripts/install-worker.sh
```diff
@@ -328,6 +341,31 @@ if [[ ! $KUBERNETES_VERSION =~ "1.19"* && ! $KUBERNETES_VERSION =~ "1.20"* && !
     sudo mv $TEMPLATE_DIR/ecr-credential-provider-config /etc/eks/ecr-credential-provider/ecr-credential-provider-config
 fi
 
+################################################################################
+### Cache Images ##############################################################
+################################################################################
+AWS_DOMAIN=$(curl -s http://169.254.169.254/latest/meta-data/services/domain)
```
guessi, line 347: If SCP enforced [IMDSv2](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html) policy applied to the Org (or IAM policy enforced), it will fail to retrieve metadata from IMDSv1. Need to implement a logic to detect if it running with v1 or v2.

bwagner5: 👍 we should absolutely be using IMDSv2 everywhere we can. I believe there are some other locations that still use v1 in the building scripts. I'll take an action to update those with this PR.

bwagner5: https://github.com/awslabs/amazon-eks-ami/pull/1013

files/bootstrap.sh
```diff
@@ -586,26 +554,23 @@ if [[ "$CONTAINER_RUNTIME" = "containerd" ]]; then
 
     sudo mkdir -p /etc/containerd
     sudo mkdir -p /etc/cni/net.d
-    mkdir -p /etc/systemd/system/containerd.service.d
```
bwagner5, line 589: this is now setup in `install-worker.sh`

files/pull-image.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
```
bwagner5, line 1: copied from `pull-sandbox-image.sh` but made generic to use in install-worker for caching other images.

scripts/install-worker.sh
```diff
@@ -328,6 +359,64 @@ if [[ ! $KUBERNETES_VERSION =~ "1.19"* && ! $KUBERNETES_VERSION =~ "1.20"* && !
     sudo mv $TEMPLATE_DIR/ecr-credential-provider-config /etc/eks/ecr-credential-provider/ecr-credential-provider-config
 fi
 
+################################################################################
+### Cache Images ###############################################################
+################################################################################
+if [[ "$CACHE_CONTAINER_IMAGES" = "true" && "$BINARY_BUCKET_REGION" != "us-iso-east-1" && "$BINARY_BUCKET_REGION" != "us-isob-east-1" ]]; then
```
bwagner5, line 365: This results in the following images cached https://gist.github.com/bwagner5/e6415ec62e9eed3c864f39077346ce9f

files/pull-image.sh
```diff
@@ -0,0 +1,26 @@
...
+
+img=$1
+region=$(echo "${img}" | cut -f4 -d ".")
+ecr_password=$(aws ecr get-login-password --region $region)
```
cartermckinnon, line 5: can you add some retries to this? we have a couple reports of rate-limits on `get-login-password`:
- https://github.com/awslabs/amazon-eks-ami/issues/1034
- https://github.com/awslabs/amazon-eks-ami/issues/990#issuecomment-1215628507

scripts/install-worker.sh
```diff
@@ -111,6 +115,19 @@ sudo systemctl restart sshd.service
 sudo mkdir -p /etc/eks
 sudo mv $TEMPLATE_DIR/iptables-restore.service /etc/eks/iptables-restore.service
 
+################################################################################
+### installing updated cli #####################################################
```
cartermckinnon, line 119: I think we should do this in a separate PR. What's the dependency on it here?

scripts/install-worker.sh
```diff
@@ -151,11 +168,16 @@ fi
 ### Containerd setup ##########################################################
 ###############################################################################
 
+# install containerd and lock version
+# if docker is installed as well, this will run twice, but it's idempotent
+sudo yum install -y containerd-${CONTAINERD_VERSION}
+sudo yum versionlock containerd-*
```
cartermckinnon, line 174: any downside to moving the `runc` installation here too, and removing this: https://github.com/awslabs/amazon-eks-ami/pull/938/files#diff-dff0aa3a99c138c812c9364e64a6f95abec4590a27f0fb2b448231342b4a3b89R143-R149 ?

scripts/install-worker.sh
```diff
@@ -328,6 +359,64 @@ if [[ ! $KUBERNETES_VERSION =~ "1.19"* && ! $KUBERNETES_VERSION =~ "1.20"* && !
...
+    PAUSE_CONTAINER="${ECR_URI}/eks/pause:${PAUSE_CONTAINER_VERSION}"
+    KUBE_PROXY_IMG="${ECR_URI}/eks/kube-proxy:v${KUBERNETES_VERSION}-${KUBE_PROXY_VERSION_SUFFIX}"
+    KUBE_PROXY_MINIMAL_IMG="${ECR_URI}/eks/kube-proxy:v${KUBERNETES_VERSION}-minimal-${KUBE_PROXY_VERSION_SUFFIX}"
+    cat /etc/eks/containerd/containerd-config.toml | sed s,SANDBOX_IMAGE,$PAUSE_CONTAINER,g | sudo tee /etc/eks/containerd/containerd-cached-pause-config.toml
+    sudo cp -v /etc/eks/containerd/containerd-cached-pause-config.toml /etc/containerd/config.toml
+
+    sudo cp -v /etc/eks/containerd/sandbox-image.service /etc/systemd/system/sandbox-image.service
+    sudo chown root:root /etc/systemd/system/sandbox-image.service
+    sudo systemctl daemon-reload
+    sudo systemctl start containerd
```
cartermckinnon, line 378: why is containerd being started?

scripts/install-worker.sh
```diff
@@ -328,6 +359,64 @@ if [[ ! $KUBERNETES_VERSION =~ "1.19"* && ! $KUBERNETES_VERSION =~ "1.20"* && !
...
+    /etc/eks/containerd/pull-image.sh "${CNI_IMG}:${LATEST_VPC_CNI_VERSION}"
+    /etc/eks/containerd/pull-image.sh "${CNI_INIT_IMG}:${LATEST_VPC_CNI_VERSION}"
+    CNI_IMGS=("${CNI_IMG}:${DEFAULT_VPC_CNI_VERSION}" "${CNI_INIT_IMG}:${DEFAULT_VPC_CNI_VERSION}")
+    CNI_IMGS+=("${CNI_IMG}:${LATEST_VPC_CNI_VERSION}" "${CNI_INIT_IMG}:${LATEST_VPC_CNI_VERSION}")
+
+    CACHED_IMGS=("${PAUSE_CONTAINER}" "${KUBE_PROXY_IMG}" "${KUBE_PROXY_MINIMAL_IMG}")
+    CACHED_IMGS+=(${CNI_IMGS[@]})
+
+    #### Tag the pulled down image for all other regions in the partition
+    for region in $(aws ec2 describe-regions --all-regions | jq -r '.Regions[] .RegionName'); do
```
cartermckinnon, line 402: I assume `AWS_DEFAULT_REGION` determines the partition?

cartermckinnon (changes requested):

cartermckinnon:
Can you add the size of the image cache after this change?

scripts/install-worker.sh
bwagner5: `describe-add-on versions` is not in the aws-cli shipped currently

scripts/install-worker.sh
bwagner5: yep

scripts/install-worker.sh
bwagner5: to load the cached images into containerd

bwagner5:
> Can you add the size of the image cache after this change?

```
du -h /var/lib/containerd/io.containerd.snapshotter.v1.overlayfs
...
1.1G	/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/
```

scripts/install-worker.sh
bwagner5: I don't think so, just need to move on the containerd config above the docker install since docker depends on containerd and runc

files/pull-image.sh
bwagner5: sure!

scripts/install-worker.sh
cartermckinnon: oh duh, thanks 😂

scripts/install-worker.sh
cartermckinnon: yep, sounds good. In 1.24, we're disallowing docker as the runtime but leaving the packages installed; but this will lay the groundwork for us to remove it altogether in 1.25.

bwagner5: Moved and tested w/ both containerd and dockerd args to bootstrap script with caching enabled in the AMI and caching disabled in the AMI.

eks-worker-al2.json
```diff
@@ -22,8 +24,10 @@
     "kernel_version": "",
     "kms_key_id": "",
     "kubernetes_build_date": null,
+    "kube_proxy_version_suffix": "eksbuild.1",
     "kubernetes_version": null,
-    "launch_block_device_mappings_volume_size": "4",
+    "launch_block_device_mappings_volume_size": "8",
```
cartermckinnon, line 29: Could we get by with `6`? (or does it even matter?)

eks-worker-al2.json
```diff
@@ -8,11 +8,13 @@
     "arch": null,
     "associate_public_ip_address": "",
     "aws_access_key_id": "{{env `AWS_ACCESS_KEY_ID`}}",
+    "aws_cli_v2_version": "2.4.10",
```
cartermckinnon, line 11: nit: can we just call this `aws_cli_version`?

files/bootstrap.sh
```diff
@@ -586,26 +554,23 @@ if [[ "$CONTAINER_RUNTIME" = "containerd" ]]; then
...
     if [[ -n "$CONTAINERD_CONFIG_FILE" ]]; then
         sudo cp -v $CONTAINERD_CONFIG_FILE /etc/eks/containerd/containerd-config.toml
     fi
     echo "$(jq '.cgroupDriver="systemd"' $KUBELET_CONFIG)" > $KUBELET_CONFIG
     sudo sed -i s,SANDBOX_IMAGE,$PAUSE_CONTAINER,g /etc/eks/containerd/containerd-config.toml
-    sudo cp -v /etc/eks/containerd/containerd-config.toml /etc/containerd/config.toml
-    sudo cp -v /etc/eks/containerd/sandbox-image.service /etc/systemd/system/sandbox-image.service
+    # Check if the same PAUSE container was used when building the AMI
+    # If different, then restart containerd w/ proper sandbox config
+    if ! cmp -s /etc/eks/containerd/containerd-config.toml /etc/containerd/config.toml; then
```
cartermckinnon, line 564: can we make this check a little more specific to the pause image? could a user make unrelated changes to these file(s) and trigger this unintentionally?

files/pull-image.sh
cartermckinnon: this file is indented with tabs, can you convert to spaces? (we need to add some linting, note to self)

cartermckinnon (changes requested):
Couple small things.

The CLI change looks like the biggest risk here, but I'm fine with it. Pinging some others to get their 2 cents before we merge...

eks-worker-al2.json
```diff
@@ -22,8 +24,10 @@
     "kernel_version": "",
     "kms_key_id": "",
     "kubernetes_build_date": null,
+    "kube_proxy_version_suffix": "eksbuild.1",
     "kubernetes_version": null,
```
guessi, line 28: Can we ensure that `eksbuild.1` always exist?
I'll worry about a user set  `kubernetes_version ` but forget to change `kube_proxy_version_suffix`?

guessi: Take current state as example, it's now on `eksbuild.2`.
ref: https://docs.aws.amazon.com/eks/latest/userguide/managing-kube-proxy.html

files/bootstrap.sh
```diff
@@ -305,8 +273,8 @@ if [[ "$MACHINE" != "x86_64" && "$MACHINE" != "aarch64" ]]; then
     exit 1
 fi
 
-PAUSE_CONTAINER_ACCOUNT=$(get_pause_container_account_for_region "${AWS_DEFAULT_REGION}")
-PAUSE_CONTAINER_IMAGE=${PAUSE_CONTAINER_IMAGE:-$PAUSE_CONTAINER_ACCOUNT.dkr.ecr.$AWS_DEFAULT_REGION.$AWS_SERVICES_DOMAIN/eks/pause}
+ECR_URI=$(/etc/eks/get-ecr-uri.sh "${AWS_DEFAULT_REGION}" "${AWS_SERVICES_DOMAIN}" "${PAUSE_CONTAINER_ACCOUNT:-}")
```
guessi, line 276: Do we need the 3rd parameter here if we already have `/etc/eks/get-ecr-uri.sh` as helper function?

scripts/install-worker.sh
```diff
@@ -119,50 +123,39 @@ sudo mkdir -p /etc/eks
...
-    # Remove all options from sysconfig docker.
-    sudo sed -i '/OPTIONS/d' /etc/sysconfig/docker
-
-    sudo mkdir -p /etc/docker
-    sudo mv $TEMPLATE_DIR/docker-daemon.json /etc/docker/daemon.json
-    sudo chown root:root /etc/docker/daemon.json
-
-    # Enable docker daemon to start on boot.
-    sudo systemctl daemon-reload
+if [[ "$BINARY_BUCKET_REGION" != "us-iso-east-1" && "$BINARY_BUCKET_REGION" != "us-isob-east-1" ]]; then
```
guessi, line 130: not sure if there will have any exception, how about...

```
if echo ${BINARY_BUCKET_REGION} | grep -q -E 'us-iso[b]?-east-1'; then
  ...
fi
```

scripts/install-worker.sh
```diff
@@ -119,50 +123,39 @@ sudo mkdir -p /etc/eks
...
+    echo "Installing updated AWS CLI v2 v${AWS_CLI_VERSION}"
+    AWSCLI_DIR=$(mktemp -d)
+      --silent \
+      --show-error \
+      --retry 10 \
+      --retry-delay 1 \
+      -L "https://awscli.amazonaws.com/awscli-exe-linux-${MACHINE}-${AWS_CLI_VERSION}.zip" -o "${AWSCLI_DIR}/awscliv2.zip"
```
guessi, line 137: missing `curl` ?

files/pull-sandbox-image.sh
```diff
@@ -1,27 +1,5 @@
...
-	fi
-	if [[ $attempt -eq $API_RETRY_ATTEMPTS ]]; then
-        exit $rc
-    fi
-    jitter=$((1 + RANDOM % 10))
-    sleep_sec="$(( $(( 5 << $((1+$attempt)) )) + $jitter))"
-    sleep $sleep_sec
-done
+sandbox_image="$(awk -F'[ ="]+' '$1 == "sandbox_image" { print $2 }' /etc/containerd/config.toml)"
+/etc/eks/containerd/pull-image.sh "${sandbox_image}"
```
guessi, line 5: _( Discussion, I'm not sure about it )_

I can see there's a `set -e` at the very beginning and retry mechanism inside `pull-image.sh` , but there's no `set -e` in `pull-image.sh`.

I'm wondering what would happen if failed on last two lines of `pull-images.sh`, could `pull-sandbox-image.sh` detect that failure?

```
ecr_password=$(retry aws ecr get-login-password --region $region)
retry sudo ctr --namespace k8s.io image pull "${img}" --user AWS:${ecr_password}`
```

scripts/install-worker.sh
```diff
@@ -335,6 +365,64 @@ if [[ ! $KUBERNETES_VERSION =~ "1.19"* && ! $KUBERNETES_VERSION =~ "1.20"* && !
...
+    DEFAULT_VPC_CNI_VERSION=$(echo "${ADDON_VERSIONS}" | jq -r '.addons[] .addonVersions[] | select(.compatibilities[] .defaultVersion==true).addonVersion')
+    LATEST_VPC_CNI_VERSION=$(echo "${ADDON_VERSIONS}" | jq -r '.addons[] .addonVersions[] .addonVersion' | sort -V | tail -n1)
+    CNI_IMG="${ECR_URI}/amazon-k8s-cni"
+    CNI_INIT_IMG="${CNI_IMG}-init"
+    /etc/eks/containerd/pull-image.sh "${CNI_IMG}:${DEFAULT_VPC_CNI_VERSION}"
+    /etc/eks/containerd/pull-image.sh "${CNI_INIT_IMG}:${DEFAULT_VPC_CNI_VERSION}"
+    /etc/eks/containerd/pull-image.sh "${CNI_IMG}:${LATEST_VPC_CNI_VERSION}"
+    /etc/eks/containerd/pull-image.sh "${CNI_INIT_IMG}:${LATEST_VPC_CNI_VERSION}"
+    CNI_IMGS=("${CNI_IMG}:${DEFAULT_VPC_CNI_VERSION}" "${CNI_INIT_IMG}:${DEFAULT_VPC_CNI_VERSION}")
+    CNI_IMGS+=("${CNI_IMG}:${LATEST_VPC_CNI_VERSION}" "${CNI_INIT_IMG}:${LATEST_VPC_CNI_VERSION}")
```
guessi, line 402: How about format like this?

```
CNI_IMGS=(
  # we could put some comment here
  "${CNI_IMG}:${DEFAULT_VPC_CNI_VERSION}"
  "${CNI_IMG}:${LATEST_VPC_CNI_VERSION}"

  # we could put some comment here
  "${CNI_INIT_IMG}:${DEFAULT_VPC_CNI_VERSION}"
  "${CNI_INIT_IMG}:${LATEST_VPC_CNI_VERSION}"
)
```

I think it would be easier to maintain if there have additional images add later on ( minimize diff line at next PR )

bwagner5: that is a lot more readable :)

scripts/install-worker.sh
bwagner5: I kind of like spelling these out here. It's a little clearer to me since it's just two.

files/pull-sandbox-image.sh
bwagner5: if there's an error on retrieving the `ecr_password`, it would try the next line, but it would just retry and fail. Since the retry func exits with the proper status code for a failure, then it will be detected in `pull-sandbox-image.sh`.

guessi: yup... I saw it, no problem for me. 💯

files/bootstrap.sh
bwagner5: Updated comment

scripts/install-worker.sh
guessi: Just pop up in my head... a _**nice to have**_ enhancement,

Previously, it only require EC2 level permission, but now it required `eks:DescribeAddonVersions`.
Should it be mentioned in somewhere for extra permission?

ref: https://docs.aws.amazon.com/eks/latest/APIReference/API_DescribeAddonVersions.html

scripts/install-worker.sh
```diff
@@ -335,6 +366,75 @@ if [[ ! $KUBERNETES_VERSION =~ "1.19"* && ! $KUBERNETES_VERSION =~ "1.20"* && !
...
+    sudo systemctl daemon-reload
+    sudo systemctl start containerd
+    sudo systemctl enable containerd sandbox-image
+    /etc/eks/containerd/pull-image.sh "${PAUSE_CONTAINER}"
+    /etc/eks/containerd/pull-image.sh "${KUBE_PROXY_IMG}"
+    /etc/eks/containerd/pull-image.sh "${KUBE_PROXY_MINIMAL_IMG}"
+
+    #### Cache VPC CNI images starting with the addon default version and the latest version
+    K8S_MINOR_VERSION=$(echo "${KUBERNETES_VERSION}" | cut -d'.' -f1-2)
+    ADDON_VERSIONS=$(aws eks describe-addon-versions --addon-name vpc-cni --kubernetes-version=${K8S_MINOR_VERSION})
```
guessi, line 393: Do we need to ensure the output would be JSON format? ... `--output json`

bwagner5: json is the default output, but went ahead and set it explicitly at the top of the file.

scripts/install-worker.sh
```diff
@@ -335,6 +367,75 @@ if [[ ! $KUBERNETES_VERSION =~ "1.19"* && ! $KUBERNETES_VERSION =~ "1.20"* && !
...
+    sudo cp -v /etc/eks/containerd/containerd-cached-pause-config.toml /etc/containerd/config.toml
+
+    sudo cp -v /etc/eks/containerd/sandbox-image.service /etc/systemd/system/sandbox-image.service
+    sudo chown root:root /etc/systemd/system/sandbox-image.service
+    sudo systemctl daemon-reload
+    sudo systemctl start containerd
+    sudo systemctl enable containerd sandbox-image
+    /etc/eks/containerd/pull-image.sh "${PAUSE_CONTAINER}"
+    /etc/eks/containerd/pull-image.sh "${KUBE_PROXY_IMG}"
+    /etc/eks/containerd/pull-image.sh "${KUBE_PROXY_MINIMAL_IMG}"
```
guessi, line 390: Tested the script with `make 1.23` and I got this error message.

```
2022-10-16T00:37:32+08:00:     amazon-ebs: ‘/etc/eks/containerd/containerd-cached-pause-config.toml’ -> ‘/etc/containerd/config.toml’
2022-10-16T00:37:32+08:00:     amazon-ebs: ‘/etc/eks/containerd/sandbox-image.service’ -> ‘/etc/systemd/system/sandbox-image.service’
2022-10-16T00:37:33+08:00:     amazon-ebs: Created symlink from /etc/systemd/system/multi-user.target.wants/containerd.service to /usr/lib/systemd/system/containerd.service.
2022-10-16T00:37:33+08:00:     amazon-ebs: Created symlink from /etc/systemd/system/multi-user.target.wants/sandbox-image.service to /etc/systemd/system/sandbox-image.service.
2022-10-16T00:37:33+08:00:     amazon-ebs:
2022-10-16T00:37:33+08:00:     amazon-ebs: Unable to locate credentials. You can configure credentials by running "aws configure".
2022-10-16T00:37:48+08:00:     amazon-ebs: Attempt 1 of 3
2022-10-16T00:37:49+08:00:     amazon-ebs:
2022-10-16T00:37:49+08:00:     amazon-ebs: Unable to locate credentials. You can configure credentials by running "aws configure".
2022-10-16T00:38:18+08:00:     amazon-ebs: Attempt 2 of 3
2022-10-16T00:38:19+08:00:     amazon-ebs:
2022-10-16T00:38:19+08:00:     amazon-ebs: Unable to locate credentials. You can configure credentials by running "aws configure".
2022-10-16T00:39:02+08:00:     amazon-ebs: Attempt 3 of 3
2022-10-16T00:39:02+08:00:     amazon-ebs:
2022-10-16T00:39:02+08:00:     amazon-ebs: Unable to locate credentials. You can configure credentials by running "aws configure".
2022-10-16T00:39:02+08:00:     amazon-ebs: /etc/eks/containerd/pull-image.sh: line 15: return: exit: numeric argument required

( hang here, must hit CTRL-C to exit )
```

Seems like there are two issues need to resolve,

1. Instance have no Instance Profile associated, so there would have no ECR permission.
2. `$rc` does not capture as expected in `pull-image.sh`.

files/pull-image.sh
```diff
@@ -0,0 +1,23 @@
...
+        [[ $rc -eq 0 ]] && break
+        [[ $attempt -eq $MAX_RETRIES ]] && return exit $rc
+        local jitter=$((1 + RANDOM % 10))
+        local sleep_sec="$(( $(( 5 << $((1+$attempt)) )) + $jitter))"
+        sleep $sleep_sec
+    done
+}
+
+ecr_password=$(retry aws ecr get-login-password --region $region)
+retry sudo ctr --namespace k8s.io image pull "${img}" --user AWS:${ecr_password}
```
guessi, line 23: With no Instance Profile configured it will get error like this ( with debug mode ON )

```
2022-10-16T01:08:59+08:00:     amazon-ebs: Unable to locate credentials. You can configure credentials by running "aws configure".
2022-10-16T01:08:59+08:00:     amazon-ebs: ++ retry aws ecr get-login-password --region us-west-2
2022-10-16T01:08:59+08:00:     amazon-ebs: ++ local rc=0
2022-10-16T01:08:59+08:00:     amazon-ebs: +++ seq 0 3
2022-10-16T01:08:59+08:00:     amazon-ebs: ++ for attempt in '`seq 0 $MAX_RETRIES`'
2022-10-16T01:08:59+08:00:     amazon-ebs: ++ rc=0
2022-10-16T01:08:59+08:00:     amazon-ebs: ++ [[ 0 -gt 0 ]]
2022-10-16T01:08:59+08:00:     amazon-ebs: ++ aws ecr get-login-password --region us-west-2
2022-10-16T01:08:59+08:00:     amazon-ebs:
2022-10-16T01:08:59+08:00:     amazon-ebs: Unable to locate credentials. You can configure credentials by running "aws configure".
2022-10-16T01:08:59+08:00:     amazon-ebs: ++ rc=253
2022-10-16T01:08:59+08:00:     amazon-ebs: ++ [[ 253 -eq 0 ]]
2022-10-16T01:08:59+08:00:     amazon-ebs: ++ [[ 0 -eq 3 ]]
2022-10-16T01:08:59+08:00:     amazon-ebs: ++ local jitter=7
2022-10-16T01:08:59+08:00:     amazon-ebs: ++ local sleep_sec=17
2022-10-16T01:08:59+08:00:     amazon-ebs: ++ sleep 17
2022-10-16T01:09:16+08:00:     amazon-ebs: ++ for attempt in '`seq 0 $MAX_RETRIES`'
2022-10-16T01:09:16+08:00:     amazon-ebs: ++ rc=0
2022-10-16T01:09:16+08:00:     amazon-ebs: ++ [[ 1 -gt 0 ]]
2022-10-16T01:09:16+08:00:     amazon-ebs: ++ echo 'Attempt 1 of 3'
2022-10-16T01:09:16+08:00:     amazon-ebs: Attempt 1 of 3
2022-10-16T01:09:16+08:00:     amazon-ebs: ++ aws ecr get-login-password --region us-west-2
2022-10-16T01:09:17+08:00:     amazon-ebs:
2022-10-16T01:09:17+08:00:     amazon-ebs: Unable to locate credentials. You can configure credentials by running "aws configure".
2022-10-16T01:09:17+08:00:     amazon-ebs: ++ rc=253
2022-10-16T01:09:17+08:00:     amazon-ebs: ++ [[ 253 -eq 0 ]]
2022-10-16T01:09:17+08:00:     amazon-ebs: ++ [[ 1 -eq 3 ]]
2022-10-16T01:09:17+08:00:     amazon-ebs: ++ local jitter=2
2022-10-16T01:09:17+08:00:     amazon-ebs: ++ local sleep_sec=22
2022-10-16T01:09:17+08:00:     amazon-ebs: ++ sleep 22
2022-10-16T01:09:39+08:00:     amazon-ebs: ++ for attempt in '`seq 0 $MAX_RETRIES`'
2022-10-16T01:09:39+08:00:     amazon-ebs: ++ rc=0
2022-10-16T01:09:39+08:00:     amazon-ebs: ++ [[ 2 -gt 0 ]]
2022-10-16T01:09:39+08:00:     amazon-ebs: ++ echo 'Attempt 2 of 3'
2022-10-16T01:09:39+08:00:     amazon-ebs: Attempt 2 of 3
2022-10-16T01:09:39+08:00:     amazon-ebs: ++ aws ecr get-login-password --region us-west-2
2022-10-16T01:09:39+08:00:     amazon-ebs:
2022-10-16T01:09:39+08:00:     amazon-ebs: Unable to locate credentials. You can configure credentials by running "aws configure".
2022-10-16T01:09:40+08:00:     amazon-ebs: ++ rc=253
2022-10-16T01:09:40+08:00:     amazon-ebs: ++ [[ 253 -eq 0 ]]
2022-10-16T01:09:40+08:00:     amazon-ebs: ++ [[ 2 -eq 3 ]]
2022-10-16T01:09:40+08:00:     amazon-ebs: ++ local jitter=6
2022-10-16T01:09:40+08:00:     amazon-ebs: ++ local sleep_sec=46
2022-10-16T01:09:40+08:00:     amazon-ebs: ++ sleep 46
2022-10-16T01:10:26+08:00:     amazon-ebs: ++ for attempt in '`seq 0 $MAX_RETRIES`'
2022-10-16T01:10:26+08:00:     amazon-ebs: ++ rc=0
2022-10-16T01:10:26+08:00:     amazon-ebs: ++ [[ 3 -gt 0 ]]
2022-10-16T01:10:26+08:00:     amazon-ebs: ++ echo 'Attempt 3 of 3'
2022-10-16T01:10:26+08:00:     amazon-ebs: Attempt 3 of 3
2022-10-16T01:10:26+08:00:     amazon-ebs: ++ aws ecr get-login-password --region us-west-2
2022-10-16T01:10:26+08:00:     amazon-ebs:
2022-10-16T01:10:26+08:00:     amazon-ebs: Unable to locate credentials. You can configure credentials by running "aws configure".
2022-10-16T01:10:26+08:00:     amazon-ebs: ++ rc=253
2022-10-16T01:10:26+08:00:     amazon-ebs: ++ [[ 253 -eq 0 ]]
2022-10-16T01:10:26+08:00:     amazon-ebs: ++ [[ 3 -eq 3 ]]
2022-10-16T01:10:26+08:00:     amazon-ebs: ++ return exit 253
2022-10-16T01:10:26+08:00:     amazon-ebs: /etc/eks/containerd/pull-image.sh: line 17: return: exit: numeric argument required
2022-10-16T01:10:26+08:00:     amazon-ebs: + ecr_password=
2022-10-16T01:10:26+08:00:     amazon-ebs: + retry sudo ctr --namespace k8s.io image pull 602401143452.dkr.ecr.us-west-2.amazonaws.com/eks/pause:3.5 --user AWS:
2022-10-16T01:10:26+08:00:     amazon-ebs: + local rc=0
2022-10-16T01:10:26+08:00:     amazon-ebs: ++ seq 0 3
2022-10-16T01:10:26+08:00:     amazon-ebs: + for attempt in '`seq 0 $MAX_RETRIES`'
2022-10-16T01:10:26+08:00:     amazon-ebs: + rc=0
2022-10-16T01:10:26+08:00:     amazon-ebs: + [[ 0 -gt 0 ]]
2022-10-16T01:10:26+08:00:     amazon-ebs: + sudo ctr --namespace k8s.io image pull 602401143452.dkr.ecr.us-west-2.amazonaws.com/eks/pause:3.5 --user AWS:
```

Need to add a line for checking `$ecr_password` existence before trying to pull images.

```
[[ -z ${ecr_password} ]] && exit 1
```

Also, need to add `timeout XXX` for `ctr image pull ...` or it will hang here, can do nothing but CTRL-C.

guessi (changes requested):

scripts/install-worker.sh
```diff
@@ -335,6 +367,75 @@ if [[ ! $KUBERNETES_VERSION =~ "1.19"* && ! $KUBERNETES_VERSION =~ "1.20"* && !
...
+        "${CNI_INIT_IMG}:${LATEST_VPC_CNI_VERSION}"
+    )
+
+    CACHED_IMGS=(
+        "${PAUSE_CONTAINER}" 
+        "${KUBE_PROXY_IMG}" 
+        "${KUBE_PROXY_MINIMAL_IMG}"
+        ${CNI_IMGS[@]}
+    )
+
```
guessi, line 419: Looks like it could be simpler if we move image pull here?

```
for img in "${CACHED_IMGS[@]}"; do
    /etc/eks/containerd/pull-image.sh
done
```

cartermckinnon: Yeah, I agree the pulls could be moved into the same loop.

files/pull-image.sh
cartermckinnon: In addition to addressing: https://github.com/awslabs/amazon-eks-ami/pull/938/files/9e7c7686efb4cfc242db4da0ce7708ffbcaa9708#r998541356

I think adding the sanity check on `ecr_password` is a good idea. I don't know if adding a timeout on `ctr pull` is necessary, because the "hang" is happening because no password was specified in the `--user` value, and `ctr` will fallback to tty input in this case.

scripts/install-worker.sh
cartermckinnon: We should mention this in our "Building the AMI" docs, yeah.

cartermckinnon: i think we should move this: https://github.com/awslabs/amazon-eks-ami/pull/938/files#diff-dff0aa3a99c138c812c9364e64a6f95abec4590a27f0fb2b448231342b4a3b89R70

Into an `else` here. If we can install the updated CLI bundle, we can avoid installing all the global python dependencies for the yum package, e.g.:
```
> yum deplist awscli
                                                  
package: awscli.noarch 1.18.147-1.amzn2.0.1
  dependency: /usr/bin/python2
   provider: python.x86_64 2.7.18-1.amzn2.0.5
  dependency: PyYAML
   provider: PyYAML.x86_64 3.10-11.amzn2.0.2
  dependency: groff-base
   provider: groff-base.x86_64 1.22.2-8.amzn2.0.2
  dependency: python(abi) = 2.7
   provider: python.x86_64 2.7.18-1.amzn2.0.5
  dependency: python-colorama
   provider: python-colorama.noarch 0.3.2-3.amzn2
  dependency: python-docutils
   provider: python-docutils.noarch 0.12-0.2.20140510svn7747.amzn2
  dependency: python-rsa
   provider: python2-rsa.noarch 3.4.1-1.amzn2.0.1
  dependency: python2-botocore = 1.18.6
   provider: python2-botocore.noarch 1.18.6-1.amzn2.0.1
  dependency: python2-s3transfer
   provider: python2-s3transfer.noarch 0.3.3-1.amzn2.0.1
```

files/pull-image.sh
guessi: agree... it should only happen whenever `$ecr_password` is empty.
... if `$ecr_password` is retrieve from aws cli, do we need to check for other exception? I'm not sure about that.

files/pull-image.sh
```diff
@@ -0,0 +1,23 @@
...
+
+function retry() {
+    local rc=0
+    for attempt in `seq 0 $MAX_RETRIES`; do
+        rc=0
+        [[ $attempt -gt 0 ]] && echo "Attempt $attempt of $MAX_RETRIES" 1>&2
+        $@
+        rc=$?;
+        [[ $rc -eq 0 ]] && break
+        [[ $attempt -eq $MAX_RETRIES ]] && return exit $rc
```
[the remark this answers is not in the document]
bwagner5: whoops, good catch

files/pull-image.sh
bwagner5: Agreed on the empty ecr_password case. The empty case should be the only thing we need to guard though since any other input would result in a response, either success or failure for some reason (unauthorized, invalid, etc). I think the response itself would provide enough information for the user to fix it if they look in the right place.

scripts/install-worker.sh
bwagner5: moved awscli install to install only if the awscli v2 can't be installed.

bwagner5: I've added docs on the image caching including permissions for the AMI.

scripts/install-worker.sh
bwagner5: guarded against empty ecr_password

eks-worker-al2.json
```diff
@@ -8,11 +8,13 @@
     "arch": null,
     "associate_public_ip_address": "",
     "aws_access_key_id": "{{env `AWS_ACCESS_KEY_ID`}}",
+    "aws_cli_version": "2.4.10",
     "aws_region": "us-west-2",
     "aws_secret_access_key": "{{env `AWS_SECRET_ACCESS_KEY`}}",
     "aws_session_token": "{{env `AWS_SESSION_TOKEN`}}",
     "binary_bucket_name": "amazon-eks",
     "binary_bucket_region": "us-west-2",
+    "cache_container_images": "true",
```
cartermckinnon, line 17: I'm ready to ship this, thanks for all your work on this @bwagner5!

My only question is: should we default this to `false` to begin with? We can enable it for our official builds, but custom AMI users may want this to be opt-in, at least initially. @mmerkes you got an opinion here?

eks-worker-al2.json
```diff
@@ -8,11 +8,13 @@
     "arch": null,
     "associate_public_ip_address": "",
     "aws_access_key_id": "{{env `AWS_ACCESS_KEY_ID`}}",
+    "aws_cli_version": "2.4.10",
```
mmerkes, line 11: Why does this need to be set? Are you just forcibly be upgrading it?

eks-worker-al2.json
bwagner5: I'd be fine with it either way. 

On one side, I'd expect the official images to use mostly defaults for reproducible builds from users. But I can also understand wanting to be cautious with the permissions updates and the additional storage overhead.

mmerkes: Ideally, we should set this to false and allow feedback to come in. However, if we think it unlikely to actually get any feedback without enabling by default, the discussion is really just whether or not we're comfortable shipping this.

eks-worker-al2.json
bwagner5: This is set to match the current v2 CLI used. I don't see any reason to necessarily keep it at v2.4.10. But was just following the prior art. 

We could make it auto-update with no parameterization, although it may be useful to pin it depending on the CLI version release timings across partitions.

eks-worker-al2.json
bwagner5: I'll update it to default to false. I think the benefits will prompt users to enable this and we can gather feedback.

README.md
```diff
@@ -91,6 +91,102 @@ Provisioner](https://www.packer.io/docs/provisioners/shell.html) runs the
 necessary configuration tasks.  Then, Packer creates an AMI from the instance
 and terminates the instance after the AMI is created.
 
+### Container Image Caching
+
+By default, some container images are cached during the AMI build process in order to reduce the latency of the node getting to a `Ready` state when launched.
```
guessi, line 96: Should it be changed after `cache_container_images` default value changed from `true` to `false`?

Commit: https://github.com/awslabs/amazon-eks-ami/pull/938/commits/9158b7a47351f08ecb794e52823c46cd9428d5cd

README.md
```diff
@@ -91,6 +91,102 @@ Provisioner](https://www.packer.io/docs/provisioners/shell.html) runs the
...
 
+### Container Image Caching
+
+By default, some container images are cached during the AMI build process in order to reduce the latency of the node getting to a `Ready` state when launched. 
+
+To turn off container image caching:
+
+```
+cache_container_images=false make 1.23
+```
```
guessi, line 102: and here.

scripts/install-worker.sh
```diff
@@ -119,50 +123,42 @@ sudo mkdir -p /etc/eks
...
+    curl \
+      --silent \
+      --show-error \
+      --retry 10 \
+      --retry-delay 1 \
+      -L "https://awscli.amazonaws.com/awscli-exe-linux-${MACHINE}-${AWS_CLI_VERSION}.zip" -o "${AWSCLI_DIR}/awscliv2.zip"
+    unzip -q "${AWSCLI_DIR}/awscliv2.zip" -d ${AWSCLI_DIR}
+    sudo "${AWSCLI_DIR}/aws/install"
+else
+    yum install -y awscli
```
guessi, line 142: missing `sudo`?

bwagner5: oo good catch! thanks

eks-worker-al2.json
bwagner5: good with me, removed version parameter so we always pull the latest.

README.md
```diff
@@ -91,6 +91,102 @@ Provisioner](https://www.packer.io/docs/provisioners/shell.html) runs the
...
+When container image caching is enabled, the following images are cached:
+ - 602401143452.dkr.ecr.<AWS_REGION>.amazonaws.com/eks/kube-proxy:<K8s-Version>-eksbuild.1
+ - 602401143452.dkr.ecr.<AWS_REGION>.amazonaws.com/eks/kube-proxy:<K8s-Version>-minimal-eksbuild.1
+ - 602401143452.dkr.ecr.<AWS_REGION>.amazonaws.com/eks/pause:3.5
+ - 602401143452.dkr.ecr.<AWS_REGION>.amazonaws.com/amazon-k8s-cni-init:<default and latest>
+ - 602401143452.dkr.ecr.<AWS_REGION>.amazonaws.com/amazon-k8s-cni:<latest and latest>
+
+Since the VPC CNI is not versioned with K8s itself, the latest version of the VPC CNI and the default version, based on the response from the EKS DescribeAddonVersions at the time of the AMI build, will be cached. 
+
+The images listed above are also tagged with each region in the partition the AMI is built in, since images are often built in one region and copied to others within the same partition. Images that are available to pull from an ECR FIPS endpoint are also tagged as such (i.e. `602401143452.dkr.ecr-fips.us-east-1.amazonaws.com/eks/pause:3.5`).
```
guessi, line 113: To avoid confusion, maybe we should put reference link to https://docs.aws.amazon.com/eks/latest/userguide/add-ons-images.html to let reader aware of that `${AWS_ACCOUNT_ID}` for registry. Since it is not all referenced to `602401143452 `.

WDYT?

README.md
```diff
@@ -91,6 +91,102 @@ Provisioner](https://www.packer.io/docs/provisioners/shell.html) runs the
...
+            "Resource": "*"
+        },
+        {
+            "Effect": "Allow",
+            "Action": [
+                "ecr:BatchGetImage",
+                "ecr:BatchCheckLayerAvailability",
+                "ecr:GetDownloadUrlForLayer"
+            ],
+            "Resource": "arn:aws:ecr:us-west-2:602401143452:repository/*"
```
guessi, line 173: and here

files/get-ecr-uri.sh
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
```
guessi, line 3: maybe we could put a reference link here as well.
https://docs.aws.amazon.com/eks/latest/userguide/add-ons-images.html

scripts/install-worker.sh
guessi: nit, most of the string compare with syntax `[[ "$a" == "xxx" ]]` but here we use `[[ "$a" = "true" ]]`.

README.md
bwagner5: it is already mentioned below the policy.

guessi (approved):

scripts/install-worker.sh
```diff
@@ -119,50 +122,42 @@ sudo mkdir -p /etc/eks
...
+      --silent \
+      --show-error \
+      --retry 10 \
+      --retry-delay 1 \
+      -L "https://awscli.amazonaws.com/awscli-exe-linux-${MACHINE}.zip" -o "${AWSCLI_DIR}/awscliv2.zip"
+    unzip -q "${AWSCLI_DIR}/awscliv2.zip" -d ${AWSCLI_DIR}
+    sudo "${AWSCLI_DIR}/aws/install"
+else
+    sudo yum install -y awscli
 fi
```
guessi, line 142: nit... not sure if it would be a problem in the future.

for region `us-iso-east-1` and `us-isob-east-1`, command `aws` comes from `yum`. Both `aws` and `sudo aws` should works. But other regions, it will be different,

```
$ whoami
ec2-user

$ aws --version
aws-cli/2.8.4 Python/3.9.11 Linux/5.4.214-120.368.amzn2.x86_64 exe/x86_64.amzn.2 prompt/off

$ sudo aws --version
sudo: aws: command not found

$ which aws
/usr/local/bin/aws

$ sudo /usr/local/bin/aws --version
aws-cli/2.8.4 Python/3.9.11 Linux/5.4.214-120.368.amzn2.x86_64 exe/x86_64.amzn.2 prompt/off
```

cartermckinnon: hm, this seems weird. The install script puts a symlink in `/usr/local/bin` so there's no reason `root` wouldn't see it, unless i'm missing something.

bwagner5:
> My Point
> Aside from caching (which I attempted in my own setup as well, btw), I'd suggest maybe considering:
> 1. putting whatever is possible into a public, read-only ECR repo and skipping the credential wait times altogether
> 2. or at least decoupling the fetching of ECR credentials from the first-boot sequence of the node by utilizing the locally cached/warmed images, and fetching ECR credentials in a non-blocking/non-gating way, in the background somewhere.

This PR does mitigate the credential fetching by utilizing the cached pause image. The sandbox-image service still runs but it is not a blocking operation for containerd.

bwagner5:
> @bwagner5 - Thank you for the reply! I'll experiment with the referenced PR. Let me gather up some notes/things I had done in my repo, and see what might be relevant. :) Happy to share.
> 
> Additionally, you mentioned VPC CNI startup improvements - I'm kinda intrigued by this. Would you be able to point me in the direction of any info, or maybe share what you mean?
> 
> Thank you! :)

- There's this PR which speeds up VPC CNI by about 2 seconds just due to some unnecessary sleeping: https://github.com/aws/amazon-vpc-cni-k8s/pull/2104
- I'm working on speeding up the VPC CNI's init container to main container switch over latency which, in my tests, shows about a 3 second latency.  I'm experimenting with removing the init container and running it as a regular container in the pod and synchronizing the init stage with a shared emptyDir volume. This will allow the container start times to be parallelized and reduce any switching latency within the kubelet.

stevehipwell:
@bwagner5 I'm also interested in this discussion; could the VPC CNI start-up speed be addressed by using an eBPF pattern either directly or via chaining another CNI such as Cillium?

bwagner5:
@stevehipwell @armenr I started a group message with you both on K8s slack. We can continue discussion there. If anyone else is interested that stumbles across this issue, ping me on K8s slack (slack username: `brandon.wagner`)

maximethebault:
Startup latency is of major interest to our organization as well. Not sure I would bring much to the conversation, but being kept up-to-date of the different initiatives and "tuning tips" to make node startup faster is definitely something we'd appreciate.

bwagner5:
> Startup latency is of major interest to our organization as well. Not sure I would bring much to the conversation, but being kept up-to-date of the different initiatives and "tuning tips" to make node startup faster is definitely something we'd appreciate.

Created this issue to track and get feedback: https://github.com/awslabs/amazon-eks-ami/issues/1099

README.md
```diff
@@ -91,6 +91,104 @@ Provisioner](https://www.packer.io/docs/provisioners/shell.html) runs the
...
+Optionally, some container images can be cached during the AMI build process in order to reduce the latency of the node getting to a `Ready` state when launched. 
+
+To turn on container image caching:
+
+```
+cache_container_images=true make 1.23
+```
+
+When container image caching is enabled, the following images are cached:
+ - 602401143452.dkr.ecr.<AWS_REGION>.amazonaws.com/eks/kube-proxy:<K8s-Version>-eksbuild.1
```
stevehipwell, line 105: @bwagner5 don't you need to know which build to use here? The current correct versions to use for EKS v1.21 to v1.24 is `eksbuild.2`.

bwagner5: the caching logic uses the managed addons API (describe-addon-versions) to find the default and latest version. 

For kube-proxy on K8s 1.21 it uses:

```
                {
                    "addonVersion": "v1.21.14-eksbuild.2",
                    "architecture": [
                        "amd64",
                        "arm64"
                    ],
                    "compatibilities": [
                        {
                            "clusterVersion": "1.21",
                            "platformVersions": [
                                "*"
                            ],
                            "defaultVersion": false
                        }
                    ]
                },
...
                {
                    "addonVersion": "v1.21.2-eksbuild.2",
                    "architecture": [
                        "amd64",
                        "arm64"
                    ],
                    "compatibilities": [
                        {
                            "clusterVersion": "1.21",
                            "platformVersions": [
                                "*"
                            ],
                            "defaultVersion": true
                        }
                    ]
                },
```

stevehipwell: Ahhh, it might be clearer if you used the form `602401143452.dkr.ecr.<AWS_REGION>.amazonaws.com/eks/kube-proxy:<K8s-Version>-eksbuild.<BUILD_VERSION>` in the docs?

guessi (approved):

cartermckinnon (approved):

PR #939, opened 2022-06-07
Support worker nodes to continue to communicate to local cluster in Outposts even when the Outpost is disconnected from the parent AWS Region

chunywan:
Support worker nodes to continue to communicate to local cluster in Outposts even when the Outpost is disconnected from the parent AWS Region.

*Issue #, if available:*

*Description of changes:*
- How the "local cluster in outpost support" is determined?
   - If `--enable-local-outpost` option is be set with `true` or `false` explicitly, it will override the auto-detection of local cluster in the outpost.
   - If `--enable-local-outpost` option is not set and `aws eks describe-clsuter` is not bypassed [1], it will rely on whether the `outpostArn` is returned in the response of `aws eks describe-clsuter`.

- How does it support for worker nodes to continue to communicate and connect to the cluster even when the Outpost is disconnected from the AWS Region?
  - If "local cluster in outpost support" is turned on, it will
    - append entries to /etc/hosts with the mapping of control plane IP addresses to API server domain name so that the domain name can be resolved to IP addresses locally. This will work for both connected and disconnected states.
    - use `aws-iam-authenticator` as bootstrap authentication in kubelet bootstrap kubeconfig to download X.509 certificate which will be used in the generated kubelet kubeconfig. With X.509 certificate, worker node’s kubelet will be able to authenticate to API server during disconnected state.
 
[1] - The DescribeCluster call happens when `B64_CLUSTER_CA` or `APISERVER_ENDPOINT` aren't defined. Note that this PR doesn't change this behavior. 

*Testing:*
Add worker nodes to local outpost cluster successfully and the status is Ready:
```txt
$ kubectl get nodes
NAME                                       STATUS   ROLES                  AGE     VERSION
ip-10-0-2-124.us-west-2.compute.internal   Ready    control-plane,master   4h      v1.21.6
ip-10-0-2-199.us-west-2.compute.internal   Ready    control-plane,master   4h2m    v1.21.6
ip-10-0-2-247.us-west-2.compute.internal   Ready    control-plane,master   3h57m   v1.21.6
ip-10-0-2-136.us-west-2.compute.internal   Ready    <none>                 64s     v1.21.12-eks-5308cf7
ip-10-0-3-210.us-west-2.compute.internal   Ready    <none>                 82s     v1.21.12-eks-5308cf7
```

### DIFF
--- a/files/bootstrap.sh
+++ b/files/bootstrap.sh
@@ -404,9 +404,9 @@ if [[ -z "${B64_CLUSTER_CA}" ]] || [[ -z "${APISERVER_ENDPOINT}" ]]; then
     fi
     
     # If the cluster id is returned from describe cluster, let us use it no matter whether cluster id is passed from option
-	if [[ ! -z "${CLUSTER_ID_IN_DESCRIBE_CLUSTER_RESULT}" ]] && [[ "${CLUSTER_ID_IN_DESCRIBE_CLUSTER_RESULT}" != "None" ]]; then
-	    CLUSTER_ID=${CLUSTER_ID_IN_DESCRIBE_CLUSTER_RESULT}
-	fi
+    if [[ ! -z "${CLUSTER_ID_IN_DESCRIBE_CLUSTER_RESULT}" ]] && [[ "${CLUSTER_ID_IN_DESCRIBE_CLUSTER_RESULT}" != "None" ]]; then
+        CLUSTER_ID=${CLUSTER_ID_IN_DESCRIBE_CLUSTER_RESULT}
+    fi
 fi
 
 if [[ -z "${IP_FAMILY}" ]] || [[ "${IP_FAMILY}" == "None" ]]; then
--- a/files/bootstrap.sh
+++ b/files/bootstrap.sh
@@ -119,10 +119,10 @@ while [[ $# -gt 0 ]]; do
             shift
             ;;
          --cluster-id)
-	        CLUSTER_ID=$2
-	        shift
-	        shift
-	        ;;
+            CLUSTER_ID=$2
+            shift
+            shift
+            ;;
         *)    # unknown option
             POSITIONAL+=("$1") # save it in an array for later
             shift # past argument
@@ -445,10 +445,10 @@ if [[ "${ENABLE_LOCAL_OUTPOST}" == "true" ]]; then
     ###   - if "aws eks describe-cluster" is bypassed, for local outpost, the value of CLUSTER_NAME parameter will be cluster id.
     ###   - otherwise, the cluster id will use the id returned by "aws eks describe-cluster".
     if [[ -z "${CLUSTER_ID}" ]]; then
-	    echo "Cluster ID is required when local outpost support is enabled"
-	    exit 1
-	else
-	    sed -i s,CLUSTER_NAME,$CLUSTER_ID,g /var/lib/kubelet/kubeconfig
+        echo "Cluster ID is required when local outpost support is enabled"
+        exit 1
+    else
+        sed -i s,CLUSTER_NAME,$CLUSTER_ID,g /var/lib/kubelet/kubeconfig
 
         ### use aws-iam-authenticator as bootstrap auth and download X.509 cert used in kubelet kubeconfig
         mv /var/lib/kubelet/kubeconfig /var/lib/kubelet/bootstrap-kubeconfig
--- a/files/bootstrap.sh
+++ b/files/bootstrap.sh
@@ -30,7 +30,8 @@ function print_help {
     echo "--container-runtime Specify a container runtime (default: dockerd)"
     echo "--ip-family Specify ip family of the cluster"
     echo "--service-ipv6-cidr ipv6 cidr range of the cluster"
-    echo "--enable-local-outpost Enable support for worker nodes to communicate with the local control plane when running on a disconnected Outpost. (true or false)"    
+    echo "--enable-local-outpost Enable support for worker nodes to communicate with the local control plane when running on a disconnected Outpost. (true or false)"
+    echo "--cluster-id Specify the id of EKS cluster"
 }
 
 POSITIONAL=()
@@ -116,7 +117,12 @@ while [[ $# -gt 0 ]]; do
             ENABLE_LOCAL_OUTPOST=$2
             shift
             shift
-            ;; 
+            ;;
+         --cluster-id)
+	        CLUSTER_ID=$2
+	        shift
+	        shift
+	        ;;
         *)    # unknown option
             POSITIONAL+=("$1") # save it in an array for later
             shift # past argument
@@ -144,6 +150,7 @@ CONTAINER_RUNTIME="${CONTAINER_RUNTIME:-dockerd}"
 IP_FAMILY="${IP_FAMILY:-}"
 SERVICE_IPV6_CIDR="${SERVICE_IPV6_CIDR:-}"
 ENABLE_LOCAL_OUTPOST="${ENABLE_LOCAL_OUTPOST:-}"
+CLUSTER_ID="${CLUSTER_ID:-}"
 
 function get_pause_container_account_for_region () {
     local region="$1"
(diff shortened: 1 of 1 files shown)


cartermckinnon (approved):

files/bootstrap.sh
```diff
@@ -360,7 +367,7 @@ if [[ -z "${B64_CLUSTER_CA}" ]] || [[ -z "${APISERVER_ENDPOINT}" ]]; then
             --region=${AWS_DEFAULT_REGION} \
             --name=${CLUSTER_NAME} \
             --output=text \
-            --query 'cluster.{certificateAuthorityData: certificateAuthority.data, endpoint: endpoint, serviceIpv4Cidr: kubernetesNetworkConfig.serviceIpv4Cidr, serviceIpv6Cidr: kubernetesNetworkConfig.serviceIpv6Cidr, clusterIpFamily: kubernetesNetworkConfig.ipFamily}' > $DESCRIBE_CLUSTER_RESULT || rc=$?
+            --query 'cluster.{certificateAuthorityData: certificateAuthority.data, endpoint: endpoint, serviceIpv4Cidr: kubernetesNetworkConfig.serviceIpv4Cidr, serviceIpv6Cidr: kubernetesNetworkConfig.serviceIpv6Cidr, clusterIpFamily: kubernetesNetworkConfig.ipFamily, outpostArn: outpostConfig.outpostArns[0], id: id}' > $DESCRIBE_CLUSTER_RESULT || rc=$?
```
suket22, line 370: Is `outpostArn` being returned by DescribeCluster today? One issue we've seen in the past is that the version of the AWS CLI that's running on our AMIs isn't regularly updated so we might not pick up a new API parameter without a model change.  

Same comment for `id`. I don't [see it here](https://docs.aws.amazon.com/eks/latest/APIReference/API_DescribeCluster.html#API_DescribeCluster_ResponseSyntax). 

What this means is that the value might continue to remain `None` until we do a subsequent AMI release with an updated CLI. cc: @cartermckinnon

files/bootstrap.sh
```diff
@@ -137,6 +143,7 @@ PAUSE_CONTAINER_VERSION="${PAUSE_CONTAINER_VERSION:-3.1-eksbuild.1}"
 CONTAINER_RUNTIME="${CONTAINER_RUNTIME:-dockerd}"
 IP_FAMILY="${IP_FAMILY:-}"
 SERVICE_IPV6_CIDR="${SERVICE_IPV6_CIDR:-}"
+ENABLE_LOCAL_OUTPOST="${ENABLE_LOCAL_OUTPOST:-}"
```
suket22, line 146: I see this in the overview as well - `If --enable-local-outpost option is not set with true or false explicitly, bootstrap.sh will auto-detect whether the cluster is running in local outpost environment.`

Auto-discovery relies on making an EKS DescribeCluster call, and I don't want to do this for two reasons - 
* It forces a latency increase on every node launch. 
* It will be backwards incompatible for anyone running in a fully private VPC, since there's no private endpoints for EKS.

Why not default this to false? Local outposts isn't a common working configuration, so I'd expect this to run false for best backwards compatibility.

files/bootstrap.sh
```diff
@@ -360,7 +367,7 @@ if [[ -z "${B64_CLUSTER_CA}" ]] || [[ -z "${APISERVER_ENDPOINT}" ]]; then
             --region=${AWS_DEFAULT_REGION} \
             --name=${CLUSTER_NAME} \
             --output=text \
```
suket22, line 369: @cartermckinnon not for this PR, but I checked and our AMIs come with jq, so I wonder if we could get rid of `output=text` and just rely on jq rather than `awk` since the line split can be so easy to mess up.

cartermckinnon: Agreed!

files/bootstrap.sh
cartermckinnon: Yep, I discussed this with @chunywan. This will require an updated version of the CLI, which isn't yet available outside of an internal patched version, AFAIK. These changes are for a private beta, so I considered the can kicked.

files/bootstrap.sh
cartermckinnon: @chunywan can you clarify here? I think @suket22 brings up a good point.

This DescribeCluster call only happens when `B64_CLUSTER_CA` or `API_SERVER_ENDPOINT` aren't defined, but AFAIK these will always be defined for non-custom AMI nodegroups. The only time they wouldn't be defined is for a custom AMI nodegroup, in which case the user could just use the `--enable-local-outpost` flag. Am I missing something? I admittedly don't have much context on Outpost.

cartermckinnon: Regardless, this isn't changing when the DescribeCluster call happens, so I don't have any compatibility concerns. I'm just not sure the auto-detection mechanism is really useful.

chunywan: @suket22 Thanks for the comments. I think I need to refine the description of PR to make it clear.

> It forces a latency increase on every node launch.

This PR doesn't change whether the DescribeCluster is called or not. 
If it is called and `--enable-local-outpost` is not defined explicitly, the auto-discovery will be used. And the change is just add the filters to DescribeCluster call. 

> It will be backwards incompatible for anyone running in a fully private VPC, since there's no private endpoints for EKS.

As I mentioned above, if DescribeCluster is bypassed for private VPC (e.g., when B64_CLUSTER_CA or API_SERVER_ENDPOINT aren't defined) , this PR change doesn't call DescribeCluster either. 

So I think I will modify the description to make it clear.

chunywan: @suket22 @cartermckinnon 
I refined the description of the PR.

files/bootstrap.sh
chunywan: Yeah, this will make the code simple.

files/bootstrap.sh
chunywan: Yep, this change was tested with current public AWS CLI for backward compatibility. For forward compatibility, it was also tested with AWS CLI preview version (which includes `id` and `outpostArn` fields in response)

files/bootstrap.sh
```diff
@@ -389,9 +405,43 @@ fi
...
+if [[ "${ENABLE_LOCAL_OUTPOST}" == "true" ]]; then
+    ### append to /etc/hosts file with shuffled mappings of "IP address to API server domain name"
+    DOMAIN_NAME=$(echo "$APISERVER_ENDPOINT" | awk -F/ '{print $3}' | awk -F: '{print $1}')
+    getent hosts "$DOMAIN_NAME" | shuf >> /etc/hosts
+
+    ### kubelet bootstrap kubeconfig uses aws-iam-authenticator with cluster id to authenticate to cluster
+    ###   - if "aws eks describe-cluster" is bypassed, for local outpost, the value of CLUSTER_NAME parameter will be cluster id.
+    ###   - otherwise, the cluster id will use the id returned by "aws eks describe-cluster".
+    CLUSTER_ID="${CLUSTER_ID:-$CLUSTER_NAME}" 
+    sed -i s,CLUSTER_NAME,$CLUSTER_ID,g /var/lib/kubelet/kubeconfig
```
suket22, line 436: I don't understand this. So if someone runs `/etc/eks/bootstrap.sh $clusterName`, then we will attempt to auto discover, and `$CLUSTER_ID` will be set. We'll wire in the clusterId rather than clusterName in kubeConfig. 

But if someone runs `/etc/eks/bootstrap.sh --api-server-endpoint-url foo --cluster-ca-data bar --enable-local-outpost true $clusterName` then we won't autoDiscover and `clusterId` won't be used in the kubeConfig at all. Is that correct?

chunywan: Yes. For the second scenario (support local outpost and bypass DescribeClustger), the cluster ID instead of clusterName should be passed to bootstrap.sh.

files/bootstrap.sh
```diff
@@ -373,12 +387,26 @@ if [[ -z "${B64_CLUSTER_CA}" ]] || [[ -z "${APISERVER_ENDPOINT}" ]]; then
...
+    if [[ -z "${OUTPOST_ARN}" ]] || [[ "${OUTPOST_ARN}" == "None" ]]; then
+        IS_LOCAL_OUTPOST_DETECTED=false
+    else
+        IS_LOCAL_OUTPOST_DETECTED=true
+    fi
+    
+    # If the cluster id is returned from describe cluster, let us use it no matter whether cluster id is passed from option
+	if [[ ! -z "${CLUSTER_ID_IN_DESCRIBE_CLUSTER_RESULT}" ]] && [[ "${CLUSTER_ID_IN_DESCRIBE_CLUSTER_RESULT}" != "None" ]]; then
+	    CLUSTER_ID=${CLUSTER_ID_IN_DESCRIBE_CLUSTER_RESULT}
+	fi
```
suket22, line 409: I think the formatting on this might be off.

chunywan: Thanks. It is fixed by replacing tab with spaces.

PR #949, opened 2022-06-15
eks-log-collector.sh: add timeout to df command

sedflix:
*Description of changes:*

In some instances, `df` command doesn't work and therefore log collection fails. 
An example scenario, where df hangs is https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/616. 

This change will make sure that our log collector works for the above issue

### DIFF
--- a/log-collector-script/linux/eks-log-collector.sh
+++ b/log-collector-script/linux/eks-log-collector.sh
@@ -210,7 +210,7 @@ is_diskfull() {
 
   # 1.5GB in KB
   threshold=1500000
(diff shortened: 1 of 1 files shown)


sedflix:
@cartermckinnon @saurav-agarwalla it would be great if we would add reviewers to this pr.

cartermckinnon:
Seems reasonable to me.

FYI @tzneal not sure if a similar hang could happen in #978

tzneal (approved):
I suspect the same can occur with #978 , I'll update it.  I pulled this and tested it on one of my nodes and it worked fine.

Issue #1007, opened 2022-08-25
`bootstrap.sh` should support setting the `ecr` endpoint type (such as `ecr-fips`)


seanorama:
**What happened**:

In an environment which requires FIPS `./files/bootstrap.sh` used the standard `ecr` endpoint, not `ecr-fips`.

This can't be easily set in this code:

https://github.com/awslabs/amazon-eks-ami/blob/d9c8938f6666d3037d126ee906c2f133f2be0fb8/files/bootstrap.sh#L351

**What you expected to happen**:

Update bootstrap.sh with:

1. A variable named like `$PAUSE_CONTAINER_ENDPOINT_TYPE` which defaults to `ecr`.
2. A switch of `--pause-container-endpoint-type` which can set the variable.
3. Update the line above to use that variable.

**Environment**:
- AWS Region: us-gov-west-1
- EKS Platform version: eks.10
- Kubernetes version: 1.21

seanorama:
Here's a PR to solve this: #1012

bryantbiggs:
I believe this is now resolved with #1458 - cc @cartermckinnon 

Ref: https://github.com/awslabs/amazon-eks-ami/pull/1458/files#diff-2a573fe0e1a474595da4e9944a8e3c95aead7ab627514e3fc3a6f6702768f8d1R72-R78

PR #1012, opened 2022-08-31
add flag to set ecr endpoint (1007)

MrLane13:
**Issue #1007 :**

**Description of changes:**
Add a flag --pause-container-endpoint-type so that the ecr endpoint can be changed to support other environments, such as ecr-fips.

### DIFF
--- a/files/bootstrap.sh
+++ b/files/bootstrap.sh
@@ -93,6 +93,11 @@ while [[ $# -gt 0 ]]; do
             shift
             shift
             ;;
+        --pause-container-endpoint-type)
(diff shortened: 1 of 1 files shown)


stevehipwell:
We're currently doing this via `sed` so this change would be very useful.

cartermckinnon:
This is covered in #1028

stevehipwell:
@cartermckinnon I'm not sure #1028 will work in regions where a FIPS instance is wanted but the FIPS ECR endpoint still hasn't been provided.

PR #1017, opened 2022-09-09
Add note on minimum Packer version

cartermckinnon:


mmerkes (approved):

anish:
@cartermckinnon is there a specific feature in 1.8 that's required in the latest release ?

cartermckinnon:
This is mostly to standardize on the version we're using to build the official AMI, but if you try to build the HEAD against `1.6.0`, for example, you'll get errors related to the `SourceAMIName` build variable. We also saw issues with throttling that are addressed in `1.6.6`, IIRC: https://github.com/hashicorp/packer/blob/main/CHANGELOG.md#improvements-13

anish:
thanks for the info. we're on 1.7 and didn't see any failures, hence wondering if we actually needed an update.

mmerkes:
AFAIK, if it's builds, you're probably gtg. If it fails, upgrade packer.

PR #1022, opened 2022-09-15
Insignificance of  /etc/sysconfig/iptables

sumanthst25:
**Issue #, if available:**

**Description of changes:**
Removing the line 568 'bash -c "/sbin/iptables-save > /etc/sysconfig/iptables" in bootstrap.sh' as it is insignificant

**Testing Done**
The line 568 bash -c "/sbin/iptables-save > /etc/sysconfig/iptables" is insignificant.
Because kubelet is starting in 592 and then kube-proxy gets activated which manages the iptables rules dynamically in an EKS worker node. Because of this the /etc/sysconfig/iptables is empty.

I have tested this by creating AL2 worker nodes and when I run $cat /etc/sysconfig/iptables it returns nothing as it is empty.
Therefore the line can be removed as it is leading to confusion because it indicates the iptables rules should be present in /etc/sysconfig/iptables which is not happening in this case.
If I execute /sbin/iptables-save > /etc/sysconfig/iptables again manually in my worker node then I see there are iptables rule in /etc/sysconfig/iptables because kube-proxy should have started by then.

### DIFF
--- a/files/bootstrap.sh
+++ b/files/bootstrap.sh
@@ -565,7 +565,6 @@ EOF
 
 elif [[ "$CONTAINER_RUNTIME" = "dockerd" ]]; then
     mkdir -p /etc/docker
-    bash -c "/sbin/iptables-save > /etc/sysconfig/iptables"
     cp -v /etc/eks/iptables-restore.service /etc/systemd/system/iptables-restore.service
     sudo chown root:root /etc/systemd/system/iptables-restore.service
     systemctl daemon-reload


cartermckinnon:
`/etc/sysconfig/iptables` might not always be empty, because users might modify their iptables before `bootstrap.sh` executes. This line saves the current iptables, and the systemd unit `iptables-restore` later restores this after docker starts but before kubelet starts. I don't see any harm in leaving this in place, given that it does have a use case (even if it's rare).

Issue #1025, opened 2022-09-16
Remove Docker from AMI for Kubernetes version 1.24+


mmerkes:
**What would you like to be added**:
[Per this issue](https://github.com/awslabs/amazon-eks-ami/issues/1024), we're removing docker support for Kubernetes version 1.24+. As a follow up to that change, we should branch all docker-related logic to be ignored for 1.24+, including not installing it.

**Why is this needed**:
We can reduce the size of the AMI and the vulnerability surface area by removing unnecessary bits needed for supporting docker for 1.24+ AMIs. Once 1.23 is deprecated, we can remove all Docker log entirely.

abhinesh-nn:
Did we make any changes for not installing docker? The changes which are there are mostly related to ignoring docker-related changes.

PR #1027, opened 2022-09-16
Resolves #1024: makes containerd the default and only runtime for k8s…

### LINKED ISSUE #1024: Use containerd as the default container runtime for Kubernetes versions 1.24+

Use containerd as the default, and only, container runtime for Kubernetes versions 1.24+. To support that change, we'll have the following behavior in 1.24+ AMIs:

1. When no `--container-runtime` is set, we will default to `containerd`.
2. When the `--container-runtime` flag is set to `dockerd`, the bootstrap script will log an error and exit with a non-zero status.
3. Deprecated docker flags will be ignore, with a warning log, when set.
  * `--enable-docker-bridge`
  * `--docker-config-json`

**Why is this needed**:
As of Kubernetes version 1.24, [Dockershim will no longer be supported](https://kubernetes.io/blog/2021/11/12/are-you-ready-for-dockershim-removal/). To support that change, we're going to use `containerd` as the only supported container runtime for the `amazon-eks-ami`.

mmerkes:
… version 1.24+

**Issue #, if available:**
#1024

**Description of changes:**
Makes containerd as the default and only runtime for k8s versions 1.24+. If dockerd is passed in to the bootstrap script, it will log an error and fail. Docker-specific flags will be ignored.

As a note, I added [a separate issue](https://github.com/awslabs/amazon-eks-ami/issues/1025) to clean up all other docker installations, configurations, etc.

**Testing Done**

1. `make test` - I added multiple setups to test and verified that logs were as expected
2. `make 1.20`
    * Joined a node to a cluster with no runtime configured
    * Joined a node to a cluster with containerd configured
3. `make 1.24` (locally added only and need to use `client.authentication.k8s.io/v1beta1` as the API version)
    * Joined a node to a cluster with no runtime configured
    * Verified that nodes failed to bootstrap when container runtime set to dockerd

### DIFF
--- a/files/bootstrap.sh
+++ b/files/bootstrap.sh
@@ -135,18 +135,46 @@ set -- "${POSITIONAL[@]}" # restore positional parameters
 CLUSTER_NAME="$1"
 set -u
 
+KUBELET_VERSION=$(kubelet --version | grep -Eo '[0-9]\.[0-9]+\.[0-9]+')
+echo "Using kubelet version $KUBELET_VERSION"
+
+function is_greater_than_or_equal_to_version() {
+    local actual_version="$1"
+    local compared_version="$2"
+
+    [ $actual_version = "`echo -e \"$actual_version\n$compared_version\" | sort -V | tail -n1`" ]
+}
+
+# As of Kubernetes version 1.24, we will start defaulting the container runtime to containerd
+# and no longer support docker as a container runtime.
+IS_124_OR_GREATER=false
+DEFAULT_CONTAINER_RUNTIME=dockerd
+if is_greater_than_or_equal_to_version $KUBELET_VERSION "1.24.0"; then
+    IS_124_OR_GREATER=true
+    DEFAULT_CONTAINER_RUNTIME=containerd
+fi
+
+# Set container runtime related variables
+DOCKER_CONFIG_JSON="${DOCKER_CONFIG_JSON:-}"
+ENABLE_DOCKER_BRIDGE="${ENABLE_DOCKER_BRIDGE:-false}"
+CONTAINER_RUNTIME="${CONTAINER_RUNTIME:-$DEFAULT_CONTAINER_RUNTIME}"
+
+echo "Using $CONTAINER_RUNTIME as the container runtime"
+
+if $IS_124_OR_GREATER && [ $CONTAINER_RUNTIME != "containerd" ]; then
+    echo "ERROR: containerd is the only supported container runtime as of Kubernetes version 1.24"
+    exit 1
+fi
+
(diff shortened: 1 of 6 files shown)


files/bootstrap.sh
```diff
@@ -135,18 +135,56 @@ set -- "${POSITIONAL[@]}" # restore positional parameters
...
+
+# Set container runtime related variables
+DOCKER_CONFIG_JSON="${DOCKER_CONFIG_JSON:-}"
+ENABLE_DOCKER_BRIDGE="${ENABLE_DOCKER_BRIDGE:-false}"
+CONTAINER_RUNTIME="${CONTAINER_RUNTIME:-$DEFAULT_CONTAINER_RUNTIME}"
+
+echo "Using $CONTAINER_RUNTIME as the container runtime"
+
+if $IS_124_OR_GREATER; then
+    if [ $CONTAINER_RUNTIME != "containerd" ]; then
```
dewjam, line 165: Is there any reason not to use some of the existing conditions?

https://github.com/awslabs/amazon-eks-ami/blob/3a7d9744f11652cada6a1a29b878a9e3e21dbb37/files/bootstrap.sh#L604-L626

test/test-harness.sh
```diff
@@ -36,12 +36,15 @@ overall_status=0
...
+    # This variable is used to override the default value in the kubelet mock
+    KUBELET_VERSION="${KUBELET_VERSION:-}"
     cp -f ${SCRIPTPATH}/../files/kubelet-config.json ${temp_dir}/kubelet-config.json
     docker run -v ${SCRIPTPATH}/../files/:/etc/eks/ \
         -v ${temp_dir}/kubelet-config.json:/etc/kubernetes/kubelet/kubelet-config.json \
         --attach STDOUT \
         --attach STDERR \
         --rm \
+        -e KUBELET_VERSION="$KUBELET_VERSION" \
         eks-optimized-ami $@
```
cartermckinnon, line 48: We need to tweak the mocking strategy to allow individual test cases to supply their own mocks. Not sure we should address it in this PR, just a call out

files/bootstrap.sh
mmerkes: That's fair. The containerd warning logs are pertinent any time you're setting them with containerd as the runtime.

test/test-harness.sh
mmerkes: Ya, that's fair. Do you hate overwriting `kubelet` as part of the test as an alternative? That might be more sensible, but seems a little wrong in its own way.

mmerkes: I didn't actually address this. Definitely something to consider down the road.

cartermckinnon (approved):

Issue #1002, opened 2022-08-23
Add pre-built FIPS AMI


stevehipwell:
**What would you like to be added**:
I'd like to be able to use a pre-built FIPS version of the AL2 EKS AMI provided by Amazon; AFAIK the spec for this is described in this [blog](https://aws.amazon.com/blogs/publicsector/enabling-fips-mode-amazon-linux-2/). I think this is required for both AMD64 & ARM64 but AMD64 support would do for now. I'd suggest the AMI name prefix of `amazon-eks-fips-node-`.

**Why is this needed**:
Everyone needing FIPS nodes needs to do the exact same thing which is the definition of toil and it should be easy to automate this as part of the AMI release process.

seanorama:
+1 to this. I know of many who have built their own Packer pipelines for building FIPS into the image. Other Cloud providers have this by default or as an option.

seanorama:
There is already the STIG option, so adding FIPS should be easy in comparison: https://aws.amazon.com/blogs/containers/building-stig-compliant-amis-for-amazon-eks/

Other Cloud providers have this by default in their k8s node images, or as a switch as part of their k8s service.

This is a major point of confusion and extra work, especially considering the blog post states that EKS meets FedRAMP High, which implies it has FIPS: https://aws.amazon.com/about-aws/whats-new/2021/04/amazon-eks-is-now-fedramp-high-compliant/

stevehipwell:
I just spent most of the day Terraforming out the infrastructure to build an image for each EKS version for each architecture with EC2 ImageBuild. I think this is pretty much the definition of toil. And onboarding each account which needs to use the images will just add to that.

As an aside, the lack of an enable FIPS component for AL2 in ImageBuild seems like a major oversight. As does no support for source AMIs from SSM parameters.

seanorama:
And very related: #898

ghostsquad:
> I just spent most of the day Terraforming out the infrastructure to build an image for each EKS version for each architecture with EC2 ImageBuild. I think this is pretty much the definition of toil. And onboarding each account which needs to use the images will just add to that.

> As an aside, the lack of an enable FIPS component for AL2 in ImageBuild seems like a major oversight. As does no support for source AMIs from SSM parameters.

Would you be able to share what you've done to build these images?

ghostsquad:
[...]

PR #898, opened 2022-04-11
Adding FIPS 140-2 Support to EKS AMI

stanhu:
*Issue #, if available:*

*Description of changes:*

This adds support for enabling FIPS 140-2 mode in the Kernel. FIPS 140-2 is required by customers looking to achieve FedRAMP and/or DoD CC SRG compliance.

This brings up to date with the latest master.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1028, opened 2022-09-16
Add FIPS Support, with Autodetection of AWS ECR Endpoint

celestialorb:
Addresses #1002 

**Description of changes:**

Adds in FIPS compliance to the kernel (based on the work done in #898) and also adds in auto-detection of the AWS ECR endpoint to utilize based on whether or not FIPS is detected. Admittedly though, I feel the implementation there could be improved significantly.

**Testing Done**

No testing has been done as of yet, I'm throwing up this PR before heading out the door on travel.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 PACKER_BINARY ?= packer
-PACKER_VARIABLES := aws_region ami_name binary_bucket_name binary_bucket_region kubernetes_version kubernetes_build_date kernel_version docker_version containerd_version runc_version cni_plugin_version source_ami_id source_ami_owners source_ami_filter_name arch instance_type security_group_id additional_yum_repos pull_cni_from_github sonobuoy_e2e_registry ami_regions
+PACKER_VARIABLES := aws_region ami_name binary_bucket_name binary_bucket_region kubernetes_version kubernetes_build_date kernel_version docker_version containerd_version runc_version cni_plugin_version source_ami_id source_ami_owners source_ami_filter_name arch instance_type security_group_id additional_yum_repos pull_cni_from_github sonobuoy_e2e_registry ami_regions enable_fips_mode
 
 K8S_VERSION_PARTS := $(subst ., ,$(kubernetes_version))
 K8S_VERSION_MINOR := $(word 1,${K8S_VERSION_PARTS}).$(word 2,${K8S_VERSION_PARTS})
@@ -61,6 +61,23 @@ k8s: validate ## Build default K8s version of EKS Optimized AL2 AMI
 .PHONY: 1.23
 1.23: ## Build EKS Optimized AL2 AMI - K8s 1.23
 	$(MAKE) k8s kubernetes_version=1.23.9 kubernetes_build_date=2022-07-27 pull_cni_from_github=true
+
+# FIPS-mode enabled versions of EKS Optimized AL2 AMIs
+.PHONY: 1.20-fips
+1.20-fips: ## Build EKS Optimized AL2 AMI - K8s 1.20
+	$(MAKE) k8s kubernetes_version=1.20.15 kubernetes_build_date=2022-07-27 pull_cni_from_github=true enable_fips_mode=true
+
+.PHONY: 1.21-fips
+1.21-fips: ## Build EKS Optimized AL2 AMI - K8s 1.21
+	$(MAKE) k8s kubernetes_version=1.21.14 kubernetes_build_date=2022-07-27 pull_cni_from_github=true enable_fips_mode=true
+
+.PHONY: 1.22-fips
+1.22-fips: ## Build EKS Optimized AL2 AMI - K8s 1.22
+	$(MAKE) k8s kubernetes_version=1.22.12 kubernetes_build_date=2022-07-27 pull_cni_from_github=true enable_fips_mode=true
+
+.PHONY: 1.23-fips
+1.23-fips: ## Build EKS Optimized AL2 AMI - K8s 1.23
+	$(MAKE) k8s kubernetes_version=1.23.9 kubernetes_build_date=2022-07-27 pull_cni_from_github=true enable_fips_mode=true
 
 .PHONY: help
 help: ## Display help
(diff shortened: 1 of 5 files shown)


scripts/upgrade_kernel.sh
```diff
@@ -34,4 +34,17 @@ else
...
+  sudo yum install -y dracut-fips openssl
+  sudo dracut -f
+
+  # enable fips in the boot command
+  sudo sed -i 's/^\(GRUB_CMDLINE_LINUX_DEFAULT=.*\)"$/\1 fips=1"/' /etc/default/grub
+
+  # rebuild grub
+  sudo grub2-mkconfig -o /etc/grub2.cfg
+fi
+
```
cartermckinnon, line 49: I'd prefer to add a new provisioner to the packer spec instead of throwing this in `upgrade_kernel.sh`.

files/bootstrap.sh
```diff
@@ -343,8 +343,14 @@ if [[ "$MACHINE" != "x86_64" && "$MACHINE" != "aarch64" ]]; then
...
 
+AWS_ECR_SUBDOMAIN="ecr"
+# If FIPS is enabled on the machine, use the FIPS endpoint for AWS ECR.
+if grep "fips=1" /etc/default/grub; then
+    AWS_ECR_SUBDOMAIN="ecr-fips"
+fi
+
 PAUSE_CONTAINER_ACCOUNT=$(get_pause_container_account_for_region "${AWS_DEFAULT_REGION}")
-PAUSE_CONTAINER_IMAGE=${PAUSE_CONTAINER_IMAGE:-$PAUSE_CONTAINER_ACCOUNT.dkr.ecr.$AWS_DEFAULT_REGION.$AWS_SERVICES_DOMAIN/eks/pause}
+PAUSE_CONTAINER_IMAGE=${PAUSE_CONTAINER_IMAGE:-$PAUSE_CONTAINER_ACCOUNT.dkr.$AWS_ECR_SUBDOMAIN.$AWS_DEFAULT_REGION.$AWS_SERVICES_DOMAIN/eks/pause}
```
cartermckinnon, line 353: nice!

Makefile
```diff
@@ -62,6 +62,23 @@ k8s: validate ## Build default K8s version of EKS Optimized AL2 AMI
...
+	$(MAKE) k8s kubernetes_version=1.21.14 kubernetes_build_date=2022-07-27 pull_cni_from_github=true enable_fips_mode=true
+
+.PHONY: 1.22-fips
+1.22-fips: ## Build EKS Optimized AL2 AMI - K8s 1.22
+	$(MAKE) k8s kubernetes_version=1.22.12 kubernetes_build_date=2022-07-27 pull_cni_from_github=true enable_fips_mode=true
+
+.PHONY: 1.23-fips
+1.23-fips: ## Build EKS Optimized AL2 AMI - K8s 1.23
+	$(MAKE) k8s kubernetes_version=1.23.9 kubernetes_build_date=2022-07-27 pull_cni_from_github=true enable_fips_mode=true
+
```
cartermckinnon, line 81: These new make targets don't seem necessary; it's a lot of duplication just to have `make 1.22-fips` instead of `make 1.22 enable_fips_mode=true`.

cartermckinnon (changes requested):

files/bootstrap.sh
celestialorb: Correct me if I'm wrong, but I don't believe that value is available in the context of `bootstrap.sh`. `enable_fips_mode` is known at pack/build/make time and not when the node is joining the cluster. We could effectively pack the image with the value (this is what's being done here albeit indirectly), or autodetect whether or not FIPS is enabled (maybe from kernel parameters?)

scripts/enable_fips.sh
```diff
@@ -0,0 +1,15 @@
...
+  sudo dracut -f
+
+  # enable fips in the boot command
+  sudo sed -i 's/^\(GRUB_CMDLINE_LINUX_DEFAULT=.*\)"$/\1 fips=1"/' /etc/default/grub
+
+  # rebuild grub
+  sudo grub2-mkconfig -o /etc/grub2.cfg
+fi
+
+sudo reboot
```
mmerkes, line 15: Why are we rebooting regardless of if we make a change or not?

celestialorb: Copy+paste error, this was taken from the changes originally made to the `upgrade_kernel.sh` script which reboots at the end. I think it would be best to keep this reboot, remove the one in `upgrade_kernel.sh` and move the `enable_fips.sh` script to run just after the `upgrade_kernel.sh` script.

celestialorb: or maybe a separate provisioning step to perform the reboot would be cleaner?

mmerkes: We definitely don't want to keep this reboot, or it might do nothing and then reboot. Given that it's a FIPs specific script, we don't want to hide anything in there like the reboot. However, I don't have an objection to creating a task that just reboots and putting all things that require reboots before it. That probably makes sense if there's multiple steps where a reboot is needed.

files/bootstrap.sh
cartermckinnon: Yep, you're right; I was confused about where this line was. I think this detection method is fine.

eks-worker-al2.json
```diff
@@ -131,6 +131,10 @@
         "ENABLE_FIPS_MODE={{user `enable_fips_mode`}}",
       ]
     },
+    {
+      "type": "shell",
+      "inline": ["sudo reboot"]
```
cartermckinnon, line 136: We'll need `expect_disconnect` set to `true`, and we should move the `pause_before` from the next provisioner to a `pause_after` in this provisioner, if possible.

celestialorb: Done, and I also removed the `expect_disconnect` from the `upgrade_kernel.sh` provisioner as there's no longer a reboot residing in that script.

cartermckinnon:
This is implemented in #1458.

PR #1030, opened 2022-09-19
increase the kube-api-server QPS from 5/10 to 10/20

### LINKED ISSUE #969: Re-evaluate and increase QPS limits for kubelet

Currently, the [various QPS limits](https://kubernetes.io/docs/reference/config-api/kubelet-config.v1beta1/) for the Kubelet are not set which means they are configured to use the default values. We should re-evaluate if we can increase these limits for EKS Optimized AMIs to provide a better throughput for our customers.

**Why is this needed**:

This will provide more throughput to customers of EKS Optimized AMIs.

tzneal:
**Issue #, if available:**
Fixes #969

**Description of changes:**

This doubles the default API QPS values for EKS v1.22+ where API Priority & Fairness is available and there is a specific queue for kubelet health.

**Testing Done**

- Created an AMI with this change and an EKS 1.23 cluster
- Launched nodes using this AMI
- Verified nodes came up ready and that pods went ready
- Verified that the change was applied to the config file on the node
```
sh-4.2$ cat /etc/kubernetes/kubelet/kubelet-config.json  | grep -i kubeAPI
  "kubeAPIQPS": 10,
  "kubeAPIBurst": 20,
```

### DIFF
--- a/test/cases/api-qps-k8s-1.22-above.sh
+++ b/test/cases/api-qps-k8s-1.22-above.sh
@@ -1,0 +1,31 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+echo "--> Should increase API server QPS for K8s 1.22+"
+exit_code=0
+TEMP_DIR=$(mktemp -d)
+KUBELET_VERSION=v1.22.0-eks-ba74326
+run ${TEMP_DIR} /etc/eks/bootstrap.sh \
+    --b64-cluster-ca dGVzdA== \
+    --apiserver-endpoint http://my-api-endpoint \
+    test || exit_code=$?
+
+if [[ ${exit_code} -ne 0 ]]; then
+    echo "❌ Test Failed: expected a non-zero exit code but got '${exit_code}'"
+    exit 1
+fi
+
+expected_api_qps="10"
+expected_api_burst="20"
+
+actual_api_qps=$(jq -r '.kubeAPIQPS' < ${TEMP_DIR}/kubelet-config.json)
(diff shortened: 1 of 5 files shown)


tzneal:
In testing increasing from 5 QPS to 10 QPS effectively halves the overall time for 100 pods to become ready on a node from 67.5 to 32.8 seconds while also halving the time for the first pod to become ready from to 22 seconds to 10 second.

image

Any change in the kubelet QPS settings is scaled by the number of nodes in each cluster. This poses a risk that we can overwhelm the kube-api-server by increasing the amount of traffic received from kubelet running on each node.  The primary mitigation to this risk is that for K8s v1.20+ the [API Priority & Fairness](https://kubernetes.io/docs/concepts/cluster-administration/flow-control/) feature is enabled in EKS.  This added to Kubernetes the concept of priority levels where requests in different priority levels cannot starve each other. In v1.22+ specifically there is a separate queue for K8s health checks which will prevent increased load from pod churn from interfering with Kubelet health reporting.  The increase here is small while realizing most of the benefit, which further decreases negative impact risks.

The change is constructed to only update `kubeAPIQPS` and `kubeAPIBurst` if they have not already been configured in the kubelet config file which allows us to respect any customer changes to these files.

test/cases/api-qps-k8s-1.22-above.sh
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
```
bwagner5, line 1: nice!

bwagner5 (approved):
lgtm

PR #522, opened 2020-08-12
Feature flag the cleanup of the image

ericlarssen-wf:
*Description of changes:*
Put the cleanup steps behind a feature flag so that the AMI produced in this image can be used as a base image of other packer builds / forks of the packer build can have steps beyond the upstream packer build.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1036, opened 2022-09-28
Move cleanup to separate provisioner, remove cleanup_image var

cartermckinnon:
**Description of changes:**

Performing the cleanup steps within `install-worker.sh` is problematic, because they can only be skipped -- they can't be invoked separately from the rest of `install-worker.sh`. This is a pain for our GPU AMI template, for example.

The `cleanup_image` var was added to the packer template in #522. The value is passed to `install-worker.sh` as an environment variable, `CLEANUP_IMAGE`, which toggles the removal of various bits from the resulting AMI.

However, this variable doesn't actually work with our packer template; if you set its value to `false`, the build will fail. This is because `validate.sh` requires the cleanup steps to be run. This toggle would only make sense if you were using `install-worker.sh` in a custom packer template (which I assume is what folks are doing). However, I don't think it makes sense to allow cleanup to be disabled in the packer template that we publish, because `validate.sh` should never be optional in this template.

Removing the variable from our template won't break users who are using `install-worker.sh` in a custom template and setting `CLEANUP_IMAGE=false`. These users wouldn't include the new `cleanup.sh` provisioner; so there is no functional change.

This *could* break users who are directly modifying our packer template at build-time, such as setting `cleanup_image` to `false` and then deleting the `validate.sh` provisioner. This type of usage is inherently fragile and we can't always avoid breaking it. It shouldn't be difficult to resolve.

**Testing Done**

Before this change, this will fail:
```
# cleanup_image was never added to PACKER_VARIABLES in the Makefile, so just setting it here
> echo $(jq '.variables.cleanup_image = "false"' eks-worker-al2.json) > eks-worker-al2.json
> make 1.23
...
2022-09-28T14:57:52-07:00: ==> amazon-ebs: Provisioning with shell script: /var/folders/5t/c97ylshx62gb8ctzrmbtbv4h0000gs/T/tmp.AUBkWKrX/amazon-eks-ami/scripts/validate.sh
2022-09-28T14:57:52-07:00:     amazon-ebs: /etc/hostname shouldn't exists
...
```

After this change, this succeeds:
```
> make 1.23
```

### DIFF
--- a/scripts/install-worker.sh
+++ b/scripts/install-worker.sh
@@ -373,37 +373,4 @@ echo vm.max_map_count=524288 | sudo tee -a /etc/sysctl.conf
 sudo mkdir -p /etc/eks/log-collector-script/
 sudo cp $TEMPLATE_DIR/log-collector-script/eks-log-collector.sh /etc/eks/log-collector-script/
 
-################################################################################
-### Cleanup ####################################################################
-################################################################################
-
-CLEANUP_IMAGE="${CLEANUP_IMAGE:-true}"
-if [[ "$CLEANUP_IMAGE" == "true" ]]; then
-    # Clean up yum caches to reduce the image size
-    sudo yum clean all
-    sudo rm -rf \
-        $TEMPLATE_DIR  \
-        /var/cache/yum
-
-    # Clean up files to reduce confusion during debug
-    sudo rm -rf \
-        /etc/hostname \
-        /etc/machine-id \
-        /etc/resolv.conf \
-        /etc/ssh/ssh_host* \
-        /home/ec2-user/.ssh/authorized_keys \
-        /root/.ssh/authorized_keys \
-        /var/lib/cloud/data \
-        /var/lib/cloud/instance \
-        /var/lib/cloud/instances \
-        /var/lib/cloud/sem \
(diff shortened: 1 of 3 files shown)


cartermckinnon:
@ericlarssen-wf if you can add details about how you've used `cleanup_image`, it'd be helpful.

scripts/install-worker.sh
```diff
@@ -373,37 +373,4 @@ echo vm.max_map_count=524288 | sudo tee -a /etc/sysctl.conf
 sudo mkdir -p /etc/eks/log-collector-script/
 sudo cp $TEMPLATE_DIR/log-collector-script/eks-log-collector.sh /etc/eks/log-collector-script/
 
-################################################################################
-### Cleanup ####################################################################
-################################################################################
-
-CLEANUP_IMAGE="${CLEANUP_IMAGE:-true}"
-if [[ "$CLEANUP_IMAGE" == "true" ]]; then
```
mmerkes, line 381: Is this an important thing?

mmerkes: NVM, looks like you address in your description.

mmerkes (approved):

PR #1044, opened 2022-10-11
Add IMDS helper

cartermckinnon:
**Description of changes:**

Adds `/usr/bin/imds` to handle the details of making IMDS requests.

**Testing Done**

Built a 1.23 AMI (`make 1.23`) and then created a cluster with it:
```
---
kind: ClusterConfig
metadata:
  name: imds-helper
  region: us-west-2
  version: "1.23"
managedNodeGroups:
  - name: nodes
    ami: ami-0a4cae360f78ad715
    ssh:
      allow: true
    minSize: 1
    maxSize: 1
    desiredCapacity: 1
    overrideBootstrapCommand: |
      #!/bin/bash
      source /var/lib/cloud/scripts/eksctl/bootstrap.helper.sh      
      /etc/eks/bootstrap.sh imds-helper \
      --kubelet-extra-args "--node-labels=${NODE_LABELS}"
```

Nodes bootstrapped, pods came up. I SSH'd into the node and did:
```
> imds /latest/meta-data/ami-id
ami-0a4cae360f78ad715
```


bwagner5 (commented):
nice! I like it!

test/test-harness.sh
```diff
@@ -34,20 +34,13 @@ docker build -t eks-optimized-ami -f "${SCRIPTPATH}/Dockerfile" "${SCRIPTPATH}/.
...
-        -v ${temp_dir}/kubelet-config.json:/etc/kubernetes/kubelet/kubelet-config.json \
+        -v "$(realpath $1):/test.sh" \
         --attach STDOUT \
         --attach STDERR \
         --rm \
-        -e KUBELET_VERSION="$KUBELET_VERSION" \
-        eks-optimized-ami $@
+        eks-optimized-ami
 }
-export -f run
```
cartermckinnon, line 44: This was getting a bit unwieldy; I think it's more straightforward to run the test script inside the container (and it makes assertions in the test scripts a lot simpler, without requiring volume mounts).

It also gives us an easier way to accomplish mocking on a per-test basis, instead of having a set of mocks that are used for every tests. For example, a test case that needs to mock kubelet could do:
```
> cat my-test-case.sh

function kubelet() {
  echo "my expected output"
  return $MY_EXPECTED_EXIT_CODE
}
export -f kubelet
/command/that/uses/kubelet --arg
```

cartermckinnon:
@bwagner5 FYI I had to rework the test harness a bit to avoid the type of workaround we had for `kubelet-config.json` assertions.

bin/imds
```diff
@@ -0,0 +1,49 @@
...
+}
+
+if [ "$#" -ne 1 ]
+then
+  log "usage: imds API_PATH"
+  exit 1
+fi
+
+# leading slashes will be removed
+API_PATH="${1#/}"
```
cartermckinnon, line 22: some usages were like `/latest/meta-data` and some were like `latest/meta-data`. The real IMDS service handles requests like `GET //latest/meta-data`, but the mock service we use in unit testing doesn't. 🥴

bwagner5 (approved):
nice! LGTM

umarizulkifli:
do we have imds bins for ubuntu?

PR #1046, opened 2022-10-11
cri: containerd for cluster required to be on kubelet<v1.24

EmagIm:
**Issue #, if available:*NA*

**Description of changes:*Changes to AMI packer scripts to build image that uses containerd as cri by default and doesn't install docker by default*

**Testing Done*Built AMI image; decoupling docker and using containerd as the default cri. On clusters that have to be on version < 1.24 for dependencies, CRs, etc.*

- Tested built image creating new cluster with kubenertes v1.22
- Tested built image creating a managed nodegroup with cluster kubernetes v1.22

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -72,7 +72,7 @@ kind: ClusterConfig
 
 metadata:
   name: basic-cluster
-  region: us-west-2
+  region: us-east-1
   version: '1.22'
 
 nodeGroups:
--- a/eks-worker-al2.json
+++ b/eks-worker-al2.json
@@ -1,6 +1,6 @@
 {
   "variables": {
-    "aws_region": "us-west-2",
+    "aws_region": "us-east-1",
     "ami_name": null,
     "creator": "{{env `USER`}}",
     "encrypted": "false",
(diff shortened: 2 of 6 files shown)


cartermckinnon:
This PR doesn't make sense to me; we've already merged changes that will cause `containerd` to be the default (and only allowed) runtime for Kubernetes 1.24+. At this time, we are not removing the installation of `docker`, to ease the transition to `containerd` as the runtime. We'll likely disable `docker` installation for the 1.25 release.

AbeOwlu:
@cartermckinnon , I understand and agree the way I pulled this request doesn't make sense, to merge this change into main upstream. mea culpa, it was late. Maybe a branch would be better for this -

The defaults currently check for cri based on kubelet version at or above and below 1.24, yes. However on versions below 1.24, docker installs by default automatically. The usecase here is for infra that has requirements to stay on versions below 1.24 but wants to migrate to containerd as the runtime interface.

Not aware there is a way to remove that config on host instance boot with an AMI, or is there? 
Currently have to manually remove docker on every node.

cartermckinnon:
If you'd like to use `containerd` as the runtime, you can use the `--container-runtime` flag to `bootstrap.sh`. More here: https://docs.aws.amazon.com/eks/latest/userguide/eks-optimized-ami.html#containerd-bootstrap

In most cases, it doesn't matter if the `docker` package is still installed on your nodes; but you can always build a custom AMI that excludes it.

PR #1049, opened 2022-10-13
configure containerd registry certificates by default

steved:
**Issue #, if available:**

_N/A_

**Description of changes:**

By default, Docker used to allow placing registry certificates on the host filesystem under `/etc/docker/certs.d`. Containerd did not support this until 1.5 and as of 1.6, it is still not enabled by default. There are some discussions around enabling it in 1.7 (https://github.com/containerd/containerd/issues/6485), but it would be helpful to have this enabled in EKS AMIs as well.

**Testing Done**

Created an EKS 1.23 cluster with containerd as the runtime. Using a docker registry deployed within the cluster with a self-signed cert, dynamically placed a CA into /etc/docker/certs.d and confirmed that a pod referencing that registry worked.

### DIFF
--- a/files/bootstrap.sh
+++ b/files/bootstrap.sh
@@ -540,6 +540,8 @@ EOF
   sudo cp -v /etc/eks/containerd/kubelet-containerd.service /etc/systemd/system/kubelet.service
   sudo chown root:root /etc/systemd/system/kubelet.service
   sudo chown root:root /etc/systemd/system/sandbox-image.service
+  # Validate containerd config
+  sudo containerd config dump > /dev/null
   systemctl daemon-reload
   systemctl enable containerd
   systemctl restart containerd
--- a/test/cases/container-runtime-defaults.sh
+++ b/test/cases/container-runtime-defaults.sh
@@ -2,7 +2,6 @@
 set -euo pipefail
 
 exit_code=0
-TEMP_DIR=$(mktemp -d)
 
 echo "--> Should allow dockerd as container runtime when below k8s version 1.24"
 # This variable is used to override the default value in the kubelet mock
--- a/files/containerd-config.toml
+++ b/files/containerd-config.toml
@@ -11,6 +11,9 @@ default_runtime_name = "runc"
 [plugins."io.containerd.grpc.v1.cri"]
 sandbox_image = "SANDBOX_IMAGE"
 
+[plugins."io.containerd.grpc.v1.cri".registry]
+config_path = "/etc/containerd/certs.d:/etc/docker/certs.d"
+
 [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc]
 runtime_type = "io.containerd.runc.v2"
 
--- a/test/Dockerfile
+++ b/test/Dockerfile
@@ -1,17 +1,19 @@
 FROM public.ecr.aws/aws-ec2/amazon-ec2-metadata-mock:v1.11.2 as aemm
 FROM public.ecr.aws/amazonlinux/amazonlinux:2
-RUN yum install -y jq && \
-    yum install -y wget && \
+RUN amazon-linux-extras enable docker && \
+    yum install -y jq containerd wget && \
     wget -qO /usr/local/bin/yq https://github.com/mikefarah/yq/releases/latest/download/yq_linux_amd64 && \
     chmod a+x /usr/local/bin/yq
 
 ENV IMDS_ENDPOINT=127.0.0.1:1338
 COPY --from=aemm /ec2-metadata-mock /sbin/ec2-metadata-mock
+RUN mkdir -p /etc/eks/containerd
+COPY files/ /etc/eks/
+COPY files/containerd-config.toml files/kubelet-containerd.service files/pull-sandbox-image.sh files/sandbox-image.service /etc/eks/containerd/
 COPY files/kubelet-config.json /etc/kubernetes/kubelet/kubelet-config.json
 COPY files/kubelet-kubeconfig /var/lib/kubelet/kubeconfig
 COPY files/ecr-credential-provider-config /etc/eks/ecr-credential-provider/ecr-credential-provider-config
 COPY test/entrypoint.sh /entrypoint.sh
-COPY files /etc/eks
 COPY files/bin/* /usr/bin/
 COPY test/mocks/ /sbin/
 ENTRYPOINT ["/entrypoint.sh"]
--- a/test/cases/containerd-config.sh
+++ b/test/cases/containerd-config.sh
@@ -1,0 +1,35 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+exit_code=0
+
+echo "--> Default containerd config file should be valid"
+STDERR_FILE=$(mktemp)
+containerd -c /etc/eks/containerd/containerd-config.toml config dump > /dev/null 2> "$STDERR_FILE" || exit_code=$?
+
+if [[ ${exit_code} -ne 0 ]]; then
+  echo "❌ Test Failed: default containerd config file is invalid! $(cat "$STDERR_FILE")"
+  exit 1
+fi
+
+echo "--> Should fail when given an invalid containerd config"
+CONTAINERD_TOML=$(mktemp containerd-XXXXX.toml)
+cat > "$CONTAINERD_TOML" << EOF
+[cgroup]
+path = "foo"
+[cgroup]
+path = "bar"
+EOF
+
+export KUBELET_VERSION=v1.24.15-eks-ba74326
+/etc/eks/bootstrap.sh \
+  --b64-cluster-ca dGVzdA== \
+  --apiserver-endpoint http://my-api-endpoint \
+  --container-runtime containerd \
+  --containerd-config-file "$CONTAINERD_TOML" \
+  test || exit_code=$?
+
+if [[ ${exit_code} -eq 0 ]]; then
+  echo "❌ Test Failed: expected a non-zero exit code but got '${exit_code}'"
+  exit 1
+fi
--- a/test/mocks/sudo
+++ b/test/mocks/sudo
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
-echo "mocking sudo with params $@"
+exec "$@"


steved:
@cartermckinnon or @mmerkes would you be able to provide input on this?

mmerkes:
This doesn't sound unreasonable to me. However, my concern would be the same as mentioned in the issue you reference:

```
This is not defaulted because it is a new option which is replacing older options.
We could see if the older options are set and set a default if they are not, maybe?
```

steved:
Since this AMI is a consumer of containerd, I didn't think that applied. How would AWS customers provide an option currently?

mmerkes:
> Since this AMI is a consumer of containerd, I didn't think that applied. How would AWS customers provide that option currently?

I' don't think it would break as they mentioned, but I think it's possible that some customers might append to `files/containerd-config.toml`, which would result in invalid TOML as there'd be duplicate keys. For example:

```
version = 2
root = "/var/lib/containerd"
state = "/run/containerd"

[grpc]
address = "/run/containerd/containerd.sock"

default_runtime_name = "runc"

[plugins."io.containerd.grpc.v1.cri".registry]
config_path = "/etc/containerd/certs.d:/etc/docker/certs.d"

[plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc.options]
SystemdCgroup = true

bin_dir = "/opt/cni/bin"
conf_dir = "/etc/cni/net.d"

[plugins."io.containerd.grpc.v1.cri".registry]
config_path = "/etc/containerd/certs.d:/etc/docker/certs.d"
```

steved:
Interesting, that makes sense. Is there a policy for breaking changes? If, for example, we only set this in 1.24 going forward would that be acceptable?

Would most people be appending during the AMI creation process (meaning we could set this option in bootstrap.sh) or at runtime?

mmerkes:
@steved This is just hypothetical. I have no idea if and when customers are setting this. *If* customers are getting into this potentially breaking situation, I would assume that the easiest way would be at runtime by setting it in their userdata. If they were doing this at build time, they'd probably need to have their own fork of this repo, and I'd be less concerned about that case.

For breaking changes, we avoid doing them unless we have a very compelling reason. In this particular case, while it's not ideal, customers can configure this in their userdata. There would be some lift by baking something into `bootstrap.sh` (by checking for the presence of an environment variable or directly as an argument), but we already support the `--containerd-config-file` flag, which gives you this in a different way and is probably the best way to set this at present.

That being said, if we decided that this was generally valuable and don't expect this to break any customers, another option would be to add some validation in `bootstrap.sh` that ensures that running `bootstrap.sh` fails in a very clear and obvious way if the `containerd` config file is invalid (i.e. someone appended to it and now there's duplicate keys).

I will mention this to my team and get some input, but there may not be appetite for it given the potential risk of breaking customers who set this themselves with the append method vs. the `--containerd-config-file` flag. Regardless, we can keep this open (or turn it into a feature request) and see if others in the community has interest in this.

cartermckinnon:
I think this change is fine as long as we can provide a clear error message in the case @mmerkes is describing.

In the bootstrap script, we could do:
```
containerd config dump
```
which will fail if duplicate keys are present in the TOML.

For example, if I have a `/etc/containerd/config.toml`:
```
[cgroup]
path = "foo"
[cgroup]
path = "bar"
```

then I get:
```
> containerd config dump
containerd: Near line 36 (last key parsed ''): Key 'cgroup' has already been defined.
```

steved:
@cartermckinnon added the config validation and a unit test (had to make a decent set of changes to get verification to work, so let me know if there's a better way).

@mmerkes any feedback from the internal team?

files/bootstrap.sh
```diff
@@ -528,11 +528,13 @@ EOF
     fi
     echo "$(jq '.cgroupDriver="systemd"' $KUBELET_CONFIG)" > $KUBELET_CONFIG
     sudo sed -i s,SANDBOX_IMAGE,$PAUSE_CONTAINER,g /etc/eks/containerd/containerd-config.toml
+
     sudo cp -v /etc/eks/containerd/containerd-config.toml /etc/containerd/config.toml
     sudo cp -v /etc/eks/containerd/sandbox-image.service /etc/systemd/system/sandbox-image.service
     sudo cp -v /etc/eks/containerd/kubelet-containerd.service /etc/systemd/system/kubelet.service
     sudo chown root:root /etc/systemd/system/kubelet.service
     sudo chown root:root /etc/systemd/system/sandbox-image.service
+    sudo containerd config dump > /dev/null
```
cartermckinnon, line 537: nit: comment indicating the intention of this
```suggestion
    # validate containerd config
    sudo containerd config dump > /dev/null
```

cartermckinnon (changes requested):
Looks fine to me; but I don't think we should move things into the `files/containerd` directory. It probably would have been a good idea early-on, but it's likely that some folks are depending on these paths, and I don't think the re-org adds enough value to break them.

steved:
> Looks fine to me; but I don't think we should move things into the `files/containerd` directory. It probably would have been a good idea early-on, but it's likely that some folks are depending on these paths, and I don't think the re-org adds enough value to break them.

Makes sense, that was leftover from trying to continue to work with the local volume mapping `-v`. Pushed e629145fcef9080863b6317de7fe36b73de0e12c.

test/cases/container-runtime-defaults.sh
```diff
@@ -96,3 +96,25 @@ if [[ ${exit_code} -ne 0 ]]; then
...
+    --b64-cluster-ca dGVzdA== \
+    --apiserver-endpoint http://my-api-endpoint \
+    --container-runtime containerd \
+    --containerd-config-file "$config_file" \
+    test || exit_code=$?
+
+if [[ ${exit_code} -eq 0 ]]; then
+    echo "❌ Test Failed: expected a non-zero exit code but got '${exit_code}'"
+    exit 1
+fi
```
cartermckinnon, line 120: We should move this test case to a different file, since it's not related to container runtime defaults. Additionally, it'd be great to add a test for the validity of the containterd config file we have in `files/`.

e.g. something like `containerd-config.sh`:
```
echo "--> Default containerd config file should be valid"
mv /etc/eks/containerd/containerd-config.toml /etc/containerd/config.toml
STDERR_FILE=$(mktemp)
containerd config dump >/dev/null 2>$STDERR_FILE || exit_code=$?

if [[ ${exit_code} -ne 0 ]]; then
    echo "❌ Test Failed: default containerd config file is invalid! $(cat $STDERR_FILE)"
    exit 1
fi

echo "--> Should fail when given an invalid containerd config"
config_file="$TEMP_DIR/containerd.toml"
cat > "$config_file" <<EOF
[cgroup]
path = "foo"
[cgroup]
path = "bar"
EOF

export KUBELET_VERSION=v1.24.15-eks-ba74326
/etc/eks/bootstrap.sh \
    --b64-cluster-ca dGVzdA== \
    --apiserver-endpoint http://my-api-endpoint \
    --container-runtime containerd \
    --containerd-config-file "$config_file" \
    test || exit_code=$?

if [[ ${exit_code} -eq 0 ]]; then
    echo "❌ Test Failed: expected a non-zero exit code but got '${exit_code}'"
    exit 1
fi
```

steved: Added. FWIW the happy path is also tested with the "Should allow containerd as container runtime when below k8s version 1.24" case.

```
--> Should allow containerd as container runtime when below k8s version 1.24
Using kubelet version 1.20.15
Using containerd as the container runtime
'/etc/eks/containerd/containerd-config.toml' -> '/etc/containerd/config.toml'
'/etc/eks/containerd/sandbox-image.service' -> '/etc/systemd/system/sandbox-image.service'
'/etc/eks/containerd/kubelet-containerd.service' -> '/etc/systemd/system/kubelet.service'
containerd: failed to load TOML: /etc/containerd/config.toml: (3, 1): parsing error: keys cannot contain new lines
Exited with error on line 537
❌ Test Failed: expected a zero exit code but got '1'
❌ ❌ container-runtime-defaults.sh Tests Failed! ❌ ❌
```

steved:
@cartermckinnon @mmerkes thoughts?

steved:
Gentle nudge @cartermckinnon, any thoughts?

cartermckinnon:
@steved this looks good to me! Can you resolve the conflicts on this branch?

steved:
done, thank you!

cartermckinnon:
Thanks for working on this!

AljoschaP:
It seems that this PR also breaks compatibility with the [deprecated way](https://github.com/containerd/containerd/blob/main/docs/cri/registry.md#configure-registry-endpoint) of configuring registry mirrors. We use the AMI with cluster-api and append the registry mirror configuration to the default configuration as mentioned above. Its not a big deal but could lead to problems when using cluster-api without a custom image and instead rely on the official one because it could break clusters over night.

lpiob:
We have encountered the problem with newest AMI including this PR on a production cluster today.
 
The reason was as @AljoschaP  and @mmerkes mentioned - duplicate `[plugins."io.containerd.grpc.v1.cri".registry]` sections, as we have been appending ours by userdata.

We have fixed our issue by removing the custom section from userdata.

I'm just posting this here as a confirmation that this PR does in fact break backwards compatibility.

dogzzdogzz:
> > Since this AMI is a consumer of containerd, I didn't think that applied. How would AWS customers provide that option currently?
> 
> I don't think it would break as they mentioned, but I think it's possible that some customers might append to `files/containerd-config.toml`, which would result in invalid TOML as there'd be duplicate keys. For example:
> 
We encountered this issue, too. Added some simple logic to fix it
```bash
  CONTAINERD_CFG=/etc/eks/containerd/containerd-config.toml
  IS_REGISTRY_EXIST=$(grep '\[plugins\."io\.containerd\.grpc\.v1\.cri"\.registry\]' $CONTAINERD_CFG)
  if [[ $IS_REGISTRY_EXIST != "" ]]; then
    echo "registry config already exists"
  else
    echo -e "\n\
[plugins.\"io.containerd.grpc.v1.cri\".registry]\n\
config_path = \"/etc/containerd/certs.d\"" >> $CONTAINERD_CFG
  fi
```

I think we can also add some kind of similar logic in AMI to prevent potential compatibility issues if we already thought it is possible.

cartermckinnon:
I'm sorry for the annoyance this has caused. I would recommend a check like @dogzzdogzz suggests if you patch the containerd config file, because appending will always have this risk. If you override the bootstrap command, there's also the safer `--containerd-config-file`.

Issue #1050, opened 2022-10-14
Create `runtime.slice` and use it for `containerd.service` & `kubelet.service` when using the systemd cgroup driver


stevehipwell:
**What would you like to be added**:

Now that using the systemd cgroup driver has been enabled for the containerd runtime I'd like the `runtime.slice` to be added and referenced from `containerd.service` and `kubelet.service`. The kubelet args `--system-reserved-cgroup=/system` & `--kube-reserved-cgroup=/runtime` should also be set.

This pattern is already present in Bottlerocket so can be copied from there

**Why is this needed**:

This is the [recommended configuration](https://kubernetes.io/docs/tasks/administer-cluster/reserve-compute-resources/).

sidewinder12s:
@stevehipwell Should this be closed as completed?

PR #1051, opened 2022-10-14
Added runtime.slice for containerd runtime

stevehipwell:
**Issue #, if available:**
#1050

**Description of changes:**
This PR adds the `runtime.slice` and makes use of it when using containerd as the runtime as is recommended in the [Kubernetes documentation](https://kubernetes.io/docs/tasks/administer-cluster/reserve-compute-resources/).

**Testing Done**

I've built the AMI and test that containerd is now running under the `runtime.slice` .

```shell
[ec2-user@ip-10-52-14-250 ~]$ sudo systemctl status containerd
● containerd.service - containerd container runtime
   Loaded: loaded (/usr/lib/systemd/system/containerd.service; enabled; vendor preset: disabled)
  Drop-In: /etc/systemd/system/containerd.service.d
           └─00-runtime-slice.conf, 10-compat-symlink.conf
   Active: active (running) since Mon 2023-01-23 16:12:04 UTC; 5s ago
     Docs: https://containerd.io
  Process: 1311 ExecStartPre=/bin/ln -sf /run/containerd/containerd.sock /run/dockershim.sock (code=exited, status=0/SUCCESS)
  Process: 1309 ExecStartPre=/sbin/modprobe overlay (code=exited, status=0/SUCCESS)
 Main PID: 1314 (containerd)
    Tasks: 8
   Memory: 43.3M
   CGroup: /runtime.slice/containerd.service
           └─1314 /usr/bin/containerd
```


stevehipwell:
@cartermckinnon could you take a look at this and see if there is anything obviously wrong with the way I've configured containerd? I've tried overwriting the main unit file but it still uses the wrong cgroup. I don't have the infrastructure setup to build and test this myself and it might be worth trying it that way?

cartermckinnon:
>  I've tried overwriting the main unit file but it still uses the wrong cgroup

You mean the `containerd` unit file, right? I modified the one on an AL2 instance (`/usr/lib/systemd/system/containerd.service`) to use `Slice=runtime.slice` and it looks correct to me:
```
> systemctl status containerd

● containerd.service - containerd container runtime
   Loaded: loaded (/usr/lib/systemd/system/containerd.service; disabled; vendor preset: disabled)
   Active: active (running) since Tue 2022-10-18 20:45:21 UTC; 5s ago
     Docs: https://containerd.io
  Process: 1090 ExecStartPre=/sbin/modprobe overlay (code=exited, status=0/SUCCESS)
 Main PID: 1122 (containerd)
    Tasks: 15
   Memory: 56.5M
   CGroup: /runtime.slice/containerd.service
           └─1122 /usr/bin/containerd
```

Can we add a dropin unit for `containerd` to set this?

stevehipwell:
@cartermckinnon this PR has the dropin file [00-runtime-slice.conf](https://github.com/awslabs/amazon-eks-ami/pull/1051/files#diff-2a36ec539602085e7a230fd6222bab3899222bcbb3ef37ad1d9e968ca612a5fa) to do this. I was testing on an active EKS node so that might have been why it wasn't working correctly. Do you have automation in place to test this PR as based on your testing it should work?

stevehipwell:
@cartermckinnon have you had a chance to look at this again? It'd be good to get it working correctly.

stevehipwell:
@cartermckinnon would it be possible for you to get this tested or do I need to setup some infrastructure to test it myself?

stevehipwell:
@cartermckinnon I've now tested this and and would like to get it merged and released ASAP, is there anything else you need from me?

CC @bwagner5

cartermckinnon:
@stevehipwell sorry for the radio silence -- this looks fine to me. I want to read through the k8s doc on this again to make sure I'm understanding the recommendations, but this should get merged by the end of the week.

Thanks for tackling it!

files/bootstrap.sh
```diff
@@ -494,6 +494,8 @@ if [[ "$CONTAINER_RUNTIME" = "containerd" ]]; then
     sudo cp -v $CONTAINERD_CONFIG_FILE /etc/eks/containerd/containerd-config.toml
   fi
   echo "$(jq '.cgroupDriver="systemd"' $KUBELET_CONFIG)" > $KUBELET_CONFIG
+  echo "$(jq '.systemReservedCgroup="/system"' $KUBELET_CONFIG)" > $KUBELET_CONFIG
+  echo "$(jq '.kubeReservedCgroup="/runtime"' $KUBELET_CONFIG)" > $KUBELET_CONFIG
   sudo sed -i s,SANDBOX_IMAGE,$PAUSE_CONTAINER,g /etc/eks/containerd/containerd-config.toml
```
cartermckinnon, line 499: I think we need to move:
```
sudo mv "${TEMPLATE_DIR}/00-runtime-slice.conf" /etc/systemd/system/containerd.service.d/00-runtime-slice.conf
```
here? Because kubelet is only configured to use this slice when containerd is the `CONTAINER_RUNTIME`.

scripts/install-worker.sh
```diff
@@ -191,6 +197,8 @@ net.bridge.bridge-nf-call-iptables = 1
 net.ipv4.ip_forward = 1
 EOF
 
+sudo mv "${TEMPLATE_DIR}/00-runtime-slice.conf" /etc/systemd/system/containerd.service.d/00-runtime-slice.conf
```
cartermckinnon, line 200: I think this should be done at runtime, so `containerd` is only configured this way when it's the `CONTAINER_RUNTIME`

cartermckinnon:
I'm not clear on the benefit of this without enabling the optional enforcement of `KubeReserved`/`SystemReserved`, I would guess some metrics are more accurate/fine-grained?

stevehipwell:
@cartermckinnon this is the documented way to set up Kubernetes with Containerd as the CRI. So that is the main benefit along with observability improvements.

It then also means that end users can decide to enforce kube reserved and/or system reserved by adding additional kubelet arguments.

scripts/install-worker.sh
stevehipwell: I think that makes sense, can you confirm that Containerd isn't started before it's configured in bootstrap?

files/bootstrap.sh
stevehipwell: Yes we can move this from the setup to here assuming we can guarantee that Containerd hasn't been started by this point.

scripts/install-worker.sh
stevehipwell: Wouldn't that be the place where it'd be most likely that the Containerd service had already run? If anything it needs to be at the top, but we still need to confirm it hasn't run yet.

stevehipwell: I've done this and tidied up the command order slightly for consistency. I agree that the if status logic looks like it needs reviewing; AFAIK installing the RPM will not enable or start the units.

cartermckinnon: > AFAIK installing the RPM will not enable or start the units.

Correct, I tested that yesterday.

---

You'll need to change this line, though; because `TEMPLATE_DIR` doesn't exist in `bootstrap.sh`. I suggested inlining the dropin because otherwise you'll have to place the file in an intermediate location during build.

stevehipwell: This is the install script change, would you like both inlined? Or we do use a temporary location for the conditional files so I could use that for the bootstrap change?

files/bootstrap.sh
```diff
@@ -490,12 +490,19 @@ if [[ "$CONTAINER_RUNTIME" = "containerd" ]]; then
 
   sudo mkdir -p /etc/containerd
   sudo mkdir -p /etc/cni/net.d
-  if [[ -n "$CONTAINERD_CONFIG_FILE" ]]; then
-    sudo cp -v $CONTAINERD_CONFIG_FILE /etc/eks/containerd/containerd-config.toml
+
+  sudo mv "${TEMPLATE_DIR}/00-runtime-slice.conf" /etc/systemd/system/containerd.service.d/00-runtime-slice.conf
```
cartermckinnon, line 494:
```suggestion
  printf '[Service]\nSlice=runtime.slice' | sudo tee /etc/systemd/system/containerd.service.d/00-runtime-slice.conf
```

scripts/install-worker.sh
cartermckinnon: Should have commented on the `bootstrap.sh` thread, this change is outdated. This is the only one that's necessary: https://github.com/awslabs/amazon-eks-ami/pull/1051#discussion_r1087104085

files/bootstrap.sh
stevehipwell: Done

stevehipwell:
@cartermckinnon I've made the suggested change, this should be good now.

cartermckinnon:
LGTM, thanks @stevehipwell 🚀

cartermckinnon (approved):

PR #599, opened 2021-01-15
check shell scripts with shellcheck & shfmt

PascalBourdier:
*Issue #, if available:*

*Description of changes:*
use shellcheck and shfmt

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1061, opened 2022-10-26
run shellcheck on test case files

steved:
**Issue #, if available:**

_N/A_

**Description of changes:**

Looks like there was an abortive attempt in #599, but it seems like a decent idea to lint the test case files.

**Testing Done**

_N/A_


steved:
cc @cartermckinnon @mmerkes

cartermckinnon:
Thanks for bringing this up! I've added `shfmt` targets to the makefile, and the `shellcheck` change will be covered in #1068 .

PR #1066, opened 2022-10-28
Install awscli v2 bundle when possible

cartermckinnon:
**Description of changes:**

This installs the latest awscli v2 bundle when possible, as the version available in the AL yum repos generally lags behind. It also includes the version of the awscli in the `binaries` section of the `version-info.json` file, because the bundle wouldn't appear in the `packages` section.

**Testing Done**

In `us-west-2`, I can see that the v2 bundle is installed:
```
> make 1.22
...
2022-10-28T08:54:58-07:00:     amazon-ebs: Installing awscli v2 bundle
2022-10-28T08:55:03-07:00:     amazon-ebs: You can now run: /usr/local/bin/aws --version
...
```
and the version is present in `version-info.json`:
```
> jq .binaries.awscli amazon-eks-node-1.22-v20221028-version-info.json 
"2.8.6"
```


scripts/install-worker.sh
```diff
@@ -118,6 +117,27 @@ sudo systemctl restart sshd.service
 sudo mkdir -p /etc/eks
 sudo mv $TEMPLATE_DIR/iptables-restore.service /etc/eks/iptables-restore.service
 
+################################################################################
+### awscli #####################################################
+################################################################################
+
+if [[ "$BINARY_BUCKET_REGION" != "us-iso-east-1" && "$BINARY_BUCKET_REGION" != "us-isob-east-1" ]]; then
+    # https://docs.aws.amazon.com/cli/latest/userguide/getting-started-install.html
+    echo "Installing awscli v2 bundle"
```
mmerkes, line 126: Is there any functional difference between v1 and v2? We going to break anybody? Doesn't look like it...

Some questions:

1. Is there any reason this won't work in China regions?
2. Is there a reason that a customer might have access to yum but doesn't have access to this URL and we need to use a flag to make this decision instead? Would it be better to just set a flag, and we'll use it in regions we need to?

cartermckinnon: > Is there any functional difference between v1 and v2?

In general, no; but the yum package installs all sorts of python odds and ends globally. Probably more risk there of something breaking, but I think it's preferable from a CVE perspective to go with v2.

> Is there any reason this won't work in China regions?

It works! This host resolves to an in-region IP, downloads in a few seconds. This was in `cn-northwest-1`:
```
> curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "awscliv2.zip"
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 45.0M  100 45.0M    0     0  13.6M      0  0:00:03  0:00:03 --:--:-- 13.6M
```

> Is there a reason that a customer might have access to yum but doesn't have access to this URL and we need to use a flag to make this decision instead? Would it be better to just set a flag, and we'll use it in regions we need to?

I suppose it's *possible* that the builder instance is firewalled in such a way; but I think it's preferable for those users to allow this host versus using the older CLI lineage.

bwagner5 (approved):
lgtm

scripts/install-worker.sh
cartermckinnon: > If we think that's a high risk

I don't. The python interpreter (both 2 and 3) are still present, and many of the global python packages are still present dependencies of other packages, e.g.:
```
  "python": "2.7.18-1.amzn2.0.5",
  "python-babel": "0.9.6-8.amzn2.0.1",
  "python-backports": "1.0-8.amzn2.0.2",
  "python-backports-ssl_match_hostname": "3.5.0.1-1.amzn2",
  "python-cffi": "1.6.0-5.amzn2.0.2",
  "python-chardet": "2.2.1-1.amzn2",
  "python-configobj": "4.7.2-7.amzn2",
  "python-devel": "2.7.18-1.amzn2.0.5",
  "python-enum34": "1.0.4-1.amzn2",
  "python-idna": "2.4-1.amzn2",
  "python-iniparse": "0.4-9.amzn2",
  "python-ipaddress": "1.0.16-2.amzn2",
  "python-jinja2": "2.7.2-3.amzn2",
  "python-jsonpatch": "1.2-4.amzn2",
  "python-jsonpointer": "1.9-2.amzn2",
  "python-jwcrypto": "0.4.2-1.amzn2",
  "python-kitchen": "1.1.1-5.amzn2",
  "python-libs": "2.7.18-1.amzn2.0.5",
  "python-markupsafe": "0.11-10.amzn2.0.2",
  "python-ply": "3.4-11.amzn2",
  "python-pycparser": "2.14-1.amzn2",
  "python-pycurl": "7.19.0-19.amzn2.0.2",
  "python-repoze-lru": "0.4-3.amzn2",
  "python-requests": "2.6.0-7.amzn2",
  "python-six": "1.9.0-2.amzn2",
  "python-urlgrabber": "3.10-9.amzn2.0.1",
  "python-urllib3": "1.25.9-1.amzn2.0.2",
  "python2-cryptography": "1.7.2-2.amzn2",
  "python2-jsonschema": "2.5.1-3.amzn2.0.1",
  "python2-oauthlib": "2.0.1-8.amzn2.0.1",
  "python2-pyasn1": "0.1.9-7.amzn2.0.1",
  "python2-rpm": "4.11.3-48.amzn2.0.2",
  "python2-setuptools": "41.2.0-4.amzn2.0.2",
  "python3": "3.7.10-1.amzn2.0.1",
  "python3-daemon": "2.2.3-8.amzn2.0.2",
  "python3-docutils": "0.14-1.amzn2.0.2",
  "python3-libs": "3.7.10-1.amzn2.0.1",
  "python3-lockfile": "0.11.0-17.amzn2.0.2",
  "python3-pip": "20.2.2-1.amzn2.0.3",
  "python3-pystache": "0.5.4-12.amzn2.0.1",
  "python3-setuptools": "49.1.3-1.amzn2.0.2",
  "python3-simplejson": "3.2.0-1.amzn2.0.2",
```

Issues 1 and 2 that you pointed out are reason enough to not install both, in my opinion.

mmerkes: Ok, if python is present, I don't think it's really an issue then.

mmerkes (approved):

guessi:
For anyone who found that `$sudo aws` don't work after updating node to [v20221112](https://github.com/awslabs/amazon-eks-ami/releases/tag/v20221112) or later version,

```
$ sudo aws --version
sudo: aws: command not found
```

I'm here to address the change of `aws` command, see the [discussion here](https://github.com/awslabs/amazon-eks-ami/pull/938#discussion_r1002052392).

If you need to execute `aws` command somewhere in your bootstrap script (or manual exection), please update your command to `sudo /usr/local/bin/aws ...` when you calling `aws` with `sudo`.

PR #1068, opened 2022-10-29
Add lint make target

cartermckinnon:
**Description of changes:**

Adds a `make` target called `lint` which will be added to the CI, failing if the source files are not of the proper format or if they have bug-checking failures.

**Testing Done**

This will fail because I haven't opened a PR standardizing the format yet :) 
```
make lint
```


Makefile
```diff
@@ -33,19 +33,25 @@ T_RESET := \e[0m
...
-	  --case-indent \
-	  --space-redirects \
-	  --keep-padding \
-	  $(MAKEFILE_DIR)
+	shfmt $(SHFMT_FLAGS) --write $(MAKEFILE_DIR)
+
+.PHONY: lint
+lint: ## Check the source files for syntax and format issues
+	shfmt  $(SHFMT_FLAGS) --diff $(MAKEFILE_DIR)
+	for FILE in $$(find $(MAKEFILE_DIR) -type f -name '*.sh'); do
```
cartermckinnon, line 52: There's not a simple way to do shebang matching  for shellcheck, but shfmt will catch some syntax issues anyway. probably good enough for the time being.

Makefile
```diff
@@ -33,19 +33,25 @@ T_RESET := \e[0m
...
+	  shellcheck --format gcc --severity error "$$FILE"
```
cartermckinnon, line 53: I started with `error` severity, because there's only one of those right now; while there are 18 warnings. I can raise the bar once we address more of them (or decide whether a lower severity bar makes sense).

Makefile
```diff
@@ -33,19 +33,23 @@ T_RESET := \e[0m
...
-	  --case-indent \
-	  --space-redirects \
-	  --keep-padding \
-	  $(MAKEFILE_DIR)
+	shfmt $(SHFMT_FLAGS) --write $(MAKEFILE_DIR)
+
+.PHONY: lint
+lint: ## Check the source files for syntax and format issues
+	shfmt  $(SHFMT_FLAGS) --diff $(MAKEFILE_DIR)
+	for FILE in $$(find $(MAKEFILE_DIR) -type f -name '*.sh'); do shellcheck --format gcc --severity error "$$FILE"; done
```
cartermckinnon, line 52: There's not a simple way to do shebang matching for shellcheck, but shfmt will catch some syntax issues anyway. probably good enough for the time being.

cartermckinnon:
I went ahead and added `make lint` to the CI workflow, so that check will need to be overridden to merge this.

I'll open two subsequent PR's:
1. Apply formatting.
2. Fix the 1 error-level `shellcheck` finding.

I don't want to cause headache for the PR's that are close to merge, so I'll wait until #938 and #1066 are merged to move forward.

mmerkes (approved):

PR #1077, opened 2022-11-01
Add fallback to shfmt in container

cartermckinnon:
**Description of changes:**

If `shfmt` isn't found on the `PATH`, just use the official Docker image to do the formatting. Makes it easier to contribute by not requiring toolchain setup.

**Testing Done**

With `shfmt` on my `PATH`:
```
> make fmt
shfmt --list --language-dialect auto --indent 2 --binary-next-line --case-indent --space-redirects --write /github/amazon-eks-ami
```
and without it on my path:
```
> make fmt
docker run --rm -v /github/amazon-eks-ami:/github/amazon-eks-ami mvdan/shfmt --list --language-dialect auto --indent 2 --binary-next-line --case-indent --space-redirects --write /github/amazon-eks-ami
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,18 +33,22 @@ T_RESET := \e[0m
 .PHONY: all
 all: 1.20 1.21 1.22 1.23 ## Build all versions of EKS Optimized AL2 AMI
 
+# ensure that these flags are equivalent to the rules in the .editorconfig
+SHFMT_FLAGS := --list \
+--language-dialect auto \
+--indent 2 \
+--binary-next-line \
+--case-indent \
+--space-redirects
+
+SHFMT_COMMAND := $(shell which shfmt)
+ifeq (, $(SHFMT_COMMAND))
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -33,18 +33,22 @@ T_RESET := \e[0m
...
+SHFMT_FLAGS := --list \
+--language-dialect auto \
+--indent 2 \
+--binary-next-line \
+--case-indent \
+--space-redirects
+
+SHFMT_COMMAND := $(shell which shfmt)
+ifeq (, $(SHFMT_COMMAND))
+SHFMT_COMMAND = docker run --rm -v $(MAKEFILE_DIR):$(MAKEFILE_DIR) mvdan/shfmt
```
mmerkes, line 46: Is this actually better? Is it a big deal to set this up?

cartermckinnon: I think it's handy because:
1. you can always use the latest `smfmt` without thinking about it
2. if you use multiple machines for dev (like we do) you don't have to do any setup/maintenance on your various boxes.
3. `shfmt` requires a working `go` setup, which isn't a given for our contributors.

mmerkes (approved):

PR #1079, opened 2022-11-02
Move variable defaults to -var-file

cartermckinnon:
**Description of changes:**

These variables appear in multiple templates internally, so this will centralize the values.

I also fixed a weird spacing issue with the `PACKER_VARIABLES` logic. The `foreach` function in `make` will join the expanded 3rd argument with a space character. When the loop evaluates for packer variables that aren't defined, it results in a ton of extra space being inserted. Example before this change:
```
packer validate -var-file=version-variables.json  -var instance_type='m4.large'      -var arch='x86_64'  -var ami_name='amazon-eks-node-.-v20221102'              -var aws_region='us-west-2'        -var binary_bucket_region='us-west-2'                                           eks-worker-al2.json
```
and after this change:
```
packer validate -var-file=version-variables.json -var instance_type='m4.large' -var arch='x86_64' -var ami_name='amazon-eks-node-.-v20221102' -var aws_region='us-west-2' -var binary_bucket_region='us-west-2' eks-worker-al2.json
```

**Testing Done**

Succeeded:
```
make 1.22
```


eks-worker-al2.json
```diff
@@ -13,10 +14,10 @@
     "aws_session_token": "{{env `AWS_SESSION_TOKEN`}}",
     "binary_bucket_name": "amazon-eks",
     "binary_bucket_region": "us-west-2",
-    "cni_plugin_version": "v0.8.6",
-    "containerd_version": "1.6.6-1.amzn2.0.2",
+    "cni_plugin_version": "",
+    "containerd_version": "",
     "creator": "{{env `USER`}}",
-    "docker_version": "20.10.17-1.amzn2.0.1",
+    "docker_version": "",
```
mmerkes, line 20: Presumably, if you set these, it would override the variables set below?

cartermckinnon: I have to iron out the order of overrides; I think the var file would actually override this value. My guess is it goes:
1. `variables` section in the template.
2. Anything in `-var-file`
3. Anything defined with `-var`.

cartermckinnon: Looks like it's probably better to define these version variables in the template with a value of `null`, so they have to be defined either in the `version-variables.json` file or with a `make` argument.
> If the default value is null, then the user variable will be required. This means that the user must specify a value for this variable or template validation will fail.
https://developer.hashicorp.com/packer/docs/templates/legacy_json_templates/user-variables

ravisinha0506: Can we remove the properties/variables altogether from eks-worker-al2.json to avoid this confusion?

cartermckinnon: I think it's better to have all the template variables called out in the template itself (and this is a requirement of HCL-format templates, which we should eventually move to).

This change also relies on them being defined in the template, and is preferable to manually maintaining this list (which has been unreliable in the past): https://github.com/awslabs/amazon-eks-ami/pull/1079/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R2

mmerkes: I'm fine with this approach. My only concern is the priority order. IMO, command line args should override everything, `eks-worker-al2.json` should override the defaults, and this new file should just include default values that get set if nothing overrides it.

eks-worker-al2.json
```diff
@@ -13,10 +14,10 @@
     "aws_session_token": "{{env `AWS_SESSION_TOKEN`}}",
     "binary_bucket_name": "amazon-eks",
     "binary_bucket_region": "us-west-2",
-    "cni_plugin_version": "v0.8.6",
-    "containerd_version": "1.6.6-1.amzn2.0.2",
+    "cni_plugin_version": null,
+    "containerd_version": null,
```
mmerkes, line 18: What if these are empty strings? It will be confusing to customers if this has to be `null` and breaks if they include empty strings. Same with the below.

eks-worker-al2.json
cartermckinnon: > eks-worker-al2.json should override the defaults

The template file *is* the default. Changing the value there to `null` means the variable has to be defined elsewhere.

The variable file overrides the template. In practice, these will be the "defaults" for our users. Users can provide their own variable file with the `PACKER_VARIABLE_FILE` make argument.

The `-var` flag overrides everything. Users can override *our* variable file with these, if they don't want to provide their own variable file.

eks-worker-al2.json
cartermckinnon: Not sure I'm following; if a customer does `make docker_version=""`, that won't work (same as today).

cartermckinnon:
After this change, possible usages include:
1. `make`. This will use the values defined in the variables file.
2. `make PACKER_VARIABLE_FILE=my-variable-file.json`. This will allow customers to define their own values for `docker_version` or any other variable in the template. If they don't include a required variable, like `docker_version`, they'll get an error.
3. `make docker_version=1.2.3`. This will override the value we have in our variable file. This is already how users should be providing these values.

mmerkes (commented):
LGTM, but let's update the README to explain how this works and how customers should override defaults.

cartermckinnon:
@mmerkes and I discussed this -- we're going to try moving as many variables into the var-file as possible, to avoid confusion about where defaults are defined. All variables that have a default value **that isn't a go template** will be moved to the variable file. New rev coming.

cartermckinnon:
Packer allows multiple `-var-file` arguments, and will merge the contents in the order that you specify them. I moved all the defaults into the variables file. Users can specify their own variables file to override a subset of the variables, or they can specify individual key-value variable overrides to `make`. (Or some combination of both).

eks-worker-al2.json
```diff
@@ -1,42 +1,43 @@
 {
+  "_comment": "All template variables are enumerated here; and most variables have their default values defined in eks-worker-al2-variables.json",
```
saurav-agarwalla, line 2: What's the benefit in duplicating the variables here? Can't `eks-worker-al2-variables.json` be the source of truth?

README.md
```diff
@@ -32,6 +32,18 @@ To build an Amazon EKS Worker AMI for a particular Kubernetes version run the fo
...
 make 1.23 ## Build a Amazon EKS Worker AMI for k8s 1.23
 ```
+
+### AMI template variables
+
+Default values for most variables are defined in [a default variable file](eks-worker-al2-variables.json).
+
+Users have the following options for specifying their own values:
+
+1. Provide a variable file with the `PACKER_VARIABLE_FILE` argument to `make`. Values in this file will override values in the default variable file. Your variable file does not need to include all possible variables, as it will be merged with the default variable file.
```
saurav-agarwalla, line 42: Would be possible to write tests for these behaviors?

saurav-agarwalla (commented):
Thanks for the change. It'll simplify a bunch of things.

README.md
saurav-agarwalla: Nevermind, I saw your comment stating that packer is the one which does this so no point in testing that. Contents of that comment might be good for this section too. :)

eks-worker-al2.json
cartermckinnon: This PR dynamically discovers the variables from the template using `packer inspect`, so they need to be present in the template. ☹️ Because we support overriding them in the Makefile via this mechanism, we have to maintain the list of variables somewhere (either in the Makefile or in the template).

We *could* pull these out with something else like `jq`, but I'd prefer to use the Packer mechanism if it's reasonable.

README.md
cartermckinnon: I'd err on the side of fewer impl details in this doc, but I don't feel too strongly. Can totally add if you think it's necessary

eks-worker-al2.json
cartermckinnon: This is the `packer inspect` usage I'm referring to: https://github.com/awslabs/amazon-eks-ami/blob/8da3272cda99241070d5d667e7b75cd3a9e5cf4d/Makefile#L2

saurav-agarwalla (approved):

PR #1070, opened 2022-10-29
Add vercmp helper.

cartermckinnon:
**Description of changes:**

This adds a utility to deal with semantic version comparisons, since we're adding more of these as times goes on (and will likely have more coming with 1.25).

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

**Testing Done**

Test cases added for each comparison operator

**Usage**

```
> vercmp

Comparison expressions for semantic versions.

usage: vercmp VERSION_A OPERATOR VERSION_B

OPERATORS

  lt   - Less than
  lteq - Less than or equal to
  eq   - Equal to
  gteq - Grater than or equal to
  gt   - Greater than

```

PR #1089, opened 2022-11-07
Update credential provider API to beta for 1.24+

mmerkes:
**Description of changes:**
As of 1.22, we [added ECR credential provider support](https://github.com/awslabs/amazon-eks-ami/commit/da570d5d2474f847d1f24bbf36fedbf437c1de65), which was an alpha feature. The feature was enabled [here](https://github.com/awslabs/amazon-eks-ami/blob/master/scripts/install-worker.sh#L286), but in my testing, those branches didn't actually catch correctly, so the changes were never applied. As of 1.24, the feature have graduated to beta, so this PR updates the API version for 1.24+.

**Testing Done**

Created 1.24 and 1.23 AMIs, verified that they joined a cluster and confirmed that `--image-credential-provider-config /etc/eks/ecr-credential-provider/ecr-credential-provider-config --image-credential-provider-bin-dir /etc/eks/ecr-credential-provider` flags were set correctly and that 1.23 set alpha APIs and 1.24 set beta APIs.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -115,6 +115,21 @@ Delete the cluster:
 eksctl delete cluster -f cluster.yaml
 ```
 
+## Troubleshooting
+
+**Tests fail with `realpath: command not found`**
+
+When running `make test`, you may see a message like below:
+
+```
+test/test-harness.sh: line 41: realpath: command not found
+/entrypoint.sh: line 13: /test.sh: No such file or directory
+```
+
+The issue is discussed in [this StackExchange post](https://unix.stackexchange.com/questions/101080/realpath-command-not-found).
+
+On OSX, running `brew install coreutils` resolves the issue.
+
 ## Finding contributions to work on
 Looking at the existing issues is a great way to find something to contribute on. As our projects, by default, use the default GitHub issue labels ((enhancement/bug/duplicate/help wanted/invalid/question/wontfix), looking at any ['help wanted'](https://github.com/aws-samples/amazon-eks-ami/labels/help%20wanted) issues is a great place to start. 
 
--- a/files/bootstrap.sh
+++ b/files/bootstrap.sh
@@ -145,6 +145,14 @@ DEFAULT_CONTAINER_RUNTIME=dockerd
 if vercmp "$KUBELET_VERSION" gteq "1.24.0"; then
   IS_124_OR_GREATER=true
   DEFAULT_CONTAINER_RUNTIME=containerd
+elif vercmp "$KUBELET_VERSION" gteq "1.22.0"; then
+  # Ensure that these exist for testing purposes
+  mkdir -p /etc/eks/ecr-credential-provider
+  touch /etc/eks/ecr-credential-provider/ecr-credential-provider-config
+  # These APIs are only available in alpha pre-1.24.
+  # This can be removed when version 1.23 is no longer supported.
+  sed -i s,kubelet.config.k8s.io/v1beta1,kubelet.config.k8s.io/v1alpha1,g /etc/eks/ecr-credential-provider/ecr-credential-provider-config
+  sed -i s,credentialprovider.kubelet.k8s.io/v1beta1,credentialprovider.kubelet.k8s.io/v1alpha1,g /etc/eks/ecr-credential-provider/ecr-credential-provider-config
 fi
 
 # Set container runtime related variables
--- a/test/Dockerfile
+++ b/test/Dockerfile
@@ -1,6 +1,9 @@
 FROM public.ecr.aws/aws-ec2/amazon-ec2-metadata-mock:v1.11.2 as aemm
 FROM public.ecr.aws/amazonlinux/amazonlinux:2
 RUN yum install -y jq
+RUN yum install -y wget
+RUN wget -qO /usr/local/bin/yq https://github.com/mikefarah/yq/releases/latest/download/yq_linux_amd64
+RUN chmod a+x /usr/local/bin/yq
 
 ENV IMDS_ENDPOINT=127.0.0.1:1338
 COPY --from=aemm /ec2-metadata-mock /sbin/ec2-metadata-mock
--- a/test/cases/ecr-credential-provider-config.sh
+++ b/test/cases/ecr-credential-provider-config.sh
@@ -1,0 +1,70 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+exit_code=0
+TEMP_DIR=$(mktemp -d)
+
+# Setup test cases
+mkdir -p /etc/eks/ecr-credential-provider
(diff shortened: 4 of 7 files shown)


scripts/install-worker.sh
```diff
@@ -316,7 +316,7 @@ fi
 ################################################################################
 ### ECR CREDENTIAL PROVIDER ####################################################
 ################################################################################
-if [[ ! $KUBERNETES_VERSION =~ "1.19"* && ! $KUBERNETES_VERSION =~ "1.20"* && ! $KUBERNETES_VERSION =~ "1.21"* ]]; then
+if [[ ! $KUBERNETES_VERSION == "1.19"* && ! $KUBERNETES_VERSION == "1.20"* && ! $KUBERNETES_VERSION == "1.21"* ]]; then
```
cartermckinnon, line 319: Bit concerned that this wasn't working before; can you elaborate on this?

cartermckinnon: Also can you review #1070 ? Would make this change moot

files/bootstrap.sh
```diff
@@ -152,6 +152,14 @@ DEFAULT_CONTAINER_RUNTIME=dockerd
 if is_greater_than_or_equal_to_version $KUBELET_VERSION "1.24.0"; then
   IS_124_OR_GREATER=true
   DEFAULT_CONTAINER_RUNTIME=containerd
+else
+  # Ensure that these exist for testing purposes and 1.21 and below
```
cartermckinnon, line 156: I'm not following, seems like this downgrade of the API versions should happen for 1.22 and 1.23 only; can we just add an `elif` for that?

cartermckinnon:
We need some test cases for this 😬

scripts/install-worker.sh
mmerkes: When I built AMIs 1.22+, I found that the configuration files didn't exist. With some testing, I verified that this version check logic doesn't work at all.

files/bootstrap.sh
mmerkes: Ya, I can do that.

.gitignore
```diff
@@ -3,3 +3,4 @@
 .idea
 *version-info.json
 .DS_Store
+files/ecr-credential-provider/
```
mmerkes, line 6: The `files` directory is mounted at `/etc/eks`, so we end up making this directory as part of the tests and it should up here. Cleaning up in the tests doesn't remove it, so just ignoring it.

CONTRIBUTING.md
```diff
@@ -115,6 +115,21 @@ Delete the cluster:
...
+When running `make test`, you may see a message like below:
+
+```
+test/test-harness.sh: line 41: realpath: command not found
+/entrypoint.sh: line 13: /test.sh: No such file or directory
+```
+
+The issue is discussed in [this StackExchange post](https://unix.stackexchange.com/questions/101080/realpath-command-not-found).
+
+On OSX, running `brew install coreutils` resolves the issue.
```
cartermckinnon, line 131: huh, must have forgotten I did this in the past.

.gitignore
cartermckinnon: we should just `COPY` this into the container image instead of mounting it, that's what we do with all the other static resources. The only thing mounting in should be the test case being run, IMO

cartermckinnon: Actually, looks like we already do that? https://github.com/awslabs/amazon-eks-ami/blob/master/test/Dockerfile#L10

Can you just remove this: https://github.com/awslabs/amazon-eks-ami/blob/master/test/test-harness.sh#L41

test/cases/ecr-credential-provider-config.sh
```diff
@@ -0,0 +1,70 @@
...
+fi
+
+expected_kubelet_config_api="kubelet.config.k8s.io/v1beta1"
+grep -q $expected_kubelet_config_api $CRED_PROVIDER_FILE || exit_code=$?
+if [[ ${exit_code} -ne 0 ]]; then
+  echo "❌ Test Failed: expected 1.24 credential provider file to contain $expected_kubelet_config_api"
+  exit 1
+fi
+
+exit_code=0
```
cartermckinnon, line 70: medium nit: tests would be a little firmer if we did some validation on the YAML itself. Maybe install `yq` in the test container and use that to verify specific fields instead of `grep`? it would also throw an error if the YAML is malformed as a result of our `sed`-s, etc.

.gitignore
mmerkes: Ya, let me see if that resolves the issue.

files/bootstrap.sh
```diff
@@ -145,6 +145,14 @@ DEFAULT_CONTAINER_RUNTIME=dockerd
 if vercmp "$KUBELET_VERSION" gteq "1.24.0"; then
   IS_124_OR_GREATER=true
   DEFAULT_CONTAINER_RUNTIME=containerd
+elif vercmp "$KUBELET_VERSION" gteq "1.22.0"; then
+  # Ensure that these exist for testing purposes
+  mkdir -p /etc/eks/ecr-credential-provider
+  touch /etc/eks/ecr-credential-provider/ecr-credential-provider-config
```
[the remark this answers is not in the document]
mmerkes: No, because we copy them in `install-worker.sh`, so when that doesn't run, they don't exist.

test/cases/ecr-credential-provider-config.sh
mmerkes: Ya, I could, though that's yet another dependency people need to install to develop on this package.

files/bootstrap.sh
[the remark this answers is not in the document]
cartermckinnon: Can you handle this in your test cases? Just seems odd to be in the bootstrap script

test/Dockerfile
```diff
@@ -1,6 +1,9 @@
 FROM public.ecr.aws/aws-ec2/amazon-ec2-metadata-mock:v1.11.2 as aemm
 FROM public.ecr.aws/amazonlinux/amazonlinux:2
 RUN yum install -y jq
+RUN yum install -y wget
+RUN wget -qO /usr/local/bin/yq https://github.com/mikefarah/yq/releases/latest/download/yq_linux_amd64
+RUN chmod a+x /usr/local/bin/yq
```
cartermckinnon, line 6: nit: chain these into a single RUN

mmerkes: I'll combine in my next PR.

files/bootstrap.sh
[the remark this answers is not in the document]
mmerkes: Ya, it is a little odd and not ideal. It's not my test cases. The issue is that I didn't want to go edit every other test case that invokes `bootstrap.sh` and have it be a thing that perpetually needs to be handled.

PR #1090, opened 2022-11-09
Short-circut sandbox image fetching

vpineda1996:
**Issue #, if available:**
N/A
**Description of changes:**
The commit short-circuits fetching the sandbox image from ECR whenever its present locally. This is needed to support edge device stability during disconnection. 

In order to bring the kubelet back online, systemd will create a sandbox in which the kubelet will run. If connection to the AWS region is down, we won't be able to reach ECR and pulling will fail. 

We don't need to call ECR to pull the image if its available locally. So this commit adds that short-circuiting.

**Testing Done**

- Built the AMI
- Tested locally with a worker node when it is disconnect and connected using a cluster running k8s 1.21.
- Ran conformance tests

### DIFF
--- a/files/pull-sandbox-image.sh
+++ b/files/pull-sandbox-image.sh
@@ -2,4 +2,10 @@
 set -euo pipefail
 
 sandbox_image="$(awk -F'[ ="]+' '$1 == "sandbox_image" { print $2 }' /etc/containerd/config.toml)"
+
+### Short-circuit fetching sandbox image if its already present
+if [[ "$(sudo ctr --namespace k8s.io image ls | grep $sandbox_image)" != "" ]]; then
+  exit 0
+fi
+
 /etc/eks/containerd/pull-image.sh "${sandbox_image}"


cartermckinnon:
@bwagner5 you got thoughts on this one?

files/pull-sandbox-image.sh
```diff
@@ -2,6 +2,13 @@
 
 ### fetching sandbox image from /etc/containerd/config.toml
 sandbox_image=$(awk -F'[ ="]+' '$1 == "sandbox_image" { print $2 }' /etc/containerd/config.toml)
+
+### Short-circuit fetching sandbox image if its already present
+if [[ $(sudo ctr -n k8s.io images check name=="$sandbox_image") != "" ]]; then
```
cartermckinnon, line 7: I think this check would always be `true`, because `ctr` always prints the headers. It also doesn't seem to indicate with the exit code whether any images matched the filters 👎 

So your `grep` solution is probably as good as anything

vpineda1996: Does it?

I tested this out. They don't print anything if there is no image

```
sh-4.2$ sudo ctr -n k8s.io images check name==602401143452.dkr.ecr.us-west-2.amazonaws.com/eks/pause:3.5
REF                                                        TYPE                                                      DIGEST                                                                  STATUS         SIZE                UNPACKED
602401143452.dkr.ecr.us-west-2.amazonaws.com/eks/pause:3.5 application/vnd.docker.distribution.manifest.list.v2+json sha256:529cf6b1b6e5b76e901abc43aee825badbd93f9c5ee5f1e316d46a83abbce5a2 complete (2/2) 290.5 KiB/290.5 KiB true
sh-4.2$ sudo ctr -n k8s.io images check name==602401143452.dkr.ecr.us-west-2.amazonaws.com/eks/pause:3.3
sh-4.2$
```

cartermckinnon: With:
```
> ctr --version
ctr github.com/containerd/containerd 1.4.13
```
I see this on stdout:
```
> sudo ctr -n k8s.io images check name==foo 2> /dev/null
REF TYPE DIGEST STATUS SIZE UNPACKED 
```

files/pull-image.sh
```diff
@@ -19,6 +19,11 @@ function retry() {
   done
 }
 
+### Short-circuit fetching sandbox image if its already present
+if [[ "$(sudo ctr --namespace k8s.io image ls | grep $sandbox_image)" != "" ]]; then
```
cartermckinnon, line 23: I think you actually want to put this in https://github.com/awslabs/amazon-eks-ami/blob/1339eda7f7a811b2fbc7ce0d5966635cf0855bf7/files/pull-sandbox-image.sh because I'm not sure we want this behavior for all images passed to `pull-image.sh`. 

(also the variable has changed in this script from `sandbox_image` to `img` after #938 )

cartermckinnon (approved):

Issue #1020, opened 2022-09-14
Upgrade API VERSION from client.authentication.k8s.io/v1alpha1 to client.authentication.k8s.io/v1beta1


mmerkes:
### What would you like to be added
Upgrade the API version from `client.authentication.k8s.io/v1alpha1` to `client.authentication.k8s.io/v1beta1`. The API version is referenced [here](https://github.com/awslabs/amazon-eks-ami/blob/v20220824/files/kubelet-kubeconfig#L18) and used by the [aws-iam-authenticator](https://github.com/kubernetes-sigs/aws-iam-authenticator) version currently used in the AMI - `0.5.0`. More details below.

### Why is this needed
The alpha API version is no longer supported in Kubernetes version 1.24.

### `aws-iam-authenticator` Note

This change requires upgrading the aws-iam-authenticator to `0.5.9` or any version with [this commit](https://github.com/kubernetes-sigs/aws-iam-authenticator/commit/0221afb8e2a5d14a20a93473edc4c2aa7676ce95). After upgrading the API version, you'll see the following error if the `aws-iam-authenticator` is not upgraded as well:

```
getting credentials: exec plugin is configured to use API version client.authentication.k8s.io/v1beta1, plugin returned version client.authentication.k8s.io/v1alpha1
```

If you're building your own AMI and not vending the `aws-iam-authenticator` from the EKS buckets, you'll need to update it. As of 9/14/2022, the `0.5.0` authenticator is stored in the binary buckets, but we'll update that as part of resolving this issue.

**Future Improvement:**

[This PR](https://github.com/awslabs/amazon-eks-ami/pull/653) makes a change that would allow us to remove the dependency on the `aws-iam-authenticator` entirely by using the AWS CLI to call `aws eks get-token` instead of calling the authenticator. However, as of 9/14/2022, the AmazonLinux yum repos use an older version of the AWS CLI that doesn't include [this change](https://github.com/aws/aws-cli/commit/99d0fa1c96f7e76a570fe671f099163ec859cde7), which is required for resolving this issue. Once the CLI is updated, we should be able to remove the dependency altogether.

cartermckinnon:
This was addressed in https://github.com/awslabs/amazon-eks-ami/releases/tag/v20221112.

PR #1095, opened 2022-11-10
Auth api update

mmerkes:
**Issue #, if available:**
#1020
**Description of changes:**
As detailed in #1020, the `client.authentication.k8s.io/v1alpha1` API has been removed and we need to start using `client.authentication.k8s.io/v1beta1`. As of `aws-iam-authenticator` `0.5.9`, the beta API version is supported. On earlier versions, nodes will fail to join the cluster due to a mismatch in our config and expectations from the authenticator. To handle that we did a few things:

1. Released binaries with `kubernetes_build_date` of `2022-10-31` that include authenticator version `0.5.10`.
2. Added a check, in this PR, to fail the AMI build if the authenticator version does not support the beta API version.
3. Update the Makefile to use the latest binary versions, which also include `kubelet` version upgrades for Kubernetes versions 1.22+.

**NOTE: We have not yet released AMIs with these binaries and run our full suite of tests on these AMIs, though we have run a portion of them, so I recommend that customers wait for an official AMI release with all of these changes before building your own AMIs with them and running production traffic.** We intend to release AMIs soon, and those AMIs will go through our usual validation.

**Testing Done**
Built 1.20-1.24 AMIs and verified that 1.20, 1.23 and 1.24 AMIs successfully joined clusters.

Also, I verified that if we don't update the binaries, we get a build failure as expected.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -78,19 +78,23 @@ k8s: validate ## Build default K8s version of EKS Optimized AL2 AMI
 
 .PHONY: 1.20
 1.20: ## Build EKS Optimized AL2 AMI - K8s 1.20
-	$(MAKE) k8s kubernetes_version=1.20.15 kubernetes_build_date=2022-07-27 pull_cni_from_github=true
+	$(MAKE) k8s kubernetes_version=1.20.15 kubernetes_build_date=2022-10-31 pull_cni_from_github=true
 
 .PHONY: 1.21
 1.21: ## Build EKS Optimized AL2 AMI - K8s 1.21
-	$(MAKE) k8s kubernetes_version=1.21.14 kubernetes_build_date=2022-07-27 pull_cni_from_github=true
+	$(MAKE) k8s kubernetes_version=1.21.14 kubernetes_build_date=2022-10-31 pull_cni_from_github=true
 
 .PHONY: 1.22
 1.22: ## Build EKS Optimized AL2 AMI - K8s 1.22
-	$(MAKE) k8s kubernetes_version=1.22.12 kubernetes_build_date=2022-07-27 pull_cni_from_github=true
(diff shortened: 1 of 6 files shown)


prasad0896:
Thanks for this, I will close my PR https://github.com/awslabs/amazon-eks-ami/pull/1048 in favor of this one then.

prasad0896 (approved):

Makefile
```diff
@@ -78,19 +78,23 @@ k8s: validate ## Build default K8s version of EKS Optimized AL2 AMI
...
+	$(MAKE) k8s kubernetes_version=1.22.15 kubernetes_build_date=2022-10-31 pull_cni_from_github=true
 
 .PHONY: 1.23
 1.23: ## Build EKS Optimized AL2 AMI - K8s 1.23
-	$(MAKE) k8s kubernetes_version=1.23.9 kubernetes_build_date=2022-07-27 pull_cni_from_github=true
+	$(MAKE) k8s kubernetes_version=1.23.13 kubernetes_build_date=2022-10-31 pull_cni_from_github=true
+
+.PHONY: 1.24
+1.24: ## Build EKS Optimized AL2 AMI - K8s 1.24
+	$(MAKE) k8s kubernetes_version=1.24.6 kubernetes_build_date=2022-10-31 pull_cni_from_github=true
```
cartermckinnon, line 97: isn't it `1.24.7`?

cartermckinnon:
```suggestion
	$(MAKE) k8s kubernetes_version=1.24.7 kubernetes_build_date=2022-10-31 pull_cni_from_github=true
```

mmerkes: Good catch. Too much back and forth in my Makefile...

PR #1107, opened 2022-11-22
Remove unused variable

cartermckinnon:
**Description of changes:**

This environment variable doesn't actually appear in `install-worker.sh`.


bwagner5 (approved):
Oops forgot to remove that after moving to use versions out of the addons api. Thanks for catching! LGTM

PR #1118, opened 2022-12-07
Upgrade 1.24 to kernel-5.10

### LINKED ISSUE #857: Upgrade Linux Kernel to 5.10

cartermckinnon:
**Issue #, if available:**

Closes #857

**Description of changes:**

This upgrades the 1.24 AMI's to use the 5.10 kernel line.

**Testing Done**

I did `make 1.23`, confirmed it's still on 5.4; then did `make 1.24` and confirmed it's now on 5.10.

@camrakin TODO


scripts/install-worker.sh
```diff
@@ -48,13 +48,6 @@ else
   exit 1
 fi
 
-################################################################################
-### Utilities ##################################################################
-################################################################################
-
-sudo chmod -R a+x $TEMPLATE_DIR/bin/
-sudo mv $TEMPLATE_DIR/bin/* /usr/bin/
-
```
cartermckinnon, line 57: I installed these in a new provisioner so they're available during `upgrade_kernel.sh`.

guessi (approved):

scripts/upgrade_kernel.sh
```diff
@@ -5,9 +5,13 @@ set -o nounset
...
-
-  echo "kernel_version is unset. Setting to $KERNEL_VERSION"
+  if vercmp "$KUBERNETES_VERSION" gteq "1.24.0"; then
+    KERNEL_VERSION=5.10
+    echo "kernel_version is unset. Setting to $KERNEL_VERSION based on Kubernetes version $KUBERNETES_VERSION."
+  else
+    KERNEL_VERSION=5.4
+    echo "kernel_version is unset. Setting to $KERNEL_VERSION."
+  fi
 fi
```
guessi, line 15: 🔽  would it be simpler like this? 

```
if [[ -z "$KERNEL_VERSION" ]]; then
  KERNEL_VERSION=5.4

  echo "kernel_version is unset. Setting to $KERNEL_VERSION"
  if vercmp "$KUBERNETES_VERSION" gteq "1.24.0"; then
    KERNEL_VERSION=5.10
  fi

  echo "kernel_version is unset. Setting to $KERNEL_VERSION based on Kubernetes version $KUBERNETES_VERSION."

fi
```

cartermckinnon: I don't think we need both the `echo`-s, but it does make sense to log the k8s version either way

guessi (approved):

cartermckinnon:
@mhulscher I don't think the GPU operator specifically supports Amazon Linux at this point, though that support is being worked on as far as I know. My understanding is that the operator installs pre-compiled kernel modules, and I assume you'd run into issues if there was no such module compiled against the 5.10/5.15 headers. Just a guess; I'd probably open an issue with the operator maintainers.

learhy:
@cartermckinnon any updates on this PR? Do you have an estimate as to when it will land in 1.24?

Thanks!

RainofTerra:
We're currently building our own AMI that pulls in this change so we'd love to see it merged.

I'll probably file a separate issue for this, but it would be nice for 5.15 to be included as an option that we can pick at build time just so I don't have to maintain a (minor) fork to test later kernels on our nodes.

cartermckinnon:
@RainofTerra yep, I plan to make that change once we get this merged 👍

stacybird:
can't wait to see this merged.  :-)

camrakin (approved):

bryantbiggs:
oh dang - is it happening?!

learhy:
Hurrah!

camrakin:
We’re ready to merge this! We compared the 5.4 and 5.10 kernels’:


1. Conformance, using Kubernete’s end-to-end integration test suite, the aws-k8s-tester, and long-running synthetic workloads.
    1. We observed no functional difference between the two kernels when it comes to Kubernetes.
2. Performance, using stress-ng and netperf. 
    1. We observed both marginal increases and decreases across the test cases; overall representing equivalent performance.

We’re excited for users to leverage the features of 5.10 in their clusters! We encourage users to monitor their workloads for differences when upgrading, and to open an issue here if something seems off.

learhy:
Thanks, all!

stevehipwell:
@cartermckinnon based on the release notes am I correct in thinking this didn't make it into v20230203?

dims:
xref - https://github.com/aws/containers-roadmap/issues/1570

cartermckinnon:
@stevehipwell Correct, this will land in the next release. Should be in the next week or so.

PR #1119, opened 2022-12-08
Handle indentation when parsing `sandbox_image`

### LINKED ISSUE #767: Sandbox image script does not work if containerd config is indented
**What happened**:
In recent releases, containerd sandbox-image service has been added and an image needs to be defined in the containerd config file. The script [pull-sandbox-image](https://github.com/awslabs/amazon-eks-ami/blob/master/files/pull-sandbox-image.sh) is not able to fetch the image in the containerd config file if it is indented. 

When using default containerd config there is no problem. However, for anyone using custom config this may break. It also breaks when used in combination with nvidia gpu-operator which reads the current config file and re-writes it according to it's specification.

[Command](https://github.com/awslabs/amazon-eks-ami/blob/master/files/pull-sandbox-image.sh#L4) should work if `sandbox-image` is indented
Sandbox image script works with config below:
```
version = 2
...
...
```

Sandbox image script does NOT works with config below ( or any other variation where `sandbox_image` is indented x times):
```
version = 2
...
[plugins]
...
```

Amazon EKS AMI v20210813 or new

I would also suggest a default sandbox image is added as it took me a while to find the correct one.

Perhaps the [region](https://github.com/awslabs/amazon-eks-ami/blob/master/files/pull-sandbox-image.sh#L5)  can be fetched another way in case the sandbox_image is from a public ECR?

cartermckinnon:
**Issue #, if available:**

Resolves #767 

**Description of changes:**

Handles indentation on the `sandbox_image` key of the `containerd` config file. Works by just converting this line to valid bash and `source`-ing it. TOML comments are ignored because they're valid bash comments.

**Testing Done**

Executed:
```
#!/usr/bin/env bash

INDENTED=$(mktemp)

echo '[plugins]
    # sandbox_image comment about the sandbox_image
    sandbox_image = "one"' >> $INDENTED

FLAT=$(mktemp)

echo '[plugins]
# sandbox_image comment about the sandbox_image
sandbox_image = "two"' >> $FLAT

source <(grep "sandbox_image" $INDENTED | tr -d ' ')
echo "$sandbox_image"

source <(grep "sandbox_image" $FLAT | tr -d ' ')
echo "$sandbox_image"
```

Output:
```
one
two
```

### DIFF
--- a/files/pull-sandbox-image.sh
+++ b/files/pull-sandbox-image.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
-sandbox_image="$(awk -F'[ ="]+' '$1 == "sandbox_image" { print $2 }' /etc/containerd/config.toml)"
+source <(grep "sandbox_image" /etc/containerd/config.toml | tr -d ' ')
 
 ### Short-circuit fetching sandbox image if its already present
 if [[ "$(sudo ctr --namespace k8s.io image ls | grep $sandbox_image)" != "" ]]; then


files/pull-sandbox-image.sh
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
-sandbox_image="$(awk -F'[ ="]+' '$1 == "sandbox_image" { print $2 }' /etc/containerd/config.toml)"
+source <(grep "sandbox_image" /etc/containerd/config.toml | tr -d ' ')
```
guessi, line 4: I though it's a danger more to use "grep" here.

Imagine there's a comment line with "sandbox_image" but it just a comment?

cartermckinnon: TOML comments use a hash symbol (#) like bash, so it "just works". I added an example of this to my testing section in the description.

https://github.com/toml-lang/toml/blob/main/toml.md#user-content-comment

bwagner5 (approved):
very clever! LGTM

PR #1120, opened 2022-12-09
fix: find latest eksbuild version when caching startup images

bwagner5:
**Issue #, if available:**

**Description of changes:**

This PR fixes an issue when caching the startup images within the AMI build. The caching logic uses the Managed Add-On versions to find the Latest and Default VPC CNI and kube-proxy versions. However, the add-on version can be slightly different than the container version tag since it is mainly the manifest version. The add-on version should always be >= to the container tagged version since any image change would result in a manifest change too. This PR implements retires if an image cannot be pulled with an eksbuild version and decrements the build version until it gets to `eksbuild.1`. If `eksbuild.1` cannot be pulled, then the version does not exist. 

**Testing Done**

Built the AMI in us-east-2 and us-west-2 for K8s 1.22 and 1.24.

### DIFF
--- a/scripts/install-worker.sh
+++ b/scripts/install-worker.sh
@@ -403,7 +403,16 @@ if [[ "$CACHE_CONTAINER_IMAGES" == "true" && "$BINARY_BUCKET_REGION" != "us-iso-
   LATEST_KUBE_PROXY_VERSION=$(echo "${LATEST_KUBE_PROXY_FULL_VERSION}" | cut -d"-" -f1)
   LATEST_KUBE_PROXY_PLATFORM_VERSION=$(echo "${LATEST_KUBE_PROXY_FULL_VERSION}" | cut -d"-" -f2)
 
-  KUBE_PROXY_IMGS=(
+  #### Cache VPC CNI images starting with the addon default version and the latest version
+  VPC_CNI_ADDON_VERSIONS=$(aws eks describe-addon-versions --addon-name vpc-cni --kubernetes-version=${K8S_MINOR_VERSION})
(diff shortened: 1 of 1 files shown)


scripts/install-worker.sh
```diff
@@ -428,24 +429,33 @@ if [[ "$CACHE_CONTAINER_IMAGES" == "true" && "$BINARY_BUCKET_REGION" != "us-iso-
...
       continue
     fi
-    /etc/eks/containerd/pull-image.sh "${img}"
+    ## Since eksbuild.x version may not match the image tag, we need to decrement the eksbuild version until we find the latest image tag within the app semver
+    eksbuild_version="1"
+    if [[ ${img} == *'eksbuild.'* ]]; then
+      eksbuild_version=$(echo "${img}" | grep -o 'eksbuild\.[0-9]\+' | cut -d'.' -f2)
+    fi
+    ## iterate through decrementing the build version each time
+    for build_version in $(seq 1 "${eksbuild_version}" | sort -r -n); do
```
cartermckinnon, line 445: I think you can swap the arguments to `seq` to get a descending sequence

cartermckinnon (commented):
Looks fine, but we need to build all the k8s versions to verify this before merging. I'll kick that off now.

scripts/install-worker.sh
bwagner5: the mac seq works that way, gnu is a bit trickier, but works better than the reverse sort: 

```
##          FIRST       INC  LAST
seq ${eksbuild_version} -1 1
```

cartermckinnon:
My 1.20-1.24 builds all succeeded, LGTM.

cartermckinnon (approved):

PR #1121, opened 2022-12-09
Lint space errors

cartermckinnon:
**Description of changes:**

Noticed we have some files with erroneous trailing spaces. This PR adds a check for them to `make lint`.


cartermckinnon:
`lint` check failed, which is good! Pushing a revision that applies the formatting next.

cartermckinnon:
`lint` check succeeded 👍

mmerkes:
This is fine, but it there not a tool that just does the thing for you?

cartermckinnon:
I could probably find *something* to pull in, but I'm not sure it'd be meaningfully better than `sed` for this purpose, and it'd just grow the BOM. Seems like this is usually a feature of text editors or language-specific linters, and we need something more flexible here.

cartermckinnon:
Rebooting this PR. Updated the `lint` impl to use `git diff-tree` instead of `grep`. Fails as expected: https://github.com/awslabs/amazon-eks-ami/actions/runs/6563096967/job/17826538956?pr=1121

mmerkes (approved):
LGTM

PR #1126, opened 2022-12-13
Reflecting actual available ephemeral-storage for nodes

mirajranpura:
**Description of changes:**
By default the OS+k8s components consumes at-least 2+Gi on any empty node. Having the default values of KubeReserved:ephemeral-storage set to **1Gi**, leads to disk-pressure for pods heavily relying on ephemeral storage while the allocatable presented by kubelet is not exactly same as the OS free-available disk. 

**Impact** Adjusting KubeReserved:ephemeral-storage to **3Gi**, not only provides the actual usable blocks to k8s but also helps to skip the user-data script modification out of box use case eg managed node group.  

Example: An empty worker node with 20 GB disk
```
df -hk /
Filesystem              1K-blocks     Used          Available    Use%   Mounted on
/dev/nvme0n1p1          20959212      4297520       16661692     21%    / 
```

**Testing Done** 

**Case 1 (current)**: ephemeral-storage with 1Gi  
``` 
kubectl describe ip-172-31-x-x.ec2.internal  
...
Allocatable:
...
  ephemeral-storage:            18242267924
...
```

**Case 2 (Requesting with this pull)**: ephemeral-storage with 3Gi  
``` 
kubectl describe ip-172-31-x-x.ec2.internal  
...
Allocatable:
...
  ephemeral-storage:             16094784276
...
```

**Issue #, if available:**
https://github.com/aws/containers-roadmap/issues/1847


MartinEmrich:
Yes, this most probably solves exactly https://github.com/aws/containers-roadmap/issues/1847, thanks!

PR #1149, opened 2023-01-11
Pass through the INSTALL_DOCKER setting to allow using base AMIs that already have docker installed

swails:
**Issue #, if available:**

**Description of changes:**

We were unable to use the original packer configuration as-is because the base AMI we wanted to use (with the necessary GPU configuration) already had docker installed which crashed in the `install-worker.sh` step.

While the `install-worker.sh` script had an environment variable to skip the docker install, there was no way to trigger that option through packer variables. This adds the packer variable (with the default set to maintain current functionality) needed to make the docker install *actually* optional.

**Testing Done**

I ran the build via `make 1.21 <arguments>` where `<arguments>` were suitable for my account.  After these changes, the AMI built successfully.


cartermckinnon:
I don't have the full history on this environment variable, but I *think* this was added to workaround a quirk of our GPU AMI template. I'm not inclined to expose it as a Packer variable at this point, because Docker is on its way out of the AMI. It would be short-lived; we plan to skip Docker installation for 1.25 AMI's, and we already disallow Docker as the container runtime in 1.24 AMI's. What Kubernetes version are you targeting?

paulwhittemore:
Currently: make 1.21 <arguments>

With the understanding that EKS 1.21 has an EOL of 15 Feb 2023. On path to 1.25.

swails:
Right now I'm targeting EKS 1.21, although that will be updating.  But as far as I can tell the install-worker.sh provisioner is not optional and so will always be executed.  So even if later versions of EKS don't require docker, the provisioner will still try to install it, so this error would occur with any EKS version you target (even 1.24 or 1.25).

So if we use a base image (say `source_ami_id=ami-0909aa7ee6b0c13fa source_ami_owners=898082745236 source_ami_filter_name="Deep Learning AMI GPU PyTorch 1.13.1*"`) that already has docker installed and available, then the `install-workers.sh` script will fail when trying to add the docker group (which already exists).

cartermckinnon:
> as far as I can tell the install-worker.sh provisioner is not optional and so will always be executed. So even if later versions of EKS don't require docker, the provisioner will still try to install it, so this error would occur with any EKS version you target (even 1.24 or 1.25).

We'll modify `install-worker.sh` to skip this block for kubelet versions >= 1.25. We'll probably remove `INSTALL_DOCKER` entirely and replace it with such a check.

One approach that might work without adding a new variable is to check if `DOCKER_VERSION` is the empty string, and if so, skip docker installation. We'll probably want to do something like this anyway, because if a user has `docker_version=1.2.3` for a 1.25 build, we shouldn't just silently skip the installation. I'd have to think through whether this would break anyone, overloading an existing variable and such.

cartermckinnon:
We settled this in #1157; this env var will continue to function until we no longer support a Kubernetes version <= 1.24 (October 2023). At that time, the Docker installation logic will be removed. I think we should have exposed this as an option early on, but I don't think it makes sense at this point in the support cycle.

PR #1157, opened 2023-01-30
Skip docker installation for Kubernetes 1.25+

cartermckinnon:
**Description of changes:**

This skips the installation of Docker by default for Kubernetes 1.25+.

If a user has set `INSTALL_DOCKER`, that value will be used regardless of the Kubernetes version.

### DIFF
--- a/scripts/install-worker.sh
+++ b/scripts/install-worker.sh
@@ -3,6 +3,7 @@
 set -o pipefail
 set -o nounset
 set -o errexit
+set -o xtrace
 IFS=$'\n\t'
 export AWS_DEFAULT_OUTPUT="json"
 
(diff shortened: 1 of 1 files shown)


scripts/install-worker.sh
```diff
@@ -203,7 +203,11 @@ EOF
 
 sudo yum install -y device-mapper-persistent-data lvm2
 
-INSTALL_DOCKER="${INSTALL_DOCKER:-true}"
+if [[ -z "$INSTALL_DOCKER" ]]; then
```
mmerkes, line 206: Did you try building a 1.25 AMI?

dims:
@cartermckinnon is there a chance that someone may trip up when they built a 1.25+ AMI with this env var set to install docker .. will the kubelet fail? (since kubelet support for dockershim is no longer there?)

cartermckinnon:
@dims we configure the container runtime separately, so no issues will arise at runtime -- this is just a toggle for installing the docker packages. Some folks may want the docker CLI, buildkit, etc. to stick around in their custom AMI's.

Once we no longer support a k8s version with dockershim support, we'll remove the docker installation logic entirely and this option will go away.

scripts/install-worker.sh
cartermckinnon: Yep, I built both 1.24 and 1.25, both succeeded. Launched instances with both and verified that 1.24 had `docker`, and 1.25 didn't.

dims:
Ah cool thanks for the context @cartermckinnon

scripts/install-worker.sh
```diff
@@ -203,7 +203,12 @@ EOF
 
 sudo yum install -y device-mapper-persistent-data lvm2
 
-INSTALL_DOCKER="${INSTALL_DOCKER:-true}"
+if [[ ! -v "INSTALL_DOCKER" ]]; then
```
xr1776, line 206: style preference: I prefer "if conditional then else", rather than "if not-conditional then else".
so we could remove the ! and swap 207 and 209

xr1776 (approved):

mmerkes (approved):

PR #1158, opened 2023-01-31
Add ALAS issue workflow

cartermckinnon:
**Description of changes:**

This adds a scheduled workflow to create an issue for every `Important` or `Critical` ALAS bulletin published for AL2.

This will:
1. Notify the maintainers of new bulletins via the standard GH flow.
2. Give users a consistent place to track fixes.

The Action being used currently only supports excluding a feed item based on the title or description matching a regex. I think it'd be ideal if we could only include items that are particularly relevant to our use case, like for bulletins concerning the kernel or container stack.

On the other hand, it'll be quick to close issues if they aren't relevant, and then we have a searchable history of those decisions.

### DIFF
--- a/.github/workflows/alas-issues.yaml
+++ b/.github/workflows/alas-issues.yaml
@@ -9,7 +9,7 @@ on:
         required: true  
   schedule:
     # once an hour, at the top of hour
-    - cron: "0 0 * * *"
+    - cron: "0 * * * *"
 permissions:
   issues: write
 jobs:
@@ -20,7 +20,7 @@ jobs:
         with:
           repo-token: "${{ secrets.GITHUB_TOKEN }}"
           feed: "https://alas.aws.amazon.com/AL2/alas.rss"
-          dry-run: "false"
+          dry-run: "true"
           lastTime: "${{ github.event.inputs.window || '24h' }}"
           labels: "alas,alas/al2"
           titleFilter: "(medium|low)"
--- a/.github/workflows/alas-issues.yaml
+++ b/.github/workflows/alas-issues.yaml
@@ -21,6 +21,6 @@ jobs:
           repo-token: "${{ secrets.GITHUB_TOKEN }}"
           feed: "https://alas.aws.amazon.com/AL2/alas.rss"
           dry-run: "false"
-          lastTime: "${{ github.event.inputs.window }}"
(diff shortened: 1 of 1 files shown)


.github/workflows/alas-issues.yaml
```diff
@@ -0,0 +1,26 @@
...
+on:
+  workflow_dispatch:
+    inputs:
+      window:
+        description: "Only consider bulletins published within this relative time window (golang Duration)"
+        default: "24h"
+        required: true  
+  schedule:
+    # once an hour, at the top of hour
+    - cron: "0 0 * * *"
```
xr1776, line 12: Do we have confidence this won't miss anything? Every 24 look at that past 24.

.github/workflows/alas-issues.yaml
```diff
@@ -0,0 +1,26 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: guilhem/rss-issues-action@0.5.2
+        with:
+          repo-token: "${{ secrets.GITHUB_TOKEN }}"
+          feed: "https://alas.aws.amazon.com/AL2/alas.rss"
+          dry-run: "false"
+          lastTime: "${{ github.event.inputs.window || '24h' }}"
+          labels: "alas,alas/al2"
+          titleFilter: "(medium|low)"
```
xr1776, line 26: I'd expect to see the string "Important" appear somewhere in this file. Why does it not?

.github/workflows/alas-issues.yaml
cartermckinnon: Every *hour* look at the past 24 hours. So even if GitHub Actions has a lengthy outage that upends all executions, we're very unlikely to miss bulletins (and can always run it manually with a different value to backfill -- that's what the `workflow_dispatch` enables).

.github/workflows/alas-issues.yaml
cartermckinnon: If the `titleFilter` matches, the feed item is *excluded* (it's a negative filter).

One improvement we could PR to this existing Action (or make in our own) would be positive filters (see my comment about only including bulletins for certain packages, though I could argue either way on that).

xr1776: Ah so "Important" includes "Critical". got it.

xr1776 (approved):

PR #1172, opened 2023-02-07
Pause after kernel upgrade

cartermckinnon:
**Description of changes:**

Fixes a race condition introduced by #1118; the `upgrade-kernel.sh` provisioner was moved *after* the `install-additional-yum-repos.sh` provisioner, which had a `pause_before` directive that gave the instance time to reboot after the kernel was upgraded. The pause was no longer having the intended effect, leading to intermittent issues in which the instance would reboot during the `install-worker.sh` provisioner.

This moves the directive to the `upgrade-kernel.sh` provisioner as a more explicit `pause_after`.

**Testing Done**

`make 1.24` succeeds consistently.

### DIFF
--- a/eks-worker-al2.json
+++ b/eks-worker-al2.json
@@ -121,7 +121,6 @@
     },
     {
       "type": "shell",
-      "pause_before": "90s",
       "remote_folder": "{{ user `remote_folder`}}",
       "inline": [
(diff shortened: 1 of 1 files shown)


dims (approved):
LGTM

PR #1171, opened 2023-02-06
add bootstrap option to create a local NVMe RAID-0 for kubelet and containerd

bwagner5:
**Issue #, if available:**
 - https://github.com/awslabs/amazon-eks-ami/issues/1131

**Description of changes:**
 - Adds an option to the bootstrap script to setup a RAID-0 XFS of any NVMe instance storage disks, moves the contents of /var/lib/kubelet and /var/lib/containerd to the new RAID and then symlink those state dirs back in the root filesystem.
 - Installs mdadm in the AMI build process so that the RAID can be created
 - Safe to run on instances without NVMe instance storage disks, since steps will just be skipped. The script is also idempotent.
 - The implementation of the script is decoupled from the bootstrap script and doesn't require any configuration, so it's possible that the RAID script lives in the repo without including the bootstrap option. 

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

**Testing Done**

### c5d.4xlarge (1 NVMe instance storage disk)

```
> aws ec2 describe-instance-types --instance-type c5d.4xlarge
...
        "InstanceStorageInfo": {
            "Disks": [
                {
                    "Count": 1,
                    "SizeInGB": 400,
                    "Type": "ssd"
                }
            ],
            "EncryptionSupport": "required",
            "NvmeSupport": "required",
            "TotalSizeInGB": 400
        },
        "InstanceStorageSupported": true,
...

> /etc/eks/bootstrap.sh my-cluster --raid-local-disks true --apiserver-endpoint <...> --b64-cluster-ca <...> --container-runtime containerd --kubelet-extra-args --node-labels=karpenter.sh/capacity-type=on-demand,karpenter.sh/provisioner-name=default

> lsblk
NAME          MAJ:MIN RM   SIZE RO TYPE  MOUNTPOINT
nvme1n1       259:0    0 372.5G  0 disk
└─md127         9:127  0 372.4G  0 raid0 /mnt/kubernetes
nvme0n1       259:1    0    20G  0 disk
├─nvme0n1p1   259:2    0    20G  0 part  /
└─nvme0n1p128 259:3    0     1M  0 part

> find /var/lib/ -type l -ls
2222112    0 lrwxrwxrwx   1 root     root           44 Feb  6 19:34 /var/lib/cloud/instance -> /var/lib/cloud/instances/i-087f6f66328338083
[...]

PR #1176, opened 2023-02-07
bootstrap: provide option to store image and container data on instance store

otterley:
**Description of changes:**

This PR adds an `--enable-instance-store` option to to the bootstrap script. 

If and only if it is set to `true` (`--enable-instance-store true`), before taking any other actions, the bootstrap script will create an ext4 filesystem on top of the available instance storage volumes. If more than one instance store volume exists, a RAID 0 volume will be created first before building the filesystem. The filesystem will be mounted at `/mnt/instance_store`. Then all data from `/var/lib/{docker,containerd,kubelet/`  will be relocated into the filesystem. Finally, a `bind` mount will be created to alias the original directories to the files on the instance store volume.

The benefit of this option is to significantly improve the latency and throughput of image unpacking and container volume access. Instance store volumes are backed by fast NVMe local storage and are orders of magnitude faster than EBS volumes.

This option can safely be used on EC2 instances that do not have instance storage, although it will not be effective and will report a notice to this effect.

This option currently only works on EC2 Nitro instances.

**Testing Done**

Testing was performed on both c6g.large and c6gd.large instance types, using a set of managed EKS nodegroups. The override-bootstrap command was set to:

```
#!/bin/bash
/etc/eks/bootstrap.sh mycluster --enable-instance-store true
```

### DIFF
--- a/files/bootstrap.sh
+++ b/files/bootstrap.sh
@@ -32,6 +32,7 @@ function print_help {
   echo "--service-ipv6-cidr ipv6 cidr range of the cluster"
   echo "--enable-local-outpost Enable support for worker nodes to communicate with the local control plane when running on a disconnected Outpost. (true or false)"
   echo "--cluster-id Specify the id of EKS cluster"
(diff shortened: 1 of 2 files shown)


cartermckinnon:
This is very similar to #1171, @bwagner5 please add your thoughts.

PR #1177, opened 2023-02-07
Version lock kernel

Issacwww:
**Issue #, if available:**

**Description of changes:**
As we introduced version lock for kernel header/devel for GPU AMI, to prevent compatibility issue from upgrading kernel, we also version locking the kernel

**Testing Done**

`make 1.25`

```
==> Builds finished. The artifacts of successful builds are:
--> amazon-ebs: AMIs were created:
us-west-2: ami-045b361401eb3adf1
```

### DIFF
--- a/scripts/install-worker.sh
+++ b/scripts/install-worker.sh
@@ -73,8 +73,6 @@ sudo yum install -y \
 # Remove any old kernel versions. `--count=1` here means "only leave 1 kernel version installed"
 sudo package-cleanup --oldkernels --count=1 -y
 
-sudo yum versionlock kernel
-
 # Remove the ec2-net-utils package, if it's installed. This package interferes with the route setup on the instance.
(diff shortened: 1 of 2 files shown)


scripts/install-worker.sh
```diff
@@ -74,6 +74,8 @@ sudo yum install -y \
 # Remove any old kernel versions. `--count=1` here means "only leave 1 kernel version installed"
 sudo package-cleanup --oldkernels --count=1 -y
 
+sudo yum versionlock kernel
```
cartermckinnon, line 77: Can you move the installation of the versionlock plugin to `upgrade_kernel.sh` and then do the versionlock there, just to have everything in one place?

saurav-agarwalla (approved):
Worth calling out that the reason we're also adding this version lock is because these AMIs are tested with the Kernel version they're shipped with. If the Kernel is unintentionally updated as part of `yum update`, it could leads to instabilities.

PR #1184, opened 2023-02-14
Add Action to generate eni-max-pods.txt

cartermckinnon:
**Description of changes:**

This adds an Action that will generate `files/eni-max-pods.txt` once a day, and open a pull-request if the contents of the file change. This Action can also be triggered manually.

### DIFF
--- a/sync-eni-max-pods.yaml
+++ b/sync-eni-max-pods.yaml
@@ -1,0 +1,40 @@
+name: '[Sync] Update eni-max-pods.txt'
+on:
+  workflow_dispatch:
+  schedule:
+    # once a day
+    - cron: '0 0 * * *'
+permissions:
+  id-token: write
+  contents: write
+  pull-requests: write
+jobs:
+  update-max-pods:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: aws-actions/configure-aws-credentials@v1
+        with:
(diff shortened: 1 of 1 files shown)


sync-eni-max-pods.yaml
```diff
@@ -0,0 +1,40 @@
...
+      - uses: actions/checkout@v3
+        with:
+          repository: aws/amazon-vpc-cni-k8s
+          ref: refs/heads/master
+          path: amazon-vpc-cni-k8s/
+      - run: |
+          #!/usr/bin/env bash
+          set -o errexit
+          cd amazon-vpc-cni-k8s/
+          make generate-limits
```
Issacwww, line 33: Could you please link the change for avoiding update the timestampe as well?

Issacwww (approved):

sync-eni-max-pods.yaml
cartermckinnon: https://github.com/aws/amazon-vpc-cni-k8s/commit/4e5b5b35ae72033087eae9e14e597f33d020ba8a

cartermckinnon: The first time this runs, it will open a PR because the timestamp has been removed; but after that, it will only open a PR if the max pods values actually change.

Issue #1193, opened 2023-02-17
Recalling AMI Release v20230211 due to it breaking Kernel upgrades


saurav-agarwalla:
### Description

**If you don't upgrade the Kernel in EKS Optimized AMIs, you can stop reading.**

If you DO upgrade the Kernel that the EKS Optimized AMIs ship with, you should know that a recent bug caused Kernel upgrades to behave unexpectedly in the AMI Release v20230211. 

The bug was introduced as a result of [this change](https://github.com/awslabs/amazon-eks-ami/commit/58bdc867afbd7808a6f78ebb1141fa33264946fd) where we added a version lock to the Kernel. This was done to make sure that customers don't accidentally end up updating the Kernel version that the AMI ships with because our tests only validate the AMI with the Kernel that it ships with and updating it to a different version could cause issues/instabilities.

What went wrong in that change was that instead of locking only the Kernel which is shipped with the AMI by default, it version locks both `4.14` and `5.4` Kernels (for the AMIs which have `5.4` as default) as seen from the output in one of the affected AMIs:
```
$ sudo yum versionlock
Loaded plugins: priorities, update-motd, versionlock
0:kernel-4.14.301-224.520.amzn2.
0:kernel-5.4.228-132.418.amzn2.*
0:runc-1.1.4-1.amzn2.*
0:containerd-1.6.6-1.amzn2.0.2.*
0:docker-20.10.17-1.amzn2.0.1.*
versionlock list done
```

As a result of this, if you upgrade the Kernel, it can end up bringing it down to `4.14`. Effectively, something like the following:
```
$ sudo amazon-linux-extras install -y kernel-5.10
Installing kernel
Loaded plugins: priorities, update-motd, versionlock
Cleaning repos: amzn2-core amzn2extra-docker amzn2extra-kernel-5.10
...
...
Resolving Dependencies
--> Running transaction check
---> Package kernel.x86_64 0:4.14.301-224.520.amzn2 will be installed
--> Finished Dependency Resolution

Dependencies Resolved

============================================================================================================================================================================================================================================================================================================
 Package                                                             Arch                                                                Version                                                                              Repository                                                               Size
============================================================================================================================================================================================================================================================================================================
Installing:
 kernel                                                              x86_64                                                              4.14.301-224.520.amzn2                                                               amzn2-core                                                               22 M

Transaction Summary
============================================================================================================================================================================================================================================================================================================
Install  1 Package

Total download size: 22 M
[...]

PR #1195, opened 2023-02-21
Validate versionlocks

cartermckinnon:
**Issue #, if available:**

#1193 

**Description of changes:**

Adds some validation on kernel versionlocks, since multiple locks can lead to unexpected behavior.

**Testing Done**

Succeeds:
```
> make 1.25
...
2023-03-06T13:02:56-08:00: ==> amazon-ebs: Provisioning with shell script: awslabs/amazon-eks-ami/scripts/validate.sh
2023-03-06T13:02:57-08:00:     amazon-ebs: Verifying that kernel version 5.10.167-147.601.amzn2.x86_64 matches ...
2023-03-06T13:02:57-08:00:     amazon-ebs: Kernel matches expected version!
2023-03-06T13:02:57-08:00:     amazon-ebs: Verifying that the package versionlocks are correct...
2023-03-06T13:02:57-08:00:     amazon-ebs: Package versionlocks are correct!
```

Fails (as expected):
```
> git checkout 8b10a76977278e7afec85b51c326d17392565fd6
> curl -L https://github.com/awslabs/amazon-eks-ami/pull/1195.patch | git apply
> make 1.24
...
2023-03-06T13:28:35-08:00: ==> amazon-ebs: Provisioning with shell script: awslabs/amazon-eks-ami/scripts/validate.sh
2023-03-06T13:28:35-08:00:     amazon-ebs: Verifying that kernel version 5.10.167-147.601.amzn2.x86_64 matches ...
2023-03-06T13:28:35-08:00:     amazon-ebs: Kernel matches expected version!
2023-03-06T13:28:35-08:00:     amazon-ebs: Verifying that the package versionlocks are correct...
2023-03-06T13:28:36-08:00:     amazon-ebs: There is no package matching the versionlock entry: 'kernel-4.14.305-227.531.amzn2.*'
2023-03-06T13:28:36-08:00: ==> amazon-ebs: Provisioning step had errors: Running the cleanup provisioner, if present...

```

### DIFF
--- a/scripts/validate.sh
+++ b/scripts/validate.sh
@@ -46,7 +46,7 @@ else
 fi
 
 echo "Verifying that the package versionlocks are correct..."
-  
+
 function versionlock-entries() {
   # the format of this output is EPOCH:NAME-VERSION-RELEASE.ARCH
   # more info in yum-versionlock(1)
@@ -55,7 +55,7 @@ function versionlock-entries() {
 }
 
 function versionlock-packages() {
-  versionlock-entries | xargs -I '{}' rpm --query  '{}' --queryformat '%{NAME}\n'
+  versionlock-entries | xargs -I '{}' rpm --query '{}' --queryformat '%{NAME}\n'
 }
 
 for ENTRY in $(versionlock-entries); do
--- a/scripts/validate.sh
+++ b/scripts/validate.sh
@@ -36,11 +36,30 @@ validate_file_nonexists '/var/log/secure'
 validate_file_nonexists '/var/log/wtmp'
 
 actual_kernel=$(uname -r)
-echo "Verifying that kernel version $actual_kernel matches $KERNEL_VERSION"
+echo "Verifying that kernel version $actual_kernel matches $KERNEL_VERSION..."
 
 if [[ $actual_kernel == $KERNEL_VERSION* ]]; then
-  echo "Kernel matches expected version"
+  echo "Kernel matches expected version!"
 else
-  echo "Kernel does not match expected version."
+  echo "Kernel does not match expected version!"
   exit 1
 fi
+
+echo "Verifying that the kernel has the correct versionlock..."
+
+# only one version of the kernel should be version locked
+if [ $(yum versionlock list --quiet | grep -c "kernel") -ne 1 ]; then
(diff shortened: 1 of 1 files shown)


mmerkes:
Did you validate that this works with the problematic AMI?

cartermckinnon:
Yep! Updated the PR description.

mmerkes (approved):

scripts/validate.sh
```diff
@@ -36,11 +36,29 @@ validate_file_nonexists '/var/log/secure'
...
+  echo "Kernel matches expected version!"
 else
-  echo "Kernel does not match expected version."
+  echo "Kernel does not match expected version!"
   exit 1
 fi
+
+echo "Verifying that the kernel has the correct versionlock..."
+
+# only one version of the kernel should be version locked
```
saurav-agarwalla, line 50: Should we generalize this to other packages too? I can't think of a reason why we'd versionlock multiple versions of the same package.

cartermckinnon: Yeah, I considered that. The format of the versionlock file isn't particularly conducive to that sort of check, though it's not impossible.

The format is `EPOCH:NAME-VERSION-RELEASE.ARCH`, more here: https://man7.org/linux/man-pages/man1/yum-versionlock.1.html
```
0:kernel-4.14.304-226.531.amzn2.*
0:kernel-5.10.165-143.735.amzn2.*
0:runc-1.1.4-1.amzn2.*
0:containerd-1.6.6-1.amzn2.0.2.*
```

We need to convert these entries to just a list of `NAME`-s. Packages often use `-` as a delimiter in their names, so we can't just split on that (e.g. `nvidia-driver-latest-dkms`). We could chop off the `EPOCH` and query the index, like:
```
> rpm -q 'kernel-4.14.304-226.531.amzn2.*' --queryformat '%{NAME}\n'
kernel
```

But this won't detect the issue we saw in 20220211, because the `kernel-4.14` package was no longer installed:
```
> rpm -q kernel-4.14
package kernel-4.14 is not installed
```

We'll need to delete any versionlocks for packages that aren't currently installed, *then* do this check on the uniqueness of the package names. I'll put up a revision.

Issacwww (approved):

PR #1197, opened 2023-02-22
Use gp3 volume_type for 1.27+

### LINKED ISSUE #626: Make gp3 the default EBS volume type

In December 2020, [Amazon EBS announced gp3 volumes](https://aws.amazon.com/about-aws/whats-new/2020/12/introducing-new-amazon-ebs-general-purpose-volumes-gp3/), the next generation of general purpose volumes. gp3 volumes are more performant, flexible and have lower costs than gp2 volumes, the current default, and they are available in all commercial and govcloud regions.

We should verify that there are no downsides to this and would not impact any existing customers, but seems like a reasonable change.

Changing it [here](https://github.com/awslabs/amazon-eks-ami/blob/850e0029c34c0dfcb8788a1433759d1cc8cef2b2/eks-worker-al2.json#L63-L71) would make it the default, and we should consider adding a new user variable to set the volume type to simplify it for customers who want to use a different volume type.

And probably update the [CFN template](https://github.com/awslabs/amazon-eks-ami/blob/f85cf415b90acaafb2aebc1cf845b6a2282a0ada/amazon-eks-nodegroup.yaml#L629).

**Why is this needed**:

gp3 volumes are the latest generation of EBS volumes and offer benefits over gp2 AMIs.

cartermckinnon:
**Issue #, if available:**

Closes #626.

**Description of changes:**

Changes the `volume_type` to `gp3` for Kubernetes 1.27+, if the user has not specified a `volume_type` either in their `PACKER_VARIABLE_FILE` or as an argument to `make`.

**Testing Done**

Succeeds and the resulting AMI is `gp2`:
```
make 1.25
```

TODO:
```
make 1.27
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,14 +16,14 @@ vercmp = $(shell $(MAKEFILE_DIR)/files/bin/vercmp "$1" "$2" "$3")
 # gp2 volumes are used by default prior to 1.26
 # TODO: remove this when 1.25 reaches EOL
 ifeq ($(call packer_variable_file_contains,volume_type), false)
-	ifeq ($(call vercmp, $(kubernetes_version), lt, 1.26.0), true)
+	ifeq ($(call vercmp,$(kubernetes_version),lt,1.26.0), true)
 		volume_type ?= gp2
 	endif
 endif
 
 # Docker is not present on 1.25+ AMI's
 # TODO: remove this when 1.24 reaches EOL
-ifeq ($(call vercmp, $(kubernetes_version), gteq, 1.25.0), true)
+ifeq ($(call vercmp,$(kubernetes_version),gteq,1.25.0), true)
 	# do not tag the AMI with the Docker version
 	docker_version ?= none
 	# do not include the Docker version in the AMI description
--- a/Makefile
+++ b/Makefile
@@ -13,11 +13,11 @@ packer_variable_file_contains = $(if $(PACKER_VARIABLE_FILE),$(shell grep -Fq $1
 
 vercmp = $(shell $(MAKEFILE_DIR)/files/bin/vercmp "$1" "$2" "$3")
 
-# gp2 volumes are used by default prior to 1.26
+# gp3 volumes are used by default for 1.26+
 # TODO: remove this when 1.25 reaches EOL
 ifeq ($(call packer_variable_file_contains,volume_type), false)
-	ifeq ($(call vercmp,$(kubernetes_version),lt,1.26.0), true)
-		volume_type ?= gp2
+	ifeq ($(call vercmp,$(kubernetes_version),gteq,1.26.0), true)
+		volume_type ?= gp3
 	endif
 endif
 
--- a/Makefile
+++ b/Makefile
@@ -6,29 +6,45 @@ K8S_VERSION_MINOR := $(word 1,${K8S_VERSION_PARTS}).$(word 2,${K8S_VERSION_PARTS
 
 MAKEFILE_DIR := $(shell dirname $(realpath $(firstword $(MAKEFILE_LIST))))
 
+# expands to 'true' if PACKER_VARIABLE_FILE is non-empty
+# and the file contains the string passed as the first argument
+# otherwise, expands to 'false'
+packer_variable_file_contains = $(if $(PACKER_VARIABLE_FILE),$(shell grep -Fq $1 $(PACKER_VARIABLE_FILE) && echo true || echo false),false)
+
+vercmp = $(shell $(MAKEFILE_DIR)/files/bin/vercmp "$1" "$2" "$3")
+
+# gp2 volumes are used by default prior to 1.26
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -6,29 +6,45 @@ K8S_VERSION_MINOR := $(word 1,${K8S_VERSION_PARTS}).$(word 2,${K8S_VERSION_PARTS
...
+# otherwise, expands to 'false'
+packer_variable_file_contains = $(if $(PACKER_VARIABLE_FILE),$(shell grep -Fq $1 $(PACKER_VARIABLE_FILE) && echo true || echo false),false)
+
+vercmp = $(shell $(MAKEFILE_DIR)/files/bin/vercmp "$1" "$2" "$3")
+
+# gp3 volumes are used by default for 1.26+
+# TODO: remove this when 1.25 reaches EOL
+ifeq ($(call packer_variable_file_contains,volume_type), false)
+	ifeq ($(call vercmp,$(kubernetes_version),gteq,1.26.0), true)
+		volume_type ?= gp3
```
ravisinha0506, line 20: We should do a region check to exclude gp3 where it is not supported ( us-iso-east/west-1 )

Makefile
```diff
@@ -6,29 +6,49 @@ K8S_VERSION_MINOR := $(word 1,${K8S_VERSION_PARTS}).$(word 2,${K8S_VERSION_PARTS
...
+packer_variable_file_contains = $(if $(PACKER_VARIABLE_FILE),$(shell grep -Fq $1 $(PACKER_VARIABLE_FILE) && echo true || echo false),false)
+
+vercmp = $(shell $(MAKEFILE_DIR)/files/bin/vercmp "$1" "$2" "$3")
+
+in_iso_region = $(if $(findstring us-iso,$(aws_region)),true,false)
+
+# gp3 volumes are used by default for 1.26+
+# TODO: remove this when 1.25 reaches EOL
+ifeq ($(call packer_variable_file_contains,volume_type), false)
+	ifeq ($(call vercmp,$(kubernetes_version),gteq,1.26.0), true)
```
mmerkes, line 21: This going to actually happen for 1.26?

Makefile
```diff
@@ -6,29 +6,49 @@ K8S_VERSION_MINOR := $(word 1,${K8S_VERSION_PARTS}).$(word 2,${K8S_VERSION_PARTS
...
+# otherwise, expands to 'false'
+packer_variable_file_contains = $(if $(PACKER_VARIABLE_FILE),$(shell grep -Fq $1 $(PACKER_VARIABLE_FILE) && echo true || echo false),false)
+
+vercmp = $(shell $(MAKEFILE_DIR)/files/bin/vercmp "$1" "$2" "$3")
+
+in_iso_region = $(if $(findstring us-iso,$(aws_region)),true,false)
+
+# gp3 volumes are used by default for 1.26+
+# TODO: remove this when 1.25 reaches EOL
+ifeq ($(call packer_variable_file_contains,volume_type), false)
```
mmerkes, line 20: No comprendo. Does this skip cases when `gp2` is set in the variable file?

cartermckinnon:
Pushing this to 1.27, as there's some service-side work to be done and 1.26 is imminent.

Makefile
cartermckinnon: If the user has something set for `volume_type`, then we don't touch it.

Makefile
mmerkes: Did you forget to push a commit again?

cartermckinnon: I was updating things as I responded, it's pushed now 👍

Makefile
```diff
@@ -6,29 +6,54 @@ K8S_VERSION_MINOR := $(word 1,${K8S_VERSION_PARTS}).$(word 2,${K8S_VERSION_PARTS
...
+
+# gp3 volumes are used by default for 1.27+
+# TODO: remove when 1.26 reaches EOL
+# TODO: remove when gp3 is supported in isolated regions
+ifneq ($(call packer_variable_file_contains,volume_type), true)
+	ifeq ($(call in_iso_region), true)
+		volume_type ?= gp2
+	else ifeq ($(call vercmp,$(kubernetes_version),lt,1.27.0), true)
+		volume_type ?= gp2
+	endif
```
cartermckinnon, line 30: `make` has an `or` function, but it treats the empty string as `false` and any non-empty string as `true`. That's easy enough to work with, but it makes the `vercmp` and `in_iso_region` a bit clunkier to work with in general, and the resulting `or` is just a longer, less readable line. 🤷

mmerkes: Not totally following here. Are you explaining why you have two `if`s that result in the first action versus combining these checks into one line?

cartermckinnon: Yeah, the repetition of `volume_type ?= gp2` isn't ideal, but this is nicer than the alternative

mmerkes (approved):

Issue #1164, opened 2023-02-01
Make awscli instalation more configurable


im-wanyama:
**What would you like to be added**:

Currently we can't run the install-worker.sh script because the base AMI we use is CIS hardened which prevents executable files from running within /tmp.

Could we make the process of installing awscli more configurable?

For example allowing the user to specify where they download/run the install executable or allowing to user to install awscli via yum, irrespective of what binary bucket they've specified.

**Why is this needed**:

This will allow users more flexibility in using the install-worker.sh script and account for variance in setup/requirements.

cartermckinnon:
Feel free to PR a change to the working directory if `/tmp` doesn't work for your use case: https://github.com/awslabs/amazon-eks-ami/blob/master/scripts/install-worker.sh#L130

Just make sure you `rm` the zip archive and extracted installer.

ShibraAmin18:
@cartermckinnon changed `/tmp` to `/home/ec2-user ` and   `--bin-dir /bin/` to `/usr/local/bin` [https://github.com/awslabs/amazon-eks-ami/blob/343e830dc4cbf0b2646026fced4cea7202a345a5/scripts/install-worker.sh#L138](https://github.com/awslabs/amazon-eks-ami/blob/343e830dc4cbf0b2646026fced4cea7202a345a5/scripts/install-worker.sh#L138
)

When used the image created with EKS node groups, caanot exec into or view pod logs. 

`Error from server: error dialing backend: dial tcp 10.10.71.57:10250: i/o timeout`

im-wanyama:
> 

Ah no problem, assign this issue to me and I'll make a PR

cartermckinnon:
Should be resolved by #1231

PR #1198, opened 2023-02-22
Updating install worker script to accept a user defined directory whe…

im-wanyama:
…re the aws install can take place.

**Issue #, if available:**
[#1164](https://github.com/awslabs/amazon-eks-ami/issues/1164)

**Description of changes:**

I've updated the install-worker script to enable users to specify where the aws install takes place. The default behavior has been left unaltered. This allows us to install aws v2 without using a tmp directory. Our base AMI is CIS hardened and doesn't allow executables with in tmpfs.

**Testing Done**

I've verified that the AMI still builds and I've tested it in our dev environment.

### DIFF
--- a/scripts/install-worker.sh
+++ b/scripts/install-worker.sh
@@ -122,15 +122,25 @@ sudo mv $TEMPLATE_DIR/iptables-restore.service /etc/eks/iptables-restore.service
 if [[ "$BINARY_BUCKET_REGION" != "us-iso-east-1" && "$BINARY_BUCKET_REGION" != "us-isob-east-1" ]]; then
   # https://docs.aws.amazon.com/cli/latest/userguide/getting-started-install.html
   echo "Installing awscli v2 bundle"
-  AWSCLI_DIR=$(mktemp -d)
-  curl \
-    --silent \
+  AWSCLI_DIR="${AWSCLI_DIR:-""}"
+  if [[ -n $AWSCLI_DIR ]]; then
+    AWSCLI_DIR=$AWSCLI_DIR
+    sudo mkdir -p $AWSCLI_DIR
+  else
+    AWSCLI_DIR=$(mktemp -d)
+  fi
+  sudo curl \
     --show-error \
     --retry 10 \
     --retry-delay 1 \
     -L "https://awscli.amazonaws.com/awscli-exe-linux-${MACHINE}.zip" -o "${AWSCLI_DIR}/awscliv2.zip"
-  unzip -q "${AWSCLI_DIR}/awscliv2.zip" -d ${AWSCLI_DIR}
-  sudo "${AWSCLI_DIR}/aws/install" --bin-dir /bin/
+  sudo unzip -q "${AWSCLI_DIR}/awscliv2.zip" -d ${AWSCLI_DIR}
+  sudo "${AWSCLI_DIR}/aws/install" --bin-dir /bin --update
+
+  if [[ -n $AWSCLI_DIR ]]; then
+    sudo rm -rf $AWSCLI_DIR
+  fi
+
 else
   echo "Installing awscli package"
   sudo yum install -y awscli
--- a/scripts/install-worker.sh
+++ b/scripts/install-worker.sh
@@ -33,6 +33,7 @@ validate_env_set KUBERNETES_BUILD_DATE
 validate_env_set PULL_CNI_FROM_GITHUB
 validate_env_set PAUSE_CONTAINER_VERSION
 validate_env_set CACHE_CONTAINER_IMAGES
+validate_env_set AWSCLI_DIR
 
 ################################################################################
 ### Machine Architecture #######################################################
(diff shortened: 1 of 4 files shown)


scripts/install-worker.sh
```diff
@@ -122,15 +122,25 @@ sudo mv $TEMPLATE_DIR/iptables-restore.service /etc/eks/iptables-restore.service
...
     --retry 10 \
     --retry-delay 1 \
     -L "https://awscli.amazonaws.com/awscli-exe-linux-${MACHINE}.zip" -o "${AWSCLI_DIR}/awscliv2.zip"
-  unzip -q "${AWSCLI_DIR}/awscliv2.zip" -d ${AWSCLI_DIR}
-  sudo "${AWSCLI_DIR}/aws/install" --bin-dir /bin/
+  sudo unzip -q "${AWSCLI_DIR}/awscliv2.zip" -d ${AWSCLI_DIR}
+  sudo "${AWSCLI_DIR}/aws/install" --bin-dir /bin --update
+
+  if [[ -n $AWSCLI_DIR ]]; then
+    sudo rm -rf $AWSCLI_DIR
```
cartermckinnon, line 141: Not sure deleting the directory is the right thing here, we should just clean up the resources we created.

scripts/install-worker.sh
```diff
@@ -122,15 +122,25 @@ sudo mv $TEMPLATE_DIR/iptables-restore.service /etc/eks/iptables-restore.service
 if [[ "$BINARY_BUCKET_REGION" != "us-iso-east-1" && "$BINARY_BUCKET_REGION" != "us-isob-east-1" ]]; then
   # https://docs.aws.amazon.com/cli/latest/userguide/getting-started-install.html
   echo "Installing awscli v2 bundle"
-  AWSCLI_DIR=$(mktemp -d)
-  curl \
-    --silent \
+  AWSCLI_DIR="${AWSCLI_DIR:-""}"
+  if [[ -n $AWSCLI_DIR ]]; then
+    AWSCLI_DIR=$AWSCLI_DIR
```
cartermckinnon, line 127: This is a no-op.

cartermckinnon (changes requested):
I don't think we should make this configurable via an environment variable, or at least not *just* for the AWS CLI.

The issue is that `/tmp` may not--in general--be the right working directory for AMI build, right? So, we could create a proper template variable for `working_directory`, for example, and then use that instead of `/tmp` throughout the build process.

Otherwise, I'd prefer to just choose a working directory for the CLI installation that works for more use cases, something like the home directory of the packer user.

scripts/install-worker.sh
im-wanyama: What do you mean by clean up, should i move the rm command to the cleanup script?

scripts/install-worker.sh
im-wanyama: The line does work, the default behaviour goes to mktemp because AWSCLI_DIR is "" by default. If you parse the environment variable then it uses the directory you parse

im-wanyama:
> I don't think we should make this configurable via an environment variable, or at least not _just_ for the AWS CLI.
> 
> The issue is that `/tmp` may not--in general--be the right working directory for AMI build, right? So, we could create a proper template variable for `working_directory` (that defaults to `/tmp`), for example, and then use that throughout the build process.
> 
> Otherwise, I'd prefer to just choose a working directory for the AWS CLI installation that's suitable for more use cases, something like the home directory of the packer user.

I can parse it as a variable if you want

scripts/install-worker.sh
cartermckinnon: I'm referring to this:
```
AWSCLI_DIR=$AWSCLI_DIR
```

Assigning a variable to itself has no effect.

im-wanyama: Ah i see, yes. I'll remove that

im-wanyama:
I believe I've actioned all your comments. I've tested in my environment and running the unit tests looks to have worked.

im-wanyama:
I've fixed the linting error.

cartermckinnon:
I don't think it makes sense to have a template variable specifically for the working directory of the CLI installation.

If you want to refactor this into a more general variable for the working directory of the wider build process, that's reasonable.

cartermckinnon (changes requested):
I have the same feedback as in earlier revisions. I don't think it makes sense to have a template variable specifically for the working directory of the CLI installation.

If you want to refactor this into a more general variable for the working directory of the wider build process, that's reasonable.

im-wanyama:
Hi, I've refactored it to use a more general template variable. I've not using it anywhere else as there isn't a needed, but it can be used in future if required.

cartermckinnon:
I've continued this effort in #1231, using the `working_dir` in the template spec as well. Give it a look if you can!

im-wanyama:
> I've continued this effort in #1231, using the `working_dir` in the template spec as well. Give it a look if you can!

Hi, had a look and left a commment. LGTM.

PR #1203, opened 2023-02-23
Use --cloud-provider=external on 1.26+

### LINKED ISSUE #1096: Use external cloud provider instead of aws
I would like the `kubelet` to be initialized & use the external cloud provider rather than AWS. We should provide the user an option to customize which provider it uses on the worker node. 

**Why is this needed**:
Using an external cloud provider is the default way that EKS creates clusters post 1.22. We should give the option to customers to use the external cloud provider rather than the in-tree aws provider.

cartermckinnon:
**Issue #, if available:**

Closes #1096.

**Description of changes:**

This will move 1.26+ AMI's away from kubelet's in-tree cloud provider. This requires specifying a `--provider-id`.

**Testing Done**

Unit tests added.

@camrakin has run 1.26 sonobuoy conformance against this branch, and it passed.

### DIFF
--- a/files/bin/provider-id
+++ b/files/bin/provider-id
@@ -6,4 +6,4 @@ set -o nounset
 AVAILABILITY_ZONE=$(imds '/latest/meta-data/placement/availability-zone')
 INSTANCE_ID=$(imds '/latest/meta-data/instance-id')
 
-echo "aws://$AVAILABILITY_ZONE/$INSTANCE_ID"
+echo "aws:///$AVAILABILITY_ZONE/$INSTANCE_ID"
--- a/files/bin/provider-id
+++ b/files/bin/provider-id
@@ -1,0 +1,9 @@
+#!/usr/bin/env bash
+
+set -o errexit
+set -o nounset
+
+AVAILABILITY_ZONE=$(imds '/latest/meta-data/placement/availability-zone')
+INSTANCE_ID=$(imds '/latest/meta-data/instance-id')
+
+echo "aws://$AVAILABILITY_ZONE/$INSTANCE_ID"
--- a/files/bootstrap.sh
+++ b/files/bootstrap.sh
@@ -462,7 +462,7 @@ if vercmp "$KUBELET_VERSION" lt "1.26.0"; then
   KUBELET_CLOUD_PROVIDER="aws"
 else
   KUBELET_CLOUD_PROVIDER="external"
-  KUBELET_ARGS="$KUBELET_ARGS --provider-id=$(provider-id)"
+  echo "$(jq ".providerID=\"$(provider-id)\"" $KUBELET_CONFIG)" > $KUBELET_CONFIG
 fi
 
 KUBELET_ARGS="$KUBELET_ARGS --cloud-provider=$KUBELET_CLOUD_PROVIDER"
--- a/files/bootstrap.sh
+++ b/files/bootstrap.sh
@@ -455,11 +455,23 @@ if [[ "$USE_MAX_PODS" = "true" ]]; then
   echo "$(jq ".maxPods=$MAX_PODS" $KUBELET_CONFIG)" > $KUBELET_CONFIG
 fi
 
+KUBELET_ARGS="--node-ip=$INTERNAL_IP --pod-infra-container-image=$PAUSE_CONTAINER --v=2"
+
+if vercmp "$KUBELET_VERSION" lt "1.26.0"; then
+  # TODO: remove this when 1.25 is EOL
+  KUBELET_CLOUD_PROVIDER="aws"
+else
+  KUBELET_CLOUD_PROVIDER="external"
+  KUBELET_ARGS="$KUBELET_ARGS --provider-id=$(provider-id)"
+fi
+
+KUBELET_ARGS="$KUBELET_ARGS --cloud-provider=$KUBELET_CLOUD_PROVIDER"
+
 mkdir -p /etc/systemd/system/kubelet.service.d
 
 cat << EOF > /etc/systemd/system/kubelet.service.d/10-kubelet-args.conf
 [Service]
-Environment='KUBELET_ARGS=--node-ip=$INTERNAL_IP --pod-infra-container-image=$PAUSE_CONTAINER --v=2'
+Environment='KUBELET_ARGS=$KUBELET_ARGS'
 EOF
 
 if [[ -n "$KUBELET_EXTRA_ARGS" ]]; then
--- a/test/cases/cloud-provider-config.sh
+++ b/test/cases/cloud-provider-config.sh
@@ -5,14 +5,20 @@ set -o pipefail
 set -o nounset
 
 KUBELET_UNIT_DIR="/etc/systemd/system/kubelet.service.d"
+KUBELET_CONFIG_FILE="/etc/kubernetes/kubelet/kubelet-config.json"
 
 function fail() {
   echo "❌ Test Failed:" "$@"
+  echo "Kubelet systemd units:"
   find $KUBELET_UNIT_DIR -type f | xargs cat
+  echo "Kubelet config file:"
+  cat $KUBELET_CONFIG_FILE | jq '.'
   exit 1
 }
 
-echo "--> Should use in-tree cloud provider when below k8s version 1.26"
+EXPECTED_PROVIDER_ID=$(provider-id)
+
+echo "--> Should use in-tree cloud provider below k8s version 1.26"
 # This variable is used to override the default value in the kubelet mock
 export KUBELET_VERSION=v1.25.5-eks-ba74326
 EXIT_CODE=0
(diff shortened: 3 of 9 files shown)


cartermckinnon:
I suppose we can go ahead and move these parameters from the CLI to the config file, I'll do that in the next revision.

files/bin/provider-id
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+set -o errexit
+set -o nounset
+
+AVAILABILITY_ZONE=$(imds '/latest/meta-data/placement/availability-zone')
+INSTANCE_ID=$(imds '/latest/meta-data/instance-id')
```
mmerkes, line 7: We should validate these. I don't know if the script will exit or not if these API calls fail, but we definitely don't want to have bogus provider IDs.

test/cases/cloud-provider-config.sh
```diff
@@ -0,0 +1,77 @@
...
+export KUBELET_VERSION=v1.25.5-eks-ba74326
+EXIT_CODE=0
+/etc/eks/bootstrap.sh \
+  --b64-cluster-ca dGVzdA== \
+  --apiserver-endpoint http://my-api-endpoint \
+  test || EXIT_CODE=$?
+if [[ ${EXIT_CODE} -ne 0 ]]; then
+  fail "expected a zero exit code but got '${EXIT_CODE}'"
+fi
+EXIT_CODE=0
```
mmerkes, line 26: Doesn't look like this is necessary because we'll fail otherwise.

test/cases/cloud-provider-config.sh
```diff
@@ -0,0 +1,77 @@
...
+grep -RFq -e "--cloud-provider=aws" $KUBELET_UNIT_DIR || EXIT_CODE=$?
+if [[ ${EXIT_CODE} -ne 0 ]]; then
+  fail "expected --cloud-provider=aws to be present in kubelet's systemd units"
+fi
+EXIT_CODE=0
+grep -RFq -e "--provider-id=$(provider-id)" $KUBELET_UNIT_DIR || EXIT_CODE=$?
```
mmerkes, line 32: Might want to just check for the flag, not the full thing, because this could succeed due to a mismatch in provider ID, not that it wouldn't cause other issues if that went wrong...

test/cases/provider-id.sh
```diff
@@ -0,0 +1,25 @@
...
+EXPECTED_PROVIDER_ID="aws://$EXPECTED_AVAILABILITY_ZONE/$EXPECTED_INSTANCE_ID"
+PROVIDER_ID=$(provider-id)
+if [ ! "$PROVIDER_ID" = "$EXPECTED_PROVIDER_ID" ]; then
+  echo "❌ Test Failed: expected provider-id=$EXPECTED_PROVIDER_ID but got '${PROVIDER_ID}'"
+  exit 1
+fi
+
+echo "--> Should fail when imds is unreachable"
+echo '#!/usr/bin/sh
+exit 1' > $(which imds)
```
mmerkes, line 19: Will this get refreshed when other test cases run? I assume so.

test/cases/provider-id.sh
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
+
+set -o nounset
+set -o errexit
+set -o pipefail
+
+echo "--> Should fetch imds details correctly"
+EXPECTED_INSTANCE_ID="i-1234567890abcdef0"
+EXPECTED_AVAILABILITY_ZONE="us-east-1a"
+EXPECTED_PROVIDER_ID="aws://$EXPECTED_AVAILABILITY_ZONE/$EXPECTED_INSTANCE_ID"
```
mmerkes, line 10: From what I can tell, our provider ID has 3 slashes. Actual provider ID: `aws:///us-west-2d/i-052b1a849f1d6c78d
`.

test/cases/provider-id.sh
mmerkes: Also, per my comment above, is this the only way the imds call will fail? Want to make sure we don't miss out on any funkiness.

files/bin/provider-id
cartermckinnon: You mean validate the format of the instance ID and AZ name? The unit test validates that we use whatever we get from IMDS, and also validates that `provider-id` fails if `imds` does.

test/cases/provider-id.sh
cartermckinnon: We use a fresh container for each test case, yeah.

The `imds` script could fail for plenty of reasons, but would exit with a non-zero code. `set -o errexit` in `provider-id` would bail if `imds` croaks.

test/cases/provider-id.sh
cartermckinnon: Ah, yeah, you're right. Will fix.

test/cases/cloud-provider-config.sh
cartermckinnon: Fair enough, I just put this before every test case out of habit.

files/bin/provider-id
mmerkes: Ya, I'm just trying to evaluate if there's a possibility that we could get junk from imds or it fails silently (like `INSTANCE_ID` could equal some error message rather than the instance ID).

cartermckinnon: Good point, `imds` wasn't exiting non-zero when `curl` got a response that wasn't 200. Fixed in the latest revision, and added tests for it.

test/cases/provider-id.sh
cartermckinnon: >The imds script could fail for plenty of reasons, but would exit with a non-zero code. set -o errexit in provider-id would bail if imds croaks.

This wasn't true when I posted it, but it's true for the latest revision 😄

files/bootstrap.sh
```diff
@@ -455,11 +455,23 @@ if [[ "$USE_MAX_PODS" = "true" ]]; then
...
+
+if vercmp "$KUBELET_VERSION" lt "1.26.0"; then
+  # TODO: remove this when 1.25 is EOL
+  KUBELET_CLOUD_PROVIDER="aws"
+else
+  KUBELET_CLOUD_PROVIDER="external"
+  echo "$(jq ".providerID=\"$(provider-id)\"" $KUBELET_CONFIG)" > $KUBELET_CONFIG
+fi
+
+KUBELET_ARGS="$KUBELET_ARGS --cloud-provider=$KUBELET_CLOUD_PROVIDER"
```
mmerkes, line 468: Why does this have to stay in the CLI?

test/cases/imds-errors.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
```
mmerkes, line 1: Should there be a happy case?

mmerkes (commented):
Overall, looks like this is going in the right direction. Before this gets merged, let's be sure to do the 1.25/1.26 testing you discuss, which should validate at a minimum:
1. Cloud provider config works in the real world
2. Verify that provider ID looks good
3. Nodes join the clusters

files/bootstrap.sh
cartermckinnon: With kubelet's in-tree cloud providers going away, the docs state that this flag will go away too:
> --cloud-provider
> The provider for cloud services. Set to empty string for running with no cloud provider. If set, the cloud provider determines the name of the node (consult cloud provider documentation to determine if and how the hostname is used). (DEPRECATED: will be removed in 1.24 or later, in favor of removing cloud provider code from Kubelet.)
> https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/

I'm not clear on kubelet's behavior once this is removed, though. Today, when you use `--cloud-provider=external`:
> kubelets specifying --cloud-provider=external will add a taint `node.cloudprovider.kubernetes.io/uninitialized` with an effect `NoSchedule` during initialization. This marks the node as needing a second initialization from an external controller before it can be scheduled work. Note that in the event that cloud controller manager is not available, new nodes in the cluster will be left unschedulable. The taint is important since the scheduler may require cloud specific information about nodes such as their region or type (high cpu, gpu, high memory, spot instance, etc).
> https://kubernetes.io/docs/tasks/administer-cluster/running-cloud-controller/#running-cloud-controller-manager

If kubelet isn't run on a cloud, it doesn't need to set that label. My assumption is we'll continue to need a boolean `--enable-cloud-provider` flag on kubelet to toggle this behavior.

I plan to bring it up in the next SIG node, maybe the docs are just outdated.

cartermckinnon: Thinking about this more, it's probably more realistic to just deprecate the acceptable values for `--cloud-provider` one-by-one, eventually leaving only `external` and the empty string.

Assuming we ship 1.26 with the external cloud provider, I'd like to get the in-tree `aws` provider removed in 1.28 (assuming it's too late for 1.27)?

test/cases/imds-errors.sh
cartermckinnon: It's covered in `imds-token-refresh.sh`, though indirectly. I'll add one here.

files/bootstrap.sh
```diff
@@ -455,11 +455,23 @@ if [[ "$USE_MAX_PODS" = "true" ]]; then
...
 fi
 
+KUBELET_ARGS="--node-ip=$INTERNAL_IP --pod-infra-container-image=$PAUSE_CONTAINER --v=2"
+
+if vercmp "$KUBELET_VERSION" lt "1.26.0"; then
+  # TODO: remove this when 1.25 is EOL
+  KUBELET_CLOUD_PROVIDER="aws"
+else
+  KUBELET_CLOUD_PROVIDER="external"
+  echo "$(jq ".providerID=\"$(provider-id)\"" $KUBELET_CONFIG)" > $KUBELET_CONFIG
```
ravisinha0506, line 465: How does node object look like when provider id isn't passed here?

test/cases/imds-errors.sh
```diff
@@ -0,0 +1,33 @@
...
+export IMDS_DEBUG=true
+imds /latest/meta-data/instance-id || EXIT_CODE=$?
+if [[ ${EXIT_CODE} -ne 0 ]]; then
+  echo "❌ Test Failed: expected a zero exit code but got: $EXIT_CODE"
+  exit 1
+fi
+
+echo "--> Should fail for unknown API"
+EXIT_CODE=0
+export IMDS_DEBUG=true
```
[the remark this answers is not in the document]
cartermckinnon: Sure, this was just test case copy-paste 😄

files/bootstrap.sh
cartermckinnon: The `Node`'s `spec.providerID` would be missing. The cloud provider works around this by resolving the node name to an instance ID, i.e.:

https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/cloud-provider/controllers/nodelifecycle/node_lifecycle_controller.go#LL219C26-L219C26

calls `instances.InstanceID`, which calls:

https://github.com/kubernetes/cloud-provider-aws/blob/master/pkg/providers/v1/aws.go#L5125

If kubelet doesn't specify the provider ID, it's set using this approach before the node's uninitialized taint is removed: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go#L527-L528C4

ravisinha0506 (approved):

PR #1205, opened 2023-02-25
Added "--update" parameter

joseeden:
**Issue #, if available:**
Script ran into an issue when it finds that AWS CLI is pre-installed onto the EC2 instance. Adding the "--update" parameter solves this.

**Description of changes:**
Adding the "--update" parameter to AWS CLI installation step.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

**Testing Done**

*[See this guide for recommended testing for PRs.](../doc/CONTRIBUTING.md#testing-changes) Some tests may not apply. Completing tests and providing additional validation steps are not required, but it is recommended and may reduce review time and time to merge.*

PR #1206, opened 2023-02-25
Added checks for AWS CLI

joseeden:
Added condition that checks if AWS CLI is pre-installed. If yes, it skips installation step.

**Issue #, if available:**
Script is interrupted when it finds AWS CLI is already pre-installed.

**Description of changes:**
Added condition that checks if AWS CLI is pre-installed. If yes, it skips installation step.

**Testing Done**
Run the script again, and its now skipping the installation steps when it sees AWS CLI is already installed.

### DIFF
--- a/scripts/install-worker.sh
+++ b/scripts/install-worker.sh
@@ -119,21 +119,26 @@ sudo mv $TEMPLATE_DIR/iptables-restore.service /etc/eks/iptables-restore.service
 ### awscli #####################################################
 ################################################################################
 
-if [[ "$BINARY_BUCKET_REGION" != "us-iso-east-1" && "$BINARY_BUCKET_REGION" != "us-isob-east-1" ]]; then
-  # https://docs.aws.amazon.com/cli/latest/userguide/getting-started-install.html
-  echo "Installing awscli v2 bundle"
-  AWSCLI_DIR=$(mktemp -d)
-  curl \
-    --silent \
-    --show-error \
-    --retry 10 \
-    --retry-delay 1 \
-    -L "https://awscli.amazonaws.com/awscli-exe-linux-${MACHINE}.zip" -o "${AWSCLI_DIR}/awscliv2.zip"
-  unzip -q "${AWSCLI_DIR}/awscliv2.zip" -d ${AWSCLI_DIR}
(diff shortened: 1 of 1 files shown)


scripts/install-worker.sh
```diff
@@ -119,21 +119,26 @@ sudo mv $TEMPLATE_DIR/iptables-restore.service /etc/eks/iptables-restore.service
...
+    echo "Installing awscli v2 bundle"
+    AWSCLI_DIR=$(mktemp -d)
+    curl \
+      --silent \
+      --show-error \
+      --retry 10 \
+      --retry-delay 1 \
+      -L "https://awscli.amazonaws.com/awscli-exe-linux-${MACHINE}.zip" -o "${AWSCLI_DIR}/awscliv2.zip"
+    unzip -q "${AWSCLI_DIR}/awscliv2.zip" -d ${AWSCLI_DIR}
+    sudo "${AWSCLI_DIR}/aws/install" --bin-dir /bin/ --update
```
cartermckinnon, line 137: `--update` has no effect here.

scripts/install-worker.sh
```diff
@@ -119,21 +119,26 @@ sudo mv $TEMPLATE_DIR/iptables-restore.service /etc/eks/iptables-restore.service
...
-    --silent \
-    --show-error \
-    --retry 10 \
-    --retry-delay 1 \
-    -L "https://awscli.amazonaws.com/awscli-exe-linux-${MACHINE}.zip" -o "${AWSCLI_DIR}/awscliv2.zip"
-  unzip -q "${AWSCLI_DIR}/awscliv2.zip" -d ${AWSCLI_DIR}
-  sudo "${AWSCLI_DIR}/aws/install" --bin-dir /bin/
+if command -v aws &> /dev/null
+then
+    echo "AWS CLI is already installed."
```
cartermckinnon, line 124: We would need to add assertions on the CLI version, but that's kind of a pain with the v1/v2 divide. For example `1.25.72`/`2.7.32` is needed for local clusters on Outpost. We also need `describe-addons-versions`. I'd probably just go with what you have in #1205 to always install the latest v2.

cartermckinnon (changes requested):

PR #1207, opened 2023-02-25
Added check if group "docker" exists.

joseeden:
Added condition that checks if the group "docker" is already created.  
If yes, it changes the GID to "1950".
If no, it creates the group with GID "1950".

**Issue #, if available:**
Script stops if the group "docker" exists.

**Description of changes:**
Added condition that check if group "docker" exists.

**Testing Done**

### DIFF
--- a/scripts/install-worker.sh
+++ b/scripts/install-worker.sh
@@ -206,7 +206,18 @@ fi
 
 if [[ "$INSTALL_DOCKER" == "true" ]]; then
   sudo amazon-linux-extras enable docker
-  sudo groupadd -og 1950 docker
+
+  if grep -q "^docker:" /etc/group
+  then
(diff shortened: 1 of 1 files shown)


cartermckinnon:
Is the use case a source AMI that already has Docker installed? If so, I'd probably just disable the entire Docker block with `INSTALL_DOCKER=false`. The changes generally look fine, but we're also considering the Docker logic in maintenance mode, and would err on the side of not touching it.

Can you combine #1208 with this and I'll bring it up with the team?

cartermckinnon:
We're going to avoid touching the Docker logic at this point in the support cycle. I recommend you use `INSTALL_DOCKER=false` and use whatever Docker setup is provided in the source AMI, or fork this template if you'll need ongoing Docker support.

PR #1210, opened 2023-03-02
Discontinue 1.21 AMI's

cartermckinnon:
**Description of changes:**

Removes 1.21. Will include other associated logic related to the 1.21/1.22 divide.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,8 +36,8 @@ T_GREEN := \e[0;32m
 T_YELLOW := \e[0;33m
 T_RESET := \e[0m
 
-.PHONY: all
-all: 1.21 1.22 1.23 1.24 1.25 ## Build all versions of EKS Optimized AL2 AMI
+.PHONY: latest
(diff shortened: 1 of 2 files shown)


Issacwww (approved):

Makefile
```diff
@@ -88,10 +88,6 @@ k8s: validate ## Build default K8s version of EKS Optimized AL2 AMI
 
 # Build dates and versions taken from https://docs.aws.amazon.com/eks/latest/userguide/install-kubectl.html
 
-.PHONY: 1.21
```
ravisinha0506, line 91: Can we also remove 1.21 from line# 40?

Issacwww: good catch, I believe we should remove it

Issacwww (changes requested):
revoke approval: please make the change to remove 1.21 from line#40

cartermckinnon:
@Issacwww the team discussed the `all` target, which we decided isn't the right default target for this Makefile. I've replaced it here with a `latest` target; take a look when you get a chance!

Issacwww (approved):

PR #1211, opened 2023-03-03
rewrite k8s.gcr.io registry to registry.k8s.io

rothgar:
**Description of changes:**
Sets a default config for containerd to redirect k8s.gcr.io registry pulls to the new registry.k8s.io. This is needed to avoid breaking users if the old registry deletes images (a current option being discussed). This has potential to break some users if they rely on network rules for external connections.

More context https://github.com/kubernetes/registry.k8s.io#stability
https://github.com/kubernetes/k8s.io/issues/4214#issuecomment-1247446149

More discussions about options are happening in #sig-infra in Kubernetes slack. We would want to clearly communicate this change to users and allow them to check if their accounts will be affected.

**Testing Done**

### DIFF
--- a/files/containerd-config.toml
+++ b/files/containerd-config.toml
@@ -23,3 +23,8 @@ SystemdCgroup = true
 [plugins."io.containerd.grpc.v1.cri".cni]
 bin_dir = "/opt/cni/bin"
 conf_dir = "/etc/cni/net.d"
+
+# Enable registry.k8s.io as the primary mirror for k8s.gcr.io
+# See: https://github.com/kubernetes/k8s.io/issues/3411
+[plugins."io.containerd.grpc.v1.cri".registry.mirrors."k8s.gcr.io"]
+  endpoint = ["https://registry.k8s.io", "https://k8s.gcr.io",]


chris-short:
This is a necessary change: https://kubernetes.io/blog/2023/02/06/k8s-gcr-io-freeze-announcement/

TerryHowe:
Outside of the scope of this, this looks fine, but should there be a mirror for registry.k8s.io?

cartermckinnon:
IIUC, this isn't strictly *necessary* for (most) EKS users because we vend the most common/necessary images in ECR. I can imagine how this would be convenient but I'm not sure it's the right default.

rothgar:
There are a lot of packages not provided directly from EKS and on older versions of EKS that come from the upstream registry. The most seamless solution would be to redirect the registry and not change local config but that had issues in the past. We should have some alternative options next week so we can hold off on deciding on this for at least a week.

Here's data transfer from the upstream repo that would be better to redirect to registry.k8s.io. Not only will image pulls be faster for customers but they likely will be more reliable because they'll come from regional s3 mirrors.
image

stevehipwell:
> Outside of the scope of this, this looks fine, but should there be a mirror for registry.k8s.io?

AFAIK [registry.k8s.io](https://github.com/kubernetes/registry.k8s.io) is a mirror with AWS requests being served from S3.

chris-short:
See this for the "how" behind this new'ish K8s registry: https://github.com/kubernetes/registry.k8s.io

cartermckinnon:
I'm going to close this in light of https://kubernetes.io/blog/2023/03/10/image-registry-redirect/. 🎉

PR #1223, opened 2023-03-17
Mount eBPF filesystem by default on 1.27+

### LINKED ISSUE #728: eBPF filesystem mounted when in use
**systemd unit for mounting eBPF, when used**:

**needed by some tools like Calico**:

Latest calico versions allow using eBPF networking, replacing all the kube-proxy functionality. But eBPF filesystem must be mounted on worker node. Can be achieved by creating simple systemd unit (installed by custom startup script, but why not to have it in distro?).

Sample systemd unit:
`[Unit]
Description=BPF mounts
DefaultDependencies=no
Before=local-fs.target umount.target
After=swap.target

[Mount]
What=bpffs
Where=/sys/fs/bpf
Type=bpf
Options=rw,nosuid,nodev,noexec,relatime,mode=700

[Install]
WantedBy=multi-user.target`

just place it in /etc/systemd/system/sys-fs-bpf.mount, and enable it.

cartermckinnon:
**Issue #, if available:**

Closes #728 

**Description of changes:**

Mounts the eBPF filesystem by default.

**Testing Done**

```
> make 1.25
# succeeds, launched an instance
> mount | grep bpf
none on /sys/fs/bpf type bpf (rw,nosuid,nodev,noexec,relatime,mode=700)
```


cartermckinnon:
I'll change this to a runtime option so that we don't cause issues if folks are already mounting this themselves.

files/bootstrap.sh
```diff
@@ -269,6 +277,10 @@ if [[ "$MACHINE" != "x86_64" && "$MACHINE" != "aarch64" ]]; then
   exit 1
 fi
 
+if [ "$MOUNT_BPF_FS" = "true" ]; then
+  sudo mount-bpf-fs
```
mmerkes, line 281: What happens if this is already mounted in the customer's user data? Or via custom AMI that still uses the bootstrap script?

Also, where is `mount-bpf-fs` coming from? Don't see it in this PR.

cartermckinnon: Whoops, missed in my last commit. This process will bail if:
1. There's already a `type bpf` mount.
2. There's already a mount at `/sys/fs/bpf`
3. There's already a file at `/etc/systemd/system/sys-fs-bpf.mount`.

files/bin/mount-bpf-fs
```diff
@@ -0,0 +1,50 @@
...
+function debug() {
+  if [ "$MOUNT_BPF_FS_DEBUG" = "true" ]; then
+    echo >&2 "DEBUG:" "$@"
+  fi
+}
+
+if mount | grep "type $FS_TYPE"; then
+  debug "$FS_TYPE filesystem already mounted!"
+  exit 0
+elif mount | grep "$MOUNT_POINT"; then
```
mmerkes, line 22: Will never get here because of the above check.

files/bootstrap.sh
```diff
@@ -16,22 +16,24 @@ function print_help {
...
+  echo "--cluster-id Specify the id of EKS cluster"
+  echo "--container-runtime Specify a container runtime (default: dockerd)"
   echo "--containerd-config-file File containing the containerd configuration to be used in place of AMI defaults."
   echo "--dns-cluster-ip Overrides the IP address to use for DNS queries within the cluster. Defaults to 10.100.0.10 or 172.20.0.10 based on the IP address of the primary interface"
+  echo "--docker-config-json The contents of the /etc/docker/daemon.json file. Useful if you want a custom config differing from the default one in the AMI"
+  echo "--enable-docker-bridge Restores the docker default bridge network. (default: false)"
+  echo "--enable-local-outpost Enable support for worker nodes to communicate with the local control plane when running on a disconnected Outpost. (true or false)"
+  echo "--ip-family Specify ip family of the cluster"
+  echo "--kubelet-extra-args Extra arguments to add to the kubelet. Useful for adding labels or taints."
+  echo "--mount-bfs-fs Mount a bpffs at /sys/fs/bpf (default: true)"
```
mmerkes, line 32: Is this the only actual change?

test/cases/mount-bpf-fs.sh
```diff
@@ -0,0 +1,50 @@
...
+set -o errexit
+set -o pipefail
+
+export MOUNT_BPF_FS_DEBUG=true
+
+echo "--> Should succeed if bpf type fs already exists"
+function mount() {
+  echo "none on /foo/bar type bpf (rw,nosuid,nodev,noexec,relatime,mode=700)"
+}
+export -f mount
```
mmerkes, line 13: Fancy trick. Wasn't familiar with this one.

files/bootstrap.sh
cartermckinnon: yep, rest is just a sort

files/bin/mount-bpf-fs
cartermckinnon: The first check is looking for `type bpf`, this check is looking for `/sys/fs/bpf`. The debug output in the unit tests looks correct to me; am I missing something?

files/bin/mount-bpf-fs
```diff
@@ -0,0 +1,47 @@
...
+
+MOUNT_BPF_FS_DEBUG=${MOUNT_BPF_FS_DEBUG:-false}
+function debug() {
+  if [ "$MOUNT_BPF_FS_DEBUG" = "true" ]; then
+    echo >&2 "DEBUG:" "$@"
+  fi
+}
+
+if mount | grep "type $FS_TYPE"; then
+  debug "$FS_TYPE filesystem already mounted!"
```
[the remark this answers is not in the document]
cartermckinnon: Fair, I'll change this to pluck the specific column for filesystem type and check against that 👍

test/cases/mount-bpf-fs.sh
cartermckinnon: Yeah I think this is the right approach to "mocking" in these test cases, but I'm not going to refactor them all in this PR 😂

mmerkes (approved):

daisuke-yoshimoto:
@cartermckinnon @mmerkes 

When this PR is completed, will "Container runtime for GurdDuty" be released soon?

mmerkes:
@daisuke-yoshimoto This is a standalone thing and has no relation to anything else on our end.

files/bootstrap.sh
```diff
@@ -178,6 +185,12 @@ SERVICE_IPV6_CIDR="${SERVICE_IPV6_CIDR:-}"
 ENABLE_LOCAL_OUTPOST="${ENABLE_LOCAL_OUTPOST:-}"
 CLUSTER_ID="${CLUSTER_ID:-}"
 
+DEFAULT_MOUNT_BPF_FS="true"
+if vercmp "$KUBELET_VERSION" lt "1.26.0"; then
```
mmerkes, line 189: This going to make it in time for 1.26?

cartermckinnon: That's the plan!

mmerkes: Cool. Let's do that and keep it simple.

mmerkes (approved):

PR #1230, opened 2023-03-24
Move file copy after kernel upgrade

### LINKED ISSUE #1229: Not able to build eks ami with CIS BENCHMARK AMAZON LINUX 2
we are not able to build eks ami with latest code with base AMI reference of  CIS BENCHMARK AMAZON LINUX 2. and it is working for plain amazon linux 2.

getting below error 

2023-03-24T16:30:28+05:30:     amazon-ebs: Ciphers aes128-ctr,aes256-ctr,aes128-gcm@openssh.com,aes256-gcm@openssh.com
2023-03-24T16:30:29+05:30:     amazon-ebs: mv: cannot stat ‘/tmp/worker/iptables-restore.service’: No such file or directory
2023-03-24T16:30:29+05:30: ==> amazon-ebs: Provisioning step had errors: Running the cleanup provisioner, if present...
2023-03-24T16:30:29+05:30: ==> amazon-ebs: Terminating the source AWS instance...
2023-03-24T16:31:30+05:30: ==> amazon-ebs: Cleaning up any extra volumes...
2023-03-24T16:31:30+05:30: ==> amazon-ebs: No volumes to clean up, skipping
2023-03-24T16:31:30+05:30: ==> amazon-ebs: Deleting temporary security group...
2023-03-24T16:31:30+05:30: ==> amazon-ebs: Deleting temporary keypair...
2023-03-24T16:31:31+05:30: Build 'amazon-ebs' errored after 6 minutes 42 seconds: Script exited with non-zero exit status: 1. Allowed exit codes are: [0]

==> Wait completed after 6 minutes 42 seconds

image

Note: we were able to build with old code with same cis benchmark ami without any issues and latest code only building with plain amazon linux amis not working for cis bench mark amazon linux 2.

kindly test and revert back as we need to build for eks ami for 1.23 and 1.24 for prod upgrades got stuck over here.

matt-corbalt:
**Description of changes:**

We ran into this issue with AL2 today where we had an error using the latest tag [v20230304](https://github.com/awslabs/amazon-eks-ami/releases/tag/v20230304) during the `install-worker.sh` step:
```
2023-03-23T20:50:09Z:     amazon-ebs: mv: cannot stat ‘/tmp/worker/iptables-restore.service’: No such file or directory
```

In some distros the `/tmp` directory is cleared upon reboot, which removes the `/tmp/worker` files needed to successfully build the AMI. Since the `kernel_upgrade.sh` script has a reboot step, the file transfer needs to happen after the reboot.

In our testing the change in order of steps fixes the above issue.

Looks like this change was previously accepted (https://github.com/awslabs/amazon-eks-ami/pull/474) and then reverted in recent changes.

Closes #1229

---

### DIFF
--- a/eks-worker-al2.json
+++ b/eks-worker-al2.json
@@ -123,6 +123,17 @@
     {
       "type": "shell",
       "remote_folder": "{{ user `remote_folder`}}",
+      "expect_disconnect": true,
+      "pause_after": "90s",
+      "script": "{{template_dir}}/scripts/upgrade_kernel.sh",
+      "environment_vars": [
+        "KUBERNETES_VERSION={{user `kubernetes_version`}}",
+        "KERNEL_VERSION={{user `kernel_version`}}"
+      ]
+    },
+    {
+      "type": "shell",
+      "remote_folder": "{{ user `remote_folder`}}",
       "inline": [
         "mkdir -p /tmp/worker/log-collector-script/"
       ]
@@ -142,17 +153,6 @@
       "inline": [
         "sudo chmod -R a+x /tmp/worker/bin/",
         "sudo mv /tmp/worker/bin/* /usr/bin/"
-      ]
-    },
-    {
-      "type": "shell",
-      "remote_folder": "{{ user `remote_folder`}}",
-      "expect_disconnect": true,
-      "pause_after": "90s",
-      "script": "{{template_dir}}/scripts/upgrade_kernel.sh",
-      "environment_vars": [
-        "KUBERNETES_VERSION={{user `kubernetes_version`}}",
-        "KERNEL_VERSION={{user `kernel_version`}}"
       ]
     },
     {


matt-corbalt:
Looks like the original change (https://github.com/awslabs/amazon-eks-ami/commit/ccf5b1525dfa7a17f5cae21f8d8395a6f21b45d8) tried to include using [`vercmp`](https://github.com/awslabs/amazon-eks-ami/blob/master/files/bin/vercmp) in the [`upgrade_kernel.sh`](https://github.com/awslabs/amazon-eks-ami/blob/master/scripts/upgrade_kernel.sh#L8) script. Unfortunately this PR breaks the `vercmp` usage in the kernel upgrade script.

```
00:02:30.295  2023-03-24T19:22:54Z:     amazon-ebs: /tmp/script_3240.sh: line 8: vercmp: command not found
00:02:30.295  2023-03-24T19:22:54Z:     amazon-ebs: kernel_version is unset. Setting to 5.4 based on Kubernetes version 1.24.10.
```

A maintainer will probably have to give their opinion on how best to handle this. We could copy the files both before and after the kernel upgrade (+ reboot), but there seems like there might be a better way to handle version comparison in the `upgrade_kernel.sh` script.

Given that this change installs 5.4 instead of 5.10 (due to a soft failure of `vercmp`) for all kubernetes versions might be not a big deal?

cartermckinnon:
I think #1231 addresses this more generally, give it a look if you can!

matt-corbalt:
> I think #1231 addresses this more generally, give it a look if you can!

@cartermckinnon That definitely looks like a much better solution. I'll try that out on Monday to confirm it works with my setup. Thanks for the quick response to this!

Issue #797, opened 2021-11-04
Install crictl for CLI interaction with containerd


benmccown:
**What would you like to be added**: Install crictl as part of the EKS AMI build. Currently the only CLI utility included with the AMI is ctr.

crictl project: https://github.com/kubernetes-sigs/cri-tools/blob/master/docs/crictl.md
crictl documentation included in official Kubernetes docs: https://kubernetes.io/docs/tasks/debug-application-cluster/crictl/

**Why is this needed**:

Let me first start with issues I have seen with `ctr` and then I will discuss why crictl is the proposed alternative.

**ctr's issues**

- Non existent documentation. If you search for "containerd ctr documentation" you'll find a blog post or two, an outdated fork of the containerd project that has an old markdown file from 2016, and that's about it. There's the `-h` help output you can access from the CLI binary itself, but I don't think this is adequate for a new containerd user who is used to `docker` cli syntax.
- Unintuitive syntax when coming from the docker cli. Some commands are relatively straight forward, but in my experience so far there's a learning curve with `ctr`. This is fine, but the curve is made even steeper by the aforementioned lack of documentation.
- Doesn't pull private registry credentials when they are [defined in containerd's config.toml](https://github.com/containerd/containerd/blob/main/docs/cri/registry.md#configure-registry-credentials). See [this comment confirming the same](https://github.com/containerd/cri/issues/1482#issuecomment-630501672). I have confirmed this myself. When registry credentials are defined in containerd's config.toml a private image can be pulled with crictl but not with ctr. If a user wants to test their containerd config by pulling a private image with the default cli (ctr) they're going to see the image fail to pull even if their config is correct, leading to unnecessary troubleshooting.

**crictl as proposed solution**

- crictl is part of the official kubernetes-sigs "cri-tools" project and was developed with Kubernetes in mind.
- crictl is well documented within the official Kubernetes documentation.
- crictl seems to honor private registry credentials provided in containerd's config.toml (issue with ctr mentioned above)

Providing a CLI utility that is part of the Kubernetes ecosystem, easier to use, and well documented will ease the transition for new users to containerd.

davidegiunchi:
Another point about "why is this needed", from the ctr's manpage:
```
ctr(8)()                                                                                                                                                                                          

       ctr is an unsupported debug and administrative client for interacting with the containerd daemon. Because it is unsupported, the commands, options, and operations are not guaranteed to be backward
       compatible or stable from release to release of the containerd project.
```

[...]

PR #1235, opened 2023-03-28
Install cri-tools

cartermckinnon:
**Issue #, if available:**

Closes https://github.com/awslabs/amazon-eks-ami/issues/797 .

**Description of changes:**

Adds cri-tools, which includes crictl and critest. The latest version of the package will be installed at build, and the package is not version-locked, because it's not in the critical path.

**Testing Done**
```
> make
...
2023-03-28T11:50:21-07:00:     amazon-ebs: Downloading packages:
2023-03-28T11:50:22-07:00:     amazon-ebs: cri-tools-1.25.0-1.amzn2.0.1.x86_64.rpm                    |  17 MB   00:00
2023-03-28T11:50:22-07:00:     amazon-ebs: Running transaction check
2023-03-28T11:50:22-07:00:     amazon-ebs: Running transaction test
2023-03-28T11:50:22-07:00:     amazon-ebs: Transaction test succeeded
2023-03-28T11:50:22-07:00:     amazon-ebs: Running transaction
2023-03-28T11:50:24-07:00:     amazon-ebs:   Installing : cri-tools-1.25.0-1.amzn2.0.1.x86_64                          1/1
2023-03-28T11:50:24-07:00:     amazon-ebs:   Verifying  : cri-tools-1.25.0-1.amzn2.0.1.x86_64                          1/1
2023-03-28T11:50:24-07:00:     amazon-ebs:
2023-03-28T11:50:24-07:00:     amazon-ebs: Installed:
2023-03-28T11:50:24-07:00:     amazon-ebs:   cri-tools.x86_64 0:1.25.0-1.amzn2.0.1
2023-03-28T11:50:24-07:00:     amazon-ebs:
2023-03-28T11:50:24-07:00:     amazon-ebs: Complete!
```

### DIFF
--- a/scripts/install-worker.sh
+++ b/scripts/install-worker.sh
@@ -343,6 +343,9 @@ sudo systemctl daemon-reload
 # Disable the kubelet until the proper dropins have been configured
 sudo systemctl disable kubelet
 
+# install crictl, critest
+sudo yum install -y cri-tools
+
 ################################################################################
 ### EKS ########################################################################
 ################################################################################


dims (approved):
/approve
/lgtm

Issacwww (approved):

scripts/install-worker.sh
```diff
@@ -343,6 +343,9 @@ sudo systemctl daemon-reload
 # Disable the kubelet until the proper dropins have been configured
 sudo systemctl disable kubelet
 
+# install crictl, critest
+sudo yum install -y cri-tools
```
mmerkes, line 347: Does this package get updated in the yum repo along with containerd? We'll undoubtedly get some CVEs from this, so want to make sure that it gets updated within reason.

mmerkes (approved):
LGTM pending conversation from other comment.

scripts/install-worker.sh
cartermckinnon: I thought this was versioned with `containerd`, but it's actually a k8s project. Seems to be versioned with k8s on major + minor. I asked around in the k8s slack, not sure if we should try to install a relevant version. Also not sure how it's ending up in the AL repos, so I don't know how diligently it's updated.

mmerkes: IMO, we need to understand how it's updated before we install it by default, or we'll just be doing scramble drill when it needs to be updated. As far as we know, that could be right away!

cartermckinnon: We'll definitely get pinged to update it, but it's just a CLI tool; shouldn't ever be an emergency. I'll see what the sig-node folks say!

mmerkes: I'm more interested in how that yum repo gets updated. If it just points to some repo that sig-node folks update, that's fine. As long as we're regularly getting the latest, seems reasonable.

cartermckinnon:
There doesn't seem to be consensus or strict guidance on whether you need to run the same major + minor version of `cri-tools` and `kubelet`, though my impression is that this is the safest route. Because we don't vend `cri-tools` with our other binaries, we can't guarantee that a needed version exists in the yum repos (we shouldn't just install what's available).

I'm going to close this PR, because a proper implementation will differ significantly.

mohag:
> There doesn't seem to be consensus or strict guidance on whether you need to run the same major + minor version of `cri-tools` and `kubelet`, though my impression is that this is the safest route. Because we don't vend `cri-tools` with our other binaries, we can't guarantee that a needed version exists in the yum repos (we shouldn't just install what's available).
> 
> I'm going to close this PR, because a proper implementation will differ significantly.

cri-tools does not get released with the rest of Kubernetes - it is possible that there is no version of cri-tools for a specific kubernetes version. (the CRI apiVersion support is the most important parameter - it should be the same as the Kubernetes version)

(with kubeadm the OS packages used to depend on the latest version of cri-tools, which broke deployment of clusters on supported containerd versions, since the cri-tools version was significantly newer, without support for some (beta) CRI APIs that the kubernetes version still supported - which meant that clusters required newer containerd versions to deploy than what was supposed to work) See [kubernetes/release/#2866](https://github.com/kubernetes/release/issues/2866) for the kubeadm issue

PR #1231, opened 2023-03-24
Add configurable working directory

cartermckinnon:
**Issue #, if available:**

Closes #1229 

**Description of changes:**

This aims to address issues brought up in #1230 and #1198.

The underlying issue is the base working directory used throughout the AMI template (`/tmp`) may not be suitable for all use cases. Resources are staged in this directory (`/tmp/worker`) before being moved into their final position (such as the contents of `files/`) and this directory is also used to stash the AWS CLI installer, among other things.

After this change:
- `remote_folder` can still be used to control the location of Packer scripts on the builder instance.
- `working_dir` will default to a sub-directory of `remote_folder`, and can be adjusted as needed. All provisioners that create files on the builder instance will do so within this directory. This directory is deleted at the end of the build.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

**Testing Done**

Defaults succeed:
```
make 1.25
```

Different `remote_folder` succeeds:
```
make 1.25 remote_folder=/home/ec2-user
# working_dir will be /home/ec2-user/worker
```

Different `working_directory` succeeds:
```
make 1.25 working_directory=/tmp/foo
```

PR #1244, opened 2023-04-01
Improve scripts readability and prevent word split issues

guessi:
**Issue #, if available:**

`n/a`

**Description of changes:**

Currently `--severity` for `shellcheck` is set to `error` which may skip common word split issues, scripts under "./scripts" directory is relative more important for building AMI, would better to have check SC2046, SC2068, SC2086.

### References:
- https://github.com/koalaman/shellcheck/wiki/SC2046
- https://github.com/koalaman/shellcheck/wiki/SC2068
- https://github.com/koalaman/shellcheck/wiki/SC2086

**Testing Results**

### AMI Build

```bash
...

--> amazon-ebs: AMIs were created:
us-east-1: ami-01234567890123456
```

### Launch Test

```bash
% kubectl get nodes ip-192-168-94-246.ec2.internal -o json | jq '.metadata.labels."eks.amazonaws.com/nodegroup-image"'
"ami-01234567890123456"
```

```bash
% kubectl get nodes ip-192-168-94-246.ec2.internal -o json | jq '.status.conditions[] | select(.type == "Ready")'
{
  "lastHeartbeatTime": "2023-04-01T11:12:18Z",
  "lastTransitionTime": "2023-04-01T10:56:38Z",
  "message": "kubelet is posting ready status",
  "reason": "KubeletReady",
  "status": "True",
  "type": "Ready"
}
```

### DIFF
--- a/scripts/generate-version-info.sh
+++ b/scripts/generate-version-info.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/usr/bin/env bash
 
 # generates a JSON file containing version information for the software in this AMI
 
@@ -13,11 +13,14 @@ fi
 OUTPUT_FILE="$1"
 
 # packages
-sudo rpm --query --all --queryformat '\{"%{NAME}": "%{VERSION}-%{RELEASE}"\}\n' | jq --slurp --sort-keys 'add | {packages:(.)}' > "$OUTPUT_FILE"
+sudo rpm --query --all --queryformat '\{"%{NAME}": "%{VERSION}-%{RELEASE}"\}\n' | jq --slurp --sort-keys 'add | {packages:(.)}' > "${OUTPUT_FILE}"
 
 # binaries
-echo $(jq ".binaries.kubelet = \"$(kubelet --version | awk '{print $2}')\"" $OUTPUT_FILE) > $OUTPUT_FILE
-echo $(jq ".binaries.awscli = \"$(aws --version | awk '{print $1}' | cut -d '/' -f 2)\"" $OUTPUT_FILE) > $OUTPUT_FILE
+KUBELET_VERSION=$(kubelet --version | cut -d' ' -f2)
+echo "$(jq ".binaries.kubelet = \"${KUBELET_VERSION}\"" "${OUTPUT_FILE}")" > "${OUTPUT_FILE}"
+
+AWSCLI_VERSION=$(aws --version | cut -d' ' -f1 | cut -d'/' -f2)
+echo "$(jq ".binaries.awscli = \"${AWSCLI_VERSION}\"" "${OUTPUT_FILE}")" > "${OUTPUT_FILE}"
 
 # cached images
-echo $(jq ".images = [ $(sudo ctr -n k8s.io image ls -q | cut -d'/' -f2- | sort | uniq | grep -v 'sha256' | xargs -r printf "\"%s\"," | sed 's/,$//') ]" $OUTPUT_FILE) > $OUTPUT_FILE
+echo "$(jq ".images = [ $(sudo ctr -n k8s.io image ls -q | cut -d'/' -f2- | sort | uniq | grep -v 'sha256' | xargs -r printf "\"%s\"," | sed 's/,$//') ]" "${OUTPUT_FILE}")" > "${OUTPUT_FILE}"
--- a/scripts/install-worker.sh
+++ b/scripts/install-worker.sh
@@ -3,6 +3,7 @@
 set -o pipefail
 set -o nounset
 set -o errexit
+
 IFS=$'\n\t'
 export AWS_DEFAULT_OUTPUT="json"
 
@@ -39,12 +40,12 @@ validate_env_set CACHE_CONTAINER_IMAGES
 ################################################################################
 
 MACHINE=$(uname -m)
-if [ "$MACHINE" == "x86_64" ]; then
+if [ "${MACHINE}" == "x86_64" ]; then
   ARCH="amd64"
-elif [ "$MACHINE" == "aarch64" ]; then
+elif [ "${MACHINE}" == "aarch64" ]; then
   ARCH="arm64"
 else
-  echo "Unknown machine architecture '$MACHINE'" >&2
+  echo "Unknown machine architecture '${MACHINE}'" >&2
   exit 1
 fi
 
@@ -75,10 +76,12 @@ sudo yum install -y \
 # Remove any old kernel versions. `--count=1` here means "only leave 1 kernel version installed"
 sudo package-cleanup --oldkernels --count=1 -y
 
-sudo yum versionlock kernel-$(uname -r)
+sudo yum versionlock "kernel-$(uname -r)"
 
 # Remove the ec2-net-utils package, if it's installed. This package interferes with the route setup on the instance.
-if yum list installed | grep ec2-net-utils; then sudo yum remove ec2-net-utils -y -q; fi
+if yum list installed | grep ec2-net-utils; then
+  sudo yum remove ec2-net-utils -y -q
+fi
 
 ################################################################################
 ### Time #######################################################################
@@ -114,13 +117,13 @@ sudo systemctl restart sshd.service
 ### iptables ###################################################################
 ################################################################################
 sudo mkdir -p /etc/eks
-sudo mv $TEMPLATE_DIR/iptables-restore.service /etc/eks/iptables-restore.service
+sudo mv "${TEMPLATE_DIR}/iptables-restore.service" /etc/eks/iptables-restore.service
 
 ################################################################################
 ### awscli #####################################################
 ################################################################################
 
-if [[ "$BINARY_BUCKET_REGION" != "us-iso-east-1" && "$BINARY_BUCKET_REGION" != "us-isob-east-1" ]]; then
+if [[ "${BINARY_BUCKET_REGION}" != "us-iso-east-1" && "${BINARY_BUCKET_REGION}" != "us-isob-east-1" ]]; then
   # https://docs.aws.amazon.com/cli/latest/userguide/getting-started-install.html
   echo "Installing awscli v2 bundle"
   AWSCLI_DIR=$(mktemp -d)
@@ -130,7 +133,7 @@ if [[ "$BINARY_BUCKET_REGION" != "us-iso-east-1" && "$BINARY_BUCKET_REGION" != "
     --retry 10 \
     --retry-delay 1 \
     -L "https://awscli.amazonaws.com/awscli-exe-linux-${MACHINE}.zip" -o "${AWSCLI_DIR}/awscliv2.zip"
-  unzip -q "${AWSCLI_DIR}/awscliv2.zip" -d ${AWSCLI_DIR}
+  unzip -q "${AWSCLI_DIR}/awscliv2.zip" -d "${AWSCLI_DIR}"
   sudo "${AWSCLI_DIR}/aws/install" --bin-dir /bin/ --update
 else
   echo "Installing awscli package"
@@ -148,11 +151,11 @@ sudo mv "${TEMPLATE_DIR}/runtime.slice" /etc/systemd/system/runtime.slice
 ###############################################################################
 
 # install runc and lock version
-sudo yum install -y runc-${RUNC_VERSION}
+sudo yum install -y "runc-${RUNC_VERSION}"
 sudo yum versionlock runc-*
 
 # install containerd and lock version
-sudo yum install -y containerd-${CONTAINERD_VERSION}
+sudo yum install -y "containerd-${CONTAINERD_VERSION}"
 sudo yum versionlock containerd-*
 
 sudo mkdir -p /etc/eks/containerd
@@ -160,13 +163,13 @@ if [ -f "/etc/eks/containerd/containerd-config.toml" ]; then
   ## this means we are building a gpu ami and have already placed a containerd configuration file in /etc/eks
   echo "containerd config is already present"
 else
-  sudo mv $TEMPLATE_DIR/containerd-config.toml /etc/eks/containerd/containerd-config.toml
+  sudo mv "${TEMPLATE_DIR}/containerd-config.toml" /etc/eks/containerd/containerd-config.toml
 fi
 
-sudo mv $TEMPLATE_DIR/kubelet-containerd.service /etc/eks/containerd/kubelet-containerd.service
-sudo mv $TEMPLATE_DIR/sandbox-image.service /etc/eks/containerd/sandbox-image.service
-sudo mv $TEMPLATE_DIR/pull-sandbox-image.sh /etc/eks/containerd/pull-sandbox-image.sh
-sudo mv $TEMPLATE_DIR/pull-image.sh /etc/eks/containerd/pull-image.sh
+sudo mv "${TEMPLATE_DIR}/kubelet-containerd.service" /etc/eks/containerd/kubelet-containerd.service
+sudo mv "${TEMPLATE_DIR}/sandbox-image.service" /etc/eks/containerd/sandbox-image.service
+sudo mv "${TEMPLATE_DIR}/pull-sandbox-image.sh" /etc/eks/containerd/pull-sandbox-image.sh
+sudo mv "${TEMPLATE_DIR}/pull-image.sh" /etc/eks/containerd/pull-image.sh
 sudo chmod +x /etc/eks/containerd/pull-sandbox-image.sh
 sudo chmod +x /etc/eks/containerd/pull-image.sh
 
@@ -194,26 +197,26 @@ EOF
 sudo yum install -y device-mapper-persistent-data lvm2
 
 if [[ ! -v "INSTALL_DOCKER" ]]; then
-  INSTALL_DOCKER=$(vercmp "$KUBERNETES_VERSION" lt "1.25.0" || true)
+  INSTALL_DOCKER=$(vercmp "${KUBERNETES_VERSION}" lt "1.25.0" || true)
 else
   echo "WARNING: using override INSTALL_DOCKER=${INSTALL_DOCKER}. This option is deprecated and will be removed in a future release."
 fi
 
-if [[ "$INSTALL_DOCKER" == "true" ]]; then
+if [[ "${INSTALL_DOCKER}" == "true" ]]; then
   sudo amazon-linux-extras enable docker
   sudo groupadd -og 1950 docker
-  sudo useradd --gid $(getent group docker | cut -d: -f3) docker
+  sudo useradd --gid "$(getent group docker | cut -d: -f3)" docker
 
   # install docker and lock version
-  sudo yum install -y docker-${DOCKER_VERSION}*
+  sudo yum install -y "docker-${DOCKER_VERSION}"*
   sudo yum versionlock docker-*
-  sudo usermod -aG docker $USER
+  sudo usermod -aG docker "${USER}"
 
   # Remove all options from sysconfig docker.
   sudo sed -i '/OPTIONS/d' /etc/sysconfig/docker
 
   sudo mkdir -p /etc/docker
-  sudo mv $TEMPLATE_DIR/docker-daemon.json /etc/docker/daemon.json
+  sudo mv "${TEMPLATE_DIR}/docker-daemon.json" /etc/docker/daemon.json
   sudo chown root:root /etc/docker/daemon.json
 
   # Enable docker daemon to start on boot.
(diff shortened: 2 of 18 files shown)


scripts/install-worker.sh
```diff
@@ -240,88 +243,87 @@ sudo mkdir -p /var/lib/kubernetes
...
-    aws s3 cp --region $BINARY_BUCKET_REGION $S3_PATH/${CNI_PLUGIN_FILENAME}.tgz .
-    aws s3 cp --region $BINARY_BUCKET_REGION $S3_PATH/${CNI_PLUGIN_FILENAME}.tgz.sha256 .
+    aws s3 cp --region "${BINARY_BUCKET_REGION}" "${S3_PATH}/${CNI_PLUGIN_FILENAME}.tgz" .
+    aws s3 cp --region "${BINARY_BUCKET_REGION}" "${S3_PATH}/${CNI_PLUGIN_FILENAME}.tgz.sha256" .
   else
     echo "AWS cli missing - using wget to fetch cni binaries from s3. Note: This won't work for private bucket."
-    sudo wget "$S3_URL_BASE/${CNI_PLUGIN_FILENAME}.tgz"
-    sudo wget "$S3_URL_BASE/${CNI_PLUGIN_FILENAME}.tgz.sha256"
+    wget "$S3_URL_BASE/${CNI_PLUGIN_FILENAME}.tgz"
+    wget "$S3_URL_BASE/${CNI_PLUGIN_FILENAME}.tgz.sha256"
```
guessi, line 304: no need to pass `sudo` here.

scripts/generate-version-info.sh
```diff
@@ -13,11 +13,19 @@ fi
 OUTPUT_FILE="$1"
 
 # packages
-sudo rpm --query --all --queryformat '\{"%{NAME}": "%{VERSION}-%{RELEASE}"\}\n' | jq --slurp --sort-keys 'add | {packages:(.)}' > "$OUTPUT_FILE"
+sudo rpm --query --all --queryformat '\{"%{NAME}": "%{VERSION}-%{RELEASE}"\}\n' | jq --slurp --sort-keys 'add | {packages:(.)}' > "${OUTPUT_FILE}"
+
+# version detection
+kubelet_version=$(kubelet --version | cut -d' ' -f2)
+awscli_version=$(aws --version | cut -d' ' -f1 | cut -d'/' -f2)
```
guessi, line 20: to avoid shellcheck mistakenly take `$1` and `$2` in awk as warning, turns to use `cut -d' ' -f2` instead.

scripts/generate-version-info.sh
```diff
@@ -13,11 +13,19 @@ fi
...
 
 # binaries
-echo $(jq ".binaries.kubelet = \"$(kubelet --version | awk '{print $2}')\"" $OUTPUT_FILE) > $OUTPUT_FILE
-echo $(jq ".binaries.awscli = \"$(aws --version | awk '{print $1}' | cut -d '/' -f 2)\"" $OUTPUT_FILE) > $OUTPUT_FILE
+# shellcheck disable=SC2046
+echo $(jq -r --arg KUBELET_VERSION "${kubelet_version}" '.binaries.kubelet = $KUBELET_VERSION' "${OUTPUT_FILE}") > "${OUTPUT_FILE}"
+
+# shellcheck disable=SC2046
+echo $(jq -r --arg AWSCLI_VERSION "${awscli_version}" '.binaries.awscli = $AWSCLI_VERSION' "${OUTPUT_FILE}") > "${OUTPUT_FILE}"
```
guessi, line 27: improve readability by utilizing `jq` built in `--arg` flag.

scripts/install-worker.sh
```diff
@@ -113,13 +116,13 @@ sudo systemctl restart sshd.service
 ### iptables ###################################################################
 ################################################################################
 sudo mkdir -p /etc/eks
-sudo mv $TEMPLATE_DIR/iptables-restore.service /etc/eks/iptables-restore.service
+sudo mv "${TEMPLATE_DIR}/iptables-restore.service" /etc/eks/iptables-restore.service
```
guessi, line 119: `TEMPLATE_DIR` is customizable variable, defined at LINE-10.
If input path contains `space` it would became a problem.

scripts/install-worker.sh
```diff
@@ -240,88 +243,87 @@ sudo mkdir -p /var/lib/kubernetes
...
 # Since CNI 0.7.0, all releases are done in the plugins repo.
 CNI_PLUGIN_FILENAME="cni-plugins-linux-${ARCH}-${CNI_PLUGIN_VERSION}"
 
-if [ "$PULL_CNI_FROM_GITHUB" = "true" ]; then
+if [ "${PULL_CNI_FROM_GITHUB}" = "true" ]; then
   echo "Downloading CNI plugins from Github"
   wget "https://github.com/containernetworking/plugins/releases/download/${CNI_PLUGIN_VERSION}/${CNI_PLUGIN_FILENAME}.tgz"
   wget "https://github.com/containernetworking/plugins/releases/download/${CNI_PLUGIN_VERSION}/${CNI_PLUGIN_FILENAME}.tgz.sha512"
-  sudo sha512sum -c "${CNI_PLUGIN_FILENAME}.tgz.sha512"
+  sha512sum -c "${CNI_PLUGIN_FILENAME}.tgz.sha512"
```
guessi, line 294: no need to pass `sudo` here.

scripts/install-worker.sh
```diff
@@ -240,88 +243,87 @@ sudo mkdir -p /var/lib/kubernetes
...
+    wget "$S3_URL_BASE/${CNI_PLUGIN_FILENAME}.tgz.sha256"
   fi
-  sudo sha256sum -c "${CNI_PLUGIN_FILENAME}.tgz.sha256"
+  sha256sum -c "${CNI_PLUGIN_FILENAME}.tgz.sha256"
+  rm "${CNI_PLUGIN_FILENAME}.tgz.sha256"
 fi
 sudo tar -xvf "${CNI_PLUGIN_FILENAME}.tgz" -C /opt/cni/bin
 rm "${CNI_PLUGIN_FILENAME}.tgz"
 
-sudo rm ./*.sha256
```
guessi, line 308: moved to Line-307.

scripts/install-worker.sh
```diff
@@ -240,88 +243,87 @@ sudo mkdir -p /var/lib/kubernetes
...
+sudo mv "${TEMPLATE_DIR}/kubelet-kubeconfig" /var/lib/kubelet/kubeconfig
 sudo chown root:root /var/lib/kubelet/kubeconfig
 
 # Inject CSIServiceAccountToken feature gate to kubelet config if kubernetes version starts with 1.20.
 # This is only injected for 1.20 since CSIServiceAccountToken will be moved to beta starting 1.21.
-if [[ $KUBERNETES_VERSION == "1.20"* ]]; then
-  KUBELET_CONFIG_WITH_CSI_SERVICE_ACCOUNT_TOKEN_ENABLED=$(cat $TEMPLATE_DIR/kubelet-config.json | jq '.featureGates += {CSIServiceAccountToken: true}')
-  echo $KUBELET_CONFIG_WITH_CSI_SERVICE_ACCOUNT_TOKEN_ENABLED > $TEMPLATE_DIR/kubelet-config.json
+if [[ "${KUBERNETES_VERSION}" == "1.20"* ]]; then
+  KUBELET_CONFIG_WITH_CSI_SERVICE_ACCOUNT_TOKEN_ENABLED=$(jq '.featureGates += {CSIServiceAccountToken: true}' "${TEMPLATE_DIR}/kubelet-config.json")
```
guessi, line 320: replace `cat FILE | jq ...` with `jq ... FILE`.

scripts/install-worker.sh
```diff
@@ -333,49 +335,49 @@ sudo systemctl disable kubelet
...
 ### Cache Images ###############################################################
 ################################################################################
-if [[ "$CACHE_CONTAINER_IMAGES" == "true" && "$BINARY_BUCKET_REGION" != "us-iso-east-1" && "$BINARY_BUCKET_REGION" != "us-isob-east-1" ]]; then
+if [[ "${CACHE_CONTAINER_IMAGES}" == "true" && "${BINARY_BUCKET_REGION}" != "us-iso-east-1" && "${BINARY_BUCKET_REGION}" != "us-isob-east-1" ]]; then
   AWS_DOMAIN=$(imds 'latest/meta-data/services/domain')
   ECR_URI=$(/etc/eks/get-ecr-uri.sh "${BINARY_BUCKET_REGION}" "${AWS_DOMAIN}")
 
   PAUSE_CONTAINER="${ECR_URI}/eks/pause:${PAUSE_CONTAINER_VERSION}"
-  cat /etc/eks/containerd/containerd-config.toml | sed s,SANDBOX_IMAGE,$PAUSE_CONTAINER,g | sudo tee /etc/eks/containerd/containerd-cached-pause-config.toml
+  sed "s,SANDBOX_IMAGE,${PAUSE_CONTAINER},g" /etc/eks/containerd/containerd-config.toml | sudo tee /etc/eks/containerd/containerd-cached-pause-config.toml
```
guessi, line 380: It is not about in-place replacement, so we could replace `cat FILE | sed ...` with `sed ... FILE`.

scripts/generate-version-info.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/usr/bin/env bash
```
guessi, line 1: keep all scripts' Shebang  aligned.

Makefile
```diff
@@ -61,11 +61,13 @@ ifeq (, $(SHELLCHECK_COMMAND))
 SHELLCHECK_COMMAND = docker run --rm -v $(MAKEFILE_DIR):$(MAKEFILE_DIR) koalaman/shellcheck:stable
 endif
 SHELL_FILES := $(shell find $(MAKEFILE_DIR) -type f -name '*.sh')
+SHELL_FILES_SCRIPTS := $(shell find $(MAKEFILE_DIR)/scripts -type f -name '*.sh')
 
 .PHONY: lint
 lint: ## Check the source files for syntax and format issues
 	$(SHFMT_COMMAND) $(SHFMT_FLAGS) --diff $(MAKEFILE_DIR)
 	$(SHELLCHECK_COMMAND) --format gcc --severity error $(SHELL_FILES)
+	$(SHELLCHECK_COMMAND) --format gcc --severity warning $(SHELL_FILES_SCRIPTS)
```
guessi, line 70: relative strict shellcheck for `scripts/*.sh`

scripts/install-worker.sh
```diff
@@ -193,26 +196,26 @@ EOF
...
-if [[ "$INSTALL_DOCKER" == "true" ]]; then
+if [[ "${INSTALL_DOCKER}" == "true" ]]; then
   sudo amazon-linux-extras enable docker
   sudo groupadd -og 1950 docker
-  sudo useradd --gid $(getent group docker | cut -d: -f3) docker
+  sudo useradd --gid "$(getent group docker | cut -d: -f3)" docker
 
   # install docker and lock version
-  sudo yum install -y docker-${DOCKER_VERSION}*
+  sudo yum install -y "docker-${DOCKER_VERSION}"*
```
guessi, line 210: Do we really need `*` here?

```bash
# yum search docker
============================================================================================================ N/S matched: docker =============================================================================================================
pcp-pmda-docker.x86_64 : Performance Co-Pilot (PCP) metrics from the Docker daemon
amazon-ecr-credential-helper.x86_64 : Amazon ECR Docker Credential Helper
docker.x86_64 : Automates deployment of containerized applications
nerdctl.x86_64 : nerdctl is a Docker-compatible CLI for containerd.
oci-add-hooks.x86_64 : Injects OCI hooks as a Docker runtime

  Name and summary matches only, use "search all" for everything.
```

scripts/install-worker.sh
```diff
@@ -240,88 +243,87 @@ sudo mkdir -p /var/lib/kubernetes
...
+    aws s3 cp --region "${BINARY_BUCKET_REGION}" "${S3_PATH}/${CNI_PLUGIN_FILENAME}.tgz.sha256" .
   else
     echo "AWS cli missing - using wget to fetch cni binaries from s3. Note: This won't work for private bucket."
-    sudo wget "$S3_URL_BASE/${CNI_PLUGIN_FILENAME}.tgz"
-    sudo wget "$S3_URL_BASE/${CNI_PLUGIN_FILENAME}.tgz.sha256"
+    wget "$S3_URL_BASE/${CNI_PLUGIN_FILENAME}.tgz"
+    wget "$S3_URL_BASE/${CNI_PLUGIN_FILENAME}.tgz.sha256"
   fi
-  sudo sha256sum -c "${CNI_PLUGIN_FILENAME}.tgz.sha256"
+  sha256sum -c "${CNI_PLUGIN_FILENAME}.tgz.sha256"
```
guessi, line 306: no need to pass `sudo` here.

scripts/validate.sh
```diff
@@ -47,27 +47,27 @@ fi
 
 echo "Verifying that the package versionlocks are correct..."
 
-function versionlock-entries() {
+versionlock-entries() {
```
cartermckinnon, line 50: We've taken the approach of including *all* the optional shell grammar for function definitions (including the `function` keyword, parentheses after the `fname` and curly braces around the compound command). Please don't remove this.

cartermckinnon (changes requested):
I don't have any large concerns here, but I'd like to see some of these changes split across PR's given the number of lines touched. For example, the majority of changes here are adding curly braces around variable expansions and wrapping arguments in quotes. We should do all of those changes in a single PR.

Other things like removing unnecessary `sudo` usage should be done in separate PR's; though unless you've crawled the git history and verified that these were in fact added for no reason, I don't see much reason to change it.

scripts/validate.sh
guessi: sure, I simply trying to make all the function-style in the same way under `./scripts/` folder.

Makefile
guessi: nope, it's just my test code left behind, I will remove it later.

guessi:
@cartermckinnon I've updated changes, please review while you are free, thanks!

scripts/generate-version-info.sh
```diff
@@ -13,11 +13,15 @@ fi
 OUTPUT_FILE="$1"
 
 # packages
-sudo rpm --query --all --queryformat '\{"%{NAME}": "%{VERSION}-%{RELEASE}"\}\n' | jq --slurp --sort-keys 'add | {packages:(.)}' > "$OUTPUT_FILE"
+sudo rpm --query --all --queryformat '\{"%{NAME}": "%{VERSION}-%{RELEASE}"\}\n' | jq --slurp --sort-keys 'add | {packages:(.)}' > "${OUTPUT_FILE}"
+
+# version detection
+KUBELET_VERSION=$(kubelet --version | cut -d' ' -f2)
+AWSCLI_VERSION=$(aws --version | cut -d' ' -f1 | cut -d'/' -f2)
```
cartermckinnon, line 20: nit: move these variables immediately prior to their usage under the `# binaries` comment.

scripts/install-worker.sh
cartermckinnon: The asterisk would allow more flexible values of the `docker_version` template variable to be used, but IMO the user should specify that, we shouldn't hard code it here.

That being said, I consider all Docker code paths to be frozen at this point. This will be removed when 1.23 is EOL and shouldn't be modified until then.

scripts/install-worker.sh
cartermckinnon: This line is intended to remove *all* checksum files, including the ones downloaded in this block: https://github.com/awslabs/amazon-eks-ami/blob/master/scripts/install-worker.sh#L259-L272

So, moving this to 307 will leave behind those other checksum files.

However, I don't think we should do cleanup in this way during the build. In #1231 , we'll remove the directory used to store all of these build-time temporary bits, we don't need to remove them all individually.

This PR will cause merge conflicts with #1231 anyway, so can you revert these changes and I'll handle it in that PR?

scripts/install-worker.sh
```diff
@@ -518,7 +520,7 @@ echo vm.max_map_count=524288 | sudo tee -a /etc/sysctl.conf
 ### adding log-collector-script ################################################
 ################################################################################
 sudo mkdir -p /etc/eks/log-collector-script/
-sudo cp $TEMPLATE_DIR/log-collector-script/eks-log-collector.sh /etc/eks/log-collector-script/
+sudo cp "${TEMPLATE_DIR}/log-collector-script/eks-log-collector.sh" "/etc/eks/log-collector-script/"
```
cartermckinnon, line 523: You don't need to quote the destination path.

scripts/validate.sh
```diff
@@ -10,7 +10,7 @@
 #   1 if a file exists, after printing an error
 validate_file_nonexists() {
   local file_blob=$1
-  for f in $file_blob; do
+  for f in ${file_blob}; do
     if [ -e "$f" ]; then
```
cartermckinnon, line 14: You don't need braces in `"$f"`?

cartermckinnon (changes requested):
I see you reverted your change to the severity of shellcheck -- I think that's fine to include in this PR, if the braces + quoting changes is sufficient for that shellcheck level to pass.

scripts/validate.sh
guessi: nice catch!

scripts/install-worker.sh
guessi: nice catch!

scripts/generate-version-info.sh
```diff
@@ -13,11 +13,17 @@ fi
...
+echo $(jq ".binaries.kubelet = \"${KUBELET_VERSION}\"" "${OUTPUT_FILE}") > "${OUTPUT_FILE}"
+
+AWSCLI_VERSION=$(aws --version | cut -d' ' -f1 | cut -d'/' -f2)
+# shellcheck disable=SC2046
+echo $(jq ".binaries.awscli = \"${AWSCLI_VERSION}\"" "${OUTPUT_FILE}") > "${OUTPUT_FILE}"
 
 # cached images
-echo $(jq ".images = [ $(sudo ctr -n k8s.io image ls -q | cut -d'/' -f2- | sort | uniq | grep -v 'sha256' | xargs -r printf "\"%s\"," | sed 's/,$//') ]" $OUTPUT_FILE) > $OUTPUT_FILE
+# shellcheck disable=SC2046
+echo $(jq ".images = [ $(sudo ctr -n k8s.io image ls -q | cut -d'/' -f2- | sort | uniq | grep -v 'sha256' | xargs -r printf "\"%s\"," | sed 's/,$//') ]" "${OUTPUT_FILE}") > "${OUTPUT_FILE}"
```
guessi, line 29: I think it is hard to make `shellcheck` happy `SC2046` here if I don't put a comment here.

scripts/install-worker.sh
guessi: I prefer to leave it as it, not going change it's logic in this PR. Just let `shellcheck` pass 😄

scripts/install-worker.sh
guessi: I see... How about `sudo rm ./*.sha256 ./*.sha512` then? will it still be a conflict?

scripts/install-worker.sh
```diff
@@ -305,23 +307,23 @@ fi
 sudo tar -xvf "${CNI_PLUGIN_FILENAME}.tgz" -C /opt/cni/bin
 rm "${CNI_PLUGIN_FILENAME}.tgz"
 
-sudo rm ./*.sha256
+sudo rm ./*.sha256 ./*.sha512
```
guessi, line 310: I see... How about `sudo rm ./*.sha256 ./*.sha512` then? will it still be a conflict?

guessi:
> I see you reverted your change to the severity of shellcheck -- I think that's fine to include in this PR, if the braces + quoting changes is sufficient for that shellcheck level to pass.

Actually, I tried to change severity from `error` to `warning` but that would made PR scope too wide.

- `test/cases/*.sh`
- `test/*.sh`
- `files/*.sh`
- `log-collector-script/linux/*.sh`

I think we should keep PR small as Phase 1 cleanup, the do it one-by-one in separated PR.

that explains why I made change to `Makefile` [before](https://github.com/awslabs/amazon-eks-ami/compare/d71aca7ab984dc2e660bffe1b94bbdc736ce68b6..daf51eb0f08190bb003a19b7abd2eb3cf2473855) trying to do `--severity error` for all, but `--severity warning` for `./scripts` only.

guessi:
I tried to make changes to scripts under `test/` and `files/` but I found it too hard to get them all fixed, as there are too many scripts not able to get `shellcheck (--severity warning)` pass.

Not sure if I should keep this PR opened, or maybe I should focus on `scripts/` only? WDYT?

cartermckinnon:
Lots of conflicts to address, please open new PR(s) with scoped-down changes if you'd still like to change the style requirements.

PR #1250, opened 2023-04-04
Remove --container-runtime kubelet flag for 1.27+

cartermckinnon:
**Description of changes:**

This flag is gone in 1.27, so it's removed in the default `systemd` service unit, and conditionally added back for 1.26-.

**Testing Done**

TODO


files/bootstrap.sh
```diff
@@ -191,6 +191,8 @@ if vercmp "$KUBELET_VERSION" lt "1.27.0"; then
 fi
 MOUNT_BPF_FS="${MOUNT_BPF_FS:-$DEFAULT_MOUNT_BPF_FS}"
 
+KUBELET_ARGS=""
```
mmerkes, line 194: Why is this needed?

cartermckinnon: I was misreading line numbers while jumping around this file -- it's not needed, for now.

PR #1252, opened 2023-04-05
Parameterize Packer template and default variable file

cartermckinnon:
**Description of changes:**

This allows us to reuse the Makefile for other AMI templates.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,10 +1,12 @@
+MAKEFILE_DIR := $(shell dirname $(realpath $(firstword $(MAKEFILE_LIST))))
+
+PACKER_DEFAULT_VARIABLE_FILE ?= $(MAKEFILE_DIR)/eks-worker-al2-variables.json
+PACKER_TEMPLATE_FILE ?= $(MAKEFILE_DIR)/eks-worker-al2.json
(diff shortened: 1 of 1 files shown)


xr1776 (commented):
There is a pattern, I would say but likely you know better, of AMI changes having unexpected impact on customers. Edge cases we would not think of where a customer is impacted because they are doing something differently.
The last time this happened I reflected on the fact that many of these changes, to my knowledge, don't have corresponding tests.

Ideally, the set of possible configurations would be defined in data and run automatically, blocking check in if failure occurs. 
1. Am I correct in saying we do not have this generally?
2. If not, do you think we should do this generally?
3. Am I correct in saying we don't have tests exercising this code change?
4. If not, do you think we should create them?

cartermckinnon:
This change only adds new variables, it's not really possible to break existing use cases.

We implicitly "test" the Makefile by using targets like `test` in the PR unit test check. The best way to test other targets is just to use them, in my opinion; "unit testing" the Makefile isn't the right approach.

xr1776 (approved):

mmerkes (approved):

PR #1255, opened 2023-04-05
[DISCUSSION ONLY][DO NOT MERGE] pull upstream binaries related to k8s

dims:
**Issue #, if available:**

**Description of changes:**

**Testing Done**


scripts/install-worker.sh
```diff
@@ -240,41 +240,54 @@ sudo mkdir -p /var/lib/kubernetes
...
-    sudo wget $S3_URL_BASE/$binary.sha256
+if [ "$PULL_K8S_FROM_UPSTREAM" = "true" ]; then
+  echo "Downloading binaries from upstream"
+  sudo wget "https://dl.k8s.io/release/v$KUBERNETES_VERSION/bin/linux/amd64/kubelet"
+  sudo wget "https://dl.k8s.io/release/v$KUBERNETES_VERSION/bin/linux/amd64/kubelet.sha256"
+  echo "$(cat kubelet.sha256) kubelet" | sha256sum -c -
+  sudo chmod +x kubelet
+  sudo mv kubelet /usr/bin/
+
+  wget -O aws-iam-authenticator "https://github.com/kubernetes-sigs/aws-iam-authenticator/releases/download/v0.6.2/aws-iam-authenticator_0.6.2_linux_amd64"
```
cartermckinnon, line 251: Hm, I think it's kind of odd that we store `aws-iam-authenticator` and `ecr-credential-provider` under each Kubernetes semver + build date prefix in the S3 bucket. Those dependencies should probably be separate (and therefore configurable).

cartermckinnon:
I think this is useful.

I threw together something similar for `containerd` and `runc` last week. It was less straightforward because we need to e.g. create the systemd units, default config files, etc. that the RPM does for us today. Not a huge deal, but more opportunity for subtle differences to be introduced with the two installation methods.

scripts/install-worker.sh
```diff
@@ -240,41 +240,54 @@ sudo mkdir -p /var/lib/kubernetes
...
+    sudo chmod +x $binary
+    sudo mv $binary /usr/bin/
+  done
+fi
 
 # Verify that the aws-iam-authenticator is at last v0.5.9 or greater. Otherwise, nodes will be
 # unable to join clusters due to upgrading to client.authentication.k8s.io/v1beta1
 iam_auth_version=$(sudo /usr/bin/aws-iam-authenticator version | jq -r .Version)
-if vercmp "$iam_auth_version" lt "v0.5.9"; then
+if vercmp "v$iam_auth_version" lt "v0.5.9"; then
```
guessi, line 290: It looks like a bug? maybe it should be handled separately...

scripts/install-worker.sh
```diff
@@ -352,7 +365,11 @@ fi
 ################################################################################
 if vercmp "$KUBERNETES_VERSION" gteq "1.22.0"; then
   ECR_BINARY="ecr-credential-provider"
-  if [[ -n "$AWS_ACCESS_KEY_ID" ]]; then
+
+  if [ "$PULL_K8S_FROM_UPSTREAM" = "true" ]; then
+    echo "copy ecr-credential-provider binary from upstream."
+    wget -O $ECR_BINARY "https://storage.googleapis.com/k8s-artifacts-prod/binaries/cloud-provider-aws/v1.26.1/linux/amd64/ecr-credential-provider-linux-amd64"
```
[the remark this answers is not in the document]
cartermckinnon: yeah, this is just a PoC, though

scripts/install-worker.sh
cartermckinnon: This is actually correct -- for the `aws-iam-authenticator` binaries in the `amazon-eks` S3 bucket. I found a discrepancy in our build process that causes the version string to look different in our builds versus the ones in the upstream GitHub releases. I have a PR open internally.

PR #1265, opened 2023-04-14
Allow runc and containerd installation skips in install-worker

baykovr:
**Description of changes:**
This patch provides two optional environment variables of 
* `SKIP_INSTALL_RUNC` 
* `SKIP_INSTALL_CONTAINERD` 

Similar in function and implementation of `INSTALL_DOCKER`, when one or both of the environment variables are set install-worker.sh will skip over the yum install and version lock steps, respectfully to the vars.

By default, the variables being undefined and optional, no change to install behavior is made.

**Justification**
In cases where containerd or runc are installed from outside of yum availability (such as from source) the current yum install steps will fail, triggering `errexit`. This option allows for greater flexibility, but is sufficiently simple as to not introduce unnecessary complexity or upkeep burden.

**Testing Done**
Followed your instruction, validated that regression to the existing installation is not introduced. (Additionally tested AMI construction on my side.)
```
$ make fmt test 
...
===============================
✅ ✅ All Tests Passed! ✅ ✅
```

### DIFF
--- a/scripts/install-worker.sh
+++ b/scripts/install-worker.sh
@@ -147,13 +147,21 @@ sudo mv "${TEMPLATE_DIR}/runtime.slice" /etc/systemd/system/runtime.slice
 ### Containerd setup ##########################################################
 ###############################################################################
 
-# install runc and lock version
-sudo yum install -y runc-${RUNC_VERSION}
-sudo yum versionlock runc-*
+if [[ ! -v "SKIP_INSTALL_RUNC" ]]; then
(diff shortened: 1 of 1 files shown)


cartermckinnon:
I think it's worthwhile to support other installation methods for `containerd` and `runc`, but I don't think we should make this change in isolation. Without exposing these options as template variables (which I don't think we should do), they can only be used by referencing this installation script in a custom Packer template, which is prone to breakage and not something we can support.

PR #1269, opened 2023-04-18
Use v1 credential provider API in 1.27+

### LINKED ISSUE #1268: Handle credentialprovider.kubelet.k8s.io/v1alpha1 -> v1 migration in 1.27+
The `ecr-credential-provider` will need to be registered with the `v1` credential provider API after this is merged: https://github.com/kubernetes/cloud-provider-aws/pull/597

We should probably go ahead and move to the v1 kubelet config API's across the board.

cartermckinnon:
**Issue #, if available:**

Closes #1268 .

**Description of changes:**

This moves to the `v1` API for `CredentialProviderRequest`/`CredentialProviderResponse`. These API's have graduated to `v1`.

**Testing Done**

You can observe the issue by placing a shim in between `kubelet` and the `ecr-credential-provider`.

1. Create `cred-provider-shim`:
```
#!/usr/bin/env bash
REQUEST=$(cat -)
LOG_FILE=/tmp/cred-provider-shim.log
function log() {
  echo "$(date)" "$@" >> $LOG_FILE
}
log "request: $REQUEST"
RESPONSE=$(mktemp)
echo "$REQUEST" | /etc/eks/ecr-credential-provider/ecr-credential-provider > $RESPONSE 2>> $LOG_FILE
log "response: $(cat $RESPONSE)"
cat "$RESPONSE"
```
2. Modify the `ecr-credential-provider-config` to use this executable.
3. Modify kubelet's `--image-credential-provider-bin-dir` flag to the directory containing the shim executable.
4. Create a test pod on the node using an image that matches the ECR pattern:
```
kubectl \
  run test-pod \
  --image 123456789123.dkr.ecr.us-west-2.amazonaws.com/foo \
  --overrides='{"apiVersion": "v1", "spec": {"nodeSelector": { "kubernetes.io/hostname": "$HOSTNAME" }}}' 
```
5. Look at `/tmp/cred-provider-shim.log`:
```
Tue Apr 18 06:12:14 UTC 2023 request: {"kind":"CredentialProviderRequest","apiVersion":"credentialprovider.kubelet.k8s.io/v1beta1","image":"123456789123.dkr.ecr.us-west-2.amazonaws.com/foo"}
E0418 06:12:14.490885   14903 main.go:161] Error running credential provider plugin: group version credentialprovider.kubelet.k8s.io/v1beta1 is not supported
Tue Apr 18 06:12:14 UTC 2023 response:
```

You can also change the `apiVersion` in `ecr-credential-provider-config` to `v1alpha1` and see a token generated successfully.

### DIFF
--- a/doc/USER_GUIDE.md
+++ b/doc/USER_GUIDE.md
@@ -318,6 +318,6 @@ sudo yum versionlock clear
 Prior to Kubernetes 1.27, the `kubelet` could obtain credentials for ECR out of the box. This legacy credential process has been removed in Kubernetes 1.27, and
 ECR credentials should now be obtained via a plugin, the `ecr-credential-provider`. This plugin is installed in the AMI at `/etc/eks/image-credential-provider/ecr-credential-provider`. More information about this plugin is available in the [`cloud-provider-aws` documentation](https://cloud-provider-aws.sigs.k8s.io/credential_provider/).
 
-Additional image credential provider plugins may be appended to `/etc/eks/image-credential-provider/config.json`. In Kubernetes versions before 1.27, all plugins in this file must support `credentialprovider.kubelet.k8s.io/v1alpha1`. In Kubernetes versions 1.27 and above, they must support `credentialprovider.kubelet.k8s.io/v1`.
+Additional image credential provider plugins may be appended to `/etc/eks/image-credential-provider/config.json`. In Kubernetes versions 1.26 and below, all plugins in this file must support `credentialprovider.kubelet.k8s.io/v1alpha1`. In Kubernetes versions 1.27 and above, they must support `credentialprovider.kubelet.k8s.io/v1`.
 
 For more information about image credential provider plugins, refer to the [Kubernetes documentation](https://kubernetes.io/docs/tasks/administer-cluster/kubelet-credential-provider/).
--- a/doc/USER_GUIDE.md
+++ b/doc/USER_GUIDE.md
@@ -11,6 +11,7 @@ This document includes details about using the AMI template and the resulting AM
 1. [AL2 and Linux kernel information](#al2-and-linux-kernel-information)
 1. [Updating known instance types](#updating-known-instance-types)
 1. [Version-locked packages](#version-locked-packages)
+1. [Image credential provider plugins](#image-credential-provider-plugins)
 
 ---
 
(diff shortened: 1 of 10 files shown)


cartermckinnon:
Something is off with our cred provider config.

The `ecr-credential-provider` only supported the `v1alpha1` API prior to this change: https://github.com/kubernetes/cloud-provider-aws/pull/597, and we changed the configuration file to specify `v1beta1` back when we launched 1.24: https://github.com/awslabs/amazon-eks-ami/commit/a521047d1b097b9c3dbb562ca9bdab5a641f347f.

We haven't disabled the in-tree credential providers by way of kubelet's `DisableKubeletCloudCredentialProviders` feature gate. This means that the in-tree ECR provider was always registered. Providers are sorted alphabetically when building the keyring, so `amazon-ecr` always outranks `ecr-credential-provider`: https://github.com/kubernetes/kubernetes/blob/master/pkg/credentialprovider/plugins.go#L68

So, we'd always fall back to the in-tree `amazon-ecr` credential provider.

If we invoked `ecr-credential-provider` with `v1alpha1`, it wouldn't fail, because it's only comparing the API group, not the version: https://github.com/kubernetes/cloud-provider-aws/blob/release-1.26/cmd/ecr-credential-provider/plugin.go#L126-L128

But, it would respond with a `v1alpha1` response and kubelet should fail: https://github.com/kubernetes/kubernetes/blob/release-1.26/pkg/credentialprovider/plugin/plugin.go#L424

cartermckinnon:
The in-tree ECR credential provider is gone in 1.27.

cartermckinnon:
This likely went unnoticed because the stdout/stderr from failed cred provider plugin invocations is discarded: https://github.com/kubernetes/kubernetes/blob/master/pkg/credentialprovider/plugin/plugin.go#L413

Throwing that output in the logs is a security risk, so IIWII.

prasad0896 (approved):

mmerkes (commented):
What kind of validation did you do to confirm that your fix works? Probably need to validation a version 1.24-1.26 and 1.27 to ensure it works correctly.

cartermckinnon:
@mmerkes We'll need to turn on the `DisableKubeletCloudCredentialProviders` feature gate to force usage of the `ecr-credential-provider`. If the standard pods (`kube-proxy`, `aws-node`, etc.) stabilize, then we're good. I'll post a cluster config file here and my results once that's been run.

cartermckinnon:
I tested this with 1.26 by:
1. Create `config.yaml`:
```
---
kind: ClusterConfig
metadata:
  name: "126-credprovider"
  region: us-west-2
  version: "1.26"
managedNodeGroups:
  - name: nodes
    ami: ami-022441ec63297a0c9
    amiFamily: AmazonLinux2
    minSize: 1
    maxSize: 1
    desiredCapacity: 1
    overrideBootstrapCommand: |
      #!/bin/bash
      echo "$(jq '.featureGates.DisableKubeletCloudCredentialProviders = true' /etc/kubernetes/kubelet/kubelet-config.json)" > /etc/kubernetes/kubelet/kubelet-config.json
      /etc/eks/bootstrap.sh 126-credprovider --kubelet-extra-args "--node-labels=eks.amazonaws.com/nodegroup=nodes,eks.amazonaws.com/nodegroup-image=ami-022441ec63297a0c9"
```
2. `eksctl create cluster --config-file config.yaml`
3. Nodes joined the cluster.
4. Created a test `Pod` with an ECR image (that is *not* cached in the AMI, to be certain).
5. Image pulls succeeded.

cartermckinnon:
@prasad0896 has tested this with 1.27.

dims:
@cartermckinnon if a feature flag in an old release is not already on, then i don't think we should go back and switch it on... 2 cents.

doc/USER_GUIDE.md
```diff
@@ -309,3 +310,14 @@ sudo yum versionlock delete $PACKAGE_NAME
 # unlock all packages
 sudo yum versionlock clear
 ```
+
+---
+
+## Image credential provider plugins
```
mmerkes, line 316: Tx for the addition here.

PR #1275, opened 2023-04-21
Fix imds setting for multiple enis on ipv6

### LINKED ISSUE #1103: bootstrap uses incorrect IP when multiple ENI exists on node
**What happened**:
when multiple ENI exists on node prior to bootstrap, the IP address of ENI other than eth0 may be used and passed to kubelet.
Which then may fail [kubelet's IP address validation](https://github.com/kubernetes/kubernetes/blob/3f823c0daa002158b12bfb2d53bcfe433516659d/pkg/kubelet/kubelet_node_status.go#L793)

we should always use eth0's IP address when generate node's IP.

keep launch nodes with 2 enis, and sometimes node won't get IP.

In https://github.com/awslabs/amazon-eks-ami/blob/master/files/bootstrap.sh#L430 we used the first MAC instead of eth0's mac.
i believe we should just use [latest/meta-data/mac](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-categories.html) (The instance's media access control (MAC) address. In cases where multiple network interfaces are present, this refers to the eth0 device (the device for which the device number is 0).)

The code change was introduced in https://github.com/awslabs/amazon-eks-ami/pull/773

**Environment**:
- AWS Region: us-west-2
- Instance Type(s): c5.9xlarge
- EKS Platform version (use `aws eks describe-cluster --name <name> --query cluster.platformVersion`):
- Kubernetes version (use `aws eks describe-cluster --name <name> --query cluster.version`):
- AMI Version: 
- Kernel (e.g. `uname -a`):

```
```

willgleich:
**Issue #, if available:**
Closes #1103 .
**Description of changes:**
This changes the imds metadata path to choose the [latest/meta-data/mac](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-categories.html) rather than the network interfaces level which can incorrectly label the node ip field of the kubelet args when an ipv6 instance has multiple ENIs. The incorrect latest/meta-data/network/interfaces/macs/ will not guarantee eth0 as the mac address choice causing the below errors. 

This will resolve the following kubectl error when trying to view pod logs on an ipv6 incorrectly configured instance:
```
Error from server: no preferred addresses found; known addresses: []
```

Or in the kubelet logs it will say:
```
Failed to set some node status fields" err="failed to validate nodeIP: node IP:
```

**Testing Done**
I've modified this line using patch on my user_data script to this new imds path to ensure the functionality works as intended.

### DIFF
--- a/files/bootstrap.sh
+++ b/files/bootstrap.sh
@@ -400,7 +400,7 @@ fi
 
 ### kubelet.service configuration
 
-MAC=$(imds 'latest/meta-data/network/interfaces/macs/' | head -n 1 | sed 's/\/$//')
+MAC=$(imds 'latest/meta-data/mac')
 
 if [[ -z "${DNS_CLUSTER_IP}" ]]; then
   if [[ "${IP_FAMILY}" == "ipv6" ]]; then


ljosyula (approved):
lgtm

Issue #1214, opened 2023-03-06
can we upgrade cni plugin version


yongzhang:
Hi, team

Seems like cni plugin version has been stick to `v0.8.6` for a long time, can we upgrade it to latest version(`v1.2.0`)?

I'm asking because we need to install `nerdctl` which requires higher version of cni plugin, thanks.

anavarreteuk:
In addition, the tag `v0.8.6` comes with CVE-2022-27664.

cartermckinnon:
The AWS VPC CNI installs these plugins itself, so we're evaluating removing them from the AMI entirely, rather than updating them. @yongzhang does your use case require them to be baked into the AMI?

If so, would it be reasonable to install them yourself, since you're already installing `nerdctl`?

anavarreteuk:
> The AWS VPC CNI installs newer versions of a subset of these plugins itself, so we're evaluating removing them from the AMI entirely, rather than updating them here.
> 
> @yongzhang does your use case require them to be baked into the AMI? If so, would it be reasonable to install them yourself, since you're already installing `nerdctl`?

Even better, thanks.

yongzhang:
> The AWS VPC CNI installs newer versions of a subset of these plugins itself, so we're evaluating removing them from the AMI entirely, rather than updating them here.
> 
> @yongzhang does your use case require them to be baked into the AMI? If so, would it be reasonable to install them yourself, since you're already installing `nerdctl`?

This should be cool, and yea I'm building a custom AMI based in this repo but using ubuntu os.

cartermckinnon:
I'm reopening this to track removal of the plugins (or an update, if it comes to that).

anavarreteuk:
> I'm reopening this to track removal of the plugins (or an update, if it comes to that).

I will be keen to create the PR once the decision is made.

cartermckinnon:
This has been resolved, we're on `1.2.0` now. 👍

PR #1279, opened 2023-04-25
update CNI plugin version

jdn5126:
**Issue #, if available:**
#1214 

**Description of changes:**
This PR upgrades the CNI plugin versions pulled into the AMI to be the latest upstream release: `v1.2.0`: https://github.com/containernetworking/plugins/releases/tag/v1.2.0

This is done for all supported k8s versions as there are no backward compatibility issues.

**Testing Done**
Verified that AMI could be built and tests passed. Built AMI with `pull_cni_from_github` set to true and false.


cartermckinnon:
Marking as draft until these binaries are available in the public `amazon-eks` bucket.

PR #1283, opened 2023-04-25
add pigz to explicitly installed packages

steved:
**Issue #, if available:** N/A

**Description of changes:**

When Docker was installed it explicitly depends on `pigz`, however in 1.25+ without Docker it is no longer installed.

```
$ yum deplist docker | grep pigz
  dependency: pigz
   provider: pigz.x86_64 2.3.4-1.amzn2.0.1
```

**Testing Done**

```
# Before
$ time ctr i pull docker.io/library/python:latest > /dev/null

real    0m14.779s
user    0m0.832s
sys     0m0.473s 

# After (requires containerd restart)
$ time ctr i pull docker.io/library/python:latest > /dev/null

real    0m10.890s
user    0m0.768s
sys     0m0.507s
```

### DIFF
--- a/scripts/install-worker.sh
+++ b/scripts/install-worker.sh
@@ -71,7 +71,8 @@ sudo yum install -y \
   wget \
   yum-utils \
   yum-plugin-versionlock \
-  mdadm
+  mdadm \
+  pigz
 
 # Remove any old kernel versions. `--count=1` here means "only leave 1 kernel version installed"
 sudo package-cleanup --oldkernels --count=1 -y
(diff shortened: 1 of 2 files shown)


cartermckinnon:
This is great, thanks @steved. 

I don't see any way to force `containerd` to use only `unpigz`. Maybe we should add a check to `scripts/validate.sh` so we can at least fail the build if this isn't installed?

steved:
Sure! Let me know what you think of 4fb3469.

cartermckinnon:
Perfect, thanks!

cartermckinnon (approved):

PR #1285, opened 2023-04-26
Allow to set pause container image repository name via bootstrap.sh

guessi:
**Issue #, if available:**

**Description of changes:**

Before this change, if a user set `--pause-container-account` and `--pause-container-version` but not knowing there's a hidden limitation `ecr repository name should be "eks/pause"` by default, node will not able to join cluster because `pause` container not ready.

**Testing Done**

### DIFF
--- a/files/bootstrap.sh
+++ b/files/bootstrap.sh
@@ -34,9 +34,10 @@ function print_help {
   echo "--kubelet-extra-args Extra arguments to add to the kubelet. Useful for adding labels or taints."
   echo "--local-disks Setup instance storage NVMe disks in raid0 or mount the individual disks for use by pods [mount | raid0]"
   echo "--mount-bpf-fs Mount a bpffs at /sys/fs/bpf (default: true, for Kubernetes 1.25+; false otherwise)"
-  echo "--pause-container-account The AWS account (number) to pull the pause container from"
-  echo "--pause-container-repository-name The AWS ECR repository name to pull the pause container from"
-  echo "--pause-container-version The tag of the pause container"
+  echo "--pause-container-account The AWS account (number) to pull the pause container from. This flag is deprecated and will be removed in a future release."
+  echo "--pause-container-repository-name The AWS ECR repository name to pull the pause container from. This flag is deprecated and will be removed in a future release."
+  echo "--pause-container-version The tag of the pause container. This flag is deprecated and will be removed in a future release."
+  echo "--pause-container-image-uri The image URI for pause image."
   echo "--service-ipv6-cidr ipv6 cidr range of the cluster"
   echo "--use-max-pods Sets --max-pods for the kubelet when true. (default: true)"
 }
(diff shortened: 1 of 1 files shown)


files/bootstrap.sh
```diff
@@ -105,7 +106,13 @@ while [[ $# -gt 0 ]]; do
       ;;
     --pause-container-account)
       PAUSE_CONTAINER_ACCOUNT=$2
-      log "INFO: --pause-container-accounte='${PAUSE_CONTAINER_ACCOUNT}'"
+      log "INFO: --pause-container-account='${PAUSE_CONTAINER_ACCOUNT}'"
```
guessi, line 109: Unrelate to this PR, but it's just a simple typo, not a big deal.

guessi: rebased

files/bootstrap.sh
```diff
@@ -35,6 +35,7 @@ function print_help {
   echo "--local-disks Setup instance storage NVMe disks in raid0 or mount the individual disks for use by pods [mount | raid0]"
   echo "--mount-bfs-fs Mount a bpffs at /sys/fs/bpf (default: true, for Kubernetes 1.27+; false otherwise)"
   echo "--pause-container-account The AWS account (number) to pull the pause container from"
+  echo "--pause-container-repository-name The AWS ECR repository name to pull the pause container from"
```
cartermckinnon, line 38: nit: this doesn't need to be an `AWS ECR repository`, it can be any image registry.

cartermckinnon:
I would prefer we add an option to just entirely override what sandbox image is used. Having options to override each component of an image reference is pretty tedious in my opinion.

Allowing an override on the image tag is reasonable, but allowing an override on the account number is strange IMO. I think we should deprecate that option and recommend folks use a new option, something like:
```
--pause-container-image
```

If this flag is defined, we should not allow `--pause-container-image-version` or `--pause-container-account` to be defined.

guessi:
Agree... having 3 flags, `account`, `repo`, `tag` is relatively hard to maintain, will update the PR later.

guessi:
@cartermckinnon I've updated the script, please review again if you have time.

I walk through the script once again and I think there are more follow-up discussions here.

If we are going to mark `PAUSE_CONTAINER_{REPOSITORY_NAME,IMAGE,VERSION}` as deprecated.

1. https://github.com/awslabs/amazon-eks-ami/blob/v20230607/files/containerd-config.toml#L11-L12
2. https://github.com/awslabs/amazon-eks-ami/blob/v20230607/scripts/install-worker.sh#L375-L386
3. https://github.com/awslabs/amazon-eks-ami/blob/v20230607/eks-worker-al2-variables.json#L23
4. https://github.com/awslabs/amazon-eks-ami/blob/v20230607/eks-worker-al2.json#L30
5. https://github.com/awslabs/amazon-eks-ami/blob/v20230607/eks-worker-al2.json#L176

I think we need to take care of these changes as well? but should these change be made (in this PR, or separated PR)?

WDYT?

cartermckinnon:
This is one of those things that we'll tackle in a rewrite of the bootstrap program; I'm not keen on adding/deprecating flags on the existing implementation at this point.

Issue #1266, opened 2023-04-14
EKS Latest kernel release 5.4.238-148.346 has no headers/devel packages


renilthomas:
**Environment**:
- AWS Region: us-east-1
- EKS Platform version (use `aws eks describe-cluster --name <name> --query cluster.platformVersion`): eks.6
- Kubernetes version (use `aws eks describe-cluster --name <name> --query cluster.version`): 1.23
- AMI Version: 1.23.17-20230406
- Kernel (e.g. `uname -a`):5.4.238-148.346.amzn2.x86_64
- Release information (run `cat /etc/eks/release` on a node):

```
BASE_AMI_ID="ami-099e00fe4091e48af"
BUILD_TIME="Thu Apr  6 01:36:29 UTC 2023"
BUILD_KERNEL="5.4.238-148.346.amzn2.x86_64"
ARCH="x86_64"
```

Why this kernel version does not seem to have a linux-headers or kernel-devel pacakge. Falco relies on this to build the driver.

HenryYanTR:
Datadog agent also crashes on this EKS AMI

ravilr:
looks like there is only `kernel-headers-5.4.238-148.347.amzn2.x86_64` and `kernel-devel-5.4.238-148.347.amzn2.x86_64` available in the rpm repo here. 

@AWS team/ @camrakin   any idea why the latest AMI release is installing `kernel-5.4.238-148.346.amzn2.x86_64` instead of `5.4.238-148.347.amzn2.x86_64`  ( .346 versus .347) ?  perhaps the versionlock of kernel is introducing this issue ?

`sudo yum versionlock delete kernel && sudo yum -y install kernel  && sudo reboot`   does align both the `kernel` and `kernel-headers` rpm to same `...347` version and unblocks the kernel module installations.

cartermckinnon:
Sorry for the late reply; `kernel-5.4.238-155.346` had a fatal issue affecting SMB mounts in which a null pointer dereference caused a panic. As a result, this package (and the associated `header`, `devel`, etc. packages) was removed from the Amazon Linux 2 repositories. The issue has been fixed in `kernel-5.4.238-155.347`.

PR #1302, opened 2023-05-15
Install kernel-headers, kernel-devel

cartermckinnon:
**Issue #, if available:**

#1266

**Description of changes:**

Software common in the k8s ecosystem requires the header + devel packages for the current kernel. This PR installs those packages at build-time, for two reasons:
1. Installing the packages at runtime adds startup latency to software that requires these packages.
2. These packages may not be available from the remote yum repository indefinitely (for example, in the case of a repository rollback).

I chose to install the packages instead of only caching the RPM's with `yum-downloadonly`, because the remote repository data would still need to be refreshed in most cases (it expires after 2 hours) to complete the installation, which wouldn't address reason 2 above.

**Testing Done**

All succeed, and the additional storage space consumed is reasonable:
```
# kernel-5.15
make 1.26 kernel_version=5.15
# Installed size: 271 M

# kernel-5.10
make 1.26
# Installed size: 266 M

# kernel-5.4
make 1.23
# Installed size: 224 M
```

### DIFF
--- a/scripts/install-worker.sh
+++ b/scripts/install-worker.sh
@@ -77,7 +77,7 @@ sudo yum install -y \
 # Remove any old kernel versions. `--count=1` here means "only leave 1 kernel version installed"
 sudo package-cleanup --oldkernels --count=1 -y
 
-sudo yum versionlock kernel-$(uname -r)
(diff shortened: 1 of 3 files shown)


mmerkes (approved):
Sounds reasonable. Let's consider if we need to do something to validate free space

PR #1320, opened 2023-06-06
Mount bpffs by default on 1.25+

cartermckinnon:
**Description of changes:**

Changes the default for `--mount-bpf-fs` to `true` for Kubernetes 1.25+.


ravisinha0506 (approved):

files/bootstrap.sh
```diff
@@ -322,7 +322,7 @@ if [[ "$MACHINE" != "x86_64" && "$MACHINE" != "aarch64" ]]; then
 fi
 
 if [ "$MOUNT_BPF_FS" = "true" ]; then
-  sudo mount-bpf-fs
+  mount-bpf-fs
```
cartermckinnon, line 325: `sudo` isn't necessary here because the bootstrap script runs as `root`. Removing it allows `mount-bpf-fs` to be mocked in the unit tests.

PR #1234, opened 2023-03-28
Install cri-tools

cartermckinnon:
**Issue #, if available:**

Closes #797 .

**Description of changes:**

Adds `cri-tools`, which includes `crictl` and `critest`. The latest version of the package will be installed at build, and the package is not version-locked, because it's not in the critical path.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

**Testing Done**

```
> make
...
2023-03-28T11:50:21-07:00:     amazon-ebs: Downloading packages:
2023-03-28T11:50:22-07:00:     amazon-ebs: cri-tools-1.25.0-1.amzn2.0.1.x86_64.rpm                    |  17 MB   00:00
2023-03-28T11:50:22-07:00:     amazon-ebs: Running transaction check
2023-03-28T11:50:22-07:00:     amazon-ebs: Running transaction test
2023-03-28T11:50:22-07:00:     amazon-ebs: Transaction test succeeded
2023-03-28T11:50:22-07:00:     amazon-ebs: Running transaction
2023-03-28T11:50:24-07:00:     amazon-ebs:   Installing : cri-tools-1.25.0-1.amzn2.0.1.x86_64                          1/1
2023-03-28T11:50:24-07:00:     amazon-ebs:   Verifying  : cri-tools-1.25.0-1.amzn2.0.1.x86_64                          1/1
2023-03-28T11:50:24-07:00:     amazon-ebs:
2023-03-28T11:50:24-07:00:     amazon-ebs: Installed:
2023-03-28T11:50:24-07:00:     amazon-ebs:   cri-tools.x86_64 0:1.25.0-1.amzn2.0.1
2023-03-28T11:50:24-07:00:     amazon-ebs:
2023-03-28T11:50:24-07:00:     amazon-ebs: Complete!
```

PR #1321, opened 2023-06-06
Install containerd/nerdctl

### LINKED ISSUE #797: Install crictl for CLI interaction with containerd

ddl-retornam:
**Issue #, if available:**

Attempts to close #797 in  https://github.com/awslabs/amazon-eks-ami/pull/1234 and https://github.com/awslabs/amazon-eks-ami/pull/1235 were stalled.  This PR is an attempt
to provide a working replacement for  the default `ctr` command bundled with `containerd` with the `nerdctl` command
which is officially supported by `containerd` 

**Description of changes:**

According to the default ctr cli man page bundled with containerd

"ctr is an unsupported debug and administrative client for interacting with the containerd daemon. Because it is unsupported, the commands, options, and operations are not guaranteed to be backward compatible or stable from release to release of the containerd project. "

The nerdctl cli program provided by containerd at
https://github.com/containerd/nerdctl

is meant to be a Docker-compatible CLI for containerd.

**Testing Done**

Installed the package on  running versions of the `amazon-eks-node-*` AMI


cartermckinnon:
Can you add some details about your use case? What do you do with `nerdctl` on your nodes?

ddl-retornam:
> Can you add some details about your use case? What do you do with `nerdctl` on your nodes?

There is sometimes the need to debug containers and images on the node directly.  The `ctr` command lacks some functionality like testing direct access to configured registries and the like.

bryantbiggs:
> There is sometimes the need to debug containers and images on the node directly. The ctr command lacks some functionality like testing direct access to configured registries and the like.

I think the key phrase there is `sometimes` - which means installing it at the time of debugging is preferred over having it installed by default on all AMIs/instances

ddl-retornam:
> I think the key phrase there is `sometimes` - which means installing it at the time of debugging is preferred over having it installed by default on all AMIs/instances


There have been prior discussions about this in https://github.com/awslabs/amazon-eks-ami/issues/797 in  and attempts were made to close it https://github.com/awslabs/amazon-eks-ami/pull/1234 and https://github.com/awslabs/amazon-eks-ami/pull/1235

I think it is prudent for every reviewer to read those prior discussions before commenting on this PR as it would lead to better conversations around this particular PR.

https://github.com/containerd/nerdctl is containterd projects suggested debugging tool instead of the bundled `ctr`


Thank you

cartermckinnon:
I tend to agree with @bryantbiggs that we should avoid installing stuff that isn't strictly necessary at runtime, but `nerdctl` *is* the blessed CLI for `containerd` these days. My only concern around installing it is the maintenance of the package in the AL2 repos -- let me follow up internally to figure out how/when that happens.

scripts/install-worker.sh
```diff
@@ -191,6 +191,13 @@ net.bridge.bridge-nf-call-iptables = 1
 net.ipv4.ip_forward = 1
 EOF
 
+###############################################################################
+### Nerdctl setup #############################################################
+###############################################################################
+
+
+sudo yum install -y nerdctl
```
cartermckinnon, line 199: I think we should install this. But I want to make sure that this package doesn't declare `containerd` as a dependency; because if we install it before `containerd`, this might not respect the `containerd_version` AMI template variable. I'll kick off the CI and see what the logs look like.

cartermckinnon: Logs look fine, this installs after `containerd`. 👍

cartermckinnon (approved):

PR #1325, opened 2023-06-13
[WIP] Add `hack` mode with in-place bootstrapping

cartermckinnon:
**Description of changes:**

This is an attempt at a new dev workflow for the AMI template, with the goal being faster end-to-end testing of changes.

Using the new `hack` template variable:
```
make 1.27 hack=true
```

This will:
1. Launch a cluster called `hack-127` with an empty self-managed nodegroup, using `eksctl`. If one already exists, reuse it.
2. Fetch details of the hacking cluster such as the node instance profile, subnets, cluster CA, and API server endpoint.
3. Enable a new provisioner at the end of the AMI build process that will join the Packer builder instance to the hacking cluster.
4. When you're done poking around, CTRL+C will continue the normal Packer cleanup process. No AMI needs to be created and you don't need to launch a separate instance/nodegroup using an AMI.

**Testing Done**

I've tried this on 1.25-1.27.

**TODO**

- SSH access is still wonky.


cartermckinnon:
The hackery is too strong with this one, not moving forward.

PR #1328, opened 2023-06-15
Use recommended clocksources

### LINKED ISSUE #249: Make TSC the default system clock source

### LINKED ISSUE #1324: Clock drift issues
Our software has issues due to clock drifts between EC2 instances using this AMI.  We see it more in region: `ap-northeast-2`.

Relevant issues: 
* #249 
* #321

cartermckinnon:
**Issue #, if available:**

Fixes #249 

**Description of changes:**

The `tsc` clocksource is the recommended default for instances on the Xen hypervisor, as it's significantly more performant than the `xen` clocksource (there's an interesting [case study](https://www.brendangregg.com/blog/2021-09-26/the-speed-of-time.html) from Netflix).

A previous attempt (in #272) to swap `xen` for `tsc` had no effect; only the current clocksource was changed, not the default. The setting was applied during the AMI build process and had no effect after a reboot.

EC2 [recommends](https://repost.aws/knowledge-center/manage-ec2-linux-clock-source):
- Xen-based instance types: `tsc`
  - In my tests, `xen` was used by default.
- Nitro-based instance types: `kvm-clock`
  - In my tests, `tsc` was used by default.
- Graviton instance types: `arch_sys_counter`
  - In my tests, this is the only clocksource available; any other `clocksource` set on the kernel command-line is ignored.

**Testing Done**

1. I built a 1.27 AMI on this branch (`make 1.27`).
2. I launched:
  - `m6i.large` (Nitro): `kvm-clock` used as expected.
  - `r7g.medium` (Graviton3): `arch_sys_counter` used as expected.
    - I built an `arm64` variant for this (`make 1.27 arch=arm64`).
  - `m4.large` (Xen): `tsc` used as expected.

### DIFF
--- a/scripts/install-worker.sh
+++ b/scripts/install-worker.sh
@@ -81,27 +81,13 @@ sudo yum versionlock kernel-$(uname -r)
 # Remove the ec2-net-utils package, if it's installed. This package interferes with the route setup on the instance.
 if yum list installed | grep ec2-net-utils; then sudo yum remove ec2-net-utils -y -q; fi
 
+sudo mkdir -p /etc/eks/
+
 ################################################################################
 ### Time #######################################################################
 ################################################################################
 
-# Make sure Amazon Time Sync Service starts on boot.
-sudo chkconfig chronyd on
-
-# Make sure that chronyd syncs RTC clock to the kernel.
-cat << EOF | sudo tee -a /etc/chrony.conf
-# This directive enables kernel synchronisation (every 11 minutes) of the
-# real-time clock. Note that it can’t be used along with the 'rtcfile' directive.
-rtcsync
-EOF
-
-# If current clocksource is xen, switch to tsc
-if grep --quiet xen /sys/devices/system/clocksource/clocksource0/current_clocksource \
-  && grep --quiet tsc /sys/devices/system/clocksource/clocksource0/available_clocksource; then
-  echo "tsc" | sudo tee /sys/devices/system/clocksource/clocksource0/current_clocksource
-else
-  echo "tsc as a clock source is not applicable, skipping."
-fi
+sudo mv $WORKING_DIR/configure-clocksource.service /etc/eks/configure-clocksource.service
 
 ################################################################################
 ### SSH ########################################################################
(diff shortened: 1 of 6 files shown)


scripts/install-worker.sh
```diff
@@ -82,28 +82,6 @@ sudo yum versionlock kernel-$(uname -r)
...
-
-# Make sure Amazon Time Sync Service starts on boot.
-sudo chkconfig chronyd on
-
-# Make sure that chronyd syncs RTC clock to the kernel.
-cat << EOF | sudo tee -a /etc/chrony.conf
-# This directive enables kernel synchronisation (every 11 minutes) of the
-# real-time clock. Note that it can’t be used along with the 'rtcfile' directive.
-rtcsync
-EOF
```
cartermckinnon, line 97: I verified that this is already present in `/etc/chrony.conf` in the latest AL2 minimal AMI. We should not configure `chrony` here to avoid diverging from the defaults provided by AL2.

scripts/install-worker.sh
```diff
@@ -82,28 +82,6 @@ sudo yum versionlock kernel-$(uname -r)
...
-
-# If current clocksource is xen, switch to tsc
-if grep --quiet xen /sys/devices/system/clocksource/clocksource0/current_clocksource \
-  && grep --quiet tsc /sys/devices/system/clocksource/clocksource0/available_clocksource; then
-  echo "tsc" | sudo tee /sys/devices/system/clocksource/clocksource0/current_clocksource
-else
-  echo "tsc as a clock source is not applicable, skipping."
-fi
```
cartermckinnon, line 105: This only sets the current clocksource, which has no effect after a reboot.

scripts/upgrade_kernel.sh
```diff
@@ -24,4 +24,10 @@ sudo grubby \
   --update-kernel=ALL \
   --args="psi=1"
 
+# use the tsc clocksource by default
+# https://repost.aws/knowledge-center/manage-ec2-linux-clock-source
+sudo grubby \
+  --update-kernel=ALL \
+  --args="clocksource=tsc tsc=reliable"
+
```
cartermckinnon, line 32: This isn't strictly necessary with the `configure-clocksource` unit in place, but the `tsc=reliable` flag is still worth setting, and we might as well configure a default here too.

bwagner5 (approved):
LGTM

PR #1212, opened 2023-03-04
[WIP] Amazon Linux 2023 proof-of-concept

cartermckinnon:
**Description of changes:**

This is an initial PoC for an AL2023-based EKS worker AMI.

**Changelog:**
- `rsa` SSH keys not supported by default on AL2023, switched to `ed25519`.
- Removed Docker from template variables.
- Removed `upgrade_kernel.sh` provisioner -- AL2023 is on kernel 6.1 by default.
- Removed `curl` in favor of `curl-minimal`.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1340, opened 2023-07-06
Amazon Linux 2023 proof-of-concept (Take 2)

dims:
Makefile has a converter that generates al2023 configs from existing al2 ones

Very minimal changes to other files to ensure al2023 works and does not get in the way of AL2 ongoing development.

Based on earlier work by @cartermckinnon in: 
https://github.com/awslabs/amazon-eks-ami/pull/1212

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,13 +26,18 @@ ifeq ($(call vercmp,$(kubernetes_version),gteq,1.25.0), true)
 	ami_component_description ?= (k8s: {{ user `kubernetes_version` }}, containerd: {{ user `containerd_version` }})
 endif
 
+OS=
+ifneq (,$(findstring al2023, $(PACKER_TEMPLATE_FILE)))
+	OS=-al2023
+endif
+
 arch ?= x86_64
 ifeq ($(arch), arm64)
 	instance_type ?= m6g.large
-	ami_name ?= amazon-eks-arm64-node-$(K8S_VERSION_MINOR)-v$(shell date +'%Y%m%d')
+	ami_name ?= amazon-eks-arm64-node$(OS)-$(K8S_VERSION_MINOR)-v$(shell date +'%Y%m%d')
 else
 	instance_type ?= m5.large
-	ami_name ?= amazon-eks-node-$(K8S_VERSION_MINOR)-v$(shell date +'%Y%m%d')
+	ami_name ?= amazon-eks-node$(OS)-$(K8S_VERSION_MINOR)-v$(shell date +'%Y%m%d')
 endif
 
 ifeq ($(aws_region), cn-northwest-1)
--- a/Makefile
+++ b/Makefile
@@ -74,6 +74,12 @@ ifeq (, $(SHELLCHECK_COMMAND))
 endif
 SHELL_FILES := $(shell find $(MAKEFILE_DIR) -type f -name '*.sh')
 
+.PHONY: transform-al2-to-al2023
+transform-al2-to-al2023:
+	PACKER_TEMPLATE_FILE=$(PACKER_TEMPLATE_FILE) \
+	PACKER_DEFAULT_VARIABLE_FILE=$(PACKER_DEFAULT_VARIABLE_FILE) \
+		hack/transform-al2-to-al2023.sh
+
 .PHONY: lint
 lint: ## Check the source files for syntax and format issues
 	$(SHFMT_COMMAND) $(SHFMT_FLAGS) --diff $(MAKEFILE_DIR)
--- a/hack/transform-al2-to-al2023.sh
+++ b/hack/transform-al2-to-al2023.sh
@@ -1,0 +1,32 @@
+#!/usr/bin/env bash
+
+set -o pipefail
+set -o nounset
+set -o errexit
+
+if [[ -z "${PACKER_TEMPLATE_FILE:-}" ]]; then
+  echo "PACKER_TEMPLATE_FILE must be set." >&2
+  exit 1
+fi
+if [[ -z "${PACKER_DEFAULT_VARIABLE_FILE:-}" ]]; then
+  echo "PACKER_DEFAULT_VARIABLE_FILE must be set." >&2
+  exit 1
+fi
+
+# rsa keys are not supported in al2023, switch to ed25519
+# delete the upgrade kernel provisioner as we don't need it for al2023
+cat "${PACKER_TEMPLATE_FILE}" \
+  | jq '._comment = "All template variables are enumerated here; and most variables have a default value defined in eks-worker-al2023-variables.json"' \
+  | jq '.variables.temporary_key_pair_type = "ed25519"' \
+  | jq 'del(.provisioners[5])' \
+    > "${PACKER_TEMPLATE_FILE/al2/al2023}"
+
+# use newer versions of containerd and runc, do not install docker
+# use al2023 6.1 minimal image
+cat "${PACKER_DEFAULT_VARIABLE_FILE}" \
+  | jq '.ami_component_description = "(k8s: {{ user `kubernetes_version` }}, containerd: {{ user `containerd_version` }})"' \
+  | jq '.ami_description = "EKS-optimized Kubernetes node based on Amazon Linux 2023"' \
+  | jq '.containerd_version = "*" | .runc_version = "*" | .docker_version = "" ' \
+  | jq '.source_ami_filter_name = "al2023-ami-minimal-2023.*-kernel-6.1-x86_64"' \
+  | jq '.volume_type = "gp3"' \
+    > "${PACKER_DEFAULT_VARIABLE_FILE/al2/al2023}"
--- a/hack/transform-al2-to-al2023.sh
+++ b/hack/transform-al2-to-al2023.sh
@@ -18,7 +18,7 @@ fi
 cat "${PACKER_TEMPLATE_FILE}" \
   | jq '._comment = "All template variables are enumerated here; and most variables have a default value defined in eks-worker-al2023-variables.json"' \
   | jq '.variables.temporary_key_pair_type = "ed25519"' \
-  | jq 'del(.provisioners[5])' \
+  | jq '.provisioners |= map(select(.script//empty|endswith("upgrade_kernel.sh")|not))' \
     > "${PACKER_TEMPLATE_FILE/al2/al2023}"
 
 # use newer versions of containerd and runc, do not install docker
--- a/scripts/install-worker.sh
+++ b/scripts/install-worker.sh
@@ -59,7 +59,6 @@ sudo yum install -y \
   aws-cfn-bootstrap \
   chrony \
   conntrack \
-  curl \
   ec2-instance-connect \
   ethtool \
   ipvsadm \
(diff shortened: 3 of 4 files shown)


scripts/upgrade_kernel.sh
```diff
@@ -4,6 +4,11 @@ set -o pipefail
 set -o nounset
 set -o errexit
 
+# Bailout if the kernel is already 6 or above (al2023)
+if [[ $(uname -r | cut -d'.' -f1) -ge 6 ]]; then
+  exit 0
+fi
+
```
cartermckinnon, line 11: Instead of changing this script, can we just delete the template provisioner that uses it, during `transform-al2-to-al2023-config`?

Makefile
```diff
@@ -74,6 +74,23 @@ ifeq (, $(SHELLCHECK_COMMAND))
...
+	# TODO: switch to minimal "al2023-ami-minimal-2023.*-kernel-6.1-x86_64"
+	cat $(PACKER_DEFAULT_VARIABLE_FILE) | \
+	jq '.ami_component_description = "(k8s: {{ user `kubernetes_version` }}, containerd: {{ user `containerd_version` }})"' | \
+	jq '.ami_description = "EKS-optimized Kubernetes node based on Amazon Linux 2023"' | \
+	jq '.containerd_version = "*" | .runc_version = "*" | .docker_version = "" ' | \
+	jq '.source_ami_filter_name = "al2023-ami-2023.*-kernel-6.1-x86_64"' | \
+	jq '.launch_block_device_mappings_volume_size = "8"' | \
+	jq '.volume_type = "gp3"' \
+	> $(subst al2,al2023,$(PACKER_DEFAULT_VARIABLE_FILE))
+
```
cartermckinnon, line 93: Can we throw this logic into a script in `hack/` to keep things a bit cleaner? (and the bash can look nicer there)

Makefile
```diff
@@ -74,6 +74,23 @@ ifeq (, $(SHELLCHECK_COMMAND))
 endif
 SHELL_FILES := $(shell find $(MAKEFILE_DIR) -type f -name '*.sh')
 
+.PHONY: lint
```
cartermckinnon, line 77:
```suggestion
.PHONY: transform-al2-to-al2023-config
```

scripts/install-worker.sh
```diff
@@ -59,7 +59,6 @@ sudo yum install -y \
   aws-cfn-bootstrap \
   chrony \
   conntrack \
-  curl \
```
cartermckinnon, line 62: I think `curl-minimal` is what we want on AL2023 -- that isn't on AL2 I assume?

scripts/validate.sh
```diff
@@ -48,10 +48,16 @@ fi
...
 
 function versionlock-entries() {
-  # the format of this output is EPOCH:NAME-VERSION-RELEASE.ARCH
-  # more info in yum-versionlock(1)
-  # rpm doesn't accept EPOCH when querying the db, so remove it
-  yum versionlock list --quiet | cut -d ':' -f2
+  if [ -f "/etc/dnf/plugins/versionlock.conf" ]; then
+    # the format of this output in al2023 is NAME-EPOCH:VERSION-RELEASE.ARCH
+    # rpm works fine with this format
+    yum versionlock list --quiet
```
cartermckinnon, line 54: this reminds me -- I don't think we need to do this versionlocking at all on AL2023. The repo itself is kind of "version locked" in that distro, and the recommendation I've heard from AL folks was to get rid of this. It's mostly been more annoying than helpful, too.

cartermckinnon (changes requested):
Love it! I'll close #1212 in favor of this PR.

scripts/install-worker.sh
stewartsmith: Correct. `curl-minimal` is only on AL2023.

scripts/validate.sh
stewartsmith: Correct. By default, AL2023 images are locked to a specific snapshot of the repo.

The only reason to use a dnf plugin (note that `yum` is not a thing on AL2023, only a symlink exists for compatibility) for locking versions is when you want to move everything *else* forward but to lock some other things... which should all be a VERY exceptional scenario.

Makefile
```diff
@@ -74,6 +74,23 @@ ifeq (, $(SHELLCHECK_COMMAND))
...
+transform-al2-to-al2023-config:
+	cat $(PACKER_TEMPLATE_FILE) | \
+	jq '._comment = "All template variables are enumerated here; and most variables have a default value defined in eks-worker-al2023-variables.json"' | \
+	jq '.variables.temporary_key_pair_type = "ed25519"' \
+	> $(subst al2,al2023,$(PACKER_TEMPLATE_FILE))
+
+	# TODO: switch to minimal "al2023-ami-minimal-2023.*-kernel-6.1-x86_64"
```
stewartsmith, line 84: Why not start out with minimal? It likely will work out-of-the-box for this use case.

scripts/install-worker.sh
```diff
@@ -73,8 +72,20 @@ sudo yum install -y \
   mdadm \
   pigz
 
-# Remove any old kernel versions. `--count=1` here means "only leave 1 kernel version installed"
-sudo package-cleanup --oldkernels --count=1 -y
+# packages that need special handling
+SPECIAL_PACKAGES=(
+  "curl"            # exists in both AL2 and AL2023 (need to overwrite curl-minimal)
```
stewartsmith, line 77: hopefully you don't actually *need* some of the functionality not in `curl-minimal`. If you do though, we want to know :)

scripts/install-worker.sh
```diff
@@ -73,8 +72,20 @@ sudo yum install -y \
...
+  "iptables-legacy" # exists in AL2023 only (needed by kubelet)
+)
+
+for package in "${SPECIAL_PACKAGES[@]}"; do
+  (yum list available | grep -qw $package) || sudo yum -y --allowerasing install $package
```
stewartsmith, line 82: s/`yum`/`dnf`/

although there is a symlink so it still works...

scripts/install-worker.sh
```diff
@@ -73,8 +72,20 @@ sudo yum install -y \
...
+done
+
+if [[ $(which package-cleanup) ]]; then
+  # Remove any old kernel versions. `--count=1` here means "only leave 1 kernel version installed"
+  sudo package-cleanup --oldkernels --count=1 -y || true
```
stewartsmith, line 87: this probably isn't needed?

Makefile
```diff
@@ -74,6 +74,23 @@ ifeq (, $(SHELLCHECK_COMMAND))
...
+	cat $(PACKER_DEFAULT_VARIABLE_FILE) | \
+	jq '.ami_component_description = "(k8s: {{ user `kubernetes_version` }}, containerd: {{ user `containerd_version` }})"' | \
+	jq '.ami_description = "EKS-optimized Kubernetes node based on Amazon Linux 2023"' | \
+	jq '.containerd_version = "*" | .runc_version = "*" | .docker_version = "" ' | \
+	jq '.source_ami_filter_name = "al2023-ami-2023.*-kernel-6.1-x86_64"' | \
+	jq '.launch_block_device_mappings_volume_size = "8"' | \
```
stewartsmith, line 90: Does `packer` create a fresh file system when you do this? Or is this really just setting the launch volume size?

stewartsmith: I ask as I've seen customer EKS worker nodes with kind of horrific file system usage patterns that would have been somewhat alleviated by having all the data on a fresh (somewhat optimized) `mkfs` of a large (e.g. 1TB) EBS volume rather than growing from the 2GB or 8GB file system that we started with.

Perhaps there's an opportunity to move to having the general advice being to have a separate EBS volume for all the k8s things?

bryantbiggs: if we change the volume size, we'll want to pick one that works for cached AMIs

> Perhaps there's an opportunity to move to having the general advice being to have a separate EBS volume for all the k8s things?

Yes - containerd and kubelet should have their own volume

Makefile
dims: figured out why minimal was not working with `instance connect`, now that i know we need `sudo yum install ec2-instance-connect` will switch to minimal image

Makefile
dims: ignore this `launch_block_device_mappings_volume_size`, am rolling this back

Makefile
dims: Done https://github.com/kubernetes-sigs/provider-aws-test-infra/pull/73

Makefile
dims: Done

scripts/upgrade_kernel.sh
dims: done

scripts/validate.sh
dims: okeydokey, adding an if condition to avoid running the versionlocking code

scripts/install-worker.sh
dims: will add a better check to run on al2 only

scripts/install-worker.sh
dims: ack, will switch back to just curl-minimal

scripts/install-worker.sh
dims: ack

scripts/install-worker.sh
dims: ack

dims:
@cartermckinnon if this PR is good enough, let's merge and iterate, that way i can throw a few CI jobs at the generated image to see how it fares.

Makefile
dims: @bryantbiggs not in scope for this PR :) we can open another issue and track if you all think that's a good idea.

hack/transform-al2-to-al2023.sh
```diff
@@ -0,0 +1,32 @@
...
+  echo "PACKER_DEFAULT_VARIABLE_FILE must be set." >&2
+  exit 1
+fi
+
+# rsa keys are not supported in al2023, switch to ed25519
+# delete the upgrade kernel provisioner as we don't need it for al2023
+cat "${PACKER_TEMPLATE_FILE}" \
+  | jq '._comment = "All template variables are enumerated here; and most variables have a default value defined in eks-worker-al2023-variables.json"' \
+  | jq '.variables.temporary_key_pair_type = "ed25519"' \
+  | jq 'del(.provisioners[5])' \
```
cartermckinnon, line 21: We change the provisioners array often enough to not hardcode this index:
```suggestion
  | jq '.provisioners |= map(select(.script//empty|endswith("upgrade_kernel.sh")|not))' \
```

dims:
@cartermckinnon thanks! applied the suggestion

cartermckinnon:
I think this is fine to get us going. Some things I'd like to see in followup PR's:
- A user guide entry on distro support (AL2023 being called out as experimental/in-progress).
- A make var to select the distro, usage like `make 1.27 distro=al2023`
- Static variable file for al2023 instead of modifying the one for al2 (once the wrinkles are ironed out).

I'm not sure what the right approach will be for the template file. We need to eventually move to the HCL template format, and AL2023 is an opportunity to do that.

cartermckinnon (approved):

Issue #1331, opened 2023-06-22
ctr failed to dial "/run/containerd/containerd.sock"


shaileshlaghane:
**What happened**:
Creating an AMI with centos image using packer, During image creation getting `ctr: failed to dial "/run/containerd/containerd.sock": context deadline exceeded: connection error: desc = "transport: error while dialing: dial unix:///run/containerd/containerd.sock: timeout"` in the logs, however image is getting created. Attached this AMI to worker node, kubelet, containerd both are up.
In `ebs-csi-node` pod logs I can see below 

```
I0620 05:48:36.367644       1 main.go:166] Version: v2.5.1
I0620 05:48:36.367687       1 main.go:167] Running node-driver-registrar in mode=registration
I0620 05:48:36.368209       1 main.go:191] Attempting to open a gRPC connection with: "/csi/csi.sock"
W0620 05:48:46.369172       1 connection.go:173] Still connecting to unix:///csi/csi.sock
W0620 05:48:56.368642       1 connection.go:173] Still connecting to unix:///csi/csi.sock
W0620 05:49:06.369161       1 connection.go:173] Still connecting to unix:///csi/csi.sock
W0620 05:49:16.369140       1 connection.go:173] Still connecting to unix:///csi/csi.sock
W0620 05:49:26.368868       1 connection.go:173] Still connecting to unix:///csi/csi.sock
W0620 05:49:36.368487       1 connection.go:173] Still connecting to unix:///csi/csi.sock
W0620 05:49:46.369019       1 connection.go:173] Still connecting to unix:///csi/csi.sock
W0620 05:49:56.369075       1 connection.go:173] Still connecting to unix:///csi/csi.sock
W0620 05:50:06.369135       1 connection.go:173] Still connecting to unix:///csi/csi.sock
W0620 05:50:16.368857       1 connection.go:173] Still connecting to unix:///csi/csi.sock
W0620 05:50:26.369265       1 connection.go:173] Still connecting to unix:///csi/csi.sock
W0620 05:50:36.368846       1 connection.go:173] Still connecting to unix:///csi/csi.sock
```
**What you expected to happen**:
All should pods should be up and running fine
**How to reproduce it (as minimally and precisely as possible)**:
Use centos 7 images instead of amazon linux
**Anything else we need to know?**:

**Environment**:
- Kubernetes version: `1.24`
- Kernel (e.g. `uname -a`): `3.10.0-1160.80.1.el7.x86_64`
- Centos 7
[...]

PR #1341, opened 2023-07-07
Generate version info for cached images only when `containerd` is active

cartermckinnon:
**Issue #, if available:**

#1331 

**Description of changes:**

The warning printed by `ctr` when `containerd` isn't active (as a result of the `cache_container_images` feature being disabled) has been confusing. This adds a condition so that image info is only gathered when `containerd` is running. If `containerd` *isn't* running, and the `cache_container_images` feature is enabled, the script exits with an error.

**Testing Done**

Succeeds:
```
make
```

### DIFF
--- a/scripts/generate-version-info.sh
+++ b/scripts/generate-version-info.sh
@@ -20,4 +20,9 @@ echo $(jq ".binaries.kubelet = \"$(kubelet --version | awk '{print $2}')\"" $OUT
 echo $(jq ".binaries.awscli = \"$(aws --version | awk '{print $1}' | cut -d '/' -f 2)\"" $OUTPUT_FILE) > $OUTPUT_FILE
 
 # cached images
-echo $(jq ".images = [ $(sudo ctr -n k8s.io image ls -q | cut -d'/' -f2- | sort | uniq | grep -v 'sha256' | xargs -r printf "\"%s\"," | sed 's/,$//') ]" $OUTPUT_FILE) > $OUTPUT_FILE
+if systemctl is-active --quiet containerd; then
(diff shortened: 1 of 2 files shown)


bwagner5 (approved):
lgtm

scripts/generate-version-info.sh
```diff
@@ -20,4 +20,9 @@ echo $(jq ".binaries.kubelet = \"$(kubelet --version | awk '{print $2}')\"" $OUT
 echo $(jq ".binaries.awscli = \"$(aws --version | awk '{print $1}' | cut -d '/' -f 2)\"" $OUTPUT_FILE) > $OUTPUT_FILE
 
 # cached images
-echo $(jq ".images = [ $(sudo ctr -n k8s.io image ls -q | cut -d'/' -f2- | sort | uniq | grep -v 'sha256' | xargs -r printf "\"%s\"," | sed 's/,$//') ]" $OUTPUT_FILE) > $OUTPUT_FILE
+if systemctl is-active --quiet containerd; then
+  echo $(jq ".images = [ $(sudo ctr -n k8s.io image ls -q | cut -d'/' -f2- | sort | uniq | grep -v 'sha256' | xargs -r printf "\"%s\"," | sed 's/,$//') ]" $OUTPUT_FILE) > $OUTPUT_FILE
+elif [ "${CACHE_CONTAINER_IMAGES}" = "true" ]; then
+  echo "containerd must be active to generate version info for cached images"
```
ravisinha0506, line 26: How will this behave in older k8s versions with docker cri?

cartermckinnon: There's no behavior difference. If `cache_container_images` is enabled, then we'll include the image details here. We don't enable this feature on versions that use Docker, though.

PR #1344, opened 2023-07-10
Add us-iso-west-1 to credential-provider

prasad0896:
**Issue #, if available:**

**Description of changes:**
Adding endpoint for `us-iso-west-1` to the credential-provider-config file so that we are able to pull images from ECR on nodes that don't have the in-tree cloud provider like 1.27

**Testing Done**

### DIFF
--- a/files/ecr-credential-provider-config.json
+++ b/files/ecr-credential-provider-config.json
@@ -8,9 +8,8 @@
         "*.dkr.ecr.*.amazonaws.com",
         "*.dkr.ecr.*.amazonaws.com.cn",
         "*.dkr.ecr-fips.*.amazonaws.com",
-        "*.dkr.ecr.us-iso-east-1.c2s.ic.gov",
-        "*.dkr.ecr.us-isob-east-1.sc2s.sgov.gov",
-        "*.dkr.ecr.us-iso-west-1.c2s.ic.gov"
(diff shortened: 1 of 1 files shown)


Issacwww (approved):

files/ecr-credential-provider-config.json
```diff
@@ -9,7 +9,8 @@
         "*.dkr.ecr.*.amazonaws.com.cn",
         "*.dkr.ecr-fips.*.amazonaws.com",
         "*.dkr.ecr.us-iso-east-1.c2s.ic.gov",
-        "*.dkr.ecr.us-isob-east-1.sc2s.sgov.gov"
+        "*.dkr.ecr.us-isob-east-1.sc2s.sgov.gov",
+        "*.dkr.ecr.us-iso-west-1.c2s.ic.gov"
```
cartermckinnon, line 13: can't we do something like line 10? so we don't have to update this for every region in these partitions?
```suggestion
        "*.dkr.ecr.*.c2s.ic.gov"
```

cartermckinnon (changes requested):

files/ecr-credential-provider-config.json
prasad0896: Yes, fair callout. I have updated both the partition endpoints.

PR #1346, opened 2023-07-11
Remove hard-coded pull_cni_from_github

cartermckinnon:
**Description of changes:**

The default value for this arg is already `true`. Hard-coding it on the `make` targets means you cannot set a different value when the target is used, which is unnecessary (and was annoying while preparing for #500).


mmerkes (approved):
LGTM

PR #1360, opened 2023-07-20
Enable discard_unpacked_layers by default

### LINKED ISSUE #1359: Should the containerd-config.toml file have discard_unpacked_layers set to true by default
**Environment**:
- AWS Region: All
- Instance Type(s): All
- EKS Platform version (use `aws eks describe-cluster --name <name> --query cluster.platformVersion`): Any version using containerd
- Kubernetes version (use `aws eks describe-cluster --name <name> --query cluster.version`): Any version using containerd
- AMI Version: Any version using containerd
- Kernel (e.g. `uname -a`): NA

```
BASE_AMI_ID="ami-044dcf122cb58da69"
BUILD_TIME="Mon Jul  3 03:32:16 UTC 2023"
BUILD_KERNEL="5.4.247-162.350.amzn2.x86_64"
ARCH="x86_64"
```

**Issue**:

The issue here is that the configuration used for containerd can result in significantly increased disk usage because containerd will store both the packed and unpacked layers which in the case where the image does not compress well as a tarball can result in increased utilization of the root volume.  To see this in action you can used these commands [1] to make a dockerfile as the random data will not compress well.

The workaround is to create your own containerd config file then add `discard_unpacked_layers = true` to the `plugins."io.containerd.grpc.v1.cri".containerd section`.  Then pass this to the bootstrap command using the --containerd-config-file after putting the file on the instance, usually by downloading from S3 although a custom AMI could also be used.  However, this seems like a lot of work just to bring containerd in line with the space usage of docker.  Which bring us to the question, should the [containerd-config.toml](https://github.com/awslabs/amazon-eks-ami/blob/master/files/containerd-config.toml) file have discard_unpacked_layers set to true by default?

[1] 
```
mkdir /var/lib/docker/cstest
cd /var/lib/docker/cstest
dd if=/dev/urandom of=testfile.img bs=$((1024*1024)) count=1007
dd if=/dev/urandom of=testfile2.img bs=$((1024*1024)) count=2015
dd if=/dev/urandom of=testfile3.img bs=$((1024*1024)) count=2015
cat > Dockerfile <<EOF
FROM ubuntu
COPY ./testfile.img /testfile.img
COPY ./testfile2.img /testfile2.img
COPY ./testfile3.img /testfile3.img
CMD ["/bin/bash","cat testfile.img"]
EOF
```

cartermckinnon:
**Issue #, if available:**

Fixes #1359.

**Description of changes:**

Sets `discard_unpacked_layers = true` in the `containerd` config file. This is a disk space optimization.

**Testing**

Verified the config key:
```
> containerd --version
containerd github.com/containerd/containerd 1.6.19 1e1ea6e986c6c86565bc33d52e34b81b3e2bc71f

> containerd config default | grep -A3 -B3 discard_unpacked_layers
      default_runtime_name = "runc"
      disable_snapshot_annotations = true
      discard_unpacked_layers = false
      ignore_rdt_not_enabled_errors = false
      no_pivot = false
      snapshotter = "overlayfs"
```

### DIFF
--- a/files/containerd-config.toml
+++ b/files/containerd-config.toml
@@ -7,10 +7,10 @@ address = "/run/containerd/containerd.sock"
 
 [plugins."io.containerd.grpc.v1.cri".containerd]
 default_runtime_name = "runc"
+discard_unpacked_layers = true
 
 [plugins."io.containerd.grpc.v1.cri"]
 sandbox_image = "SANDBOX_IMAGE"
-discard_unpacked_layers = true
 
 [plugins."io.containerd.grpc.v1.cri".registry]
 config_path = "/etc/containerd/certs.d:/etc/docker/certs.d"
--- a/files/containerd-config.toml
+++ b/files/containerd-config.toml
@@ -10,6 +10,7 @@ default_runtime_name = "runc"
 
 [plugins."io.containerd.grpc.v1.cri"]
 sandbox_image = "SANDBOX_IMAGE"
+discard_unpacked_layers = true
 
 [plugins."io.containerd.grpc.v1.cri".registry]
 config_path = "/etc/containerd/certs.d:/etc/docker/certs.d"


bwagner5 (approved):
nice! LGTM

dineshp-09:
@cartermckinnon After this change we are having issue with image export in containerd:
https://github.com/containerd/containerd/issues/9277
https://github.com/containerd/nerdctl/issues/2588
when we disable this flag manually, image export is successful. 
Would you have any thoughts on this?

cartermckinnon:
Why do you need to export images from the node’s container runtime?

dineshp-09:
@cartermckinnon we need to create an image tar and do some processing in that tar.
basically, the image export functionality should work.

dineshp-09:
@cartermckinnon got your point. Thanks for the crane tool suggestion.

hellt:
👋 @cartermckinnon 
I wonder if you know a workaround for a challenge we get with this setting being the new default.

In our case we need to export an image from the k8s worker node to the pod. The reason we need to it via the node is that we want k8s users to leverage Pull Secrets and get the image on the nodes using their credentials that our pod has no idea about.

Once the image is pulled, we wanted to export it from the node's image store to the POD by bind mounting the containerd.sock, but it seems that it is not possible anymore, because of the missing layer.

Would appreciate if anyone found a workaround for this.

cartermckinnon:
In general, I would recommend a tool like [crane](https://github.com/google/go-containerregistry/blob/main/cmd/crane/doc/crane.md) for working with container images inside a pod. Mounting the node's container runtime socket into a pod is risky because 1) it gives the pod the ability to control any other pod on the node and 2) lets the pod circumvent its resource limits by consuming CPU/memory/disk that it has not been allocated (which can interfere with kubelet).

That being said, you know your workload better than I do. 😄 We're working on #1630, which will allow you to disable this setting in the `containerd` config.

hellt:
Thanks for your comments, @cartermckinnon 

We looked at crane and skopeo, but we couldn't still export the image that was pulled by a kubelet and extensive googling failed me as well.

We are not concerned about the security limitations or resource shunting as, indeed, the clusters we created are ephemeral, not multi-tenanted and specifically are made for CI.

The only reason we need the export capability is to export the image pulled by the k8s and import it in the docker-in-docker pod we are running. This works great when `discard_unpacked_layers` is not set.

Would be amazing to have this setting exposed tot he EKS users, but also I would be super super curious to know why a simple layer packing can't be done when layers are missing, since the image is run-worthy, why can't one pack it up?

zaidmnsr:
You can disable it via setting it to false using below strategy.

I updated the config.toml using below command as eks bootstrap script is using this file as the base file.

`sed -i 's/discard_unpacked_layers = true/discard_unpacked_layers = false/g'  /etc/eks/containerd/containerd-config.toml`

Once you update this, you can verify this using below command from your worker node.

`containerd config dump`

hellt:
Hi @zaidmnsr 
would you mind sharing more details about this approach? At what step you change the config.toml?

zaidmnsr:
Hi @hellt 

You can use this command in userdata script. You can modify userdata script in launch template by creating a new version of launch template.

PR #1363, opened 2023-07-21
Cleanup /var/log/audit

cartermckinnon:
**Description of changes:**

Some traces of the build process/initial instance launch are left behind in various log files (such as SSH logins by Packer). This whacks everything under `/var/log`.

**Testing Done**

`make` succeeds.


prasad0896:
Would we ever have a need for these files for debugging if AMI build fails?

cartermckinnon:
Nope. If there's a failure in the build, we won't make it to this step in the first place.

prasad0896 (approved):

PR #1370, opened 2023-08-01
Upgrade amazon-ssm-agent version to 3.2.1377.0

### LINKED ISSUE #1369: SSM agent v3.2.x support
Hello,

It seems that the amazon-ssm-agent version is frozen at 3.1.1732.0 for the latest releases.
Do you think it would be possible to update the installed version of amazon-ssm-agent to 3.2.x?

**Why is this needed**:
Installing a version of amazon-ssm-agent higher than 3.2.582.0 would allow us to manage eks instances automatically by configuring [Default Host Management Configuration](https://docs.aws.amazon.com/systems-manager/latest/userguide/managed-instances-default-host-management.html).

vincentmrg:
**Issue #, if available:**
closes #1369

**Description of changes:**

* Use amazon-ssm-agent rpm from s3

* Set ssm-agent version configurable

* Upgrade ssm-agent to latest version (3.2.1377.0)

**Testing Done**

I have successfully launched the tests, built and published an AMI with the `make` command, then launched an instance from this ami and verified that the `amazon-ssm-agent` was correctly installed and running.

### DIFF
--- a/eks-worker-al2-variables.json
+++ b/eks-worker-al2-variables.json
@@ -30,6 +30,7 @@
     "source_ami_owners": "137112412989",
     "ssh_interface": "",
     "ssh_username": "ec2-user",
+    "ssm_agent_version": "3.2.1377.0",
     "subnet_id": "",
     "temporary_security_group_source_cidrs": "",
     "volume_type": "gp2",
--- a/eks-worker-al2-variables.json
+++ b/eks-worker-al2-variables.json
@@ -30,7 +30,6 @@
     "source_ami_owners": "137112412989",
     "ssh_interface": "",
     "ssh_username": "ec2-user",
-    "ssm_agent_rpm_bucket_region": "us-west-2",
     "ssm_agent_version": "3.2.1377.0",
     "subnet_id": "",
     "temporary_security_group_source_cidrs": "",
--- a/eks-worker-al2-variables.json
+++ b/eks-worker-al2-variables.json
@@ -30,7 +30,7 @@
     "source_ami_owners": "137112412989",
     "ssh_interface": "",
     "ssh_username": "ec2-user",
-    "ssm_agent_version": "3.2.1377.0",
+    "ssm_agent_version": "latest",
     "subnet_id": "",
     "temporary_security_group_source_cidrs": "",
     "volume_type": "gp2",
--- a/eks-worker-al2-variables.json
+++ b/eks-worker-al2-variables.json
@@ -30,6 +30,7 @@
     "source_ami_owners": "137112412989",
     "ssh_interface": "",
     "ssh_username": "ec2-user",
+    "ssm_agent_rpm_bucket_region": "us-west-2",
     "ssm_agent_version": "3.2.1377.0",
     "subnet_id": "",
     "temporary_security_group_source_cidrs": "",
--- a/scripts/install-worker.sh
+++ b/scripts/install-worker.sh
@@ -32,7 +32,6 @@ validate_env_set PULL_CNI_FROM_GITHUB
 validate_env_set PAUSE_CONTAINER_VERSION
 validate_env_set CACHE_CONTAINER_IMAGES
 validate_env_set WORKING_DIR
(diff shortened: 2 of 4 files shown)


scripts/install-worker.sh
```diff
@@ -473,7 +474,7 @@ fi
 ### SSM Agent ##################################################################
 ################################################################################
 
-sudo yum install -y amazon-ssm-agent
+sudo yum install -y https://s3.amazonaws.com/ec2-downloads-windows/SSMAgent/${SSM_AGENT_VERSION}/linux_amd64/amazon-ssm-agent.rpm
```
bryantbiggs, line 477: I think you'll want to look at how other artifacts are pulled from region specific buckets https://github.com/awslabs/amazon-eks-ami/blob/55cbc55f69a029dc821d16081a7c75c206a5d1cd/scripts/install-worker.sh#L245-L255

vincentmrg: I've just added a variable to configure the region.
Note that according [the documentation](https://docs.aws.amazon.com/systems-manager/latest/userguide/agent-install-al2.html#custom-url-al2), the supported regions are the same as those defined on [Systems Manager service endpoints](https://docs.aws.amazon.com/general/latest/gr/ssm.html#ssm_region).

bryantbiggs: you should be able to re-use the already existing `$BINARY_BUCKET_REGION` and `$S3_DOMAIN` variables instead of adding new. and the domain will need to be updated as well

vincentmrg: Are you sure about this?
Looking at the documentation I put in this comment thread, the regions cn-north-1 / cn-northwest-1 / us-iso-east-1 / us-iso-west-1 / us-isob-east-1 are not supported.
That's why I chose to use a dedicated `SSM_AGENT_RPM_BUCKET_REGION` variable.

cartermckinnon: @vincentmrg SSM is definitely supported in those partitions, the docs are just elsewhere. E.g. https://docs.amazonaws.cn/en_us/aws/latest/userguide/endpoints-Beijing.html

vincentmrg: Thanks for the explanation, I'm not familiar with these particular regions, I've just pushed the requested changes.

bryantbiggs: Looks correct, but @cartermckinnon will have the final say

eks-worker-al2-variables.json
```diff
@@ -30,6 +30,7 @@
     "source_ami_owners": "137112412989",
     "ssh_interface": "",
     "ssh_username": "ec2-user",
+    "ssm_agent_version": "3.2.1377.0",
```
cartermckinnon, line 33: We should default this to `latest`

scripts/install-worker.sh
```diff
@@ -473,7 +474,7 @@ fi
 ### SSM Agent ##################################################################
 ################################################################################
 
-sudo yum install -y amazon-ssm-agent
+sudo yum install -y https://s3.${BINARY_BUCKET_REGION}.${S3_DOMAIN}/amazon-ssm-${BINARY_BUCKET_REGION}/${SSM_AGENT_VERSION}/linux_${ARCH}/amazon-ssm-agent.rpm
```
cartermckinnon, line 477: As we did with the AWS CLI, we need to fall back to installing from the `yum` repos in isolated regions: https://github.com/awslabs/amazon-eks-ami/blob/master/scripts/install-worker.sh#L122-L139

cartermckinnon (changes requested):
Couple small things. Thanks for putting this up!

eks-worker-al2-variables.json
vincentmrg: done

scripts/install-worker.sh
vincentmrg: done

cartermckinnon (approved):
LGTM. Thanks @vincentmrg !

PR #1383, opened 2023-08-08
[WIP] Handle eventually-consistent PrivateDnsName

cartermckinnon:
**Description of changes:**

Waits until the `PrivateDnsName` returned by EC2 is not empty. In a small percentage of cases, EC2 may return an empty `PrivateDnsName` for a period of time after the instance transitions to the `running` state.

**Testing Done**

*TODO*


bwagner5 (approved):
nice!! lgtm

files/bin/private-dns-name
```diff
@@ -0,0 +1,47 @@
...
+
+ATTEMPT=0
+while true; do
+  PRIVATE_DNS_NAME=$(aws ec2 describe-instances --region $REGION --instance-ids $INSTANCE_ID | jq -r '.Reservations[].Instances[].PrivateDnsName')
+  if [ ! "${PRIVATE_DNS_NAME}" = "" ] || [ ${ATTEMPT} -ge ${PRIVATE_DNS_NAME_MAX_ATTEMPTS} ]; then
+    break
+  fi
+  ATTEMPT=$((ATTEMPT + 1))
+  JITTER=$(seq "-${PRIVATE_DNS_NAME_JITTER}" "${PRIVATE_DNS_NAME_JITTER}" | shuf -n1)
+  DELAY=$((PRIVATE_DNS_NAME_ATTEMPT_INTERVAL + JITTER))
```
ravisinha0506, line 35: Can we just wait for regular interval, in this way we would know how long we have waited for before giving up?

cartermckinnon: I don't feel too strongly either way; but there is a case in which the describe-instances call is throttled, and the jitter will help in that case. If you think the debuggability of a consistent interval is more worthwhile, I'll remove this

ravisinha0506: If we want to handle the throttling scenarios, then we should do back-off retry. This isn't a blocking comment from my end. I am approving this PR for now.

ravisinha0506 (approved):

files/bin/private-dns-name
cartermckinnon: That's fair. I'll just remove the jitter to keep things simple, and we can add exponential backoff later if we decide it's necessary.

PR #1384, opened 2023-08-09
Install latest runc 1.1.*

cartermckinnon:
**Description of changes:**

Expands `runc` package selection to `1.1.*`.

### DIFF
--- a/eks-worker-al2-variables.json
+++ b/eks-worker-al2-variables.json
@@ -23,7 +23,7 @@
     "pause_container_version": "3.5",
     "pull_cni_from_github": "true",
     "remote_folder": "/tmp",
-    "runc_version": "1.1.5-1.amzn2",
+    "runc_version": "1.1.*",
     "security_group_id": "",
     "source_ami_filter_name": "amzn2-ami-minimal-hvm-*",
     "source_ami_id": "",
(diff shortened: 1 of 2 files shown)


eks-worker-al2-variables.json
```diff
@@ -23,7 +23,7 @@
     "pause_container_version": "3.5",
     "pull_cni_from_github": "true",
     "remote_folder": "/tmp",
-    "runc_version": "1.1.5-1.amzn2",
+    "runc_version": "1.1.*",
```
ravisinha0506, line 26: What's the version which would get install if we don't add anything here?

cartermckinnon: If we specified `*` here, we would always get the latest. I've scoped it down a bit so we don't install, e.g. `2.0` without warning and break the build.

cartermckinnon: The latest version available is `runc-1.1.7-2.amzn2`, which will be pulled in by `1.1.*` obvs

cartermckinnon:
Oh nice, the doc linter did it's thing 👍 https://github.com/awslabs/amazon-eks-ami/actions/runs/5812932554/job/15759400309?pr=1384

camrakin (approved):

PR #1391, opened 2023-08-16
Build with latest binaries available by default

cartermckinnon:
**Description of changes:**

This removes the hard-coded Kubernetes binary versions and build dates from the Makefile, and adds a script to discover the latest binaries available at build time.

This aims to remove a circular dependency in the current AMI build and release process. Instead of referencing specific binaries in version control, the binaries that were used to build each AMI are present in the release notes.

**Testing Done**

```
> hack/latest-binaries.sh 1.23
kubernetes_version=1.23.17 kubernetes_build_date=2023-08-15

> make 1.23
make k8s kubernetes_version=1.23.17 kubernetes_build_date=2023-08-15
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,8 +53,12 @@ T_GREEN := \e[0;32m
 T_YELLOW := \e[0;33m
 T_RESET := \e[0m
 
-.PHONY: latest
-latest: 1.28 ## Build EKS Optimized AL2 AMI with the latest supported version of Kubernetes
+# default to the latest supported Kubernetes version
+k8s=1.28
+
+.PHONY: build
+build: ## Build EKS Optimized AL2 AMI
+	$(MAKE) k8s $(shell hack/latest-binaries.sh $(k8s))
 
 # ensure that these flags are equivalent to the rules in the .editorconfig
 SHFMT_FLAGS := --list \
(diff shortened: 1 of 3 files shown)


hack/latest-binaries.sh
```diff
@@ -0,0 +1,18 @@
...
+if [ "$#" -ne 1 ]; then
+  echo "usage: $0 KUBERNETES_VERSION"
+  exit 1
+fi
+
+K8S_VERSION="${1}"
+
+LATEST_K8S_VERSION=$(aws s3 ls "s3://amazon-eks/${K8S_VERSION}" | awk '{print $2}' | rev | cut -c2- | rev | sort -V -r | head -n1)
+
+LATEST_DATE=$(aws s3 ls "s3://amazon-eks/$LATEST_K8S_VERSION/" | awk '{print $2}' | rev | cut -c2- | rev | sort -r | head -n1)
```
cartermckinnon, line 16: I could do this in one call with `s3api list-objects`, but it's not really any faster and a lot uglier.

mmerkes:
Chatted offline. I'm good with the idea of this change, as it's a best practice and is a change at build time, not runtime. However, I think that we should update the user guide to have information on this.

Makefile
```diff
@@ -109,27 +109,25 @@ k8s: validate ## Build default K8s version of EKS Optimized AL2 AMI
 	@echo "$(T_GREEN)Building AMI for version $(T_YELLOW)$(kubernetes_version)$(T_GREEN) on $(T_YELLOW)$(arch)$(T_RESET)"
 	$(PACKER_BINARY) build -timestamp-ui -color=false $(PACKER_VAR_FLAGS) $(PACKER_TEMPLATE_FILE)
 
-# Build dates and versions taken from https://docs.aws.amazon.com/eks/latest/userguide/install-kubectl.html
-
 .PHONY: 1.23
 1.23: ## Build EKS Optimized AL2 AMI - K8s 1.23
-	$(MAKE) k8s kubernetes_version=1.23.17 kubernetes_build_date=2023-08-16
+	$(MAKE) k8s $(shell hack/latest-binaries.sh 1.23)
```
ndbaker1, line 114: since these all pretty much look the same is there anything stopping us from generating these targets using the [`kubernetes-versions.json`](https://github.com/awslabs/amazon-eks-ami/blob/master/kubernetes-versions.json)? i think it might look like

```Make
# generate k8s version targets from versions json
K8S_VERSIONS=$(shell cat kubernetes-versions.json | jq -r -c 'join(" ")')
.PHONY: $(K8S_VERSIONS)
$(K8S_VERSIONS): %:
	$(MAKE) k8s $(shell hack/latest-binaries.sh $@)
```

and once this is rebased we'd need to add `1.28` to the `kubernetes-version.json`!

cartermckinnon: We could, yeah; but that JSON file controls which versions are used for CI, which in turn depends on `eksctl` supporting the version, so we can't couple the make targets with that list.

We could avoid a dependency on the JSON list if we go with a make variable instead of a target, i.e.:
```
make k8s=1.27
```

instead of:
```
make 1.27
```

ndbaker1: gotcha, so is there any validation required around that input? or is it just based on whether version/binaries can be found for a given `k8s=*`, and if so then it can be built

cartermckinnon: Yep, we can just bail if there are no binaries available. I added this in the latest rev 👍

Issacwww (approved):

doc/USER_GUIDE.md
```diff
@@ -66,50 +66,68 @@ Users have the following options for specifying their own values:
...
 ---
 
-## Building against other versions of Kubernetes binaries
-To build an Amazon EKS Worker AMI with other versions of Kubernetes that are not listed above run the following AWS Command
-Line Interface (AWS CLI) commands to obtain values for KUBERNETES_VERSION, KUBERNETES_BUILD_DATE, PLATFORM, ARCH from S3
+## Choosing Kubernetes binaries
+
+When building the AMI, binaries such as `kubelet`, `aws-iam-authenticator`, and `ecr-credential-provider` are installed.
+
+### Using the latest binaries
```
mmerkes, line 73: Tx for updating the docs!

hack/latest-binaries.sh
```diff
@@ -0,0 +1,24 @@
...
+set -o nounset
+
+if [ "$#" -ne 1 ]; then
+  echo "usage: $0 KUBERNETES_MINOR_VERSION"
+  exit 1
+fi
+
+MINOR_VERSION="${1}"
+
+LATEST_BINARIES=$(aws s3api list-objects-v2 --bucket amazon-eks --prefix "${MINOR_VERSION}" --query 'Contents[*].[Key]' --output text | cut -d'/' -f-2 | sort -V -r | uniq | head -n1)
```
mmerkes, line 14: Can provide a comment here that explains what this is doing and what the directory should look like?

PR #1395, opened 2023-08-17
Fetch new IMDS token for every request

cartermckinnon:
**Description of changes:**

IMDS throttling should not be a concern with our call volume, so caching the tokens isn't necessary.

**Testing**

I built a 1.27 AMI and launched a nodegroup with it, nodes joined the cluster, and I see no errors in the bootstrap log. all good!


mmerkes:
[We should clean this up too](https://github.com/awslabs/amazon-eks-ami/blob/865f9f2b6990e8acc407f4d0456411c16cf90964/scripts/cleanup.sh#L28).

I see 5 calls to IMDS: [1/2](https://github.com/awslabs/amazon-eks-ami/blob/865f9f2b6990e8acc407f4d0456411c16cf90964/files/bin/provider-id#L6-L7) [3/4](https://github.com/awslabs/amazon-eks-ami/blob/865f9f2b6990e8acc407f4d0456411c16cf90964/files/bootstrap.sh#L311-L312) [5](https://github.com/awslabs/amazon-eks-ami/blob/865f9f2b6990e8acc407f4d0456411c16cf90964/files/bin/configure-clocksource#L30)

If we think that IMDS throttling could potentially be an issue for large scale ups, we could just explicitly get the token in the scripts and re-use for each call.

mmerkes:
Did you validate that the AMIs continue to function as expected?

mmerkes (approved):
LGTM. Add your notes about the successful test before closing this PR.

PR #1405, opened 2023-08-25
don't set default cgroup option if reserved-cpus provided via kubelet…

raghs-aws:
… args

**Issue #, if available:**
Kubelet fails to start with --reserved-cpus argument due to conflict with system-reserved-cgroups being always specified in the kubelet configuration file. Bootstrap.sh shall allow use of reserved-cpus argument since that option is required for telco NFV workloads. Per k8s official documentation:

This option is specifically designed for Telco/NFV use cases where uncontrolled interrupts/timers may impact the workload performance. you can use this option to define the explicit cpuset for the system/kubernetes daemons as well as the interrupts/timers, so the rest CPUs on the system can be used exclusively for workloads, with less impact from uncontrolled interrupts/timers.

**Description of changes:**
Change is made to look for  --reserved-cpus options in KUBELET_EXTRA_ARGS. If reserved-cpus are requested then don't set systemReservedCgroup and kubeReservedCgroup, as reserved cpus will reserve the cpus for system and kube.
if the --reserved-cpus  parameter is  not present then the default handling would be retained which will allow setting kubeReservedCgroup and kubeReservedCgroup

**Testing Done**

Yes, validated with and without  --reserved-cpus in kubelet-extra-args. 

Launch self managed EKS worker node group using latest EKS AMI (post Jan 2023) which sets systemReservedCgroup and kubeReservedCgroup in the kubelet configuration file by default. Add --reserved-cpus flag to BootstrapArguments for the node group. Kubelet will fail to start, if the above changes are not added. Once the changes are added as per this PR, the kubelet will come up with reserved-cpus.


nebm-aws (approved):
Looks good. Tested with karpenter.

raghs-aws:
@cartermckinnon what are the next steps, I see that PR is open as code is not merged and is pending an review.  Let me know, if there is anything else I need to trigger.

cartermckinnon:
You'll need to fix the unit tests because of the change in behavior, and please add a new test for this. The linter also failed, so run make fmt.

files/bootstrap.sh
```diff
@@ -565,8 +572,12 @@ if [[ "$CONTAINER_RUNTIME" = "containerd" ]]; then
   sudo sed -i s,SANDBOX_IMAGE,$PAUSE_CONTAINER,g /etc/eks/containerd/containerd-config.toml
 
   echo "$(jq '.cgroupDriver="systemd"' "${KUBELET_CONFIG}")" > "${KUBELET_CONFIG}"
-  echo "$(jq '.systemReservedCgroup="/system"' "${KUBELET_CONFIG}")" > "${KUBELET_CONFIG}"
-  echo "$(jq '.kubeReservedCgroup="/runtime"' "${KUBELET_CONFIG}")" > "${KUBELET_CONFIG}"
+  ##allow --reserved-cpus options via kubelet arg directly and not set default cgroup option
+  if [[ "${USE_RESERVED_CPUS}" = false ]]; then
+    log "INFO: skipping setting of .systemReservedCgroup and .kubeReservedCgroup!"
```
cartermckinnon, line 577: This log line is incorrect; within this `if` we *are* setting the cgroups. You already state that it'll be skipped in the other log line, so just remove this

files/bootstrap.sh
```diff
@@ -222,6 +222,13 @@ ENABLE_LOCAL_OUTPOST="${ENABLE_LOCAL_OUTPOST:-}"
 CLUSTER_ID="${CLUSTER_ID:-}"
 LOCAL_DISKS="${LOCAL_DISKS:-}"
 
+##allow --reserved-cpus options via kubelet arg directly and not set default cgroup option
+USE_RESERVED_CPUS=false
+if [[ ${KUBELET_EXTRA_ARGS} == *'--reserved-cpus'* ]]; then
+  USE_RESERVED_CPUS=true
+  log "INFO: received reserved-cpus options via kubelet arg: ${KUBELET_EXTRA_ARGS} .cgroup settings will not be set"
```
cartermckinnon, line 229:
```suggestion
  log "INFO: --kubelet-extra-args includes --reserved-cpus, so kube/system-reserved cgroups will not be used."
```

test/cases/reserved-cpus-kubelet-arg.sh
```diff
@@ -0,0 +1,98 @@
...
+if [[ ${exit_code} -ne 0 ]]; then
+  echo "❌ Test Failed: expected a non-zero exit code but got '${exit_code}'"
+  exit 1
+fi
+
+retCode=$(
+  grep -q systemReservedCgroup /etc/kubernetes/kubelet/kubelet-config.json
+  echo $?
+)
+if [[ ${retCode} -eq 0 ]]; then
```
cartermckinnon, line 23: You can write this sort of thing like:
```suggestion
if grep -q systemReservedCgroup /etc/kubernetes/kubelet/kubelet-config.json; then
```

I would also use a variable instead of repeating the kubelet config path several times.

cartermckinnon (changes requested):
Few small things.

test/cases/reserved-cpus-kubelet-arg.sh
raghs-aws: @cartermckinnon committed changes and resolved all the comments

files/bootstrap.sh
```diff
@@ -222,6 +222,13 @@ ENABLE_LOCAL_OUTPOST="${ENABLE_LOCAL_OUTPOST:-}"
 CLUSTER_ID="${CLUSTER_ID:-}"
 LOCAL_DISKS="${LOCAL_DISKS:-}"
 
+##allow --reserved-cpus options via kubelet arg directly and not set default cgroup option
+USE_RESERVED_CPUS=false
```
cartermckinnon, line 226: I think we should invert the semantics of this variable, to disable the cgroup settings (e.g. `USE_RESERVED_CGROUPS`) instead of enabling reserved CPU's. Other than that, this PR LGTM.

raghs-aws: makes sense, this way if we have some other usecases which needs disabling of cgroup, then we can reuse the same variable.  let me push the changes,

cartermckinnon:
@raghs-aws you need to make your test case script executable.

raghs-aws:
> @raghs-aws you need to make your test case script executable (`chmod +x`). Run `make test` to check locally.



> @raghs-aws you need to make your test case script executable (`chmod +x`). Run `make test` to check locally.

somehow the script got duplicated the original one passed but the copy didnt have the +x, I will delete that test case file.

PR #1434, opened 2023-09-21
Set pid_max to 4194304

### LINKED ISSUE #737: Max pids is set to 32768
**What happened**:

Nodes crashed because we didn't set a max pids per pod. 

Question is why this is set to 32768? Which seems to be the max value for 32-bit.
Can we raise it? I'll make the PR. 

Run a lot of Java :')

cartermckinnon:
**Issue #, if available:**

Fixes #737

**Description of changes:**

The current default value of 32768 is from the 32-bit days and is defunct. Newer versions of `systemd` (244+) set this value to the maximum-possible, 4194304. That's what is currently being used on AL2023, for example, because it uses `systemd` 252.

**Testing Done**

Confirmed on the 1.27 CI image:
```
> cat /proc/sys/kernel/pid_max 
4194304
```


bryantbiggs (approved):
🚀

dims (approved):
LGTM

PR #1437, opened 2023-09-21
Fix containerd slice configuration

### LINKED ISSUE #1436: Follow general guideline: containerd and kubelet should share the same kubereserved cgroup
The kubernetes General Guideline explains that the kubelet and the container runtime should share a cgroup slice. 
The design proposal for node-allocatable resources also suggests that both kubelet and containerd are to be put into the same cgroup, and that kubelet should share it's reserved resources with the container runtime.

> Following is the recommended cgroup configuration for Kubernetes nodes. All OS system daemons are expected to be placed under a top level SystemReserved cgroup. Kubelet and Container Runtime are expected to be placed under KubeReserved cgroup. The reason for recommending placing the Container Runtime under KubeReserved is as follows:
>
> A container runtime on Kubernetes nodes is not expected to be used outside of the Kubelet.
    It's resource consumption is tied to the number of pods running on a node.

Right now I'm seeing that this is not the case with the EKS AMI when running with default userdata: The cgroup tree is roughly like this:

```
/
├─runtime.slice
│   └─kubelet.service
└─system.slice
     └─containerd.service
```

While it should look like this instead:

```
/
└──runtime.slice
    ├─containerd.service
    └─kubelet.service
```

Related docs:
https://kubernetes.io/docs/tasks/administer-cluster/reserve-compute-resources/#general-guidelines
https://github.com/kubernetes/design-proposals-archive/blob/main/node/node-allocatable.md#recommended-cgroups-setup

reegnz:
The containerd systemd slice cannot be changed once containerd has
started and systemd has realized the cgroup for the unit.

Therefore we need to set the slice during the AMI build and not during
the cloud-init process when contanerd has already started.

Fixes #1436

**Issue #, if available:**

**Description of changes:**

**Testing Done**

I can't build an AMI, and this change should be tested with a new AMI build.


cartermckinnon:
Looks fine in the CI cluster:
```
─runtime.slice
│ ├─containerd.service
│ │ ├─2869 /usr/bin/containerd
│ │ ├─3113 /usr/bin/containerd-shim-runc-v2 -namespace k8s.io -address /run/containerd/containerd.sock -publish-binary /usr/bin/containerd -id 94a63d8790cbf4fafda
│ │ └─3114 /usr/bin/containerd-shim-runc-v2 -namespace k8s.io -address /run/containerd/containerd.sock -publish-binary /usr/bin/containerd -id 632fe29132d415b3467
│ └─kubelet.service
│   └─3012 /usr/bin/kubelet --config /etc/kubernetes/kubelet/kubelet-config.json --kubeconfig /var/lib/kubelet/kubeconfig --container-runtime-endpoint unix:///run
└
```

I'm not sure what was different when I was poking around on a node 🤷 thanks @reegnz

reegnz:
@cartermckinnon  That looks awesome.

scripts/install-worker.sh
```diff
@@ -175,6 +175,11 @@ sudo chmod +x /etc/eks/containerd/pull-sandbox-image.sh
 sudo chmod +x /etc/eks/containerd/pull-image.sh
 
 sudo mkdir -p /etc/systemd/system/containerd.service.d
+cat << EOF | sudo tee /etc/systemd/system/containerd.service.d/00-runtime-slice.conf
+[Service]
+Slice=runtime.slice
+EOF
```
cartermckinnon, line 181: We still need to add a condition around this for k8s 1.24+, otherwise this LGTM

reegnz: IMHO the advice around using the runtime.slice also applies to the Docker runtime (which is where the current installation of containerd comes from as a dependency).

reegnz: Anyway, will add a conditional if that makes you feel more comfortable.

reegnz: @cartermckinnon by that condition did you mean the slice should be configured with all versions above and including 1.24.0? Isn't that the lowest kubernetes version the AMI is built for?

Sorry I don't get the exact requirement, I'd appreciate if you'd elaborate:
* why is the condition needed
* what exact version range are we talking about where the runtime.slice should be configured for for containerd

reegnz: Edit, ah OK, the EOL for 1.23 is October 11 I see. 

@cartermckinnon Well maybe let's just wait 2 weeks before we merge this PR and then no condition is necessary anymore to merge the change since 1.23 is then EOL.

reegnz: @cartermckinnon I gues this latest announcement means you won't drop 1.23 support soon: https://aws.amazon.com/blogs/containers/amazon-eks-extended-support-for-kubernetes-versions-available-in-preview/

So I'll add the version check.

reegnz: Added the version check.

reegnz:
@cartermckinnon I added the version check, but couldn't verify if it works yet. I hope we'll see it in CI.

scripts/install-worker.sh
```diff
@@ -172,7 +172,15 @@ sudo mv $WORKING_DIR/pull-image.sh /etc/eks/containerd/pull-image.sh
 sudo chmod +x /etc/eks/containerd/pull-sandbox-image.sh
 sudo chmod +x /etc/eks/containerd/pull-image.sh
 
-sudo mkdir -p /etc/systemd/system/containerd.service.d
+CONFIGURE_CONTAINERD_SLICE=$(vercmp "$KUBERNETES_VERSION" gteq "1.24.0" || true)
+if [ "$CONFIGURE_CONTAINERD_SLICE" == "true" ]; then
+  sudo mkdir -p /etc/systemd/system/containerd.service.d
```
cartermckinnon, line 177: This needs to be moved outside the `if`, CI failed on 1.23:
```
2023-10-13T05:52:06Z:     amazon-ebs: tee: /etc/systemd/system/containerd.service.d/10-compat-symlink.conf: No such file or directory
```

reegnz: whoops. fixed.

reegnz:
@cartermckinnon can you please relaunch the CI please.

PR #1463, opened 2023-10-10
Add named arguments to bot commands

cartermckinnon:
**Description of changes:**

The `/ci` command accepts a "goal" as an argument. The goals make up a DAG, `build` -> `launch` -> `test`.

We sometimes need to kick off a CI run with specific arguments for one of these goals. For example, you might want to achieve the `test` goal, but pass an additional argument to the `build` goal.

This PR adds a new syntax to pass "named arguments" to a command. For the above example, you would do:
```
/ci test
```

**Testing Done**

Tested in my fork: https://github.com/cartermckinnon/amazon-eks-ami/pull/6864#issuecomment-1756238390

I can see the arg passed correctly to the workflow in the logs.

### DIFF
--- a/.github/actions/bot/index.js
+++ b/.github/actions/bot/index.js
@@ -25,7 +25,7 @@ async function bot(core, github, context, uuid) {
         commands = parseCommands(uuid, payload, payload.comment.body);
     } catch (error) {
         console.log(error);
-        const reply = `@${author} I didn't understand [that](${payload.comment.html_url}! 🤔\nTake a look at my [logs](https://github.com/${github.repository}/actions/runs/${github.run_id}).`
+        const reply = `@${author} I didn't understand [that](${payload.comment.html_url})! 🤔\n\nTake a look at my [logs](${getBotWorkflowURL(payload, context)}).`
         replyToCommand(github, payload, reply);
         return;
     }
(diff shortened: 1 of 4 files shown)


.github/actions/bot/index.js
```diff
@@ -49,17 +49,32 @@ async function bot(core, github, context, uuid) {
...
+                if (previousCommand) {
+                    if (typeof previousCommand.addNamedArguments === 'function') {
+                        previousCommand.addNamedArguments(namedArguments.name, namedArguments.args);
+                    } else {
+                        console.log(`Parsed named arguments but previous command (${previousCommand.constructor.name}) does not support arguments: ${JSON.stringify(namedArguments)}`)
+                    }
+                } else {
+                    console.log(`Parsed named arguments with no previous command: ${JSON.stringify(namedArguments)}`)
+                }
+            }
```
Issacwww, line 77: Should `previousCommand` be a required? can we validate the there is at least one line to reduce the nest if condition?

cartermckinnon: I've been just logging and exiting the bot when things like this happen, but I'll add a function to reply to the user saying something went wrong 👍

cartermckinnon: @Issacwww added in latest rev, see my fork for example usage: https://github.com/cartermckinnon/amazon-eks-ami/pull/6864#issuecomment-1758655539

mmerkes (approved):

PR #1465, opened 2023-10-11
Pull eksctl supported versions

ndbaker1:
**Issue #, if available:**

**Description of changes:**

update the manual ci workflow to pull supported k8s versions [directly from eksctl](https://github.com/eksctl-io/eksctl/pull/2126#issuecomment-626837847)

**Testing Done**

Identical output to what is used currently:

```bash
# setup
git clone https://github.com/awslabs/amazon-eks-ami
cd amazon-eks-ami
git checkout -q 1c9e032
# execution
echo "kubernetes_versions=$(cat kubernetes-versions.json | jq -c .)"
```

with the following:

```bash
wget --no-verbose -O eksctl.tar.gz "https://github.com/weaveworks/eksctl/releases/latest/download/eksctl_Linux_amd64.tar.gz"
tar xf eksctl.tar.gz && chmod +x ./eksctl
echo "kubernetes_versions=$(eksctl version --output json | jq -c .EKSServerSupportedVersions)"
```


.github/workflows/ci-manual.yaml
```diff
@@ -36,14 +36,17 @@ jobs:
...
     steps:
-    - uses: actions/checkout@v3
     - id: variables
       run: |
         echo "git_sha_short=$(echo ${{ inputs.git_sha }} | rev | cut -c-7 | rev)" >> $GITHUB_OUTPUT
         echo "workflow_run_url=https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}" >> $GITHUB_OUTPUT
-        echo "kubernetes_versions=$(cat kubernetes-versions.json | jq -c .)" >> $GITHUB_OUTPUT
+        # grab supported versions directly from eksctl
+        wget --no-verbose -O eksctl.tar.gz "https://github.com/weaveworks/eksctl/releases/latest/download/eksctl_Linux_amd64.tar.gz"
+        tar xf eksctl.tar.gz && chmod +x ./eksctl
```
cartermckinnon, line 45: This fails for me without `z` because the tarball is GZIP'd:
```
tar: Hmm, this doesn't look like a tar archive
tar: Skipping to next file header
tar: Only read 5853 bytes from archive eksctl.tar.gz
tar: Error is not recoverable: exiting now
```
```suggestion
        tar xzf eksctl.tar.gz && chmod +x ./eksctl
```

cartermckinnon:
Nice! 🙌

.github/workflows/ci-manual.yaml
ndbaker1: oh i guess that's not autodetected in every case :/, should this be updated in [.github/actions/ci/launch/action.yaml](https://github.com/awslabs/amazon-eks-ami/blob/master/.github/actions/ci/launch/action.yaml#L25) as well just in case?

cartermckinnon: Maybe a BSD vs. GNU thing, I was testing on macOS. We should probably centralize the eksctl setup in an Action, but that doesn’t need to be in this PR.

PR #1467, opened 2023-10-12
Add CHANGELOG workflow

cartermckinnon:
**Description of changes:**

This adds a workflow that is triggered on publication of a new release. The notes of the release will be inserted in the `CHANGELOG.md` file at the placeholder: https://github.com/awslabs/amazon-eks-ami/blob/915ce2222e692a4d7b5904ab020d3c28e1e3e511/CHANGELOG.md?plain=1#L3

And the result will be pull-requested.

**Testing Done**

Tested in my fork (with some tweaks -- this type of workflow won't run in a fork without modification).


.github/workflows/update-changelog.yaml
```diff
@@ -0,0 +1,57 @@
+name: '[Release] Update CHANGELOG.md'
+on:
+  release:
+    types: [released]
```
mmerkes, line 4: Will this run when we generate the pre-release or only the final release?

.github/workflows/update-changelog.yaml
```diff
@@ -0,0 +1,57 @@
...
+      - uses: actions/checkout@v3
+        with:
+          repository: awslabs/amazon-eks-ami
+          ref: refs/heads/master
+          path: amazon-eks-ami/
+      - uses: actions/github-script@v6
+        with:
+          script: |
+            const fs = require('fs');
+            const changelogPath = './amazon-eks-ami/CHANGELOG.md';
```
mmerkes, line 33: We should verify that the changelog doesn't already have the release before we add it to ensure that this workflow is idempotent. If it fails, we'd ideally open a github issue.

.github/workflows/update-changelog.yaml
cartermckinnon: Only after the final release. The `prereleased` activity type has some caveats: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release

> he prereleased type will not trigger for pre-releases published from draft releases, but the published type will trigger. If you want a workflow to run when stable and pre-releases publish, subscribe to published instead of released and prereleased.

So I just used released to keep it simple.

.github/workflows/update-changelog.yaml
cartermckinnon: I added a check for the release title. We'll get an email if this workflow fails, I *could* open an issue if you feel strongly about it

mmerkes: Not really. I think that the changelog has limited value, if any, since it's already in the release notes and now it's just a formatted list of commits and version information.

mmerkes (approved):

cartermckinnon:
Test succeeded on latest rev: https://github.com/cartermckinnon/amazon-eks-ami/pull/6867

PR #1468, opened 2023-10-12
get-ecr-uri.sh falls back to use another region in partition if regio…

mmerkes:
…n unconfigured

**Issue #, if available:**
N/A

**Description of changes:**
The `get-ecr-uri/sh` script is used to get the ECR URI for the pause container in the region. I've updated that script to fall back to another region in the same partition if the region is unknown. This is necessary to automate support for new regions. Eventually, we should still add an entry for the specific region, but this will enable new regions without a change.

**Testing Done**

New unit tests added for the script. Also, I removed the entry for `eu-south-1`, which is an opt-in region, and validated that the nodes were still able to join the clusters in that region.

### DIFF
--- a/files/get-ecr-uri.sh
+++ b/files/get-ecr-uri.sh
@@ -39,14 +39,14 @@ else
     af-south-1)
       acct="877085696533"
       ;;
-    eu-south-1)
-      acct="590381155156"
-      ;;
     ap-southeast-3)
       acct="296578399912"
       ;;
     me-central-1)
       acct="759879836304"
+      ;;
+    eu-south-1)
+      acct="590381155156"
       ;;
     eu-south-2)
       acct="455263428931"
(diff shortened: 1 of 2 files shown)


files/get-ecr-uri.sh
```diff
@@ -39,15 +59,15 @@ else
     af-south-1)
```
Issacwww, line 59: Changes looks good to me! Out of curiosity, so does it mean we can actually not maintain this account list anymore but just using one account per partition?

zaf6862:
LGTM.

zaf6862 (approved):

files/get-ecr-uri.sh
mmerkes: No. We should always add a new account when there's a new region, but this means that there's no rush. We can start building the region, and do this at our leisure.

Issacwww (approved):

cartermckinnon:
@mmerkes looks fine to me, the linter just wants some indentation tweaks.

mmerkes:
@cartermckinnon Ya, I forget that we have the linter and pushed and logged off before it finish. I just updated it. Should be gtg.

files/get-ecr-uri.sh
```diff
@@ -63,10 +63,55 @@ else
...
+      us-east-2 | \
+      us-west-1 | \
+      us-west-2)
       acct="602401143452"
       ;;
-  esac
+    *)
+      # If the region is not mapped to an account, let's try to choose another region
+      # in that partition.
+      case $aws_domain in
```
cartermckinnon, line 91: wait a minute -- this value is a TLD, not a partition name. E.g. for `aws-cn`, it's `amazonaws.com.cn`, right?

cartermckinnon (changes requested):

files/get-ecr-uri.sh
mmerkes: Doh! Good catch. I'm just going to use the region name since this is just a fallback anyway.

Issacwww (approved):

cartermckinnon (approved):

PR #1472, opened 2023-10-13
Force delete CI clusters, don't wait for pod eviction

cartermckinnon:
**Description of changes:**

The janitor is stuck on a CI cluster because a pod can't be evicted: https://github.com/awslabs/amazon-eks-ami/actions/runs/6512608996/job/17690593749

This PR:
1. Simplifies the cluster sweeper. Previously, a CI cluster may be shared by multiple PR's, so there was separate handling for nodegroup deletion. We create a new cluster + nodegroup for every CI run now, so this removes the nodegroup handling.
2. We now force-delete the cluster, and we disable pod eviction.


mmerkes (approved):
LGTM

Issue #1482, opened 2023-10-24
hack/latest-binaries.sh should use binary_bucket_name and binary_bucket_region, when specified


donovanrost:
**What happened**:
Builds fail when in GovCloud environment because `hack/latest-binaries.sh` can't can't query the `amazon-eks` bucket.

**What you expected to happen**:
I expect to be able to build this project in a GovCloud account. 

**How to reproduce it (as minimally and precisely as possible)**:
1. Have credentials set for a GovCloud account
2. Attempt to `make ...`
3. Observe Failure

**Anything else we need to know?**:

**Environment**:
- AWS Region: us-gov-west-1

PR #1483, opened 2023-10-24
#1482 Use AWS CLI profile, binary_bucket_region and binary_bucket_name in hack/latest-binaries.sh

donovanrost:
**Issue #, if available:**
#1482 
**Description of changes:**
Wire in bucket_name, bucket_region, and profile into hack/latest_binaries.sh

**Testing Done**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,6 +51,19 @@ ifeq ($(aws_region), us-gov-west-1)
 	source_ami_owners ?= 045324592363
 endif
 
+# The following _FLAG vars are to be passed into hack/latest-binaries.sh 
+ifdef aws_profile
+	PROFILE_FLAG := --profile $(aws_profile)
+endif
+
+ifdef binary_bucket_name
+	BUCKET_FLAG := --bucket $(binary_bucket_name)
+endif
+
+ifdef binary_bucket_region
+	REGION_FLAG := --region $(binary_bucket_region)
+endif
+
 T_RED := \e[0;31m
 T_GREEN := \e[0;32m
 T_YELLOW := \e[0;33m
@@ -61,7 +74,7 @@ k8s=1.28
 
 .PHONY: build
 build: ## Build EKS Optimized AL2 AMI
-	$(MAKE) k8s $(shell hack/latest-binaries.sh $(k8s))
+	$(MAKE) k8s $(shell hack/latest-binaries.sh $(k8s) $(PROFILE_FLAG) $(BUCKET_FLAG) $(REGION_FLAG))
 
 # ensure that these flags are equivalent to the rules in the .editorconfig
 SHFMT_FLAGS := --list \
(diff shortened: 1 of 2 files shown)


hack/latest-binaries.sh
```diff
@@ -4,23 +4,75 @@ set -o errexit
 set -o pipefail
 set -o nounset
 
+
+BUCKET="amazon-eks"
+REGION="us-west-2"
+PROFILE=""
```
cartermckinnon, line 10: The region and profile would already be pulled in with the standard CLI environment variables (`AWS_PROFILE`, `AWS_REGION`).

hack/latest-binaries.sh
```diff
@@ -4,23 +4,75 @@ set -o errexit
 set -o pipefail
 set -o nounset
 
+
+BUCKET="amazon-eks"
```
cartermckinnon, line 8: Why don't you just allow overriding this environment variable directly?
```suggestion
BINARY_BUCKET=${BINARY_BUCKET:-amazon-eks}
```

cartermckinnon (changes requested):

hack/latest-binaries.sh
donovanrost: I can see the case for AWS_PROFILE. But `aws_region` is already is exposed as a `PKR_VAR_` and so by not explicitly using `--region <region>` in `latest-binaries` it creates confusing behavior

hack/latest-binaries.sh
donovanrost: Because the entrypoint into this project is `make` and one of the exposed variables is `binary_bucket_name`. I chose this approach to keep the bucket name consistent with this. 

```
make 1.28 binary_bucket_name=my-bucket
```
vs 

```
BINARY_BUCKET=my-bucket; make 1.28 binary_bucket_name=my_bucket
```

Unless I'm misunderstanding how 
`BINARY_BUCKET=${BINARY_BUCKET:-amazon-eks}` would work.

hack/latest-binaries.sh
cartermckinnon: You can call the script in the Makefile like:
```
AWS_REGION=$(aws_region) hack/latest-binaries.sh
```
I think this is just as clear as:
```
hack/latest-binaries.sh --region $(aws_region)
```
But with the former you don't need to add CLI parsing to this script.

hack/latest-binaries.sh
cartermckinnon: You can `export` variables within the Makefile to child commands, so you could use `binary_bucket_region` and `binary_bucket_name` directly in the script as well.

hack/latest-binaries.sh
donovanrost: I agree those are equally readable. Now, as I go about making these changes. Do you have a preference on how to conditionally export `AWS_REGION` and `BINARY_BUCKET_NAME` in the Makefile?

One of the reasons I took this approach is that my Make isn't that strong, and this was the most readable to approach

Makefile
```diff
@@ -51,6 +51,19 @@ ifeq ($(aws_region), us-gov-west-1)
...
+endif
+
+ifdef binary_bucket_name
+	BUCKET_FLAG := --bucket $(binary_bucket_name)
+endif
+
+ifdef binary_bucket_region
+	REGION_FLAG := --region $(binary_bucket_region)
+endif
+
```
cartermckinnon, line 66: I think this should do what you want:
```suggestion
# make available in hack/latest-binaries.sh 
export binary_bucket_region
export binary_bucket_name
```

donovanrost: It looks like the exported variables are not being picked up by$(shell ....)

donovanrost:
I'm having trouble navigating the idiosyncrasies of Make, to make this work via the environment or macro expansion. I think the submitted solution is perfectly fine. It's readable, extensible, doesn't clobber any environment variables, and doesn't break any existing workflows.

PR #1484, opened 2023-10-25
Fail fast if we cannot determine kubelet version

dims:
kubelet is likely to fail when there is a mismatch with GLIBC that is in the image vs the one golang uses to build the kubelet. So fail the image right away when this happens as this specific kubelet binary will NOT work in any instance started with this image.
```
2023-10-25T10:11:38-04:00:     amazon-ebs: kubelet: /lib64/libc.so.6: version `GLIBC_2.32' not found (required by kubelet)
2023-10-25T10:11:38-04:00:     amazon-ebs: kubelet: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by kubelet)
```

**Issue #, if available:**

**Description of changes:**

**Testing Done**

### DIFF
--- a/scripts/generate-version-info.sh
+++ b/scripts/generate-version-info.sh
@@ -16,8 +16,19 @@ OUTPUT_FILE="$1"
 sudo rpm --query --all --queryformat '\{"%{NAME}": "%{VERSION}-%{RELEASE}"\}\n' | jq --slurp --sort-keys 'add | {packages:(.)}' > "$OUTPUT_FILE"
 
 # binaries
-echo $(jq ".binaries.kubelet = \"$(kubelet --version | awk '{print $2}')\"" $OUTPUT_FILE) > $OUTPUT_FILE
-echo $(jq ".binaries.awscli = \"$(aws --version | awk '{print $1}' | cut -d '/' -f 2)\"" $OUTPUT_FILE) > $OUTPUT_FILE
+KUBELET_VERSION=$(kubelet --version | awk '{print $2}')
+if [ "$?" != 0 ]; then
+  echo "unable to get kubelet version"
+  exit 1
+fi
+echo $(jq ".binaries.kubelet = \"$KUBELET_VERSION\"" $OUTPUT_FILE) > $OUTPUT_FILE
+
+CLI_VERSION=$(aws --version | awk '{print $1}' | cut -d '/' -f 2)
+if [ "$?" != 0 ]; then
+  echo "unable to get aws cli version"
+  exit 1
+fi
+echo $(jq ".binaries.awscli = \"$CLI_VERSION\"" $OUTPUT_FILE) > $OUTPUT_FILE
 
 # cached images
 if systemctl is-active --quiet containerd; then


dims:
Not a theoretical problem... see https://storage.googleapis.com/kubernetes-jenkins/logs/ci-cgroupv1-containerd-node-e2e-ec2-eks/1717075170234994688/artifacts/i-0a28d6418f87301ac-system.log

dims:
cc @cartermckinnon

scripts/generate-version-info.sh
```diff
@@ -16,8 +16,17 @@ OUTPUT_FILE="$1"
 sudo rpm --query --all --queryformat '\{"%{NAME}": "%{VERSION}-%{RELEASE}"\}\n' | jq --slurp --sort-keys 'add | {packages:(.)}' > "$OUTPUT_FILE"
 
 # binaries
-echo $(jq ".binaries.kubelet = \"$(kubelet --version | awk '{print $2}')\"" $OUTPUT_FILE) > $OUTPUT_FILE
-echo $(jq ".binaries.awscli = \"$(aws --version | awk '{print $1}' | cut -d '/' -f 2)\"" $OUTPUT_FILE) > $OUTPUT_FILE
+KUBELET_VERSION=$(kubelet --version | awk '{print $2}')
+if [ "$?" != 0 ]; then
+  echo "unable to get kubelet version"
+fi
+echo $(jq ".binaries.kubelet = \"$KUBELET_VERSION\"" $OUTPUT_FILE) > $OUTPUT_FILE
```
mmerkes, line 23: Isn't this just failing in the same way, but with a different log?

scripts/generate-version-info.sh
```diff
@@ -16,8 +16,17 @@ OUTPUT_FILE="$1"
 sudo rpm --query --all --queryformat '\{"%{NAME}": "%{VERSION}-%{RELEASE}"\}\n' | jq --slurp --sort-keys 'add | {packages:(.)}' > "$OUTPUT_FILE"
 
 # binaries
-echo $(jq ".binaries.kubelet = \"$(kubelet --version | awk '{print $2}')\"" $OUTPUT_FILE) > $OUTPUT_FILE
-echo $(jq ".binaries.awscli = \"$(aws --version | awk '{print $1}' | cut -d '/' -f 2)\"" $OUTPUT_FILE) > $OUTPUT_FILE
+KUBELET_VERSION=$(kubelet --version | awk '{print $2}')
+if [ "$?" != 0 ]; then
```
mmerkes, line 20: Will we ever get here since the script is set to fail on errors?

scripts/generate-version-info.sh
dims: currently it is not failing when `kubelet --version` fails as it is earlier in the pipe (even though we set `pipefail`), the `errexit` does not kick in either. so we have to check this explicitly

scripts/generate-version-info.sh
dims: script doesn't fail, that's the problem. (also see above)

mmerkes: Right, but if it gets here, it's still not going to fail, right?

dims: sorry missed the exit (didn't push it)

mmerkes (approved):
LGTM

Issue #1486, opened 2023-10-26
Have cri-tools installed and configured


reegnz:
**What would you like to be added**:
Install cri-tools and configure it with /etc/crictl.yaml.

Previously there was an issue that requested this (#797) but it was closed by a change that introduced nerdctl instead.
AFAICT amazon-linux does have a cri-tools package, so installing it should be trivial.

**Why is this needed**:

Crictl is the officially documented tooling for kubernetes to troubleshoot nodes: https://kubernetes.io/docs/tasks/debug/debug-cluster/crictl/

Nerdctl does not have a concept of pods, while crictl does. This allows continuing to use pod semantics when troubleshooting within on a node using crictl vs using nerdctl and figuring out which containers belong to which pod (not terribly difficult, but still an extra step).
 
Having nerdctl on the nodes is nice and I think it belongs with any containerd installation by default, but the nodes are k8s nodes, missing capabilities that crictl does add, which is pod-aware semantics during troubleshooting.

Additional notes:

nerdctl is not properly configured on a node, `nerdctl ps` doesn't list anything, you have to use `nerdctl --namespace="k8s.io" ps`. I'll open a separate ticket to fix that as well, because on a k8s node one would expect nerdctl to default to the kubernetes pods namespace.

alvaroaleman:
+1 on this

`crictl` is the tool recommended by upstream Kubernetes to debug node issues: https://kubernetes.io/docs/tasks/debug/debug-cluster/crictl/

It is hard to understand that EKS nodes by default have a non-working version of a non-standard tool but do not have the standard tool recommended by upstream.

cartermckinnon:
1. This was already added: https://github.com/awslabs/amazon-eks-ami/blob/c5132e9b1cc342de477b7347ec62efed82d42ceb/templates/al2/provisioners/install-worker.sh#L152
2. `nerdctl` works fine, its default config was improved in: https://github.com/awslabs/amazon-eks-ami/pull/1488

`nerdctl` is an official CLI for `containerd` with (IMO) much better ergonomics than `crictl`; but it can definitely be helpful to have `crictl`. 👍

mihuross:
AL2023 is missing the cri-tools

cartermckinnon:
@mihuross fixing that in https://github.com/awslabs/amazon-eks-ami/pull/2162 👍

cartermckinnon:
[...]

PR #1489, opened 2023-10-26
Add cri-tools

reegnz:
This allows using the crictl cli tool on worker nodes.

**Issue #, if available:**

#1486

**Description of changes:**

Installing cri-tools ensures that crictl is available on the worker nodes.
crictl is the official 1st party kubernetes tool to interact with the container runtime in a runtime-agnostic way.
It is also pod-aware so troubleshooting pods running on a node is made easier.

**Testing Done**

Performed the same commands to install and configure crictl on a worker.
Running `crictl pods` and `crictl ps` works as expected.


cartermckinnon:
There's more discussion in #1235.

The `cri-tools` README was updated since then to have more firm guidance around compatibility: https://github.com/kubernetes-sigs/cri-tools#compatibility-matrix-cri-tools--kubernetes

> It's recommended to use the same cri-tools and Kubernetes minor version, because new features added to the Container Runtime Interface (CRI) may not be fully supported if they diverge.

With k8s >= 1.27 needing cri-tools >= 1.27, and earlier k8s needing cri-tools <= 1.26.

So we can't just install the latest version of cri-tools from the AL repo. We also can't guarantee that the necessary version of cri-tools is available in the repo for current or future k8s versions.

reegnz:
Nice catch @cartermckinnon , had that in the back of my mind but lost that by the time I opened the PR. 😅

Will adjust the logic to take that into consideration.

reegnz:
So the package being installed currently is coming from the `amzn2-core` repository:
```
[root@ip-10-80-140-61 /]# yum --showduplicates list cri-tools | expand
Available Packages
cri-tools.x86_64                  1.25.0-1.amzn2.0.1                  amzn2-core
cri-tools.x86_64                  1.26.1-1.amzn2.0.1                  amzn2-core
cri-tools.x86_64                  1.26.1-1.amzn2.0.2                  amzn2-core
```

@cartermckinnon it seems like the amzn2-core repo doesn't provide a version for 1.27 and up. Could you please escalate to the amazon-linux team so they also package up the 1.27 version?

Alternatively, we could bring in google's repo, which has all versions of cri-tools available: https://packages.cloud.google.com/yum/repos/kubernetes-el7-x86_64

As for version pinning, I'm not 100% familiar with how yum does it.

AFAICT you can install with `yum install cri-tools-1.26.1` but then you don't get updates if there's a hotfix. Alternatively you can `yum instl cri-tools < 1.27` but I'm unsure how it works with upgrades, although during AMI build it would likely always pick the latest version matching the constraint if that version is available.

reegnz:
Ironically, the google yum repo also doesn't have a 1.27 package. 😅 
So I think this is a hard blocker to use cri-tools if nobody else has made any yum packages available either...

PR #1274, opened 2023-04-21
Include user-data in log collection bundle

cartermckinnon:
**Description of changes:**

Diagnosing issues with node bootstrapping almost always requires the details of the user-data. This eliminates a round-trip email in the support cycle.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1504, opened 2023-11-04
Document how to collect UserData

edmondceausu:
**Issue #, if available:** None

**Description of changes:**

Requesting to add userdata collection in the EKS log collection. This is very useful when troubleshooting EKS Nodes not joining their cluster or any Kubelet configuration issues that may occur when calling `bootstrap.sh`.

This would option would be opt-in

**Testing Done**

Tested this in my own environment to collect the Userdata on an EKS node. Does not interfere with other other functions.


cartermckinnon:
We've considered this in the past: #1274 and decided against it because of the sensitive nature of this data. If you want to add a section to the log collector README about retrieving the user data, that could be a reference for when this info is requested when troubleshooting; but we should not include it in the general log bundle.

edmondceausu:
> We've considered this in the past: #1274 and decided against it because of the sensitive nature of this data. If you want to add a section to the log collector README about retrieving the user data, that could be a reference for when this info is requested during troubleshooting; but we should not include it in the general log bundle.

Sure thing. I will update the PR to add this information to the README within the log bundles.

cartermckinnon (approved):

cartermckinnon:
Thanks @edmondceausu!

PR #1513, opened 2023-11-14
add sudo to avoid error out with UMASK set to 027

leowenlu:
**Issue #, if available:**

When source_ami is using With UMASK set to 027, `kubelet` and `aws` will error out with permissions issue.

 The following is what I have for both `kubelet` and `aws`
```
amazon-ebs: -rwxr-x---. 1 root root 120671264 Nov 7 09:14 /usr/bin/kubelet
amazon-ebs: -rwxr-x---. 1 root root 6692768 Nov 13 20:11 /usr/local/aws-cli/v2/2.13.34/dist/aws
```

Errors I had: 
```
==> amazon-ebs: Provisioning with shell script: /work/amazon-eks-ami/scripts/generate-version-info.sh
    amazon-ebs: /home/ec2-user/script_5180.sh: line 19: /usr/bin/kubelet: Permission denied
==> amazon-ebs: Provisioning step had errors: Running the cleanup provisioner, if present...
```

**Description of changes:**
Add 'sudo' in front of both commands, which will avoid errors listed above

**Testing Done**

Tested with UMASK 027 images with 'sudo', worked as expected.
Tested with vanilla AWS AL2 latest image, worked as expected.

### DIFF
--- a/scripts/generate-version-info.sh
+++ b/scripts/generate-version-info.sh
@@ -16,14 +16,14 @@ OUTPUT_FILE="$1"
 sudo rpm --query --all --queryformat '\{"%{NAME}": "%{VERSION}-%{RELEASE}"\}\n' | jq --slurp --sort-keys 'add | {packages:(.)}' > "$OUTPUT_FILE"
 
 # binaries
-KUBELET_VERSION=$(kubelet --version | awk '{print $2}')
+KUBELET_VERSION=$(sudo kubelet --version | awk '{print $2}')
 if [ "$?" != 0 ]; then
   echo "unable to get kubelet version"
   exit 1
 fi
 echo $(jq ".binaries.kubelet = \"$KUBELET_VERSION\"" $OUTPUT_FILE) > $OUTPUT_FILE
 
-CLI_VERSION=$(aws --version | awk '{print $1}' | cut -d '/' -f 2)
+CLI_VERSION=$(sudo aws --version | awk '{print $1}' | cut -d '/' -f 2)
 if [ "$?" != 0 ]; then
   echo "unable to get aws cli version"
   exit 1


cartermckinnon:
Is a different `ssh_username` an option for you?

leowenlu:
No, the test I have done set  `"ssh_username": "ec2-user"`.
Best Regards,
*Leo Li*

leowenlu:
> Is a different `ssh_username` an option for you?

@cartermckinnon

As you could see that we are restricting /usr/ folder which is access only to 'root', see the folder permissions

`amazon-ebs: -rwxr-x---. 1 root root 120671264 Nov 7 09:14 /usr/bin/kubelet
amazon-ebs: -rwxr-x---. 1 root root 6692768 Nov 13 20:11 /usr/local/aws-cli/v2/2.13.34/dist/aws`

By adding `sudo`, it will solve my issue, also I believe adding `sudo` will not introduce any harm to the existing code at all.

leowenlu:
@cartermckinnon 
looks like there is another case:
https://github.com/awslabs/amazon-eks-ami/commit/967fb36528818af12a9cf2dcfc4097dec2e1e2b3

cartermckinnon:
> As you could see that we are restricting /usr/ folder which is access only to 'root', see the folder permissions

Sure; my suggestion is to just build the AMI as `root`. Set the `username` to `root` in your Packer template.

leowenlu:
Hi @cartermckinnon 

Thanks for your reply, and the suggestion is valid, if we have to build the AMI as `root`, it's kind of missing the point that we could specify a user to back AMI unless only `root` will be able to back the AMI in the road map which is the the way we are heading to.

I am also curious what harm it will introduce if we add 'sudo' ?

CC @varun-armis

PR #1528, opened 2023-11-24
Miscellaneous fixes from AL2023 testing

dims:
- Add `"` around `B64_CLUSTER_CA` so a multi-line value can be saved correctly into the file
- In AL2 we had `ec2-net-utils` which is now named `amazon-ec2-net-utils` so handle them separately/explicitly (so it's obvious!)
-  Patch `MACAddressPolicy` which defaults to `persistent` in AL2023 and [causes issues](https://github.com/aws/amazon-vpc-cni-k8s/pull/2118)

### DIFF
--- a/scripts/install-worker.sh
+++ b/scripts/install-worker.sh
@@ -88,13 +88,19 @@ fi
 if cat /etc/*release | grep "al2023" > /dev/null 2>&1; then
   # exists in al2023 only (needed by kubelet)
   sudo yum install -y iptables-legacy
+
+  # Remove the amazon-ec2-net-utils package, if it's installed. This package interferes with the route setup on the instance.
+  if yum list installed | grep amazon-ec2-net-utils; then sudo yum remove amazon-ec2-net-utils -y -q; fi
+
+  # Temporary fix for https://github.com/aws/amazon-vpc-cni-k8s/pull/2118
+  sed -i "s/^MACAddressPolicy=.*/MACAddressPolicy=none/" /usr/lib/systemd/network/99-default.link
 else
   # curl-minimal already exists in al2023 so install curl only on al2
   sudo yum install -y curl
-fi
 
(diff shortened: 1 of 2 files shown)


jayanthvn (approved):

scripts/install-worker.sh
```diff
@@ -88,13 +88,19 @@ fi
 if cat /etc/*release | grep "al2023" > /dev/null 2>&1; then
   # exists in al2023 only (needed by kubelet)
   sudo yum install -y iptables-legacy
+
+  # Remove the amazon-ec2-net-utils package, if it's installed. This package interferes with the route setup on the instance.
+  if yum list installed | grep amazon-ec2-net-utils; then sudo yum remove amazon-ec2-net-utils -y -q; fi
```
bryantbiggs, line 93: Is this still valid on AL2023 or is it just a carryover from AL2? Based on this statement in the project https://github.com/amazonlinux/amazon-ec2-net-utils#usage - have we reached out to the EC2 net utils project/team to figure out the proper solution instead of simply removing the package?

dims: @bryantbiggs not doing it idly :) been struggling with this one for 2 days now and finally figured out that the package name was different! https://testgrid.k8s.io/amazon-ec2#Conformance%20-%20EC2/EKS/AL2023%20-%20master&width=20

jayanthvn: This https://github.com/awslabs/amazon-eks-ami/pull/368#issuecomment-788417027 has the TL;DR on why `ec2-net-utils` is not compatible with aws-vpc-cni..

dims: @jayanthvn @bryantbiggs opened an issue on the other side: https://github.com/amazonlinux/amazon-ec2-net-utils/issues/100

PR #1529, opened 2023-11-25
fix Permission denied for 99-default.link

dims:
Let's try `sudo` and also throw in `|| true` for good measure!

```
2023-11-25T04:27:26Z:     amazon-ebs: sed: couldn't open temporary file /usr/lib/systemd/network/sedelA3MU: Permission denied
2023-11-25T04:27:26Z: ==> amazon-ebs: Provisioning step had errors: Running the cleanup provisioner, if present...
```


cartermckinnon:
@dims if you set your awslabs org membership to public, you can use the CI commands: https://github.com/orgs/awslabs/people?query=dims

change Lgtm!

/ci

cartermckinnon (approved):

dims:
> @dims if you set your awslabs org membership to public, you can use the CI commands: https://github.com/orgs/awslabs/people?query=dims

TIL!!! i didn't know i had it private thanks Carter

PR #1531, opened 2023-11-28
Install SSM agent from AL core repo by default

cartermckinnon:
**Description of changes:**

This changes the default installation method for `amazon-ssm-agent` back to the AL2 core repository, instead of the S3 bucket provided by the SSM agent project. This ensures that the version installed by default in the EKS AMI is compliant with any security bulletins issued by Amazon Linux.

### DIFF
--- a/doc/USER_GUIDE.md
+++ b/doc/USER_GUIDE.md
@@ -58,7 +58,7 @@ Users have the following options for specifying their own values:
 | `source_ami_owners` | ```137112412989``` |  |
 | `ssh_interface` | `""` |  |
 | `ssh_username` | ```ec2-user``` |  |
-| `ssm_agent_version` | ```latest``` |  |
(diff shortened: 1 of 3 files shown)


mmerkes (approved):
LGTM

Issacwww:
Build failed due to validation of SSM_AGENT_VERSION, should we remove it?

cartermckinnon:
Yeah not sure what's up with that -- we use the same approach for `kernel_version` and it works fine. Looking into it.

Issacwww:
guess this line might be the problem. https://github.com/awslabs/amazon-eks-ami/blob/master/scripts/install-worker.sh#L35

cartermckinnon:
Yep, removed it. Forgot that was a thing, we should probably get rid of that altogether at some point.

cartermckinnon:
Looks good in CI:
```
2023-11-29T22:26:05Z:     amazon-ebs: Installing amazon-ssm-agent from AL core repository
```

PR #1544, opened 2023-12-13
Enable graceful node shutdown for 1.29+

cartermckinnon:
**Description of changes:**

Configures a node shutdown period of 45 seconds, with the last 15 seconds reserved for critical pods, for Kubernetes versions 1.29 and above.

### DIFF
--- a/files/bootstrap.sh
+++ b/files/bootstrap.sh
@@ -482,6 +482,11 @@ fi
 KUBELET_CONFIG=/etc/kubernetes/kubelet/kubelet-config.json
 echo "$(jq ".clusterDNS=[\"$DNS_CLUSTER_IP\"]" $KUBELET_CONFIG)" > $KUBELET_CONFIG
 
+if vercmp "$KUBELET_VERSION" gteq "1.29.0"; then
+  echo "$(jq ".shutdownGracePeriod=\"45s\"" $KUBELET_CONFIG)" > $KUBELET_CONFIG
+  echo "$(jq ".shutdownGracePeriodCriticalPods=\"15s\"" $KUBELET_CONFIG)" > $KUBELET_CONFIG
+fi
+
 if [[ "${IP_FAMILY}" == "ipv4" ]]; then
   INTERNAL_IP=$(imds 'latest/meta-data/local-ipv4')
 else


cartermckinnon:
CI is broken after Packer finally pulled the plug on builtin plugins. Should be fixed by #1545 .

cartermckinnon:
/ci

let's see if that did it...

cartermckinnon:
In my tests, this config didn't seem sufficient to actually block shutdown. I think we need to increase logind's `InhibitDelayMaxSec` for this to have any effect. Kubelet will attempt to increase it itself, but maybe this doesn't work on AL2?

youwalther65:
@cartermckinnon EBS CSI FAQ troubleshooting section [here](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/docs/faq.md#what-steps-can-be-taken-to-mitigate-this-issue) uses `InhibitDelayMaxSec` as well and following it I got: 
```
# systemd-inhibit --list
     Who: kubelet (UID 0/root, PID 1011931/kubelet)
    What: shutdown
     Why: Kubelet needs time to handle node shutdown
    Mode: delay

1 inhibitors listed.
```

cartermckinnon:
/ci
+workflow:os_distros al2

cartermckinnon:
/ci
+workflow:os_distros al2

cartermckinnon:
/ci
+workflow:os_distros al2

cartermckinnon:
/ci
+workflow:os_distros al2
+workflow:k8s_versions 1.30

youwalther65:
> > But I still wonder how Karpenter or MNG node drain plays together with kubelet graceful shutdown.
> 
> MNG and Karpenter will evict all the pods from a node before shutting it down, so you don't really need kubelet's graceful shutdown feature by the time the shutdown signal arrives (though some daemonsets could still benefit from it). The graceful shutdown feature is intended to honor your pods' own termination grace periods and cleanly shut them down when e.g. there is no orchestrator handling that for you (such as when the shutdown is unexpected).

Agree, especially DaemonSets like EBS CSI would benefit from kubelet graceful node shutdown and the team even recommends it in its [FAQ](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/docs/faq.md#what-steps-can-be-taken-to-mitigate-this-issue)

cartermckinnon:
/ci
+workflow:k8s_versions 1.30

cartermckinnon:
Continued in #2390.

PR #1545, opened 2023-12-13
Install amazon packer plugin for CI

cartermckinnon:
**Description of changes:**

Newer versions of Packer don't ship with plugins anymore, so we have to install it ourselves.


mmerkes (approved):
LGTM

PR #1546, opened 2023-12-15
reorganize with operating system and k8s version dimensions

ndbaker1:
**Description of changes:**

* organize packer templates based on operating system base, and k8s versions.
  * each os distro has a template, and each k8s version has a set of variable defaults 
* moved and renamed source files  `src/`
* moved misc Makefile logic into wrapper scripts and updated relative paths in packer templates to provisioning scripts
* reads the `kubernetes_version` and `os_distro` in the Makefile to pull the correct packer template and default variables from the `templates` file-tree.

file-tree view:
```
src/templates/
├── al2
│   ├── 1.23
│   │   └── variables.json
│   ├── 1.24
│   │   └── variables.json
│   ├── 1.25
│   │   └── variables.json
│   ├── 1.26
│   │   └── variables.json
│   ├── 1.27
│   │   └── variables.json
│   ├── 1.28
│   │   └── variables.json
│   └── template.json
└── al2023
    ├── 1.23
    │   └── variables.json
    ├── 1.24
    │   └── variables.json
    ├── 1.25
    │   └── variables.json
    ├── 1.26
    │   └── variables.json
    ├── 1.27
    │   └── variables.json
    ├── 1.28
    │   └── variables.json
    └── template.json
```

**Testing Done**

```
make test
```

built all versions 1.23 - 1.28 for al2 and al2023

```
# eg:
make 1.28 os_distro=al2023
```


ndbaker1:
gpu variants will likely come into this hierarchy soon enough, which will produce a dimension looking like `cpu|nvidia|tranium`.

cartermckinnon (changes requested):
Few high level things:
1. Can we rename the `packer` directory to something like `src` or `templates`?
2. This is a good opportunity to rename the JSON files, I've never loved the `eks-worker` prefix. They could just  be `template.json` and `defaults.json` or something.
3. I think some of the Makefile gymastics could be moved down into scripts, we're accumulating a lot of esoteric stuff in there.
4. I don't love `packer-gen.sh`. If it's just something used during dev to do some code gen, I think that's okay; but I'd prefer to avoid generating stuff at runtime. Also maybe could be 2 scripts, which more descriptive names?

ndbaker1:
1. `templates` sgtm 👍 
2. agreed, trying out `template.json` and `variables.json`
3. a few changes and cleanups here. one catch on how we need to know the `kubernetes_version` to pickup the right template & variables, but if a user passes these through a packer var-file then the makefile will not pick it up, so I checked for that case explicitly at the start of the Makefile.
4. moved to a full template + variables at the leaf of each dimension/file-tree since it better reflects how versioned branches would look

also not worrying about the gpu variant topic yet, since the main differences are installed packages, and we can navigate that by passing flags to provisioners

Makefile
```diff
@@ -1,66 +1,53 @@
...
-vercmp = $(shell $(MAKEFILE_DIR)/files/bin/vercmp "$1" "$2" "$3")
-
-# Docker is not present on 1.25+ AMI's
-# TODO: remove this when 1.24 reaches EOL
-ifeq ($(call vercmp,$(kubernetes_version),gteq,1.25.0), true)
-	# do not tag the AMI with the Docker version
-	docker_version ?= none
-	# do not include the Docker version in the AMI description
-	ami_component_description ?= (k8s: {{ user `kubernetes_version` }}, containerd: {{ user `containerd_version` }})
+al_variant ?= al2
```
cartermckinnon, line 15: I think this variable could be a bit more generic, such as `OS_DISTRO`. We could feasibly add `ubuntu22.04`, `rhel9`, etc.  or something that isn't Amazon Linux in the future.

Makefile
```diff
@@ -1,66 +1,53 @@
...
+VALID_AL_VARIANTS := al2 al2023
+ifneq ($(filter $(al_variant),$(VALID_AL_VARIANTS)), $(al_variant))
+	$(error Invalid al_variant: $(al_variant). Must be one of [$(VALID_AL_VARIANTS)])
```
cartermckinnon, line 18: You could just bail if you don't find a directory called `templates/$OS_DISTRO`, instead of maintaining this list of valid values.

Makefile
```diff
@@ -1,66 +1,53 @@
...
-ifeq ($(aws_region), us-gov-west-1)
+else ifeq ($(aws_region), us-gov-west-1)
 	source_ami_owners ?= 045324592363
 endif
 
-T_RED := \e[0;31m
-T_GREEN := \e[0;32m
-T_YELLOW := \e[0;33m
-T_RESET := \e[0m
-
```
cartermckinnon, line 58: yes, *please* get rid of this 😂

Makefile
```diff
@@ -149,6 +131,7 @@ lint-docs: ## Lint the docs
 clean:
 	rm *-manifest.json
 	rm *-version-info.json
+	rm -rf .staging/
```
cartermckinnon, line 134: is this still needed?

cartermckinnon (changes requested):
I think we need to move `scripts` and `files` under a `src` directory (and pick better names).

e.g.
```
.
└── src
    ├── boot | runtime-resources | etc.
    ├── build | build-scripts | etc.
    └── templates
```

cartermckinnon:
I think we should also consider merging this PR to a new `main` branch. We've heard in the past that some folks grab random scripts/resource files out of this source tree, so it'd prevent breakage if we just used the `master` -> `main` rename to leave `master` as-is. We'd update the `README` on `master` to say it's deprecated, development has continued on `main`, etc.

@mmerkes WDYT?

hack/generate-template-variable-doc.py
```diff
@@ -8,11 +8,11 @@
 os.chdir(os.path.dirname(whereami))
 
 template = {}
-with open('../eks-worker-al2.json') as template_file:
+with open('../templates/al2/1.28/template.json') as template_file:
     template = json.load(template_file)
 
 default_vars = {}
-with open('../eks-worker-al2-variables.json') as default_var_file:
+with open('../templates/al2/1.28/variables.json') as default_var_file:
```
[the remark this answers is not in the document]
cartermckinnon: Do we actually need separate templates for each k8s version? Seems like we could have a single `template.json` for each distro, and then default variable files for each k8s version?

Makefile
cartermckinnon: I mean, Amazon Linux *is* an OS 😉 We could use `al_version`, but then I'd expect values like `2` and `2023`. Maybe `base_distro`?

doc/USER_GUIDE.md
```diff
@@ -22,7 +22,7 @@ Users have the following options for specifying their own values:
 | Variable | Default value | Description |
 | - | - | - |
 | `additional_yum_repos` | `""` |  |
-| `ami_component_description` | ```(k8s: {{ user `kubernetes_version` }}, docker: {{ user `docker_version` }}, containerd: {{ user `containerd_version` }})``` |  |
+| `ami_component_description` | ```(k8s: {{ user `kubernetes_version` }}, containerd: {{ user `containerd_version` }})``` |  |
```
Issacwww, line 25: How this get updated for 1.25 and below versions?

hack/generate-template-variable-doc.py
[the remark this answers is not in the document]
ndbaker1: would we just use the latest k8s for the default variables, or omit the defaults entirely? relating to https://github.com/awslabs/amazon-eks-ami/pull/1546#discussion_r1440810649

Makefile
ndbaker1: `os_distro` has grown on me the more i fiddle with names 🤔, ill try rolling with it

doc/USER_GUIDE.md
ndbaker1: using a pin to 1.28 for now, so that following https://github.com/awslabs/amazon-eks-ami/pull/1546#discussion_r1440703896 at least stably produces one table per variant

ndbaker1:
> I think we need to move scripts and files under a src directory (and pick better names).

I bit the bullet and did all the moving/renaming necessary for this. let me know if the names still need changin

ndbaker1:
> I think we should also consider merging this PR to a new `main` branch

i have the github references updates to `main` already, so we can go through with this instead of merging this once its all approved

cartermckinnon:
I think this is close. Going to see how badly the CI breaks :wink: 

/ci

doc/USER_GUIDE.md
cartermckinnon: I think omitting the defaults make sense. We can just add a note that the default depends on the Kubernetes version, and can call out notable k8s version boundaries  in the `Description` column where it matters.

Makefile
Issacwww, line 1: Like the clean up in this file! Thanks!

Issacwww (approved):

cartermckinnon:
We discussed organizing runtime resources (at least somewhat) similarly to their final placement in the AMI's filesystem; do we still want to do that?

e.g.
```
.
└── runtime
    ├── bin
    └── etc
        ├── eks
        │   └── eni-max-pods.txt
        └── systemd
            └── system
                └── containerd.service.d
```

We don't have to go too far in this direction, but some kind of structure seems preferable to what we've had (a junk drawer)

ndbaker1:
I still think its preferable to organize the `runtime` structure that to match the filesystem and can be something we do as a next step.

> Similarly, how will the runtime directory look when we merge in the new bootstrap impl?

originally I had a new set of `runtime` files live with the bootstrap code

PR #1547, opened 2023-12-16
Fix typo opt names in `bootstrap.sh` logging

ketozhang:
```
log "INFO: --cluster-ca or --api-server-endpoint is not defined, describing cluster..."
```

Fixes these two opts as it doesn't exist.


cartermckinnon:
Thanks @ketozhang !

cartermckinnon (approved):

PR #1553, opened 2023-12-23
[WIP] Optionally require IMDS v2 on instances launched from AMI

xdu31:
**Issue #, if available:**

**Description of changes:**

Enable IMDS v2

There is a bug in packer,  IMDS v2 is not enabled in the AMI produced as documented with the option of:
```
"http_tokens": "required",
```
See more details in https://github.com/hashicorp/packer-plugin-amazon/issues/357
As indicated in the issue, the solution is:
1) explicitly enable IMDS v2 with `"imds_support": "v2.0",`
2) this requires the installation of packer 1.8.6+
3) this also requires the installation of packer amazon-ebs plugin 1.2.7+ 

It is noted that the Packer team has moved away from bundling updated plugins with Packer since 1.8.x in favor of using packer plugins install for JSON users. This was called out in the [1.9.2 Packer release](https://github.com/hashicorp/packer/blob/main/CHANGELOG.md#notes-1)

**Testing Done**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -102,8 +102,9 @@ lint: lint-docs ## Check the source files for syntax and format issues
 test: ## run the test-harness
 	test/test-harness.sh
 
-# include only variables which have a defined value
-PACKER_VARIABLES := $(foreach packerVar,$(AVAILABLE_PACKER_VARIABLES),$(if $($(packerVar)),$(packerVar)))
+# include only variables which have a defined value or the variable is imds_support
+PACKER_IMDS_SUPPORT_VAR := imds_support
+PACKER_VARIABLES := $(foreach packerVar,$(AVAILABLE_PACKER_VARIABLES),$(if $(or $($(packerVar)), $(filter $(packerVar),$(PACKER_IMDS_SUPPORT_VAR))),$(packerVar)))
 PACKER_VAR_FLAGS := -var-file $(PACKER_DEFAULT_VARIABLE_FILE) \
 $(if $(PACKER_VARIABLE_FILE),-var-file=$(PACKER_VARIABLE_FILE),) \
 $(foreach packerVar,$(PACKER_VARIABLES),-var $(packerVar)='$($(packerVar))')
--- a/Makefile
+++ b/Makefile
@@ -102,9 +102,8 @@ lint: lint-docs ## Check the source files for syntax and format issues
 test: ## run the test-harness
 	test/test-harness.sh
 
-# include only variables which have a defined value or the variable is imds_support
-PACKER_IMDS_SUPPORT_VAR := imds_support
-PACKER_VARIABLES := $(foreach packerVar,$(AVAILABLE_PACKER_VARIABLES),$(if $(or $($(packerVar)), $(filter $(packerVar),$(PACKER_IMDS_SUPPORT_VAR))),$(packerVar)))
+# include only variables which have a defined value
+PACKER_VARIABLES := $(foreach packerVar,$(AVAILABLE_PACKER_VARIABLES),$(if $($(packerVar)),$(packerVar)))
 PACKER_VAR_FLAGS := -var-file $(PACKER_DEFAULT_VARIABLE_FILE) \
 $(if $(PACKER_VARIABLE_FILE),-var-file=$(PACKER_VARIABLE_FILE),) \
 $(foreach packerVar,$(PACKER_VARIABLES),-var $(packerVar)='$($(packerVar))')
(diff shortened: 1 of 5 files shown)


cartermckinnon:
This is a breaking change for any customer depending on IMDSv1, so we've put off doing it for a while. I'll discuss with the team.

cc: @mmerkes @ravisinha0506

mmerkes:
There are plans to eventually make this change, but it is a breaking change. Our product team is working to communicate this to customers and set a timeline.

In the meantime, I'd either pass on this PR or set it behind a `require-imds-v2` flag so that customers using custom AMIs can require it, if they choose.

xdu31:
> There are plans to eventually make this change, but it is a breaking change. Our product team is working to communicate this to customers and set a timeline.
> 
> In the meantime, I'd either pass on this PR or set it behind a `require-imds-v2` flag so that customers using custom AMIs can require it, if they choose.

@cartermckinnon @mmerkes I added a new variable `imds_support` so the following cmd can build with IMDSv2:
```
make build kubernetes_version=1.23.9 kubernetes_build_date=2022-07-27 arch=x86_64 imds_support_version="v2.0"
```
and the original cmd is still building with IMDSv1.

Makefile
```diff
@@ -102,8 +102,9 @@ lint: lint-docs ## Check the source files for syntax and format issues
 test: ## run the test-harness
 	test/test-harness.sh
 
-# include only variables which have a defined value
-PACKER_VARIABLES := $(foreach packerVar,$(AVAILABLE_PACKER_VARIABLES),$(if $($(packerVar)),$(packerVar)))
+# include only variables which have a defined value or the variable is imds_support
+PACKER_IMDS_SUPPORT_VAR := imds_support
```
cartermckinnon, line 106: What's the intention of this?

doc/USER_GUIDE.md
```diff
@@ -107,6 +108,14 @@ aws s3 ls s3://amazon-eks/1.23.9/2022-07-27/bin/linux/x86_64/
 
 To build using the example binaries above:
 ```bash
+# Using IMDS v2
+make k8s \
+  kubernetes_version=1.23.9 \
+  kubernetes_build_date=2022-07-27 \
+  arch=x86_64 \
+  imds_support='v2.0'
```
cartermckinnon, line 116: I don't think we need these examples, the variable doc table should be sufficient.

eks-worker-al2.json
```diff
@@ -42,7 +42,8 @@
     "subnet_id": null,
     "temporary_security_group_source_cidrs": null,
     "volume_type": null,
-    "working_dir": null
+    "working_dir": null,
+    "imds_support": null
```
cartermckinnon, line 46: Please insert this in alphabetic order

cartermckinnon (changes requested):

Makefile
xdu31: > What's the intention of this?

The valid values for `imds_supports` are "v2" and "".

Currently, the behavior is, when a packer value is assigned with "" (empty value), it is a noop.

However, we want to change this behavior for `imds_support`, in case of `imds_support=""`, it should be treated as a valid assignment to disable IMDSv2 on the AMI.

cartermckinnon: Did you try it without this extra logic in the Makefile? I think it would work the way you intend. If the template variable has the value of `""`, then `{{user `imds_support`}}` is going to resolve to the empty string as expected. We set the default value of `""` in our default variable file. A user doesn't need to additionally define `imds_support` in their `make` command or in their own variable file, because our default would apply.

xdu31: @cartermckinnon 
> If a user wants to leave IMDSv1 enabled, they don't need to define imds_support in their make command or in their own variable file, because our default would apply.
This is correct

However, imagine that we will make IMDSv2 enabled by default in the future, and the customer wants to use IMDSv1 for whatever reason, it won't work without this change.
When it sees `make k8s imds_support=''`, it won't have the effect to set to IMDSv1, it will just ignore this, it is a noop (`''` and `'v2.0'` are the only 2 valid values for `imds_support`).

cartermckinnon: I think we’re unlikely to need that flexibility. We don’t ever have to change the default, we could just always pass v2.0 for official builds. After a while, we can remove the variable altogether and hardcode v2.0 in the template.

xdu31: @cartermckinnon Makefile change reverted

eks-worker-al2-variables.json
```diff
@@ -35,5 +35,6 @@
     "subnet_id": "",
     "temporary_security_group_source_cidrs": "",
     "volume_type": "gp2",
-    "working_dir": "{{user `remote_folder`}}/worker"
+    "working_dir": "{{user `remote_folder`}}/worker",
+    "imds_support": ""
```
cartermckinnon, line 39: Last nit: can you sort this file?

xdu31: > Last nit: can you sort this file?

done

cartermckinnon:
This is the default for AL2023, but we don't intend to change this on AL2 today. We can always reopen this PR if/when that changes.

PR #1561, opened 2024-01-10
Make k8s-specific variable overrides optional

cartermckinnon:
**Description of changes:**

Few of the variables actually differ between k8s versions, so this only utilizes a k8s-specific variables file when necessary, which will be merged with the default variable values.


ndbaker1 (approved):
nice! much cleaner than the way i was trying to descend the tree and find variables in the first version of the reorg 😅

PR #1563, opened 2024-01-10
Share only certain provisioners across templates

cartermckinnon:
**Description of changes:**

Moves the provisioner scripts and runtime resources into the template directories, and removes all the template-specific conditionals as a result.

Also gets rid of some cruft from the `al2023` template as a result (the additional YUM repo thing, which I don't think anyone actually uses, the `kernel_version` template parameter, any mention of Docker, etc.).

Also creates a `shared` template directory for resources that can actually be shared between multiple templates.


src/templates/al2023/runtime/bin/mount-bpf-fs
ndbaker1, line 1: this type of script also probably falls in the `shared` category?

src/templates/al2023/runtime/docker-daemon.json
ndbaker1, line 1: should also delete

ndbaker1:
i like this setup, we just need to organize more of the shared runtime files. also very happy that it flattens the paths a little :)

once it goes through i can cleanup the al2023 `runtime` for the new bootstrap.

cartermckinnon:
@ndbaker1 Yup I expect `src/templates/al2023/runtime` to just contain `nodeadm`, but I wanted to keep this branch buildable for now.

src/templates/al2023/runtime/bin/mount-bpf-fs
cartermckinnon: Some of these little utilities could be in `templates/shared/runtime`, yeah; but this specific one is only necessary on AL2. The bpffs is mounted by default with newer systemd versions

src/templates/shared/runtime/eni-max-pods.txt
ndbaker1, line 1: theres some extra housework like updating the sync ci job to recognize this path, but we can always make more doing cleanup PRs

ndbaker1 (approved):

src/templates/shared/runtime/eni-max-pods.txt
cartermckinnon: ack! We'll have plenty of things to tweak under `.github`, the CI flow needs an overhaul to capture AL2023 variants as well

PR #1572, opened 2024-01-13
Speed up nodeadm build

cartermckinnon:
**Description of changes:**

Skips codegen, `go vet`, etc. for the `build` target, which can be used during AMI build to speed things up.

With:
```
make k8s=1.28 os_distro=al2023
```

Before:
```
==> Wait completed after 15 minutes 36 seconds
```

After:
```
==> Wait completed after 9 minutes 35 seconds
```

Additional miscellany:
- Removed the `git`-based version string from the CLI. This doesn't have it's own versioning or release process, so isn't necessary (and doesn't work in the kaniko build approach anyway).


ndbaker1 (approved):
👍

PR #1573, opened 2024-01-16
Build nodeadm without Dockerfile

cartermckinnon:
**Description of changes:**

Uses the EKS-D golang image directly, instead of doing a nested build. Only downside is the `nodeadm` build steps aren't codified in a Dockerfile, but it's just `make build` anyway.

**Testing Done**

```
make os_distro=al2023 k8s=1.28
```

### DIFF
--- a/nodeadm/Dockerfile
+++ b/nodeadm/Dockerfile
@@ -1,3 +1,0 @@
-FROM public.ecr.aws/eks-distro-build-tooling/golang:1.21
-WORKDIR /workspace
-RUN make build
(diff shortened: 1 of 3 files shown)


ndbaker1 (changes requested):
cool this works, could we remove the `KANIKO_IMAGE` env var in the template too?

cartermckinnon:
@ndbaker1 yup, done

ndbaker1 (approved):
🚀

PR #1574, opened 2024-01-16
Fix nodeadm build on macOS

cartermckinnon:
**Description of changes:**

Adds no-op impl's for systemd funcs to hush the compiler on macOS.

TODO: the systemd abstraction is pretty leaky right now, the `kubelet` package is making calls like "write a systemd drop-in unit", which is weird.

**Testing Done**

Succeeds on macOS:
```
make
```


ndbaker1 (approved):
ack 🤣

PR #1576, opened 2024-01-16
convert yum to dnf for al2023

ndbaker1:
**Description of changes:**

update `yum` calls to `dnf` to stay up to date with al2023 practices

**Testing Done**

```
make k8s=1.28 os_distro=al2023
```


templates/al2023/provisioners/install-worker.sh
```diff
@@ -276,15 +274,15 @@ EOF
...
   if ! [[ -z "${SSM_AGENT_VERSION}" ]]; then
     echo "Installing amazon-ssm-agent@${SSM_AGENT_VERSION} from S3"
-    sudo yum install -y https://s3.${BINARY_BUCKET_REGION}.${S3_DOMAIN}/amazon-ssm-${BINARY_BUCKET_REGION}/${SSM_AGENT_VERSION}/linux_${ARCH}/amazon-ssm-agent.rpm
+    sudo dnf install -y https://s3.${BINARY_BUCKET_REGION}.${S3_DOMAIN}/amazon-ssm-${BINARY_BUCKET_REGION}/${SSM_AGENT_VERSION}/linux_${ARCH}/amazon-ssm-agent.rpm
   else
     echo "Installing amazon-ssm-agent from AL core repository"
-    sudo yum install -y amazon-ssm-agent
+    sudo dnf install -y amazon-ssm-agent
   fi
 fi
```
cartermckinnon, line 287: Does the AL2023 base already have the SSM agent?

ndbaker1: it does, but locked to [`3.2.1705.0-1.amzn2023`](https://docs.aws.amazon.com/linux/al2023/release-notes/all-packages-AL2023.3.html)

Issacwww: Is there any version diff between yum repo v.s. dnf repo?

Issacwww (approved):

templates/al2023/provisioners/install-worker.sh
ndbaker1: found that yum [already points to dnf for al2023](https://docs.aws.amazon.com/linux/al2023/ug/package-management.html), so there won't be a difference, just whatever al2023 vends

cartermckinnon: Yeah `yum` is just an alias for `dnf`. If the SSM agent is already present in the base AMI, we should just remove all this logic IMO. That can happen in a separate PR though

cartermckinnon (approved):

templates/al2023/provisioners/install-worker.sh
bryantbiggs: FYI - the minimal AMI does not have the SSM agent nor does it have the yum alias that points to dnf. so if you use the "full" AMI, you'll have the SSM agent and you can use `yum` commands, but on the minimal AMI you'll need to make some adjustments

ndbaker1: ah thanks for the catch, i should've been looking [here](https://docs.aws.amazon.com/linux/al2023/ug/image-comparison.html). in that case we'll keep the SSM logic as is currently

PR #1584, opened 2024-01-18
Add nodeadm usage, config example

cartermckinnon:
**Description of changes:**

Adds some more detail to the README, to aid in discussion with other projects in this AMI's orbit.

### DIFF
--- a/nodeadm/README.md
+++ b/nodeadm/README.md
@@ -32,7 +32,7 @@ spec:
     cidr: 10.100.0.0/16
 ```
 
-Typically, you'll provide this configuration in your EC2 instance's user data, within a MIME multi-part document:
+You'll typically provide this configuration in your EC2 instance's user data, either as-is or embedded within a MIME multi-part document:
 ```
 Content-Type: multipart/mixed; boundary="BOUNDARY"
 MIME-Version: 1.0
--- a/nodeadm/README.md
+++ b/nodeadm/README.md
@@ -1,3 +1,52 @@
 # nodeadm
 
-nodeadm is a bootstrap script for EKS worker nodes
+Initializes a node in an EKS cluster.
+
+---
+
+## Usage
+
+To initialize a node:
+```
+nodeadm init
+```
+
+**Note** that this happens automatically, via a `systemd` service, on AL2023-based EKS AMI's.
+
+---
+
+## Configuration source
+
+`nodeadm` uses a YAML configuration schema that will look familiar to Kubernetes users.
+
+This is an example of the minimum required parameters:
+```yaml
+---
+apiVersion: node.eks.aws/v1alpha1
+kind: NodeConfig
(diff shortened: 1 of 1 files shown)


nodeadm/README.md
```diff
@@ -1,3 +1,52 @@
...
+Content-Type: multipart/mixed; boundary="BOUNDARY"
+MIME-Version: 1.0
+
+--BOUNDARY
+Content-Type: application/node.eks.aws
+MIME-Version: 1.0
+
+---
+apiVersion: node.eks.aws/v1alpha1
+kind: NodeConfig
```
bryantbiggs, line 46: how does this interface with managed nodegroups - are the details to join to the cluster contained in this same config, can multiple configs be provided and merged, etc?

Mainly - will this get around the current issues with managed nodegroups today https://github.com/awslabs/amazon-eks-ami/issues/844

nodeadm/README.md
```diff
@@ -1,3 +1,52 @@
...
+spec:
+  cluster:
+    name: my-cluster
+    apiServerEndpoint: https://example.com
+    certificateAuthority: Y2VydGlmaWNhdGVBdXRob3JpdHk=
+    cidr: 10.100.0.0/16
+```
+
+Typically, you'll provide this configuration in your EC2 instance's user data, within a MIME multi-part document:
```
ndbaker1, line 35: just a nit that i want to make it clear the user data doesn't *have* to be MIME multipart

```suggestion
You'll typically provide this configuration in your EC2 instance's user data, either as-is or embedded within a MIME multi-part document:
```

nodeadm/README.md
cartermckinnon: You're talking about passing this config object in a custom launch template, *without* specifying an AMI in the LT, right? i.e. a "custom LT" nodegroup instead of a "custom AMI" nodegroup?

We have the ability to implement that with this approach, yes; because managed nodegroups can add the important bits as necessary, instead of only tacking on a MIME part.

ndbaker1 (approved):

nodeadm/README.md
bryantbiggs: >  i.e. a "custom LT" nodegroup instead of a "custom AMI" nodegroup?

Correct

The most basic, contrived example would be a managed nodegroup using the `AL2_x86_64` AMI type (EKS AMI, not custom AMI), and the user provides a custom launch template where they want to somehow do the equivalent of `--local-disks raid0`. Today this is a bit of a challenge since MNGs will append the bootstrap script in the last MIME doc, so you have to find various hacky ways for that invocation to pick up this config.  So with either Karpenter or MNG, if there was a way to do something where users can pass in a config <somewhere> that does not conflict with the node joining the cluster logic, that would be ideal. something that is probably merged in with whatever defaults you all set

PR #1586, opened 2024-01-19
move image-credential-provider configuration to nodeadm

ndbaker1:
**Description of changes:**

the `image-credential-provider` related kubelet flags are currently in the worker installation script, so i moved them back into the bootstrap/nodeadm since the properties depend on kubelet version (and is considerably part of the kubelet config)

however, the install/download of the credential provider binary (`ecr-credential-provider`) still should belong in the install script, since it is a static asset.

I configured the default bin path pointers in `nodeadm` to match what is in `install-worker.sh`, and for extensibility i made this passable via an environment variable 

also, collapse some `kubeconfig.go` logic

**Testing Done**

```
make test && make test-e2e
```

also launched nodegroup with the built amis for v1.28

### DIFF
--- a/nodeadm/internal/kubelet/kubeconfig.go
+++ b/nodeadm/internal/kubelet/kubeconfig.go
@@ -30,15 +30,18 @@ func (k *kubelet) writeKubeconfig(cfg *api.NodeConfig) error {
 	if err != nil {
 		return err
 	}
+	if err := os.MkdirAll(kubeconfigRoot, kubeconfigPerm); err != nil {
+		return err
+	}
 	if enabled := cfg.Spec.Cluster.EnableOutpost; enabled != nil && *enabled {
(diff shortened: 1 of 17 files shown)


nodeadm/internal/kubelet/kubeconfig.go
```diff
@@ -68,10 +71,3 @@ func generateKubeconfig(cfg *api.NodeConfig) ([]byte, error) {
 	}
 	return buf.Bytes(), nil
 }
-
```
Issacwww, line 71: might worth move this to util is we have future changes that need to write files

ndbaker1: the extra calls are not too noisy, but it is pretty common so i can add a wrapper/helper

Issacwww (approved):

PR #1587, opened 2024-01-19
decouple kubelet environment from systemd drop-in 

ndbaker1:
**Description of changes:**

writes the kubelet environment to a file that is picked up by systemd unit rather than directly writing drop-ins

### DIFF
--- a/templates/al2023/runtime/rootfs/etc/systemd/system/kubelet.service
+++ b/templates/al2023/runtime/rootfs/etc/systemd/system/kubelet.service
@@ -6,6 +6,7 @@ Requires=containerd.service
 
 [Service]
 Slice=runtime.slice
+EnvironmentFile=/etc/eks/kubelet/environment
 ExecStartPre=/sbin/iptables -P FORWARD ACCEPT -w 5
 ExecStart=/usr/bin/kubelet $NODEADM_KUBELET_ARGS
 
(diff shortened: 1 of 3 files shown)


cartermckinnon:
This should be an option in the NodeConfig, we want to centralize this sort of thing

ndbaker1:
> This should be an option in the NodeConfig, we want to centralize this sort of thing

right, I came full circle with ensuring that users in custom-land can still add extra args, but there's still a multitude of ways to do it without an extra env 🤷

On the other hand i repurposed this PR to remove the tight environment coupling from kubelet to systemd (via drop-in) by using `EnvironmentFile` instead. lmk what you think @cartermckinnon @Issacwww

templates/al2023/runtime/rootfs/etc/systemd/system/kubelet.service
```diff
@@ -6,6 +6,7 @@ Requires=containerd.service
 
 [Service]
 Slice=runtime.slice
+EnvironmentFile=/etc/eks/kubelet-environment
```
cartermckinnon, line 9: Can we make this path /etc/eks/kubelet/environment?

ndbaker1: done 👍

cartermckinnon (approved):

PR #1590, opened 2024-01-23
add NodeConfig api support for node labels and taints

ndbaker1:
**Issue #, if available:**

**Description of changes:**

extend `NodeConfig` spec to include label and taint fields that pass through to kubelet `--node-labels` and `--register-with-taints` respectively.

**Testing Done**

add new e2e tests for labels and taints


nodeadm/api/v1alpha1/nodeconfig_types.go
```diff
@@ -29,8 +30,10 @@ type NodeConfigList struct {
 }
 
 type NodeConfigSpec struct {
-	Cluster      ClusterDetails  `json:"cluster,omitempty"`
-	FeatureGates map[string]bool `json:"featureGates,omitempty"`
+	Cluster      ClusterDetails    `json:"cluster,omitempty"`
+	Labels       map[string]string `json:"labels,omitempty"`
+	Taints       []v1.Taint        `json:"taints,omitempty"`
```
ndbaker1, line 35: these are applied by kubelet, but they are "node" configuration so not sure if they feel best here or in a `Spec.Kubelet.{Labels,Taints}`

cartermckinnon: I think it belongs in `Kubelet`

ndbaker1: updated 👍

nodeadm/internal/kubelet/config.go
```diff
@@ -268,6 +267,21 @@ func (k *kubelet) GenerateKubeletConfig(cfg *api.NodeConfig) (*kubeletSubConfig,
 		kubeletConfig.withDefaultReservedResources()
 	}
 
+	if len(cfg.Spec.Kubelet.Labels) > 0 {
+		var labelStrings []string
+		for labelKey, label := range cfg.Spec.Kubelet.Labels {
+			labelStrings = append(labelStrings, fmt.Sprintf("%s=%s", labelKey, label))
+		}
+		k.additionalArguments["node-labels"] = strings.Join(labelStrings, ",")
```
Issacwww, line 275: Would like to double confirm these keys are supported for all versions?
`node-labels` and `register-with-taints`

ndbaker1: checked [changelogs](https://github.com/kubernetes/kubernetes/tree/master/CHANGELOG) and only found a couple of relevant details:
* [1.23 : `--register-with-taints` option is now available via the Kubelet config file field registerWithTaints](https://github.com/kubernetes/kubernetes/blob/7bb00356f06332e63a9f06acd42f1bdd8fc559d6/CHANGELOG/CHANGELOG-1.23.md?plain=1#L2401)
  * seems safe to use for now in the case that a cluster beyond 1.23 is spun up
* no important changes for `--node-labels`

ndbaker1: however, does ^ incentivize us to put this in `Kubelet.Config.RegisterWithTaints` instead to keep on the edge of kubelet config adoption?

nodeadm/api/v1alpha1/nodeconfig_types.go
```diff
@@ -30,6 +31,7 @@ type NodeConfigList struct {
 
 type NodeConfigSpec struct {
 	Cluster      ClusterDetails  `json:"cluster,omitempty"`
+	Kubelet      KubeletOptions  `json:"kubelet,omitempty"`
 	FeatureGates map[string]bool `json:"featureGates,omitempty"`
```
ellistarn, line 35: Are feature gates scoped to our API, or is this kubelet feature gates?

nodeadm/internal/kubelet/config.go
ellistarn: Here's what we do in Karpenter, which has worked out very well: https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/apis/v1beta1/nodeclaim.go#L70

nodeadm/api/v1alpha1/nodeconfig_types.go
ndbaker1: yep its ours not kubelet's; to opt-in for a particular "preset" of the options

nodeadm/internal/kubelet/config.go
ndbaker1: think this is a good idea, and will treat the config as first class. i.e. write the `--register-with-taints` argument if necessary based on a version check. similar to sichao's [container runtime catch](https://github.com/awslabs/amazon-eks-ami/blob/al2023/nodeadm/internal/kubelet/config.go#L209)

nodeadm/api/v1alpha1/nodeconfig_types.go
ellistarn: Is there an example of how to configure this, or is it something we think well use in the future. If the latter, I'd recommend only introducing it when we need it. Sounds like the kind of thing that's design worthy in its own right.

ndbaker1:
some branch shenanigans happened while updating the scope of this PR, so redirecting to https://github.com/awslabs/amazon-eks-ami/pull/1593

PR #1593, opened 2024-01-24
support KubeletConfiguration and flag passthrough in NodeConfig 

ndbaker1:
**Issue #, if available:**

**Description of changes:**

supporting [KubeletConfiguration](https://pkg.go.dev/k8s.io/kubelet/config/v1beta1#KubeletConfiguration) parameters through the `NodeConfig` as pass through as well as the kubelet cli flags.

**Testing Done**

added e2e tests to cover the additional configuration surface area


cartermckinnon:
Chatted offline, I think we have to find a more generic way to accept kubelet config customizations. Clients like Karpenter want to pass a defined subset of the options, but clients like `eksctl` want to set arbitrary options: https://github.com/eksctl-io/eksctl/blob/9575570b554610129382d0a181645a3806cea98f/pkg/apis/eksctl.io/v1alpha5/types.go#L1247

We shouldn't need to make a code change here for every kubelet config option that every user wants to modify, it's toil for the maintainers and an artificial limitation for users.

I don't think we can do any meaningful validation of these options; when `nodeadm` runs, it's already too late to tell the user whether or not their kubelet config is valid. Kubelet will tell us itself. :)

ellistarn:
> We shouldn't need to make a code change here for every kubelet config option that every user wants to modify, it's toil for the maintainers and an artificial limitation for users.

I'm curious to interrogate this assumption. I think that there's value in knowing how customers are using your API, and intentionally constraining things. I think we can pretty trivially fine examples of kubeletconfiguration that would never be used.

cartermckinnon:
> knowing how customers are using your API

I think this data is valuable, but I think asking folks to cut an issue/PR here, get it merged, and wait for a release is a lot of friction just to permit them to set a field in the kubelet config file. If it were me, I'd just hack around it myself.

> I think we can pretty trivially fine examples of kubeletconfiguration that would never be used.

If the knob exists, someone wants (or perhaps less often, needs) to turn it. 😉

nodeadm/api/v1alpha1/nodeconfig_types.go
```diff
@@ -29,19 +30,31 @@ type NodeConfigList struct {
...
+	// Flags is a string map that takes command-line kubelet arguments without
+	// leading dashes. These arguments override any of generated defaults
+	// https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/
+	Flags map[string]string `json:"flags,omitempty"`
+	// Labels is a map of labels to apply to the node when the kubelet
+	// registers itself
+	Labels map[string]string `json:"labels,omitempty"`
+	// Labels is a map of labels to apply to the node when the kubelet
+	// registers itself
+	Taints []v1.Taint `json:"taints,omitempty"`
```
ndbaker1, line 59: Is this a premature addition of the `labels` and `taints` fields at the top level? they are technically configurable through `inline` config or `flags`, and i dont think we need to do any special validation with them.

Many users might expect `labels` and `taints` coming from the previous bootstrap.sh, but there's not a need to uphold that image

cartermckinnon: Managed nodegroups has passed these with the `--kubelet-extra-args` flag, so I think it's fine to let orchestrators use the generic kubelet options directly.

mmerkes:
> I think that there's value in knowing how customers are using your API, and intentionally constraining things

For the Karpenter use case, I think this makes sense because there's some inherent expectations that you can make. However, in the case of these AMIs as a whole, the use cases are pretty vast. I don't think that we can realistically decide what parameters are and are not useful to customers without just allowing a majority of them to be set regardless, and this code will be significantly less maintainable because any new (or unsupported) config that needs to be added will require a PR. Given, may be a minor PR, but will still need to be cut and then released. Some people will just hope to hack their user data with `jq` to get it done, which is less desirable. There will be undoubtedly be niche use cases where very few actually need to support something, but it may be very important to them.

I like the idea of making it easy for customers to configure the fields they want and easily adopt new config options.

nodeadm/api/v1alpha1/nodeconfig_types.go
ndbaker1: ill pull back on these for now, since an operator can still fully configure kubelet without them and the contracts for these via `--node-labels`, `--register-with-taints` isn't going to break anytime soon

nodeadm/internal/api/net.go
```diff
@@ -39,7 +40,7 @@ func GetIMDSInstanceDetails(ctx context.Context, imdsClient *imds.Client) (*Inst
 }
 
 // Derive the default ClusterIP of the kube-dns service from EKS built-in CoreDNS addon
-func (details *ClusterDetails) GetClusterDns() (string, error) {
+func GetClusterDns(details *v1alpha1.ClusterDetails) (string, error) {
```
cartermckinnon, line 43: What's the rationale for this? Seems like it should be working with the internal API types

nodeadm/internal/api/types.go
```diff
@@ -43,15 +39,6 @@ type InstanceDetails struct {
...
 
-type ClusterDetails struct {
-	Name                 string `json:"name,omitempty"`
-	APIServerEndpoint    string `json:"apiServerEndpoint,omitempty"`
-	CertificateAuthority []byte `json:"certificateAuthority,omitempty"`
-	CIDR                 string `json:"cidr,omitempty"`
-	EnableOutpost        *bool  `json:"enableOutpost,omitempty"`
-	ID                   string `json:"id,omitempty"`
-}
-
```
cartermckinnon, line 54: Ah, you'll have to add this back, the internal API should be separate from the public API types. The internal/api/bridge package maps multiple supported versions of the public API onto a single internal representation, that should be used everywhere. That package is the only place under internal/ where the public API types should be imported

nodeadm/internal/kubelet/environment.go
```diff
@@ -13,21 +13,28 @@ const (
...
-		fmt.Sprintf("%s=%s", kubeletArgsEnvironmentName, strings.Join(kubeletArgs, " ")),
+	// transform kubelet arguments into a string and write them to the kubelet
+	// environment variable
+	var kubeletFlags []string
+	for flag, value := range k.additionalArguments {
+		kubeletFlags = append(kubeletFlags, fmt.Sprintf("--%s=%s", flag, value))
 	}
+
+	var kubeletEnvironment []string
+	kubeletEnvironment = append(kubeletEnvironment, fmt.Sprintf("%s=%s", kubeletArgsEnvironmentName, strings.Join(kubeletFlags, " ")))
```
cartermckinnon, line 32: Does the value of this environment variable need to be quoted?

i.e.
```
KUBELET_ARGS=--foo --bar
```
or
```
KUBELET_ARGS='--foo --bar'
```

nodeadm/test/e2e/cases/kubeconfig/config.yaml
```diff
@@ -0,0 +1,11 @@
+---
+apiVersion: node.eks.aws/v1alpha1
+kind: NodeConfig
+metadata:
+  name: example
```
cartermckinnon, line 5: nit: this isn't necessary, right? we could omit this from all the test assets going forward

nodeadm/test/e2e/cases/kubelet-flags/config.yaml
```diff
@@ -0,0 +1,14 @@
...
+  name: example
+spec:
+  cluster:
+    name: my-cluster
+    apiServerEndpoint: https://example.com
+    certificateAuthority: Y2VydGlmaWNhdGVBdXRob3JpdHk=
+    cidr: 10.100.0.0/16
+  kubelet:
+    flags:
+      verbosity: '5'
```
cartermckinnon, line 14: This usage looks like of odd. Why not use the container args approach:
```
flags:
  - -v=7
  - --cloud-provider=external
  - --image-credential-provider-config=/foo/bar
```

cartermckinnon (changes requested):

nodeadm/internal/api/net.go
ndbaker1: i wanted to work with the external types as much as possible to avoid having too an excess of duplicated types, but wasn't sure if it was a good idea to move this to the api package or just swap the signature like so

nodeadm/internal/kubelet/environment.go
ndbaker1: didnt see any hard requirements for it to be quoted, let me triple check

nodeadm/internal/api/types.go
ndbaker1: oh gotcha, i saw other projects using public types in their impls and misinterpreted this, i can quickly flip them all back

nodeadm/test/e2e/cases/kubelet-flags/config.yaml
ndbaker1: The only real difference is that you are going to see all of the argument that nodeadm writes as well as the user provided arguments, when with the map you are actually overriding the value in-place; if users are more familiar with this approach then it makes sense to me. There shouldn't be a functional difference as long as the flags are appended to the end.

nodeadm/test/e2e/cases/kubeconfig/config.yaml
ndbaker1: true 👍

nodeadm/internal/kubelet/environment.go
ndbaker1: going to fix this to double quotes per https://github.com/markstos/systemd/blob/master/man/systemd.exec.xml#L345-L363

nodeadm/test/e2e/cases/kubelet-flags/config.yaml
cartermckinnon: The k8s components use spf13/pflag under the covers: https://github.com/kubernetes/component-base/blob/2065849162a4fdce3f03b9e255475ef0ef04390d/cli/flag/flags.go#L23

Which has the behavior that if a flag which only supports a single value is defined twice, the later value is used. So as long as what the user defines here follows our defaults (which is the case on AL2), it'll work as expected as far as overrides go.

nodeadm/test/e2e/cases/node-taints/run.sh
```diff
@@ -18,4 +18,8 @@ assert::file-contains /etc/eks/kubelet/environment "--register-with-taints=foo=:
 
 mock::kubelet 1.23.0
 nodeadm init --skip run --config-source file://config.yaml
-assert::json-files-equal /etc/kubernetes/kubelet/config.json expected-kubelet-config.json
+assert::file-contains /etc/eks/kubelet/environment "--register-with-taints=foo=:NoSchedule,foo2=baz:NoSchedule"
+
+mock::kubelet 1.28.0
+nodeadm init --skip run --config-source file://config.yaml
```
ellistarn, line 24: Would be sweet if we could do 
```
cat <<EOF | kubectl init -f -
... inline file
EOF
```

nodeadm/test/e2e/cases/kubelet-flags/config.yaml
```diff
@@ -0,0 +1,12 @@
+---
+apiVersion: node.eks.aws/v1alpha1
+kind: NodeConfig
+spec:
+  cluster:
+    name: my-cluster
+    apiServerEndpoint: https://example.com
+    certificateAuthority: Y2VydGlmaWNhdGVBdXRob3JpdHk=
+    cidr: 10.100.0.0/16
```
ellistarn, line 9: Does this cause NodeConfig to inject kubelet flags? What if they conflict with user defined flags?

ndbaker1: [carter just answered this one](https://github.com/awslabs/amazon-eks-ami/pull/1593#discussion_r1467083198) while we determined the input style; we will generate flags in nodeadm and then user-provided flags can override any of these

nodeadm/test/e2e/cases/node-taints/run.sh
ndbaker1: most users wont have a need to actually call `nodeadm` based on how its booted in the ami, but im a fan of the style

nodeadm/test/e2e/cases/kubelet-flags/config.yaml
ndbaker1: done, makes sense to use the typical command args approach

nodeadm/api/v1alpha1/nodeconfig_types.go
```diff
@@ -29,19 +29,26 @@ type NodeConfigList struct {
...
+	CIDR                 string `json:"cidr,omitempty"`
+	EnableOutpost        *bool  `json:"enableOutpost,omitempty"`
+	ID                   string `json:"id,omitempty"`
+}
+
+type KubeletOptions struct {
+	// InlineConfig is a raw document of a kubelet config that can be provided
+	// by the user to override default generated configurations
+	// https://kubernetes.io/docs/reference/config-api/kubelet-config.v1/
+	InlineConfig string `json:"inline,omitempty"`
```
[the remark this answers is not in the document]
ndbaker1: im going with `config` for the brevity 😬

cartermckinnon (approved):

PR #1594, opened 2024-01-24
Init change for scaning dependency vulnerability 

Issacwww:
**Issue #, if available:**

**Description of changes:**
- Add dependabot.yml for existing dependencies, [reference](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file)
- Add CI actions to scan dependency for PR, [reference](https://github.com/actions/dependency-review-action)

**Testing Done**


.github/workflows/ci-auto.yaml
```diff
@@ -38,3 +38,8 @@ jobs:
     steps:
     - uses: actions/checkout@v3
     - run: cd nodeadm && make test-e2e
+  dependency-review:
+    runs-on: ubuntu-latest
+    steps:
```
ndbaker1, line 43: do we need to [make the other steps dependent](https://docs.github.com/en/actions/using-jobs/using-jobs-in-a-workflow#defining-prerequisite-jobs) on this to prevent from running potentially bad code?

.github/dependabot.yml
```diff
@@ -0,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
+      interval: "weekly"
```
cartermckinnon, line 6: dependabot is incredibly annoying, I think it'd be better to do something like this: https://github.com/kubernetes/cloud-provider-aws/blob/master/.github/workflows/update-deps.yaml

.github/workflows/ci-auto.yaml
ndbaker1, line 43: maybe we could do the same thing that vpc-cni proj does with dependency review and govulcheck? https://github.com/aws/amazon-vpc-cni-k8s/blob/b0ff9b904a54a9bec276b7510a92d14ea6ab2d90/.github/workflows/deps.yml

seems like they would need to check the same boxes

.github/dependabot.yml
Issacwww: Seems vpc-cni also use the dependabot. https://github.com/aws/amazon-vpc-cni-k8s/blob/b0ff9b904a54a9bec276b7510a92d14ea6ab2d90/.github/dependabot.yml, when you say annoying, is it sending notification frequently?

while the workflow solution is already used in cloud-provider-aws repo, just double confirm the plugins used like peter-evans/create-pull-request, are approved by AppSec?

Issacwww:
Attempted to add gosec check as well, but it always fail with cannot import package error that would block the CI wf. thus remove it
https://github.com/awslabs/amazon-eks-ami/actions/runs/7677361909/job/20926010246

.github/workflows/update-dependency.yaml
```diff
@@ -0,0 +1,47 @@
...
+      pull-requests: write
+    steps:
+      - uses: actions/setup-go@4.0.0
+        with:
+          go-version: '1.21.5'
+      - uses: actions/checkout@v4
+      - name: Update Dependencies
+        id: update_deps
+        run: |
+          make update-deps
```
ndbaker1, line 27: we don't have this target yet right? yank from https://github.com/kubernetes/cloud-provider-aws/blob/master/hack/update-deps.sh?

Issacwww: Added!

hack/update-deps.sh
```diff
@@ -0,0 +1,36 @@
...
+go get $(go list -f '{{if not (or .Main .Indirect)}}{{.Path}}{{end}}' -mod=mod -m all)
+go mod tidy
+
+# Update dependencies in the e2e test directory
+if [ -d "${TARGET_DIR}/test/e2e" ]; then
+  cd "${TARGET_DIR}/test/e2e"
+  go mod tidy
+else
+  echo "e2e test directory does not exist"
+fi
```
ndbaker1, line 36: we dont have the same go e2e mods, so do we need this?

hack/update-deps.sh
```diff
@@ -0,0 +1,36 @@
...
+#!/bin/bash
+
+set -o errexit
+set -o pipefail
+
+# The first argument is the path, defaulting to the current directory if not provided
+TARGET_DIR="${1:-.}"
+
+# Update the main project dependencies
+cd "${TARGET_DIR}"
```
ndbaker1, line 26: feels weird, does it not work to just use `cd nodeadm; make update-deps` due to some Make assumptions about path?

hack/update-deps.sh
cartermckinnon, line 1: I think this could all go in the nodeadm Makefile? Shouldn't need to copy this script wholesale from k/cloud-provider-aws.

.github/workflows/update-dependency.yaml
```diff
@@ -0,0 +1,48 @@
...
+          commit-message: Update dependencies
+          committer: github-actions <actions@github.com>
+          author: github-actions <actions@github.com>
+          branch: dependencies/update
+          branch-suffix: timestamp
+          base: master
+          delete-branch: true
+          labels: |
+            ok-to-test
+            release-note-none
```
cartermckinnon, line 46: these labels aren't valid for our repo, we need to add `changelog/exclude` though.

cartermckinnon (changes requested):

cartermckinnon (approved):

PR #1596, opened 2024-01-29
handle pod-infra-container-image on older versions

ndbaker1:
**Issue #, if available:**

**Description of changes:**

some k8s versions will require the `--pod-infra-container-image` flag to avoid the sandbox image getting GC'd. [Kubelet options](https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/) state this will be removed in 1.27.

**Testing Done**

e2e test cases

### DIFF
--- a/nodeadm/internal/kubelet/config.go
+++ b/nodeadm/internal/kubelet/config.go
@@ -240,6 +240,35 @@ func (ksc *kubeletConfig) withDefaultReservedResources() {
 	ksc.KubeReservedCgroup = ptr.String("/runtime")
 }
 
+// The '--pod-infra-container-image' flags is added so that the sandbox image is
+// not garbage collected. There are several way in which we could remove this:
+//   - wait until a minimum supported version of kubernetes which implements the
+//     image pinning CRI support: https://github.com/kubernetes/kubernetes/pull/118544
+//   - update to containerd 2.0, which reworks the abstraction and no longer
+//     requires sandbox image
+func (ksc *kubeletConfig) withPodInfraContainerImage(cfg *api.NodeConfig, kubeletVersion string, flags map[string]string) error {
+	if semver.Compare(kubeletVersion, "v1.27.0") < 0 {
+		awsDomain, err := util.GetAwsDomain(context.TODO(), imds.New(imds.Options{}))
+		if err != nil {
+			return err
+		}
+		ecrUri, err := util.GetEcrUri(util.GetEcrUriRequest{
+			Region:    cfg.Status.Instance.Region,
+			Domain:    awsDomain,
+			AllowFips: true,
+		})
+		if err != nil {
+			return err
+		}
+		pauseContainerImage, err := util.GetPauseContainer(ecrUri)
+		if err != nil {
+			return err
+		}
+		flags["pod-infra-container-image"] = pauseContainerImage
+	}
+	return nil
+}
+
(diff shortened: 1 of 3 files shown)


nodeadm/internal/kubelet/config.go
```diff
@@ -240,6 +240,35 @@ func (ksc *kubeletConfig) withDefaultReservedResources() {
...
+			AllowFips: true,
+		})
+		if err != nil {
+			return err
+		}
+		pauseContainerImage, err := util.GetPauseContainer(ecrUri)
+		if err != nil {
+			return err
+		}
+		flags["pod-infra-container-image"] = pauseContainerImage
```
cartermckinnon, line 267: Wait, so we weren't passing this flag at all prior to this PR?

ndbaker1: we weren't :/

> The flag still exists in 1.29 AFAIK

hmm my bad based on the [PRs](https://github.com/kubernetes/kubernetes/pull/118544) the timeline for the support doesn't match the deprecation, so i'll make this the regular behavior

cartermckinnon:
@ndbaker1 do you intend to add the `if` statement back here to exclude the flag on 1.29?

ndbaker1:
> @ndbaker1 do you intend to add the `if` statement back here to exclude the flag on 1.29? If it's a no-op, I think it's confusing to leave it there...

@cartermckinnon sorry bout that, updated the `if` + e2e test (since this reads IMDS) + comments with more context

nodeadm/internal/kubelet/config.go
```diff
@@ -240,6 +240,38 @@ func (ksc *kubeletConfig) withDefaultReservedResources() {
...
 
+// withPodInfraContainerImage determines whether to add the
+// '--pod-infra-container-image' flag, which is used to ensure the sandbox image
+// is not garbage collected.
+//
+// TODO: revisit once the minimum supportted version catches up or the container
+// runtime is moved to containerd 2.0
+func (ksc *kubeletConfig) withPodInfraContainerImage(cfg *api.NodeConfig, kubeletVersion string, flags map[string]string) error {
+	// the flag is a noop on 1.29+, since the behavior was changed to use the
+	// CRI image pinning behavior and no longer considers the flag value.
```
Issacwww, line 251: IIUC, we also need to add https://github.com/awslabs/amazon-eks-ami/pull/1605 into al2023 to help with 1.29 right?

ndbaker1: yea, it looks like we will need the `crictl` for sure in some equivalent fashion. Am i missing any go libraries to pull the image through the CRI versus `crictl`?

ndbaker1: i could use the same upstream resources to pull this to remove the cri-tools install,
it would look something like:
```
imageManager, err := remote.NewRemoteImageService("unix:///run/containerd/containerd.sock", 2*time.Second, nil)
imageSpec := v1.ImageSpec{Image: sandboxImage}
authConfig := v1.AuthConfig{Auth: ecrUserToken}
imageRef, err := imageManager.PullImage(context.TODO(), &imageSpec, &authConfig, nil)
```

cartermckinnon: Chatted offline, I think it's a good idea for `nodeadm` to bake in this functionality but we need to find the right CRI client library to use (pulling in kubelet internals should be a last resort)

cartermckinnon (approved):

PR #1602, opened 2024-01-30
Update CI to use k8s arg to make command

cartermckinnon:
**Description of changes:**

Instead of using the k8s version stub targets, use the `k8s` make variable.


ndbaker1 (approved):
🚀

PR #1603, opened 2024-01-30
write default kubeletConfig to config path in drop-in dir versions

ndbaker1:
**Issue #, if available:**

**Description of changes:**

* reduce the number of defaults written in the kubeletconfig by scoping down the `logging` configuration to a custom type.
* write the default kubelet config to the `config` path rather than the `config-dir`. This gives the correct `KubeletConfiguration` defaults when parsing rather than using the [raw json](https://github.com/kubernetes/kubernetes/pull/121193).

**Testing Done**

launch AL2023 1.29 ami and verify nodes don't enounter `panic: non-positive interval for NewTicker` error

### DIFF
--- a/nodeadm/internal/kubelet/config.go
+++ b/nodeadm/internal/kubelet/config.go
@@ -20,7 +20,6 @@ import (
 
 	"k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	logsapi "k8s.io/component-base/logs/api/v1"
 	k8skubelet "k8s.io/kubelet/config/v1beta1"
 
 	"github.com/aws/aws-sdk-go-v2/feature/ec2/imds"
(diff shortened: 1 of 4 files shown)


nodeadm/internal/kubelet/config.go
```diff
@@ -79,6 +79,10 @@ type kubeletConfig struct {
 	RegisterWithTaints       []v1.Taint                       `json:"registerWithTaints,omitempty"`
 }
 
+type loggingConfiguration struct {
+	Verbosity logsapi.VerbosityLevel `json:"verbosity"`
```
cartermckinnon, line 83: is `VerbosityLevel` just an alias for an integer type? Unless we get some validation during serialization with this, we should just use `int` here.

cartermckinnon (changes requested):

nodeadm/internal/kubelet/config.go
cartermckinnon: Ah it's in component-base.

Thinking about this more -- if we write out the entire default `logsapi.LoggingConfiguration`, wouldn't kubelet still apply its own defaults on top of that? is this change actually necessary?

ndbaker1: it is yea, that's fair to change

ndbaker1: there's [one hairy field](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/component-base/logs/api/v1/types.go#L112C2-L112C39), plus i think it's no issue to just minimize the output json with this change

ndbaker1:
@cartermckinnon the description's been updated to match the behavior unless we are pending anything else

cartermckinnon (approved):

PR #1604, opened 2024-01-31
Add gosec

Issacwww:
**Issue #, if available:**

**Description of changes:**
Add gosec to dependency review. Attempted to add [this previously](https://github.com/awslabs/amazon-eks-ami/pull/1594#issuecomment-1913084855) and turns out the docker based  approach `securego/gosec@master` cannot configure the path to the nodeadm module and it cannot get all the required dependencies. 
Move to a cli way to install and run gosec to address above issue

**Testing Done**

### DIFF
--- a/.github/workflows/dependency-review.yaml
+++ b/.github/workflows/dependency-review.yaml
@@ -12,6 +12,14 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: actions/dependency-review-action@v4
+  gosec:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - run: cd nodeadm
+      - uses: securego/gosec@master
+        with:
+          args: -exclude=G101,G103,G204 ./...
   govulncheck:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/dependency-review.yaml
```diff
@@ -16,7 +16,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: actions/setup-go@4.0.0
+      - uses: actions/setup-go@v5
```
ndbaker1, line 19: `setup-go` version is the culprit? 🤔

Issacwww: no, version is a small part. 
The real problem is the `securego/gosec@master` does not support `working-directory` and it cannot load our dependencies, [example failures](https://github.com/awslabs/amazon-eks-ami/actions/runs/7729953567/job/21074270364) and the cli way works

ndbaker1: oh got it, got confused based on just looking at `commit <-> checks`

ndbaker1 (approved):

PR #1608, opened 2024-01-31
Stop pulling aws-iam-authenticator binary

ndbaker1:
**Issue #, if available:**

**Description of changes:**

we are using `aws eks get-token` over `aws-iam-authenticator` since the `aws` cli version in [AL2023 packages](https://docs.aws.amazon.com/linux/al2023/release-notes/all-packages-AL2023.3.html) supports this feature generally.

**Testing Done**

WIP testing successfully nodegroup launch

### DIFF
--- a/templates/al2023/provisioners/install-worker.sh
+++ b/templates/al2023/provisioners/install-worker.sh
@@ -170,7 +170,6 @@ S3_PATH="s3://$BINARY_BUCKET_NAME/$KUBERNETES_VERSION/$KUBERNETES_BUILD_DATE/bin
 
 BINARIES=(
   kubelet
-  aws-iam-authenticator
 )
 for binary in ${BINARIES[*]}; do
   if [[ -n "$AWS_ACCESS_KEY_ID" ]]; then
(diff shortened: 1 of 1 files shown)


bryantbiggs:
Does the kubelet kubeconfig need to be updated https://github.com/awslabs/amazon-eks-ami/blob/master/files/kubelet-kubeconfig

Or is nodeadm generating a different config with the awscli used instead?

ndbaker1:
> is nodeadm generating a different config with the awscli used instead?

yep, there's an updated kubeconfig generated by nodeadm which doesn't rely on `aws-iam-authenticator`:
https://github.com/awslabs/amazon-eks-ami/blob/al2023/nodeadm/internal/kubelet/kubeconfig.template.yaml

> Does the kubelet kubeconfig need to be updated https://github.com/awslabs/amazon-eks-ami/blob/master/files/kubelet-kubeconfig

this behavior is just for AL2023 using nodeadm, so the one for AL2 will still use the linked file

cartermckinnon (approved):

PR #1609, opened 2024-01-31
implement NodeConfig merging for ec2 user-data

ndbaker1:
**Issue #, if available:**

**Description of changes:**

implement merge semantics for `NodeConfig`, which originally returned an error when encountering more than one `NodeConfig` in a userdata MIME multipart document.

* the blanket merge strategy for the `NodeConfig` is fields in the **patch/source** document will overwrite corresponding fields in the **base/destination** document.
* the merge strategy for `Spec.Kubelet.Config`, the inline kubelet config, is a full json-merge between the two unless one is empty. In the latter case, use the content from the non-empty field.
* the merge strategy for `Spec.Kubelet.Flags`, the kubelet flag arguments, is to append the source flags to the destination flags. Following the semantics of the underlying flag parsing library, https://github.com/spf13/pflag, single-value flags that come last have higher precedence, and multiple-value flags are merged with the library's underlying strategy (which obeys user expectations rather than us using some heuristic)

**Testing Done**


nodeadm/internal/api/merge.go
```diff
@@ -0,0 +1,83 @@
...
+			if err := json.Unmarshal([]byte(src.String()), &srcConfigMap); err != nil {
+				return err
+			}
+			if err := mergo.Merge(&dstConfigMap, &srcConfigMap, mergo.WithOverride); err != nil {
+				return err
+			}
+			config, err := json.MarshalIndent(dstConfigMap, "", "    ")
+			if err != nil {
+				return err
+			}
```
cartermckinnon, line 78: This looks really similar to the merging logic in #1612, lets make sure we can share the impl

nodeadm/internal/configprovider/userdata.go
```diff
@@ -83,23 +88,27 @@ func parseMultipart(userDataReader *multipart.Reader) (*internalapi.NodeConfig,
...
 				if err != nil {
 					return nil, err
 				}
+				if config != nil {
+					if err := config.Merge(decodedConfig); err != nil {
+						return nil, err
+					}
+				} else {
+					config = decodedConfig
+				}
```
cartermckinnon, line 111: I think it'd be a bit clearer to just collect the decoded configs in a slice and then merge them all at the end.

cartermckinnon (changes requested):
Couple small things

nodeadm/internal/api/merge.go
ndbaker1: put this into a `DocumentMerge` util, which basically uses marshalled representations of `any`-types to the do underlying merge. not beautiful but its not like we're we need to hunt for optimizations just yet 🤷

nodeadm/internal/configprovider/userdata.go
ndbaker1: done!

cartermckinnon (approved):

PR #1610, opened 2024-02-01
remove unneeded cni installation

ndbaker1:
**Issue #, if available:**

**Description of changes:**

if customer's will install 3rd party CNI do we need to continue installing this?

**Testing Done**


ndbaker1:
will this break things that log-analyzer is looking for?

bryantbiggs:
will this increase node startup latency; if so, do we know by how much? I assume VPC CNI will have to pull these upon instance creation

cartermckinnon:
> if customer's will install 3rd party CNI do we need to continue installing this?

Generally I would say no. If a CNI implementation is utilizing these binaries, it should include them in its distribution. Calico does this as well: https://github.com/projectcalico/calico/tree/d7934529b7f233c93c93b3ef9efc0ecc55de2a3d/cni-plugin

templates/al2023/template.json
```diff
@@ -183,7 +182,6 @@
         "KUBERNETES_BUILD_DATE={{user `kubernetes_build_date`}}",
         "KUBERNETES_VERSION={{user `kubernetes_version`}}",
         "PAUSE_CONTAINER_VERSION={{user `pause_container_version`}}",
-        "PULL_CNI_FROM_GITHUB={{user `pull_cni_from_github`}}",
```
cartermckinnon, line 186: We need to get rid of `CNI_PLUGIN_VERSION` and the associated template var as well

ndbaker1: ack, done

ndbaker1:
Is this rollable into to AL2, or should this just be an AL2023 change since we don't know who might depend on this?

cartermckinnon:
We'll leave this as-is in AL2 👍

cartermckinnon (approved):

PR #1612, opened 2024-02-01
support embedded kubelet config via YAML and JSON

ndbaker1:
**Issue #, if available:**

**Description of changes:**

rather than take an inline config, represented as a `string`, use documents embedded into the `NodeConfig` yaml in any supported format (YAML/JSON).

**Testing Done**


nodeadm/api/v1alpha1/nodeconfig_types.go
```diff
@@ -43,10 +44,10 @@ type ClusterDetails struct {
 }
 
 type KubeletOptions struct {
-	// Config is a raw document of a kubelet config that can be provided
-	// by the user to override default generated configurations
+	// Config is a kubelet config that can be provided by the user to override
+	// default generated configurations
 	// https://kubernetes.io/docs/reference/config-api/kubelet-config.v1/
-	Config string `json:"config,omitempty"`
+	Config map[string]runtime.RawExtension `json:"config,omitempty"`
```
ndbaker1, line 50: `map[string]interface{}` isn't allowed by conversion-gen, so this is on of the next best suggestions to achieve the same end usage. the internal usages are a bit more complex based on instances where u cannot directly merge `map[string]interface{}` with `map[string]runtime.RawExtension`

nodeadm/test/e2e/cases/kubelet-config-dir/config.kubelet-config-json.yaml
```diff
@@ -8,12 +8,11 @@ spec:
...
-    config: |
-      {
-        "kind": "KubeletConfiguration",
-        "apiVersion": "kubelet.config.k8s.io/v1beta1",
-        "clusterDNS": [
-          "0.0.0.0",
-          "1.1.1.1"
-        ]
-      }
+    config: {
```
jmdeal, line 11: Is this an intended use case, embedding json inside the yaml? My impression was that users could supply the **entire** `NodeConfig` as either json or yaml, not a subset. I didn't realize this was even valid yaml to begin with but if the tests run and parse I guess it is.

ndbaker1: for drop-in use cases I can see users doing it for the `kubelet.config` even though converting it wouldn't take much effort

cartermckinnon: All JSON is valid YAML, so I guess this is allowed per the specs. Looks pretty odd though 😄

nodeadm/api/v1alpha1/nodeconfig_types.go
cartermckinnon: This looks like the way to go. If merging this type with mergo is cumbersome, can we write a custom handler or something? Marshalling to-and-from JSON will work for now but would be nice to clean that up

ndbaker1: sounds good 👍, as long as the use case works we can hash this out

PR #1614, opened 2024-02-02
default http-put-response-hop-limit to 1 for al2023

Issacwww:
**Issue #, if available:**

**Description of changes:**
this option is default to 2, per this [doc](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-IMDS-new-instances.html#configure-IMDS-new-instances-ami-configuration). 
And suggested by AppSec, since we supported [pod identity](https://aws.amazon.com/about-aws/whats-new/2023/11/amazon-eks-pod-identity/), we should default it to 1

**Testing Done**
Build 1.28 AMI and tested with kubetest2 and conformance

### DIFF
--- a/templates/al2/template.json
+++ b/templates/al2/template.json
@@ -113,7 +113,8 @@
       "ami_name": "{{user `ami_name`}}",
       "ami_description": "{{ user `ami_description` }}, {{ user `ami_component_description` }}",
       "metadata_options": {
-        "http_tokens": "required"
+        "http_tokens": "required",
+        "http_put_response_hop_limit": 1
       }
     }
   ],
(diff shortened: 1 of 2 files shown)


cartermckinnon:
Do the `metadata_options` apply to Packer's builder instance, to the resulting AMI, or both?

My understanding is the `http_tokens` bit only applied for the builder instance. What does this hop limit apply to?

Issacwww:
> Do the `metadata_options` apply to Packer's builder instance, to the resulting AMI, or both?
Ah, you are right, this is for packer builder instance, we still need config the resulting AMI

Issacwww:
Discard this change as the IMDS Hop should be set when launch ec2 instance, thus need change this config in our launch template. This should be set for MNG, for self MNG cx should set it to 1 in their own LT

PR #1615, opened 2024-02-02
Remove sandbox image from build-time cache

cartermckinnon:
**Description of changes:**

This removes the sandbox image from the build-time cache. This image needs to be pulled via the CRI gRPC server of containerd in order to be properly labeled as "pinned" in order to be excluded from kubelet's image garbage collection. The multi-regional tagging behavior of the image cache is not amenable to this, and the sandbox image is very small -- it is not expensive to pull at runtime.


cartermckinnon:
/ci

mmerkes (approved):

dims (approved):
LGTM

PR #1616, opened 2024-02-02
use annotation to ignore gosec rule

Issacwww:
**Issue #, if available:**

**Description of changes:**
Suggested by AppSec to remove the exempt from the CI action.
For G101, G103, the used case was reviewed and can be muted.
 
For G204, attempted replace `exec.Command` with `syscall.Exec` but still got G204 findings. And some of our command need to get the cli output, while syscall.Exec won’t return it. Thus keep using current implementation and apply the annotation to mute it.

**Testing Done**
`gosec ./...` and no issue reported


.github/workflows/dependency-review.yaml
```diff
@@ -20,7 +20,7 @@ jobs:
         with:
           go-version: '1.21.6'
       - run: go install github.com/securego/gosec/v2/cmd/gosec@latest
-      - run: gosec -exclude=G101,G103,G204 ./...
+      - run: gosec -exclude-generated ./...
```
ndbaker1, line 23: nice!

nodeadm/test/e2e/cases/kubeconfig-outpost/run.sh
```diff
@@ -0,0 +1,14 @@
...
+set -o pipefail
+
+source /helpers.sh
+
+mock::imds
+wait::dbus-ready
+
+mock::kubelet 1.28.0
+nodeadm init --skip run --config-source file://config.yaml
+assert::file-contains /etc/hosts 'example.com'
```
ndbaker1, line 14: wait, this is hitting the real `example.com`? can we use `localhost` or mock a hostname and then include the whole `'IP HOST'` string in the assert?

ndbaker1 (approved):
LGTM

PR #1619, opened 2024-02-02
Label cached sandbox images as pinned

cartermckinnon:
**Description of changes:**

Adds the `io.cri-containerd.pinned=pinned` label to sandbox container images that are cached during the AMI build process. This will exclude all tags of the cached image layers from being garbage collected by kubelet.


cartermckinnon:
/ci

cartermckinnon:
Going with #1615 for now, may revisit this.

PR #1621, opened 2024-02-02
fix arm64 ami build

Issacwww:
**Issue #, if available:**

**Description of changes:**
run into the following issue when testing arm64 ami build
```
> make 1.28 os_distro=al2023 arch=arm64 
---
The architecture 'x86_64' of the specified instance type does not match the architecture 'arm64' of the specified AMI. Specify an instance type and an AMI that have matching architectures, and try again. You can use 'describe-instance-types' or 'describe-images' to discover the architecture of the instance type or AMI.
```

**Testing Done**
build arm64 ami succeed `ami-000353e26e1f56762`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,14 +16,15 @@ AMI_VARIANT ?= amazon-eks
 AMI_VERSION ?= v$(shell date '+%Y%m%d')
 os_distro ?= al2
 arch ?= x86_64
-instance_type ?= m5.large
 
 ifeq ($(os_distro), al2023)
 	AMI_VARIANT := $(AMI_VARIANT)-al2023
 endif
 ifeq ($(arch), arm64)
-	instance_type = m6g.large
+	instance_type ?= m6g.large
 	AMI_VARIANT := $(AMI_VARIANT)-arm64
+else
+	instance_type ?= m5.large
 endif
 ifeq ($(enable_fips), true)
 	AMI_VARIANT := $(AMI_VARIANT)-fips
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -22,7 +22,7 @@ ifeq ($(os_distro), al2023)
 	AMI_VARIANT := $(AMI_VARIANT)-al2023
 endif
 ifeq ($(arch), arm64)
-	instance_type ?= m6g.large
+	instance_type = m6g.large
```
cartermckinnon, line 25: What's the reason for this change?

Issacwww: `instance_type` will only been updated if it is not already set when using `?=`

thus run into the following error initially 
```
 make 1.28 os_distro=al2023 arch=arm64         

PACKER_ARGS: -var-file templates/al2023/variables-default.json  **-var instance_type='m5.large'** -var arch='arm64'  -var kubernetes_version='1.28.5'
```

ndbaker1: oh but we want to avoid locking in `instance_type`, any conflicts with this?
```Make
ifeq ($(arch), arm64)
	instance_type ?= m6g.large
	AMI_VARIANT := $(AMI_VARIANT)-arm64
else
	instance_type ?= m5.large
endif
```

Issacwww: ah, it would be better, although it has been defaulted to m5.large at line 19 already 😄
will update it in next rev shortly

ndbaker1 (approved):

PR #1605, opened 2024-01-31
Use crictl to pull sandbox image

cartermckinnon:
**Issue #, if available:**

Fixes #1597 on `containerd` 1.7.3+.

**Description of changes:**

The containerd CRI server will only report an image as "pinned" if it is *pulled* with the CRI server. The pinned label will not be applied if it is pulled using the bare `containerd` server with `ctr`.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

**Testing Done**

On `containerd-1.7.11` with `sandbox_image = "registry.k8s.io/pause:3.9"`:
```
> sudo ctr --namespace k8s.io image pull registry.k8s.io/pause:3.9
> sudo crictl inspecti registry.k8s.io/pause:3.9 | grep pinned
    "pinned": false

> sudo crictl rmi --all
Deleted: registry.k8s.io/pause:3.9

> sudo crictl pull registry.k8s.io/pause:3.9
Image is up to date for sha256:e6f1816883972d4be47bd48879a08919b96afcd344132622e4d444987919323c
> sudo crictl inspecti registry.k8s.io/pause:3.9 | grep pinned
    "pinned": true
```

PR #1626, opened 2024-02-05
Specify region for local zones in sandbox image ecr auth

ndbaker1:
**Issue #, if available:**

**Description of changes:**

Fix aws cli behavior pertaining to issue https://github.com/aws/aws-cli/issues/7043 for sandbox image pulling script (https://github.com/awslabs/amazon-eks-ami/pull/1605)

**Testing Done**

### DIFF
--- a/files/pull-sandbox-image.sh
+++ b/files/pull-sandbox-image.sh
@@ -8,6 +8,10 @@ if [[ "$(sudo ctr --namespace k8s.io image ls | grep $sandbox_image)" != "" ]]; 
   exit 0
 fi
 
+# use the region that the sandbox image comes from for the ecr authentication,
+# also mitigating the localzone isse: https://github.com/aws/aws-cli/issues/7043
+region=$(echo "${sandbox_image}" | cut -f4 -d ".")
+
 MAX_RETRIES=3
 
 function retry() {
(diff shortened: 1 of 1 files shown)


files/pull-sandbox-image.sh
```diff
@@ -8,6 +8,10 @@ if [[ "$(sudo ctr --namespace k8s.io image ls | grep $sandbox_image)" != "" ]];
   exit 0
 fi
 
+# use the region that the sandbox image comes from for the ecr authentication,
+# also mitigating the localzone isse: https://github.com/aws/aws-cli/issues/7043
+region=$(echo "${sandbox_image}" | cut -f4 -d ".")
```
ndbaker1, line 13: unlike `pull-image.sh` this will use whichever sandbox image has been set with containerd, so should we default this to the instance's region for values like `registry.k8s.io/pause:3.9`?

ndbaker1: reusing this snippet
```bash
imds 'latest/dynamic/instance-identity/document' | jq .region -r
```

dims: if the customer updates the config.toml to their own image (could even be outside of AWS), will this service keep failing?

dims: (because there may not be a region found?)

cartermckinnon: I think the behavior will be equivalent, we previously pulled the sandbox image from the containerd config file and passed it to pull-image.sh

cartermckinnon (approved):

Issacwww (approved):

PR #1627, opened 2024-02-06
config evictionHard and kubeReserved

Issacwww:
**Issue #, if available:**

**Description of changes:**
Add support for 
- evictionHard and kubeReserved configuration
- auto generate `eni-max-pods.go`

**Testing Done**
`make test && make test-e2e`


hack/generate_eni_max_pod_mapping.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+inputFile=$1
+outputFile=$2
+
+echo "package util" > $outputFile
+echo "" >> $outputFile
+echo "// Instance_types_map generated from eni-max-pods.txt DO NOT EDIT" >> $outputFile
+echo "var Instance_types_map = map[string]int{" >> $outputFile
```
ndbaker1, line 9: one small thing, can we use a CamelCase name like `InstanceTypeMaxPods`

nodeadm/internal/kubelet/config.go
```diff
@@ -249,9 +258,18 @@ func (ksc *kubeletConfig) withCloudProvider(cfg *api.NodeConfig, flags map[strin
...
 	ksc.SystemReservedCgroup = ptr.String("/system")
 	ksc.KubeReservedCgroup = ptr.String("/runtime")
+	maxPods, ok := getMaxPod(cfg.Status.Instance.Type)
+	if !ok {
+		return
+	}
+	ksc.KubeReserved = map[string]string{
+		"cpu":               strconv.Itoa(getCPUMillicoresToReserve()),
+		"ephemeral-storage": "1Gi",
+		"memory":            strconv.Itoa(getMemoryMebibytesToReserve(maxPods)),
```
ndbaker1, line 271: can you format string these to have the units? (`Mi`, `m`)
https://github.com/awslabs/amazon-eks-ami/blob/master/files/bootstrap.sh#L522

nodeadm/internal/util/eni-max-pods.go
cartermckinnon, line 1: We don't need to transform the text file into golang like this. You can do something like this to embed the `eni-max-pods.txt` into the go binary, and then just parse the text into whatever go type you want:

Also, we shouldn't put stuff like this in `util`. This is a `kubelet` thing, so just put in that package.
```
package kubelet

import _ "embed"

//go:embed eni-max-pods.txt
var eniMaxPods string

var MaxPodsPerInstanceType map[string]int

func init() {
    // parse the text data and populate the map
}
```

nodeadm/test/e2e/run.sh
```diff
@@ -18,6 +18,7 @@ for CASE_DIR in $(ls -d test/e2e/cases/*); do
   CONTAINER_ID=$(docker run \
     -d \
     --rm \
+    --cpus "4" \
```
cartermckinnon, line 21: what's this about?

Issacwww: This is for the e2e testing, the calculation for kubeReserved.cpu is depending on the return value of `runtime.NumCPU()`, which might change in different setup(my local docker engine use 8), thus trying to use this as a pin to get consistent e2e test result

cartermckinnon: Ah I see. `runtime.NumCPU()` will give you the number of cores available to the current process, the same as `GOMAXPROCS`, not the total number of physical/virtual cores on the system.

It looks like `kubelet` gets this information (which is reported as the allocatable resources on the node) with cadvisor: https://github.com/kubernetes/kubernetes/blob/2fb99eb9e81ec9fc92d8e694227772928f2eecc5/pkg/kubelet/cadvisor/util.go#L42

That ultimately comes from here in cadvisor: https://github.com/google/cadvisor/blob/35dd21c64581692aae5fcaf3a470e8a0fe54fcf6/utils/sysinfo/sysinfo.go#L203

I don't love the idea of adding a dependency on all of cadvisor, but we can either copy the relevant impl or pull it in as a last resort.

dims: copy and strip down please. cadvisor is `ktlo`

nodeadm/internal/kubelet/config.go
```diff
@@ -54,29 +54,32 @@ func (k *kubelet) writeKubeletConfig(cfg *api.NodeConfig) error {
...
 	ContainerRuntimeEndpoint string                           `json:"containerRuntimeEndpoint"`
+	EvictionHard             map[string]string                `json:"evictionHard,omitempty"`
 	FeatureGates             map[string]bool                  `json:"featureGates"`
 	HairpinMode              string                           `json:"hairpinMode"`
+	KubeAPIBurst             *int                             `json:"kubeAPIBurst,omitempty"`
+	KubeAPIQPS               *int                             `json:"kubeAPIQPS,omitempty"`
+	KubeReserved             map[string]string                `json:"kubeReserved,omitempty"`
+	KubeReservedCgroup       *string                          `json:"kubeReservedCgroup,omitempty"`
+	Logging                  loggingConfiguration             `json:"logging"`
+	MaxPods                  int32                            `json:"maxPods,omitempty"`
```
Issacwww, line 73: new attribute added, the rest just sort

nodeadm/internal/util/sysinfo.go
```diff
@@ -0,0 +1,286 @@
...
+	packageIDFilePath = sysFsCPUTopology + "/physical_package_id"
+)
+
+type Core struct {
+	Id       int   `json:"core_id"`
+	Threads  []int `json:"thread_ids"`
+	SocketID int   `json:"socket_id"`
+}
+
+func init() {
```
Issacwww, line 34: need this as the mock sys files cannot be copied into the /sys during docker image build, this might be a hack, let me know if there is better way to set up this mock

.github/workflows/sync-eni-max-pods.yaml
```diff
@@ -34,11 +34,14 @@ jobs:
           cd amazon-vpc-cni-k8s/
           make generate-limits
           cp misc/eni-max-pods.txt ../amazon-eks-ami/templates/shared/runtime/eni-max-pods.txt
+          cp misc/eni-max-pods.txt ../amazon-eks-ami/nodeadm/internal/kubelet/eni-max-pods.txt
```
Issacwww, line 37: go:embed need the file to be within the go module

nodeadm/internal/kubelet/config.go
```diff
@@ -437,3 +455,42 @@ func getNodeIp(ctx context.Context, imdsClient *imds.Client, cfg *api.NodeConfig
...
+
+// getResourceToReserveInRange calculates the CPU resources to reserve for a given range.
+func getResourceToReserveInRange(totalCPU, startRange, endRange, percentage int) int {
+	if totalCPU > startRange {
+		if totalCPU < endRange {
+			return (totalCPU - startRange) * percentage / 10000
+		}
+		return (endRange - startRange) * percentage / 10000
+	}
+	return 0
```
ndbaker1, line 486: super nit, but this isn't instantly readable to me, could it be 
```suggestion
	if totalCPU <= startRange {
	    return 0
	}
	reserved := totalCPU
	if reserved > endRange {
	    reserved = endRange
	}
	return (reserved - startRange) * percentage / 10000
```

nodeadm/internal/kubelet/config.go
```diff
@@ -437,3 +455,42 @@ func getNodeIp(ctx context.Context, imdsClient *imds.Client, cfg *api.NodeConfig
...
+}
+
+func getMemoryMebibytesToReserve(maxPods int) int {
+	return 11*maxPods + 255
+}
+
+func getMaxPod(instanceType string) (int, bool) {
+	maxPod, ok := MaxPodsPerInstanceType[instanceType]
+	return maxPod, ok
+}
```
ndbaker1, line 496: is this indirection to `MaxPodsPerInstanceType` needed?

Issacwww: good catch, removed

nodeadm/internal/kubelet/config.go
Issacwww: ack, updated

nodeadm/test/e2e/infra/Dockerfile
```diff
@@ -23,9 +23,12 @@ COPY test/e2e/infra/aemm-default-config.json /etc/aemm-default-config.json
 COPY --from=nodeadm-build /nodeadm /usr/local/bin/nodeadm
 COPY test/e2e/infra/systemd/kubelet.service /usr/lib/systemd/system/kubelet.service
 COPY test/e2e/infra/systemd/containerd.service /usr/lib/systemd/system/containerd.service
+COPY test/e2e/infra/mock/ /sys_devices_system_mock/
 COPY test/e2e/helpers.sh /helpers.sh
 
 RUN mkdir -p /etc/eks/image-credential-provider/
 RUN touch /etc/eks/image-credential-provider/ecr-credential-provider
+ENV CPU_DIR /sys_devices_system_mock/cpu
+ENV NODE_DIR //sys_devices_system_mock/node
```
ndbaker1, line 32: accidental extra character?
```suggestion
ENV NODE_DIR /sys_devices_system_mock/node
```

nodeadm/internal/util/sysinfo.go
```diff
@@ -0,0 +1,286 @@
...
+	isOnline, err := isCpuIDOnline(cpuOnlinePath, cpuID)
+	if err != nil {
+		zap.L().Info(fmt.Sprintf("Unable to get online CPUs list: %s", err))
+		return false
+	}
+	return isOnline
+}
+
+func getCPUID(dir string) (uint16, error) {
+	regex := regexp.MustCompile("cpu([0-9]+)")
```
ndbaker1, line 230: is this effectively the same as `cpuDirRegExp` on line 14?

nodeadm/internal/kubelet/config.go
```diff
@@ -249,9 +257,19 @@ func (ksc *kubeletConfig) withCloudProvider(cfg *api.NodeConfig, flags map[strin
...
-func (ksc *kubeletConfig) withDefaultReservedResources() {
+func (ksc *kubeletConfig) withDefaultReservedResources(cfg *api.NodeConfig) {
 	ksc.SystemReservedCgroup = ptr.String("/system")
 	ksc.KubeReservedCgroup = ptr.String("/runtime")
+	maxPods, ok := MaxPodsPerInstanceType[cfg.Status.Instance.Type]
+	if !ok {
+		return
+	}
+	ksc.MaxPods = int32(maxPods)
+	ksc.KubeReserved = map[string]string{
```
ndbaker1, line 268: Ideally we should have some fall back for the `kubeReserved` fields in rare cases we don't have `maxPods`. 

I don't feel like rushing to add the dynamic max pods calculation (and adding an ec2 call) is the best thing, but we could do it. 

Maybe we use the default `110` for max pods in the memory-to-reserve calc, but not set it in the kubelet config for now. This is just a rare edge case but i want to have a sane default 🙃

nodeadm/internal/util/sysinfo.go
ndbaker1: Do you mean they get wiped on boot or is it a perm issue? Sorry I haven't tested myself yet

Would a bash helper that u call in the test cases to copy them into the dir work?

nodeadm/internal/util/sysinfo.go
Issacwww: good catch, I took a closer look, and we can remove `getCPUID` entirely and use `getMatchedInt`

nodeadm/internal/kubelet/config.go
Issacwww: I think this rare edge case would happen when new instance type launched and our AMI does not have the latest max pod.
Make sure I understand this:
- Do you mean instead of default both `MaxPods` and the `kubeReserved.memory`, we just use `110` for kubeReserved.memory calc when the instance type is missing?
- Without setting MaxPods, what would be the upper limit of it?
- Where this 110 value comes from?

As I think if we want to add default value, we should default both of them

ndbaker1: i think the case only gets hit when an old AMI is used on a new instance type which isn't in the version's packaged `eni-max-pods`. just making sure we don't completely give up in that scenario

cartermckinnon: > I don't feel like rushing to add the dynamic max pods calculation (and adding an ec2 call) is the best thing, but we could do it.

I don't see much benefit in removing this fallback, and I think it'd be a surprising (and annoying) difference between AL2 and AL2023. It's easy for us to implement. If someone doesn't want to give their node role the `ec2:DescribeInstanceTypes` permission, we'll get an error and *then* we can fallback to `110`.

ndbaker1: if we *are* going to implement it, should we just be generating a file with all the instance descriptors like `EniLimit`, `CPUCount`, etc , similar to the [vpc_limits file](https://github.com/aws/amazon-vpc-cni-k8s/blob/master/pkg/vpc/vpc_ip_resource_limit.go)?

Issacwww: I think the eni-max-pods would be enough for most of the cases and don't  really need to replicate instance descriptors for all types. And inflight calculation for new types should be enough and the eni-max-pods should get updated in next AMI

ndbaker1 (approved):

PR #1637, opened 2024-02-07
support user containerd config in NodeConfig

ndbaker1:
**Issue #, if available:**

**Description of changes:**

add support for containerd options via `NodeConfig` so that users can have control over containerd configuration without workarounds

At this point in time the config is provided via an inline string, so if more than one `NodeConfig` is parsed then the value in the last config will win during the merging process.

**Testing Done**

unit & e2e tests


nodeadm/internal/containerd/config.go
```diff
@@ -4,18 +4,21 @@ import (
...
+	"go.uber.org/zap"
 )
 
 const ContainerRuntimeEndpoint = "unix:///run/containerd/containerd.sock"
 
 const (
-	containerdConfigFile = "/etc/containerd/config.toml"
-	containerdConfigPerm = 0644
+	containerdConfigFile      = "/etc/containerd/config.toml"
+	containerdConfigImportDir = "/etc/containerd/config.d"
```
Issacwww, line 20: how will this drop-in been pickup by containerd?
And is there any override to the config.toml?

ndbaker1: the required change in the base containerd config happened previously so that's why u don't see it here 😅 
https://github.com/awslabs/amazon-eks-ami/blob/e5426f60dea1dcf0256ba5a3d249ce0fd7e892d8/nodeadm/internal/containerd/config.template.toml#L4-L7

ndbaker1: no changes happen to the config.toml itself

Issacwww (approved):

PR #1640, opened 2024-02-08
Reorganize template docs, add CRD docs

cartermckinnon:
**Description of changes:**

Splits the AL2 and AL2023 bits of the docs into separate documents. We'll cover the template and runtime details in each.


nodeadm/doc/format.sh
ndbaker1, line 1: is this file used?

hack/generate-template-variable-doc.py
```diff
@@ -53,12 +51,10 @@ def update_doc(doc: str, boundary: str, template_path: str) -> str:
...
-doc = update_doc(doc, al2_boundary, '../templates/al2/template.json')
-doc = update_doc(doc, al2023_boundary, '../templates/al2023/template.json')
-
-with open(doc_file_name, 'w') as doc_file:
-    doc_file.write(doc)
+for template in ['al2', 'al2023']:
+    doc_file_name = f'../doc/usage/{template}.md'
+    with open(doc_file_name) as doc_file:
+        doc = doc_file.read()
+        doc = update_doc(doc, f'../templates/{template}/template.json')
```
ndbaker1, line 58: while we're in the business of house cleaning for new distros, maybe we can just discover these based on a directory walk. 
just one less thing to keep updating 🤷

ndbaker1:
glad we can do cleanup/reorgs like this 😌
i think the format is good but not sure if you're still adding content or if we can merge now and add the meat as we go

nodeadm/doc/format.sh
cartermckinnon: thx, this was hackery that ended up being unnecessary

cartermckinnon:
>  i think the format is good but not sure if you're still adding content or if we can merge now and add the meat as we go

I intend to merge this and isolate net-new content in their own PR's.

ndbaker1 (approved):

PR #1646, opened 2024-02-09
Fix sandbox image pull retries

cartermckinnon:
**Description of changes:**

 Fixes the retry mechanism for sandbox image pulls.


ndbaker1:
is this to prevent the error breaking the service unit so that if the image isnt an ECR registry then it can safely exit?

cartermckinnon:
covered in #1649

PR #1648, opened 2024-02-09
[WIP] Defensive code in pull-sandbox-image.sh for custom AMI(s)

dims:
- Bailout if `sandbox_image` is not even present (or commented out) in a custom config.toml (created by customer on top of our AMI as base or injected via user data)
- Bailout if the sandbox_image points to something where there is no region! (could be a public community pause image or a public/private/air-gapped repository of some sort - basically non ECR)

### DIFF
--- a/files/pull-sandbox-image.sh
+++ b/files/pull-sandbox-image.sh
@@ -3,14 +3,25 @@ set -euo pipefail
 
 source <(grep "sandbox_image" /etc/containerd/config.toml | tr -d ' ')
 
+### skip if we don't have a sandbox_image set in config.toml
+if [[ -z ${sandbox_image:-} ]]; then
+  echo >&2 "Skipping ... missing sandbox_image from /etc/containerd/config.toml"
+  exit 0
+fi
+
 ### Short-circuit fetching sandbox image if its already present
-if [[ "$(sudo ctr --namespace k8s.io image ls | grep $sandbox_image)" != "" ]]; then
+if [[ $(sudo ctr --namespace k8s.io image ls | grep "${sandbox_image}") != "" ]]; then
   exit 0
 fi
 
 # use the region that the sandbox image comes from for the ecr authentication,
 # also mitigating the localzone isse: https://github.com/aws/aws-cli/issues/7043
 region=$(echo "${sandbox_image}" | cut -f4 -d ".")
+### skip if we don't find a region in the sandbox image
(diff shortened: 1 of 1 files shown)


dims:
cc @ndbaker1 @cartermckinnon

files/pull-sandbox-image.sh
```diff
@@ -3,14 +3,25 @@ set -euo pipefail
...
-if [[ "$(sudo ctr --namespace k8s.io image ls | grep $sandbox_image)" != "" ]]; then
+if [[ $(sudo ctr --namespace k8s.io image ls | grep "${sandbox_image}") != "" ]]; then
   exit 0
 fi
 
 # use the region that the sandbox image comes from for the ecr authentication,
 # also mitigating the localzone isse: https://github.com/aws/aws-cli/issues/7043
 region=$(echo "${sandbox_image}" | cut -f4 -d ".")
+### skip if we don't find a region in the sandbox image
+if [[ -z ${region:-} ]]; then
```
ndbaker1, line 21: just because we parse a non-empty region doesn't mean its valid right?
`registry.k8s.io/pause:3.9` is going to parse `9`

files/pull-sandbox-image.sh
```diff
@@ -29,7 +40,7 @@ function retry() {
   done
 }
 
-ecr_password=$(retry aws ecr get-login-password --region $region)
+ecr_password=$(retry aws ecr get-login-password --region "${region}")
```
ndbaker1, line 43: if we add carter's fix https://github.com/awslabs/amazon-eks-ami/pull/1646/files and don't break the whole workflow on this error, could we have a branch for empty/non-empty `ecr_password` that will make the image pull with/without creds?

something roughly like
```bash
if [[ -z ${ecr_password} ]]; then
  retry sudo crictl pull "${sandbox_image}"
else
  retry sudo crictl pull --creds "AWS:${ecr_password}" "${sandbox_image}"
fi
```

files/pull-sandbox-image.sh
dims: correct, we need more validation!

files/pull-sandbox-image.sh
dims: agree

dims:
@ndbaker1 @cartermckinnon feel free to folks this PR/commit into something you are already working on. opened this as WIP just to highlight

PR #1649, opened 2024-02-09
harden pull-sandbox-image script

ndbaker1:
**Issue #, if available:**

**Description of changes:**

* remove the `pipefail` so that authentication errors don't fast fail the workflow since it is ok if the ecr call fails.
* add additional bash shell hardening.

**Testing Done**

an empty credential pull for public repositories:
```bash
$ crictl pull --creds "AWS:" registry.k8s.io/pause:3.9
Image is up to date for sha256:e6f1816883972d4be47bd48879a08919b96afcd344132622e4d444987919323c
```

in the case where a region is malformed or incorrect, this will still succeed for a public image:
```bash
$ grep sandbox_image /etc/containerd/config.toml
sandbox_image = "registry.k8s.io/pause:3.9"

$ /etc/eks/containerd/pull-sandbox-image.sh
[root@ip-192-168-5-111 bin]# /etc/eks/containerd/pull-sandbox-image.sh

Provided region_name '9' doesn't match a supported format.
Attempt 1 of 3

Provided region_name '9' doesn't match a supported format.
Attempt 2 of 3

Provided region_name '9' doesn't match a supported format.
Attempt 3 of 3

Provided region_name '9' doesn't match a supported format.
Unable to retrieve the ECR password. Image pull may not be properly authenticated.

Image is up to date for sha256:e6f1816883972d4be47bd48879a08919b96afcd344132622e4d444987919323c

# verify "caching" works
$ /etc/eks/containerd/pull-sandbox-image.sh
Skipping ... sandbox_image 'registry.k8s.io/pause:3.9' is already present
```

existing ecr uris:
```bash
$ grep sandbox_image /etc/containerd/config.toml
sandbox_image = "602401143452.dkr.ecr.us-west-2.amazonaws.com/eks/pause:3.5"

$ /etc/eks/containerd/pull-sandbox-image.sh
Image is up to date for sha256:6996f8da07bd405c6f82a549ef041deda57d1d658ec20a78584f9f436c9a3bb7

$ /etc/eks/containerd/pull-sandbox-image.sh
Skipping ... sandbox_image '602401143452.dkr.ecr.us-west-2.amazonaws.com/eks/pause:3.5' is already present
```

### DIFF
--- a/files/pull-sandbox-image.sh
+++ b/files/pull-sandbox-image.sh
@@ -1,16 +1,25 @@
 #!/usr/bin/env bash
-set -euo pipefail
 
 source <(grep "sandbox_image" /etc/containerd/config.toml | tr -d ' ')
 
-### Short-circuit fetching sandbox image if its already present
-if [[ "$(sudo ctr --namespace k8s.io image ls | grep $sandbox_image)" != "" ]]; then
+### skip if we don't have a sandbox_image set in config.toml
+if [[ -z ${sandbox_image:-} ]]; then
+  echo >&2 "Skipping ... missing sandbox_image from /etc/containerd/config.toml"
   exit 0
 fi
 
-# use the region that the sandbox image comes from for the ecr authentication,
-# also mitigating the localzone isse: https://github.com/aws/aws-cli/issues/7043
-region=$(echo "${sandbox_image}" | cut -f4 -d ".")
+### Short-circuit fetching sandbox image if its already present
+if [[ -n $(sudo ctr --namespace k8s.io image ls | grep "${sandbox_image}") ]]; then
+  echo >&2 "Skipping ... sandbox_image '${sandbox_image}' is already present"
+  exit 0
+fi
+
+# if the sandbox image is provided by the bootstrap script, then the region is
+# guaranteed to come from this data source.
+# see: https://github.com/awslabs/amazon-eks-ami/blob/baef6f0860f60dbec366de30853e47418e3fb430/files/bootstrap.sh#L320-L338
+# if the image is customer provided, then this is just a sane default for the
+# region when attempting to get ecr credentials.
+region=$(imds 'latest/dynamic/instance-identity/document' | jq .region -r)
 
 MAX_RETRIES=3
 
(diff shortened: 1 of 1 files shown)


ndbaker1:
built on top of https://github.com/awslabs/amazon-eks-ami/pull/1648 and https://github.com/awslabs/amazon-eks-ami/pull/1646
@dims @cartermckinnon can you guys check me that this covers all the bases?

files/pull-sandbox-image.sh
```diff
@@ -1,16 +1,25 @@
...
 
-# use the region that the sandbox image comes from for the ecr authentication,
-# also mitigating the localzone isse: https://github.com/aws/aws-cli/issues/7043
-region=$(echo "${sandbox_image}" | cut -f4 -d ".")
+# if the sandbox image is provided by the bootstrap script, then the region is
+# guaranteed to come from this data source.
+# see: https://github.com/awslabs/amazon-eks-ami/blob/baef6f0860f60dbec366de30853e47418e3fb430/files/bootstrap.sh#L320-L338
+# if the image is customer provided, then this is just a sane default for the
+# region when attempting to get ecr credentials.
+region=$(imds 'latest/dynamic/instance-identity/document' | jq .region -r)
```
dims, line 22: one problem here is ... if the sandbox_image is specified from another region ( different from where the node is running in :( )

dims: could we check if the echo/cut version (from the sandbox_image) is present in `aws account list-regions`? and skip if not?

ndbaker1: oh but this would require new permissions on the role 😖

ndbaker1: are there docs stating required naming conventions of ecr repos? I'm not 100% sure what i can/can't check for

ndbaker1: > if the sandbox_image is specified from another region

but this is only for user provided images right (based on the assumptions for the eks generated uri), which I'm not sure had a solution even prior to this 🤔

dims: then we can let it slide and leave it to the `imds` variation. (cross fingers!)

cartermckinnon: I'm hesitant to touch this in this PR. Cutting the region out of the image reference has it's caveats, for sure; but that logic has been in prod for over a year and it's proven solid. I think we should consider this change in a different PR, since this one is mostly for mitigation

ndbaker1: > that logic has been in prod for over a year and it's proven solid.

This logic has existed yes (https://github.com/awslabs/amazon-eks-ami/blame/master/files/pull-image.sh#L4), but its only been used in context where the image source was already known and decided by us right?

The only prior usage i know if is in
https://github.com/awslabs/amazon-eks-ami/blob/baef6f0860f60dbec366de30853e47418e3fb430/scripts/install-worker.sh#L484

cartermckinnon: `pull-sandbox-image.sh` used to call `pull-image.sh` instead of using `crictl`

ndbaker1: 🤦 that escaped my mind (and my github searches)
yea i think pulling it out makes sense then since it won't break the expectation beyond whats already changed in last commit

cartermckinnon (approved):

PR #1658, opened 2024-02-12
Add API versioning policy doc

cartermckinnon:
**Description of changes:**

Adds an "API Concepts" doc that goes over the versioning policy.

### DIFF
--- a/nodeadm/doc/api-concepts.md
+++ b/nodeadm/doc/api-concepts.md
@@ -15,8 +15,7 @@ There are three levels of stability and support:
 ### Beta
 - Example: `v3beta4`.
 - The software is well-tested, and production use is considered safe.
-- Support for a beta API will remain for at least one release following deprecation. That is, support for two successive beta API versions will overlap by at least one release.
-- Features will not be dropped, though the details may change.
+- Support for a beta API will remain for at least one release following its deprecation.
 - Subsequent beta or stable API versions may include incompatible changes, and migration instructions will be provided.
 
 ### Stable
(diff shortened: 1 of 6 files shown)


nodeadm/doc/api-concepts.md
```diff
@@ -0,0 +1,23 @@
...
+
+### Beta
+- Example: `v3beta4`.
+- The software is well-tested, and production use is considered safe.
+- Support for a beta API will remain for at least one release following its deprecation.
+- Subsequent beta or stable API versions may include incompatible changes, and migration instructions will be provided.
+
+### Stable
+- Example: `v5`
+- Support for a stable API will remain at least until the release of the next major version of Amazon Linux.
```
ndbaker1, line 23: sounds like a reasonable promise, could there ever be a reason to break it?

nodeadm/doc/api-concepts.md
```diff
@@ -0,0 +1,23 @@
+# API Concepts
+
+## Versioning
+
+The API types for `nodeadm` (the `node.eks.aws` API group) are versioned in a similar manner to the [Kubernetes API](https://kubernetes.io/docs/reference/using-api/#api-versioning).
+
+There are three levels of stability and support:
+
+### Alpha
+- Example: `v1alpha2.`
```
ndbaker1, line 10: > - Example: `v1alpha2.`
> - Example: `v3beta4`.
> - Example: `v5`

punctuation is consistently different 👀

nodeadm/doc/api-concepts.md
cartermckinnon: I think I can clarify this, the idea is that a stable version of the API will be supported for the life of the AL major version at its time of introduction. E.g. `v1` will be supported for AL2023, but there's no guarantee it will be supported for AL2025. When AL2025 is released, we are permitted to require `v2` for AL2025 nodes.

ndbaker1: ah ok gotcha

ndbaker1 (approved):

Issue #1554, opened 2024-01-03
Encountering Permission Error and AWS command not found Building AMI


voidteatime:
**What happened**: 
Encountered errors on `generate-version-info.sh` when verifying kubelet and aws binary versions.
![image](https://github.com/awslabs/amazon-eks-ami/assets/81941914/9e8980dd-dc63-4969-8f10-bbc1517765a0)
AWS command not found:
![image](https://github.com/awslabs/amazon-eks-ami/assets/81941914/2c2efab6-b0ca-4937-b422-a93589912c80)

**What you expected to happen**:
Can complete the `generate-version-info.sh` without any issues for kubelet and aws cli version checking

**How to reproduce it (as minimally and precisely as possible)**:
I tried using RHEL 8.8 or the latest AML2, edited the variables to use containerd 1.6.* (for some reason 1.7.* isn't available for AML2)
cloned the latest repo and did `make 1.28`. Error was encountered on generate-version-info.sh

**Anything else we need to know?**:
The base image was AML2 CIS Level 2 hardened.
After investigating, inside the `install-worker.sh` script, the `chmod +x` command does not work properly.
Changed this to `chmod 755` and it works for the permission issue.
No workaround yet for the missing AWS command error. Have checked manually the instance and awscli v2 is indeed installed using ec2-user.

**Environment**:
- AWS Region: ap-southeast-1
- Instance Type(s): m5.large
- EKS Platform version (use `aws eks describe-cluster --name <name> --query cluster.platformVersion`): eks.6
- Kubernetes version (use `aws eks describe-cluster --name <name> --query cluster.version`): 1.28
- AMI Version: v20240103
- Kernel (e.g. `uname -a`):  4.18.0-513.9.1.el8_9.x86_64 
- Release information (run `cat /etc/eks/release` on a node):

```
```

voidteatime:
Adding more information here, I have a copy of the repo from last week of october, seems like the problems with `generate-version-info.sh` were allowed to continue before:
![image](https://github.com/awslabs/amazon-eks-ami/assets/81941914/5d8b2b34-3626-499a-b964-6df4f2c9dd06)

commits from 967fb36 were causing this issue.

fletcherw:
Closing this since it seems like an issue specifically with AL2. On AL23 we preserve the selinux context when installing the binary, so it should still be executable without sudo. Please feel free to reopen if the issue is still happen on AL23.

PR #1661, opened 2024-02-13
Fix build issue when using hardened AMIs as base images

dawidpichen:
**Issue #, if available:**
https://github.com/awslabs/amazon-eks-ami/issues/1554

**Description of changes:**
Two minor yet important changes:

1. I changed default version to latest 1.29 in `Makefile`;
2. I corrected `generate-version-info.sh` script so that it does not fail when Packer uses hardened AMIs. Refer to  https://github.com/awslabs/amazon-eks-ami/issues/1554 for more info.

**Testing Done**

Successfully built v1.29 EKS AMI based on hardened (CIS) version of AL2 Golden AMI.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -142,7 +142,7 @@ k8s: validate ## Build default K8s version of EKS Optimized AL2 AMI
 	$(MAKE) k8s $(shell hack/latest-binaries.sh 1.28)
 
 .PHONY: 1.29
-1.29: ## Build EKS Optimized AL2 AMI - K8s 1.28
+1.29: ## Build EKS Optimized AL2 AMI - K8s 1.29
 	$(MAKE) k8s $(shell hack/latest-binaries.sh 1.29)
 
 .PHONY: lint-docs
--- a/Makefile
+++ b/Makefile
@@ -141,10 +141,6 @@ k8s: validate ## Build default K8s version of EKS Optimized AL2 AMI
 1.28: ## Build EKS Optimized AL2 AMI - K8s 1.28
 	$(MAKE) k8s $(shell hack/latest-binaries.sh 1.28)
 
-.PHONY: 1.29
-1.29: ## Build EKS Optimized AL2 AMI - K8s 1.29
-	$(MAKE) k8s $(shell hack/latest-binaries.sh 1.29)
-
 .PHONY: lint-docs
 lint-docs: ## Lint the docs
 	hack/lint-docs.sh
--- a/Makefile
+++ b/Makefile
@@ -57,7 +57,7 @@ T_YELLOW := \e[0;33m
 T_RESET := \e[0m
 
 # default to the latest supported Kubernetes version
-k8s=1.28
+k8s=1.29
 
 .PHONY: build
 build: ## Build EKS Optimized AL2 AMI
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -141,6 +141,10 @@ k8s: validate ## Build default K8s version of EKS Optimized AL2 AMI
 1.28: ## Build EKS Optimized AL2 AMI - K8s 1.28
 	$(MAKE) k8s $(shell hack/latest-binaries.sh 1.28)
 
+.PHONY: 1.29
+1.29: ## Build EKS Optimized AL2 AMI - K8s 1.29
+	$(MAKE) k8s $(shell hack/latest-binaries.sh 1.29)
+
 .PHONY: lint-docs
```
cartermckinnon, line 148: We're not adding stub targets for every version going forward, you can do:
```
make k8s=1.29
```

scripts/generate-version-info.sh
```diff
@@ -16,14 +16,14 @@ OUTPUT_FILE="$1"
 sudo rpm --query --all --queryformat '\{"%{NAME}": "%{VERSION}-%{RELEASE}"\}\n' | jq --slurp --sort-keys 'add | {packages:(.)}' > "$OUTPUT_FILE"
 
 # binaries
-KUBELET_VERSION=$(kubelet --version | awk '{print $2}')
+KUBELET_VERSION=$(sudo kubelet --version | awk '{print $2}')
```
cartermckinnon, line 19: Can you just call the entire script with `sudo` in the template? https://github.com/awslabs/amazon-eks-ami/blob/8bb63f545e74e00eb3a2b919f147ad5cfce51725/eks-worker-al2.json#L227

dawidpichen: @cartermckinnon Unfortunately, it will not work correctly if you execute the template with `sudo`. Why? Because, the `generate-version-info.sh` would generate the `version-info.json` file but it will be owned by root. The following step will fail to download the file because of it's permissions (600). Keep in mind the I use hardened base AMI.

Execution logs:
```
2024-02-14T11:07:02Z: ==> amazon-ebs: Provisioning with shell script: /codebuild/output/src3189660426/src/repo/scripts/generate-version-info.sh
2024-02-14T11:07:04Z: ==> amazon-ebs: Downloading /tmp/worker/version-info.json => RYR-AL2-EKS-1.23-GOLDEN-AMI-v20240214-version-info.json
2024-02-14T11:07:05Z: ==> amazon-ebs: Download failed: scp: /tmp/worker/version-info.json: Permission denied
==> amazon-ebs:
2024-02-14T11:07:05Z: ==> amazon-ebs: Provisioning step had errors: Running the cleanup provisioner, if present...
2024-02-14T11:07:05Z: ==> amazon-ebs: Terminating the source AWS instance...
```

Makefile
dawidpichen: @cartermckinnon  Understood. Do you want me to remove others as well?

cartermckinnon: We don't want to break anyone by removing them, they'll just go away as the k8s versions fall out of support 👍

scripts/generate-version-info.sh
cartermckinnon: Ah, gotcha. I'm fine with adding `sudo` to the commands, just looks silly to `sudo` just to run `aws --version`...

Makefile
dawidpichen: Alright, in my most recent commit I removed _only_ the target added by me, i.e. 1.29.

scripts/generate-version-info.sh
dawidpichen: I agree, looks stilly but does the job, ha ha.

varun-armis:
@cartermckinnon @dawidpichen what's the status of this PR? It seems like all that's left is to do a rebase right?

This PR would also address others opened against the master branch: https://github.com/awslabs/amazon-eks-ami/pull/1513

dawidpichen:
@cartermckinnon Are you going to proceed with this PR?

varun-armis (approved):
bump

PR #1670, opened 2024-02-15
Allow setup/configure of daemons separately

dims:
Allows nodeadm to configure/run kubelet and containerd daemons separately. So this utility can be used for CI jobs outside of EKS (kubelet setup we do here is very specific to EKS but containerd is not).

### DIFF
--- a/nodeadm/cmd/nodeadm/init/init.go
+++ b/nodeadm/cmd/nodeadm/init/init.go
@@ -28,7 +28,7 @@ const (
 func NewInitCommand() cli.Command {
 	init := initCmd{}
 	init.cmd = flaggy.NewSubcommand("init")
-	init.cmd.StringSlice(&init.daemons, "d", "daemon", "specify one or more of `containerd` and `kubelet`")
+	init.cmd.StringSlice(&init.daemons, "d", "daemon", "specify one or more of `containerd` and `kubelet`. This is intended for testing and should not be used in a production environment.")
 	init.cmd.StringSlice(&init.skipPhases, "s", "skip", "phases of the bootstrap you want to skip")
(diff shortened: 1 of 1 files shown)


dims:
cc @ndbaker1 @cartermckinnon

ndbaker1 (approved):
i like it 👍

nodeadm/cmd/nodeadm/init/init.go
```diff
@@ -28,6 +28,7 @@ const (
 func NewInitCommand() cli.Command {
 	init := initCmd{}
 	init.cmd = flaggy.NewSubcommand("init")
+	init.cmd.StringSlice(&init.daemons, "d", "daemon", "specify one or more of `containerd` and `kubelet`")
```
cartermckinnon, line 31: I'm a bit hesitant to add this to the CLI, can we add a disclaimer that you probably don't want to do this? 

```suggestion
	init.cmd.StringSlice(&init.daemons, "d", "daemon", "specify one or more of `containerd` and `kubelet`. This is intended for testing and should not be used in a production environment.")
```

ndbaker1: > So this utility can be used for CI jobs outside of EKS

probing for more context, what specific usage do you have in mind?

ndbaker1:
> So this utility can be used for CI jobs outside of EKS

probing for more context, what specific usage do you have in mind?

dims:
@ndbaker1 LOL i already have a bunch of CI jobs that reuse this AMI and run tests in upstream k8s CI - https://testgrid.k8s.io/amazon-ec2-al2023

cartermckinnon (approved):

PR #1679, opened 2024-02-20
Add iam instance profile support

AlexSchultz-clumio:
**Issue #, if available:**

**Description of changes:**
In order to use SSM for building rather than SSH, we need to be able to configure the `iam_instance_profile`.  This change adds `iam_instance_profile` that can be used when setting `ssh_interfance` to `session_manager`.

When using `session_manager` you will need to use a non-minimal version of the source ami to ensure it has the SSM agent
available at launch. You may also need to increase the volume size from 4 depending on which source AMI you use.

**Testing Done**

Successful build with `make k8s=1.29 ssh_interface=session_manager iam_instance_profile=ssm-profile aws_region=us-east-2 source_ami_filter_name="amzn2-ami-hvm-*" launch_block_device_mappings_volume_size=8`

### DIFF
--- a/doc/usage/al2.md
+++ b/doc/usage/al2.md
@@ -26,6 +26,7 @@
 | `docker_version` | Docker is not installed on Kubernetes v1.25+ |
 | `enable_fips` | Install openssl and enable fips related kernel parameters |
 | `encrypted` |  |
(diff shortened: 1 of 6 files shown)


AlexSchultz-clumio:
@cartermckinnon please review when you have a moment

cartermckinnon (approved):

PR #1680, opened 2024-02-21
Add allowed-unsafe-sysctls option for bootstrap.sh

Issacwww:
**Issue #, if available:**

**Description of changes:**
Seeing customer reporting issue that they cannot use some sysctls via
`kubelet --allowed-unsafe-sysctls 'net.ipv4.tcp_timestamps'`

This is due to `--allowed-unsafe-sysctls ` is deprecated 
```
Flag --allowed-unsafe-sysctls has been deprecated, This parameter should be set via the config file specified by the Kubelet's --config flag. See https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/ for more information.
```
Thus moving add this flag to config it in AL2. 
AL2023 has no impact, adding tests to ensure the upstream configs can be correctly pick up.

And this should also align with Karpenter, https://github.com/aws/karpenter-provider-aws/issues/1279

**Testing Done**
for bootstrap.sh: `make test` 
for nodeadm: `make test && make test-e2e`

### DIFF
--- a/nodeadm/test/e2e/cases/kubelet-upstream-config-options/config.kubelet-config-json.yaml
+++ b/nodeadm/test/e2e/cases/kubelet-upstream-config-options/config.kubelet-config-json.yaml
@@ -1,0 +1,23 @@
+---
+apiVersion: node.eks.aws/v1alpha1
+kind: NodeConfig
+spec:
+  cluster:
+    name: my-cluster
+    apiServerEndpoint: https://example.com
+    certificateAuthority: Y2VydGlmaWNhdGVBdXRob3JpdHk=
+    cidr: 10.100.0.0/16
+  kubelet:
+    config: {
+      "logging": {
+        "verbosity": 5
+      },
+      "clusterDNS": [
+        "0.0.0.0",
+        "1.1.1.1"
+      ],
+      "allowedUnsafeSysctls": [
+        "kernel.msg*",
+        "net.core.somaxconn"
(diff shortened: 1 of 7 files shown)


nodeadm/test/e2e/cases/kubelet-upstream-config-options/config.kubelet-config-json.yaml
ndbaker1, line 1: We decided before that we dont need `json` configs for the rest of the tests. from the `kubelet-config` and `kubelet-config-dir` cases we confirm this works so you can just focus on the new fields.

nevertheless, do we need a separate test or can you just add this field to the existing config tests?

ndbaker1:
is there an example of the issue, i can't seem to find changelog notes or anything aside from the docs that mentions the behavior in config vs flags

nodeadm/test/e2e/cases/kubelet-upstream-config-options/config.kubelet-config-json.yaml
Issacwww: ah, lemme move it to existing cases

Issacwww:
Closing this PR as verified this is flag is still working, I was able to reproduce this locally and by using the launch template with `--kubelet-extra-args '--allowed-unsafe-sysctls=net.ipv4.tcp_timestamps'` the testing pod come into running

PR #1685, opened 2024-02-23
fixes #1684 support multiple clusterDNS for node-local when using ipvs

### LINKED ISSUE #1684: clusterDNS is overwritten via bootstrap.sh and does not support multiple IPs
**What happened**:
clusterDNS is overwritten via bootstrap.sh

support multiple IPs passing in the arg for cluster-dns.

**Environment**:
- AWS Region:
- Instance Type(s):
- EKS Platform version (use `aws eks describe-cluster --name <name> --query cluster.platformVersion`):
- Kubernetes version (use `aws eks describe-cluster --name <name> --query cluster.version`):
- AMI Version:
- Kernel (e.g. `uname -a`):

```
```

ajohnstone:
**Issue #, if available:**
#1684

**Description of changes:**
Support a comma separated list rather than a single scalar value.
bootstrap.sh will override clusterDNS in the kubelet config.

### Use case:
This will allow enabling IPVS and node-local by first having node-local and then fall back to the service ip for coredns.

Refer to docs:
https://kubernetes.io/docs/tasks/administer-cluster/nodelocaldns/

> If using kube-proxy in IPVS mode, --cluster-dns flag to kubelet needs to be modified to use <node-local-address> that NodeLocal DNSCache is listening on. Otherwise, there is no need to modify the value of the --cluster-dns flag, since NodeLocal DNSCache listens on both the kube-dns service IP as well as <node-local-address>.

**Testing Done**
Modified bootstrap and executed on a re-run.
Tests pass, should these not be part of the build workflow?
```
+ [[ 0 -eq 0 ]]
+ echo '✅ ✅ All Tests Passed! ✅ ✅'
✅ ✅ All Tests Passed! ✅ ✅
+ exit 0
```

### DIFF
--- a/templates/al2/runtime/bootstrap.sh
+++ b/templates/al2/runtime/bootstrap.sh
@@ -480,7 +480,7 @@ else
 fi
 
 KUBELET_CONFIG=/etc/kubernetes/kubelet/kubelet-config.json
-echo $(jq --arg DNS_CLUSTER_IP "$DNS_CLUSTER_IP"  '.clusterDNS=($DNS_CLUSTER_IP|split(","))' $KUBELET_CONFIG) > $KUBELET_CONFIG
+echo $(jq --arg DNS_CLUSTER_IP "$DNS_CLUSTER_IP" '.clusterDNS=($DNS_CLUSTER_IP|split(","))' $KUBELET_CONFIG) > $KUBELET_CONFIG
 
 if [[ "${IP_FAMILY}" == "ipv4" ]]; then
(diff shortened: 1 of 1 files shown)


cartermckinnon:
Sounds reasonable to me. The linter found an extra space that needs to be removed, @ajohnstone

ajohnstone:
@cartermckinnon thanks removed extra whitespace :)

ajohnstone:
@cartermckinnon mind triggering the tests and approval thanks :)

PR #1686, opened 2024-02-23
Do not prepare local disks by default

cartermckinnon:
**Description of changes:**

This makes the `setup-local-disks` service unit on AL2023 a no-op by default. A followup PR will add an option to `NodeConfig` to set the local disk strategy (or otherwise configure the underlying unit).

### DIFF
--- a/templates/al2023/runtime/rootfs/etc/systemd/system/setup-local-disks.service
+++ b/templates/al2023/runtime/rootfs/etc/systemd/system/setup-local-disks.service
@@ -4,7 +4,8 @@ Before=containerd.service kubelet.service
 
 [Service]
 Type=oneshot
-ExecStart=/usr/bin/setup-local-disks raid0
+Environment=LOCAL_DISK_STRATEGY=none
+ExecStart=/usr/bin/setup-local-disks $LOCAL_DISK_STRATEGY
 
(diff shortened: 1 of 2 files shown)


Issacwww (approved):

templates/al2023/runtime/rootfs/etc/systemd/system/setup-local-disks.service
```diff
@@ -4,7 +4,8 @@ Before=containerd.service kubelet.service
 
 [Service]
 Type=oneshot
-ExecStart=/usr/bin/setup-local-disks raid0
+Environment=LOCAL_DISK_STRATEGY=none
+ExecStart=/usr/bin/setup-local-disks $LOCAL_DISK_STRATEGY
```
ndbaker1, line 8: sounds like we are gonna do another `EnvironmentFile` created via nodeadm?

ndbaker1 (approved):

templates/al2023/runtime/rootfs/etc/systemd/system/setup-local-disks.service
cartermckinnon: Yep!

PR #1687, opened 2024-02-23
Add ability to specify a user data file

AlexSchultz-clumio:
**Issue #, if available:**

**Description of changes:**

Sometimes we need to execute some actions on the instance along with the eks build script but currently there is not a way to inject these additional steps as part of the existing build process. This change adds the ability to specify `user_data_file` which can be passed to the build instance on startup.

**Testing Done**

Built a new image with `make k8s=1.29 user_data_file=test.yaml`

The `test.yaml` contained several `bootcmd` and `runcmd` options to tweak the base image being used:
```
#cloud-config
bootcmd:
  # remove possible fips incompatible ciphers
  - sed -i -e 's/curve25519-sha256@libssh.org[,]*//' -e 's/curve25519-sha256[,]*//' /etc/ssh/sshd_config
  # remove ip_forward disabling from CIS image
  - sed -i -e '/net.ipv4.ip_forward/d' /etc/sysctl.conf
runcmd:
  # open access to kubelet
  - iptables -A INPUT -p tcp -m tcp --dport 10250 -m state --state NEW -j ACCEPT
  # open access for NodePorts
  - iptables -A INPUT -p tcp -m multiport --dports 30000:32767 -j ACCEPT
  # save updated iptables
  - bash -c "/sbin/iptables-save > /etc/sysconfig/iptables"
```

Deployed built AMI to cluster and verified pods were able to be schedule and use the node. Also verified the actions were run during the build process and the results showed up in the built AMI.

### DIFF
--- a/doc/usage/al2.md
+++ b/doc/usage/al2.md
@@ -45,6 +45,7 @@
 | `ssm_agent_version` | Version of the SSM agent to install from the S3 bucket provided by the SSM agent project, such as ```latest```. If empty, the latest version of the SSM agent available in the Amazon Linux core repositories will be installed. |
 | `subnet_id` |  |
 | `temporary_security_group_source_cidrs` |  |
+| `user_data_file` | Path to a file that will be used for the user data when launching the instance. |
 | `volume_type` |  |
 | `working_dir` | Directory path for ephemeral resources on the builder instance |
 <!-- template-variable-table-boundary -->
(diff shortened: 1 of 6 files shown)


cartermckinnon:
I think this is a great idea. Way better than patching in a custom provisioner or maintaining a full fork of the template.

@ndbaker1 @Issacwww please weigh in

AlexSchultz-clumio:
@cartermckinnon since this is an acceptable improvement, can we merge this?

AlexSchultz-clumio:
Got. Will fix thanks

cartermckinnon (approved):

PR #1689, opened 2024-02-24
Migrate CI to `kubetest2`, add AL2023

cartermckinnon:
**Description of changes:**

This updates the CI bot + workflows to use the EKS `kubetest2` deployers.


.github/workflows/ci-manual.yaml
ndbaker1, line 1: we're still using `eksctl` to [pull supported k8s versions](https://github.com/awslabs/amazon-eks-ami/pull/1689/files#diff-9d2ab28a7718195cd81484529ba18bb65e4f012aad6a2f57204be4472b60ac9eR48), which is fine IMO, but it made more sense when we were using `eksctl` to run tests. Do we have another way to pull k8s versions via an api yet?

ndbaker1 (approved):
LGTM, feels better removing the `launch` step middle-man 👏

Issacwww (approved):

PR #1690, opened 2024-02-27
[WIP] Validate external types with CRD machinery

cartermckinnon:
**Description of changes:**

This adds validation for the external API types using the same mechanism that the API server uses for CRD's. Validation directives like `required` are only expressed in the CRD spec, so we need to leverage the core validation in the API server to leverage it.


Issacwww (approved):

nodeadm/internal/api/bridge/decode.go
```diff
@@ -28,6 +28,9 @@ func DecodeNodeConfig(data []byte) (*internalapi.NodeConfig, error) {
 	if gvk.Group != api.GroupName {
 		return nil, fmt.Errorf("failed to decode %q, unexpected group: %s", gvk.Kind, gvk.Group)
 	}
+	if err := ValidateExternalType(obj, *gvk); err != nil {
```
ndbaker1, line 31: based on tests, this is breaking how we merge partial configs 😞
will we have to do the merging in json/yaml like kubelet? https://github.com/kubernetes/kubernetes/blob/master/cmd/kubelet/app/server.go#L314-L345

Issue #1536, opened 2023-12-04
Error fetching EKS binaries  from GovCloud


Sandeepsac:
**What happened**:

releated to #762
@cartermckinnon 
In gov-cloud we facing issue while using instance profile that while connecting with S3 in region specific we getting 403 forbitten

 
EKS version 1.27

![image (23)](https://github.com/awslabs/amazon-eks-ami/assets/136812799/093a1a54-cebb-4bde-be15-c4a5e554f581)

donovanrost:
I also run into this with GovCloud. The last I looked into it, the binaries are not available in a GovCloud bucket and you can't reach into the commerial `amazon-eks` bucket from a GovCloud region. To solve this on my end, I presync the binaries into a my own bucket in a GovCloud region using the following
```bash
#!/usr/bin/env bash

LOCAL_SYNC_DIR="bucket"
REMOTE_SOURCE_BUCKET="amazon-eks"
REMOTE_TARGET_BUCKET="amazon-eks-binaries"
KUBERNETES_MINOR_VERSION="1.28"
OS="linux"
ARCH="amd64"
FORCE_SYNC="false"


POSITIONAL_ARGS=()

while [[ $# -gt 0 ]]; do
  case $1 in
    -sb|--remote-source-bucket)
      REMOTE_SOURCE_BUCKET="$2"
      shift 
      shift 
      ;;
    -tb|--remote-target-bucket)
      REMOTE_TARGET_BUCKET="$2"
      shift 
      shift
      ;;
    -l|--local-sync-dir)
      LOCAL_SYNC_DIR="$2"
      shift
      shift
      ;;
    -m|--kubernetes-minor-version)
      KUBERNETES_MINOR_VERSION="$2"
      shift
      shift
      ;;
    -d|--dry-run)
      DRY_RUN="true"
      shift # past argument
      ;;
    -sp|--source-profile)
      SOURCE_PROFILE="$2"
      shift
      shift
      ;;
    -tp|--target-profile)
      TARGET_PROFILE="$2"
      shift
      shift
      ;;
    -sr|--source-region)
      SOURCE_REGION="$2"
      shift
      shift
      ;;
    -tr|--target-region)
      TARGET_REGION="$2"
      shift
      shift
      ;;
    -f|--force-sync)
      FORCE_SYNC="true"
      shift
      ;;
    -*|--*)
      echo "Unknown option $1"
      exit 1
      ;;
    *)
[...]

PR #1641, opened 2024-02-08
Support building for Gov.cloud only accounts

jtnord:
Fixes #1536, #1482 
obsoletes #1537

**Description of changes:**

if a user only has access to an AWS account in GovCloud then obtaining the binary versions from S3 will fail.

As this bucket is public we do not need any authentication so add --no-sign-request  and hardcode the region (which would not be the default region for any govCloud).

chose `us-west-2` as the region to use as this appears to be the
canonical location.   As it is just metdata using a region on the other
side of the world should not matter too much when the overhead of the
packer build is taken into account

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

**Testing Done**

ran ` make k8s=1.25 enable_fips=true source_ami_owners=045324592363 aws_region=us-gov-east-1` and verified make completed without error (and the previous error about kubernetes versions being unavailable was no longer present).

in order to repo the original change you need `aws`  cli setup with **only** an account in gov.cloud.

prior to this change

```
jnord@pc-jnord:/mnt/d/source/github/awslabs/amazon-eks-ami$ make k8s=1.25 enable_fips=true source_ami_owners=045324592363 aws_region=us-gov-east-1

An error occurred (AccessDenied) when calling the ListObjectsV2 operation: Access Denied
make k8s
make[1]: Entering directory '/mnt/d/source/github/awslabs/amazon-eks-ami'
packer validate -var-file /mnt/d/source/github/awslabs/amazon-eks-ami/eks-worker-al2-variables.json  -var aws_region='us-gov-east-1' -var source_ami_owners='045324592363' -var ami_component_description='(k8s: {{ user `kubernetes_version` }}, containerd: {{ user `containerd_version` }})' -var arch='x86_64' -var enable_fips='true' -var instance_type='m5.large' -var ami_name='amazon-eks-fips-node-.-v20240208' -var docker_version='none' /mnt/d/source/github/awslabs/amazon-eks-ami/eks-worker-al2.json
Error:

2 errors occurred:
        * required variable not set: kubernetes_version
        * required variable not set: kubernetes_build_date




make[1]: *** [Makefile:113: validate] Error 1
make[1]: Leaving directory '/mnt/d/source/github/awslabs/amazon-eks-ami'
make: *** [Makefile:64: build] Error 2
```

after this change

```
[...]

PR #1537, opened 2023-12-04
Adjusted install-worker to redirect GovCloud

reelieuglie:
**Issue #, if available:**
[1536](https://github.com/awslabs/amazon-eks-ami/issues/1536)

**Description of changes:**

Currently, the S3 bucket for the kubelet binary in us-gov-west-1 (and us-gov-east-1) returns an 403 - `Access Denied` HTTP Code when attempting to download it, when the `binary_bucket_region` variable is set to `us-gov-west-1` or `us-gov-east-1` instead of the default `us-west-2`. 

I added a redirect from `us-gov-west-1` and `us-gov-east-1` to `us-west-2` and `us-east-2` repectively. 

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

Step 1 - Adjusted file as noted above and in the diff (line numbers included)
```
279 # Updating binary_bucket_region variable to redirect from us-gov-west-1 and us-gov-east-1 to us-west-2 and us-east-2 respectively, as the buckets in GovCloud regions is returning Access Denieds
280 # https://github.com/awslabs/amazon-eks-ami/issues/1536
281 elif [ "$BINARY_BUCKET_REGION" = "us-gov-west-1" ]; then
282   BINARY_BUCKET_REGION="us-west-2"
283 elif [ "$BINARY_BUCKET_REGION" = "us-gov-east-1" ]; then
284   BINARY_BUCKET_REGION="us-east-2"
```
Without Line Numbers:
```
# Updating binary_bucket_region variable to redirect from us-gov-west-1 and us-gov-east-1 to us-west-2 and us-east-2 respectively, as the buckets in GovCloud regions is returning Access Denieds
# https://github.com/awslabs/amazon-eks-ami/issues/1536
elif [ "$BINARY_BUCKET_REGION" = "us-gov-west-1" ]; then
  BINARY_BUCKET_REGION="us-west-2"
elif [ "$BINARY_BUCKET_REGION" = "us-gov-east-1" ]; then
  BINARY_BUCKET_REGION="us-east-2"
fi
```
Step 2 - Change the `bucket_binary_region` variable in `eks-worker-al2-variables.json` (example provided assuming that the `amazon-eks-ami` the current working directory)
```
sed -i 's/\"binary_bucket_region\": \"us-west-2\"/\"binary_bucket_region\": \"us-gov-west-1\"/g' eks-worker-al2-variables.json 
```

[...]

PR #1718, opened 2024-03-08
allow s3 commands to work from non commercial (ie us-gov) environments

joelddiaz:
**Issue #, if available:**

N/A

**Description of changes:**

when running under something like us-gov-east-1, the latest-binaries.sh script fails b/c it tries to access the us-west-2 bucket from the wrong endpoints.

this can be avoided by setting AWS_ENDPOINT_URL_S3 to point to
us-west-2, but you still end up trying to use the gov-cloud creds in the
request which would fail with:

$ ./hack/latest-binaries.sh 1.29

An error occurred (InvalidToken) when calling the ListObjectsV2 operation:
The provided token is malformed or otherwise invalid.

so, specify to perform an unauthenticated s3 api request b/c the
govcloud creds wouldn't work against the commercial cloud endpoints.

in other places in the install-worker.sh script, there are 'aws s3'
commands that would fail if running under something like the
us-gov-east-1 environment.

similar to the changes to the latest-binaries.sh script, update the
'aws' cli calls to ensure the requests are unsinged (to avoid trying
to use us-gov creds against a non-gov endpoint).

**Testing Done**

Tried using the ./hack/latest-binaries.sh when using govcloud credentials to see things not error out.

Performed a full end-to-end AMI build when running in us-gov-east-1.

### DIFF
--- a/hack/latest-binaries.sh
+++ b/hack/latest-binaries.sh
@@ -13,6 +13,8 @@ MINOR_VERSION="${1}"
 
 # retrieve the available "VERSION/BUILD_DATE" prefixes (e.g. "1.28.1/2023-09-14")
 # from the binary object keys, sorted in descending semver order, and pick the first one
+# TODO: ideally we want to use the value of $binary_bucket_region instead of hard-coding us-west-2
+# since setting the binary_bucket_region value really is the indication of where the binaries are located.
 LATEST_BINARIES=$(aws s3api list-objects-v2 --bucket amazon-eks --prefix "${MINOR_VERSION}" --query 'Contents[*].[Key]' --output text --region us-west-2 --no-sign-request | cut -d'/' -f-2 | sort -Vru | head -n1)
 
 if [ "${LATEST_BINARIES}" == "None" ]; then


joelddiaz:
@ndbaker1 i went ahead and updated (and tested) the al2023 install-worker.sh script, and some entries in the docs

doc/usage/al2.md
```diff
@@ -14,6 +14,7 @@
 | `arch` |  |
 | `associate_public_ip_address` |  |
 | `aws_access_key_id` |  |
+| `aws_endpoint_url_s3` | Use specified S3 endpoint for S3 calls (useful when running in non-commercial AWS environments).  |
```
ndbaker1, line 17:
```suggestion
| `aws_endpoint_url_s3` | Use specified S3 endpoint for S3 calls (useful when running in non-commercial AWS environments). |
```

doc/usage/al2023.md
```diff
@@ -13,6 +13,7 @@
 | `arch` |  |
 | `associate_public_ip_address` |  |
 | `aws_access_key_id` |  |
+| `aws_endpoint_url_s3` | Use specified S3 endpoint for S3 calls (useful when running in non-commercial AWS environments).  |
```
ndbaker1, line 16:
```suggestion
| `aws_endpoint_url_s3` | Use specified S3 endpoint for S3 calls (useful when running in non-commercial AWS environments). |
```

ndbaker1 (commented):
thanks 👍 alas, the linter strikes again

joelddiaz:
hopefully the linter gods are pacified this time

ndbaker1:
/ci build
+workflow:os_distros al2,al2023
+workflow:k8s_versions 1.29

ndbaker1 (changes requested):

hack/latest-binaries.sh
```diff
@@ -13,7 +13,7 @@ MINOR_VERSION="${1}"
 
 # retrieve the available "VERSION/BUILD_DATE" prefixes (e.g. "1.28.1/2023-09-14")
 # from the binary object keys, sorted in descending semver order, and pick the first one
-LATEST_BINARIES=$(aws s3api list-objects-v2 --bucket amazon-eks --prefix "${MINOR_VERSION}" --query 'Contents[*].[Key]' --output text | cut -d'/' -f-2 | sort -Vru | head -n1)
+LATEST_BINARIES=$(aws s3api list-objects-v2 --bucket amazon-eks --prefix "${MINOR_VERSION}" --query 'Contents[*].[Key]' --output text --no-sign-request | cut -d'/' -f-2 | sort -Vru | head -n1)
```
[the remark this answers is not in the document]
joelddiaz: i added it for the hack/latest-binaries.sh. do you want this change for all s3 commands in install-worker.sh as well?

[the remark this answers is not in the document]
ndbaker1: oh, no the `$BINARY_BUCKET_REGION` is rightfully set in those cases. this one is static 👍

ndbaker1 (approved):

hack/latest-binaries.sh
```diff
@@ -13,7 +13,7 @@ MINOR_VERSION="${1}"
 
 # retrieve the available "VERSION/BUILD_DATE" prefixes (e.g. "1.28.1/2023-09-14")
 # from the binary object keys, sorted in descending semver order, and pick the first one
-LATEST_BINARIES=$(aws s3api list-objects-v2 --bucket amazon-eks --prefix "${MINOR_VERSION}" --query 'Contents[*].[Key]' --output text | cut -d'/' -f-2 | sort -Vru | head -n1)
+LATEST_BINARIES=$(aws s3api list-objects-v2 --bucket amazon-eks --prefix "${MINOR_VERSION}" --query 'Contents[*].[Key]' --output text --region us-west-2 --no-sign-request | cut -d'/' -f-2 | sort -Vru | head -n1)
```
cartermckinnon, line 16: Let's add a big `TODO` outlining the long term solution (being able to use `--region $binary_bucket_region`)

templates/al2023/template.json
```diff
@@ -176,6 +177,7 @@
       "script": "{{template_dir}}/provisioners/install-worker.sh",
       "environment_vars": [
         "AWS_ACCESS_KEY_ID={{user `aws_access_key_id`}}",
+	"AWS_ENDPOINT_URL_S3={{user `aws_endpoint_url_s3`}}",
```
cartermckinnon, line 180: The `BINARY_BUCKET_REGION` should be `us-west-2` in govcloud for the time being. I think we need to stop using that variable as a proxy for `AWS_REGION` (the runtime region) when making other partition-specific decisions

hack/latest-binaries.sh
cartermckinnon: This is also forcing customers in `aws-cn` to query the bucket in `aws`, which isn't great.

templates/al2023/template.json
joelddiaz: yeah, now that the region is explicitly set there is no real reason to need to override the S3 endpoint. i've added a commit to remove all the AWS_ENDPOINT_URL_S3 pieces

bradwatsonaws:
I certainly appreciate the GovCloud support! But one thing I don't understand about this process is that the hack/latest-binaries.sh script relies on the AWS CLI for execution while the install-worker.sh scripts all attempt to install the AWS CLI long after the fact. If the AWS CLI isn't installed ahead of time, the process will fail from the start in GovCloud. @cartermckinnon had a very good workaround [here](https://github.com/awslabs/amazon-eks-ami/pull/1537#pullrequestreview-1818730169) to use curl in the event of the AWS CLI command failing.

Is the AWS CLI installation section of the install-worker.sh script really just an attempt to upgrade vs install? If so - ok that makes more sense. From my experience I see that the Al2023 minimal AMIs come with the AWS CLI preinstalled but the AL2 minimal AMIs DO NOT, meaning this approach would not work in GovCloud for AL2. Am I off here?

joelddiaz:
> I certainly appreciate the GovCloud support! But one thing I don't understand about this process is that the hack/latest-binaries.sh script relies on the AWS CLI for execution while the install-worker.sh scripts all attempt to install the AWS CLI long after the fact. If the AWS CLI isn't installed ahead of time, the process will fail from the start in GovCloud. @cartermckinnon had a very good workaround [here](https://github.com/awslabs/amazon-eks-ami/pull/1537#pullrequestreview-1818730169) to use curl in the event of the AWS CLI command failing.
> 
> Is the AWS CLI installation section of the install-worker.sh script really just an attempt to upgrade vs install? If so - ok that makes more sense. From my experience I see that the AL2023 minimal AMIs come with the AWS CLI preinstalled but the AL2 minimal AMIs DO NOT, meaning this approach would not work in GovCloud for AL2. Am I off here?

I feel like I'm missing something here regarding the questions...

hack/latest-binaries.sh is a piece that is only used to run locally before kicking of the packer build step(s) for the sole purpose of passing some variables into the packer environment, and the install-worker.sh is the software installation that happens inside the instance/AMI we're building.

I've mostly been working with AL2, and indeed there is no AWS CLI pre-installed on the base-AMI images I've been using.

bradwatsonaws:
Ah yes, you are correct @joelddiaz. I conflated the usage. My apologies and again appreciate the GovCloud love here!

hack/latest-binaries.sh
jtnord: per https://github.com/awslabs/amazon-eks-ami/pull/1641/files#r1486814106

whilst this is not "great" it is also not bad.  This is just retrieving metadata - and even if that takes an extra 5 seconds, it disappears into the time taken to actually build the image.

jtnord:
would presumably 
Fix #1536, #1482
and obsolete #1537, #1641

mebays:
Is this still actively getting worked on. I would support this change. The buckets seem to only be in us-west-2.  The aws cli errors off in gov cloud because it won't find a solution for amazon-eks.s3.us-gov-east-1.amazonaws.com, because there is no redirect or bucket in gov cloud. If hitting the endpoint directly with curl to another region amazon-eks.s3.us-east-1.amazonaws.com that ends up being in us-west-2 regardless because it returns a Permanent Redirect to us-west-2.  

If you care to look at the debug logs for awscli `aws s3api list-objects-v2 --region us-east-1 --bucket amazon-eks --debug` will attempt the us-east-1 region get the redirect message grab the Endpoint url (amazon-eks.s3-us-west-2.amazonaws.com) from the message and retry with the retrieved endpoint.

bradwatsonaws:
Any updates on this PR? hack/latest-binaries.sh has been broken in GovCloud for over 9 months now without using a workaround. @joelddiaz @ndbaker1 @cartermckinnon

fletcherw:
Closing this in favor of #2621.

PR #1717, opened 2024-03-08
update install-worker.sh to chmod under more restrictive uname configurations

joelddiaz:
**Issue #, if available:**

N/A

**Description of changes:**

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

when running under a more restrictive umask, the `chmod +x` wouldn't set the r-x bits for the 'other' group (or potentially the group permissions too depending on the umask). make the chmod more explicit.

for example `chmod +x` under a more restrictive umask ends up without execute pemissions which is likely not what was intended:

```
[jo.diaz@\ ~]$ umask 0027
[jo.diaz@\ ~]$ touch testfile
[jo.diaz@\ ~]$ chmod +x testfile
[jo.diaz@\ ~]$ ls -l testfile
-rwxr-x--- 1 jo.diaz jo.diaz 0 Mar  8 11:39 testfile
```

where under a more typical umask things look more familiar:

```
[jo.diaz@\ ~]$ umask 0002
[jo.diaz@\ ~]$ touch permissive
[jo.diaz@\ ~]$ chmod +x permissive
[jo.diaz@\ ~]$ ls -l permissive
-rwxrwxr-x 1 jo.diaz jo.diaz 0 Mar  8 11:40 permissive
```

so, to accommodate using a base AMI with more a restrictive umask, update the `chmod +x`'s to be more explicit about what is intended.

without these changes, we have seen things error out during the build

```
2024-03-08T12:02:27-05:00: ==> amazon-ebs: Provisioning with shell script:
/home/jo.diaz/projects/amazon-eks-ami/templates/al2/../shared/provisioners/generate-version-info.sh
2024-03-08T12:02:28-05:00:     amazon-ebs:
/home/ec2-user/script_2147.sh: line 19: /usr/bin/kubelet: Permission
denied
2024-03-08T12:02:28-05:00: ==> amazon-ebs: Provisioning step had errors:
Running the cleanup provisioner, if present...
```

**Testing Done**

Tried a build with an AMI base image with a restrictive umask, and avoided the errors described above.

PR #1719, opened 2024-03-08
account for the possibility of there being no aws cli binary

joelddiaz:
**Issue #, if available:**

N/A

**Description of changes:**

the generate-version-info.sh script presently assumes that the aws binary will be present. but this isn't always the case depending on the base AMI being used. allow for the possibility that we will not have an aws binary so the script doesn't error out.

otherwise we end up with errors like so:

2024-03-08T13:48:36-05:00: ==> amazon-ebs: Provisioning with shell script:
/home/jo.diaz/projects/amazon-eks-ami/templates/al2/../shared/provisioners/generate-version-info.sh
2024-03-08T13:48:37-05:00:     amazon-ebs:
/home/ec2-user/script_1364.sh: line 26: aws: command not found
2024-03-08T13:48:37-05:00: ==> amazon-ebs: Provisioning step had errors:
Running the cleanup provisioner, if present...
2024-03-08T13:48:37-05:00: ==> amazon-ebs: Terminating the source AWS
instance...

**Testing Done**

Ran a build with a base AMI that didn't include the aws CLI w/o running into this issue.

### DIFF
--- a/templates/shared/provisioners/generate-version-info.sh
+++ b/templates/shared/provisioners/generate-version-info.sh
@@ -23,12 +23,14 @@ if [ "$?" != 0 ]; then
 fi
 echo $(jq ".binaries.kubelet = \"$KUBELET_VERSION\"" $OUTPUT_FILE) > $OUTPUT_FILE
 
-CLI_VERSION=$(aws --version | awk '{print $1}' | cut -d '/' -f 2)
+CLI_VERSION=$(aws --version | awk '{print $1}' | cut -d '/' -f 2 || echo "unknown")
 if [ "$?" != 0 ]; then
   echo "unable to get aws cli version"
   exit 1
 fi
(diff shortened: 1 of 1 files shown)


cartermckinnon:
A previous provisioner installs the AWS CLI, so I think you're actually running into a `PATH` problem. We *do* want the build to fail if the CLI isn't able to be installed for some reason -- it's a runtime dependency of the bootstrap scripts.

joelddiaz:
> A previous provisioner installs the AWS CLI, so I think you're actually running into a `PATH` problem. We _do_ want the build to fail if the CLI isn't installed successfully -- it's a runtime dependency of the bootstrap scripts.

I see now what was going on. Installing the aws CLI from the downloaded zip file (instead of from the yum repo) leaves the aws CLI unusable because of the default umask on the AMI base image being used. I've updated the PR #1717 with the other umask changes to also include changes so that the aws CLI isn't installed in a broken state.

Closing this one out..

PR #1726, opened 2024-03-14
support os_distro and version for CI job

Issacwww:
**Issue #, if available:**

**Description of changes:**
Extend the CI job to accept os_distro and version to be frugal 

**Testing Done**

Tested in my folk https://github.com/Issacwww/amazon-eks-ami/pull/3

### DIFF
--- a/.github/workflows/ci-manual.yaml
+++ b/.github/workflows/ci-manual.yaml
@@ -25,9 +25,18 @@ on:
         options:
           - "build"
           - "test"
+      os_distro:
+        description: 'Operating System Distributions (comma-separated, e.g., al2,al2023)'
+        required: false
+        type: string
+      kubernetes_versions:
+        description: 'Kubernetes Versions (comma-separated, e.g., 1.21,1.22)'
+        required: false
+        type: string
       build_arguments:
         required: false
         type: string
+
 jobs:
   setup:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


cartermckinnon:
I like the change, but don't love the syntax 😛 . What if we used the `+` syntax like we do for goal-specific arguments, and handled a `workflow` "goal" or something else to pass arbitrary arguments to the `ci-manual.yaml` workflow?

```
/ci
+workflow matrix_os_distros=al2
+workflow matrix_k8s_versions=1.21,1.22
```

.github/workflows/ci-manual.yaml
```diff
@@ -25,9 +25,18 @@ on:
...
           - "build"
           - "test"
+      os_distro:
+        description: 'Operating System Distributions (comma-separated, e.g., al2,al2023)'
+        required: false
+        type: string
+      kubernetes_versions:
+        description: 'Kubernetes Versions (comma-separated, e.g., 1.21,1.22)'
+        required: false
+        type: string
```
cartermckinnon, line 35: I think we want defaults to be defined here instead of in the bot

.github/workflows/ci-manual.yaml
```diff
@@ -77,7 +85,12 @@ jobs:
       fail-fast: false
       matrix:
         k8s_version: ${{ fromJson(needs.setup.outputs.kubernetes_versions) }}
-        os_distro: [al2, al2023]
+        os_distro: ${{ fromJson(needs.setup.outputs.os_distro) }}
+        exclude:
+          - os_distro: al2023
+            k8s_version: 1.21
+          - os_distro: al2023
+            k8s_version: 1.22
```
cartermckinnon, line 93: what's this for?

Issacwww:
Fair fair, I will update the syntax

.github/workflows/ci-manual.yaml
Issacwww: This is going to skip the unsupported combination from the matrix. So we don't need such validation in the bot

Issacwww:
I think this format would be good, as the later arguments would override the earlier if we pass then in multiple lines, [here](https://github.com/Issacwww/amazon-eks-ami/blob/69a35707693a56310b23b864ae955b88075abe0e/.github/actions/bot/index.js#L165). Change this to a list is not that necessary as `build_arguments` also take multi args in one line
```
/ci
+workflow matrix_os_distros=al2,al2023 matrix_k8s_versions=1.21,1.22,1.23
```

.github/workflows/ci-manual.yaml
cartermckinnon: Can you add a comment explaining that al2023 isn't supported on these k8s versions?

Issacwww:
Updated new revision with multi-line arguments, tested [here](https://github.com/Issacwww/amazon-eks-ami/pull/3#issuecomment-1998634781)

cartermckinnon (approved):

PR #1731, opened 2024-03-19
Add InstanceIdNodeName feature gate

Issacwww:
**Issue #, if available:**

**Description of changes:**
Opt-in Instance Id as Node name for Launch template case

**Testing Done**


Issacwww:
/ci
+workflow:os_distros al2023

nodeadm/internal/kubelet/config.go
```diff
@@ -249,9 +249,14 @@ func (ksc *kubeletConfig) withCloudProvider(kubeletVersion string, cfg *api.Node
...
-		// the name of the Node object must equal the EC2 PrivateDnsName
-		// see: https://github.com/awslabs/amazon-eks-ami/pull/1264
-		flags["hostname-override"] = cfg.Status.Instance.PrivateDNSName
+		if cfg.Spec.Instance.NodeName.Strategy == api.InstanceIDName {
+			zap.L().Info("Opt-in Instance Id naming strategy")
+			flags["hostname-override"] = cfg.Status.Instance.ID
+		} else {
+			// the name of the Node object default to EC2 PrivateDnsName
+			// see: https://github.com/awslabs/amazon-eks-ami/pull/1264
+			flags["hostname-override"] = cfg.Status.Instance.PrivateDNSName
```
ndbaker1, line 258: should we also remove the api call to propogate the private dns name when nodename strategy is instanceID?

nodeadm/api/v1alpha1/nodeconfig_types.go
```diff
@@ -79,6 +79,7 @@ type ContainerdOptions struct {
 // InstanceOptions determines how the node's operating system and devices are configured.
 type InstanceOptions struct {
 	LocalStorage LocalStorageOptions `json:"localStorage,omitempty"`
+	NodeName     NodeNameOptions     `json:"nodeName,omitempty"`
```
cartermckinnon, line 82: I think we should avoid making this so strongly-typed. We don't want this to be an option in the long term, we just need a way to opt-in to the behavior before it becomes the *only* behavior on AL2025.

Something like @ndbaker1 's original feature flags approach makes more sense to me, because those values are not codified in the schema.

nodeadm/internal/feature-gates/feature-gates.go
```diff
@@ -0,0 +1,18 @@
+package featuregates
+
+type FeatureGate = string
+
+const (
+	// InstanceIdNamedNode When this flag is enabled, override the hostname with instanceId
+	InstanceIdNamedNode = FeatureGate("instanceIdNamedNode")
```
ndbaker1, line 7: nit: are feature gate keys usually lowercase or uppercase? kubelet's are uppercase

Issacwww: ack, updating it

nodeadm/internal/api/status.go
```diff
@@ -30,9 +31,12 @@ func GetInstanceDetails(ctx context.Context, imdsClient *imds.Client, ec2Client
 		return nil, err
 	}
 
-	privateDNSName, err := getPrivateDNSName(ec2Client, instanceIdenitityDocument.InstanceID)
-	if err != nil {
-		return nil, err
+	privateDNSName := ""
+	if featuregates.DefaultTrue(featuregates.InstanceIdNamedNode, featureGates) {
```
ndbaker1, line 35: wait a second, this is an opt-in behavior so we should be using `DefaultFalse` in both usages

```suggestion
	if !featuregates.DefaultFalse(featuregates.InstanceIdNamedNode, featureGates) {
```

Issacwww: kk, `! + defaultFalse` should be more readable

nodeadm/internal/feature-gates/feature-gates.go
ndbaker1, line 7: im curious, should we further abstract these to avoid thinking about the mode in several places?

something like 
```golang
type FeatureGate struct {
  Key string
  Verifier func(map[string]bool) bool
}

func (fg FeatureGate) IsEnabled(featureGates map[FeatureGate]bool) bool {
  return fg.Verifier(fg.Key, featureGates)
}

var InstanceIdNamedNode = FeatureGate{
  Key: "instanceIdNamedNode",
  Verifier: DefaultTrue,
}
```

nodeadm/internal/api/status.go
ndbaker1: its also behavioral since `DefaultTrue == !set || enabled` != `!DefaultFalse == !(set && enabled) == !set || !enabled` 😢

nodeadm/internal/feature-gates/feature-gates.go
Issacwww: good idea, updated

ndbaker1: this works for now, but its also just a suggestion. not sure if we'll need more complicated handling like differences across versions

ndbaker1 (commented):
LGTM but im not confident i like the name `InstanceIdNamedNode`

Issacwww:
How about `EnableEC2InstanceIdNaming`?

ndbaker1:
> How about `EnableEC2InstanceIdNaming`?

I think it was already close, like `InstanceIdNodeName`

just referencing [kubelet docs](https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/) and seems like the pattern is usually `*Node*`

cartermckinnon:
+1 for `InstanceIdNodeName`

nodeadm/internal/api/status.go
```diff
@@ -3,6 +3,7 @@ package api
 import (
 	"context"
 	"fmt"
+	featuregates "github.com/awslabs/amazon-eks-ami/nodeadm/internal/feature-gates"
```
cartermckinnon, line 6: if you name the directory `featuregates` or just `features` instead of `feature-gates` then you don't need to rewrite the import every time 😉

nodeadm/internal/kubelet/config.go
```diff
@@ -249,9 +250,14 @@ func (ksc *kubeletConfig) withCloudProvider(kubeletVersion string, cfg *api.Node
...
-		// the name of the Node object must equal the EC2 PrivateDnsName
-		// see: https://github.com/awslabs/amazon-eks-ami/pull/1264
-		flags["hostname-override"] = cfg.Status.Instance.PrivateDNSName
+		if featuregates.InstanceIdNodeName.IsEnabled(cfg.Spec.FeatureGates) {
+			zap.L().Info("Opt-in Instance Id naming strategy")
+			flags["hostname-override"] = cfg.Status.Instance.ID
+		} else {
+			// the name of the Node object default to EC2 PrivateDnsName
+			// see: https://github.com/awslabs/amazon-eks-ami/pull/1264
+			flags["hostname-override"] = cfg.Status.Instance.PrivateDNSName
```
cartermckinnon, line 259:
```suggestion
		var nodeName string
		if featuregates.InstanceIdNodeName.IsEnabled(cfg.Spec.FeatureGates) {
			zap.L().Info("Opt-in Instance Id naming strategy")
			nodeName = cfg.Status.Instance.ID
		} else {
			// the name of the Node object default to EC2 PrivateDnsName
			// see: https://github.com/awslabs/amazon-eks-ami/pull/1264
			nodeName = cfg.Status.Instance.PrivateDNSName
		}
		flags["hostname-override"] = nodeName
```

cartermckinnon (commented):
Couple small things, overall looks fine to me.

nodeadm/api/v1alpha1/nodeconfig_types.go
```diff
@@ -29,10 +29,11 @@ type NodeConfigList struct {
...
 type NodeConfigSpec struct {
-	Cluster    ClusterDetails    `json:"cluster,omitempty"`
-	Containerd ContainerdOptions `json:"containerd,omitempty"`
-	Instance   InstanceOptions   `json:"instance,omitempty"`
-	Kubelet    KubeletOptions    `json:"kubelet,omitempty"`
+	Cluster      ClusterDetails    `json:"cluster,omitempty"`
+	Containerd   ContainerdOptions `json:"containerd,omitempty"`
+	Instance     InstanceOptions   `json:"instance,omitempty"`
+	Kubelet      KubeletOptions    `json:"kubelet,omitempty"`
+	FeatureGates map[string]bool   `json:"featureGates,omitempty"`
```
ndbaker1, line 36: needs a doc description?

Issacwww: Yes, doc for `FeatureGates` will be easy, but I think we will also need a one for the configurable features and another guidance doc on how to use this flag for node name, I'm still working on test and writing up that doc.

Issacwww:
/ci
+workflow:os_distros al2023

nodeadm/internal/featuregates/featuregates.go
```diff
@@ -0,0 +1,35 @@
...
+	enabled, set := featureGates[key]
+	return set && enabled
+}
+
+func (fg FeatureGate) IsEnabled(featureGates map[Feature]bool) bool {
+	return fg.Verifier(fg.Key, featureGates)
+}
+
+// InstanceIdNodeNameGate controls whether to use instance ID as node name for AL2023 node group.
+// By default, this feature is disabled, and the private DNS Name will be used.
```
ndbaker1, line 31:
```suggestion
// InstanceIdNodeNameGate controls whether to use instance ID as the node's name.
// By default, this feature is disabled, and the private DNS Name will be used.
```

I think the distro specific details are unrelated to the feature gate

nodeadm/internal/featuregates/featuregates.go
```diff
@@ -0,0 +1,35 @@
...
+var InstanceIdNodeNameGate = FeatureGate{
+	Key:      InstanceIdNodeName,
```
ndbaker1, line 33: do we need the const to `InstanceIdNodeName`? I don't see it used and we can always use the `FeatureGate`'s `.Key` field

nodeadm/api/v1alpha1/nodeconfig_types.go
```diff
@@ -98,3 +100,12 @@ const (
 	// LocalStorageMount will mount each local disk individually
 	LocalStorageMount LocalStorageStrategy = "Mount"
 )
+
+// Feature specifies which feature gate should be toggle on
```
ndbaker1, line 104:
```suggestion
// Feature specifies which feature gate should be toggled
```

ndbaker1 (commented):
LGTM, some nits for docs

nodeadm/doc/examples.md
```diff
@@ -61,6 +61,38 @@ spec:
 
 The configuration objects will be merged in the order they appear in the MIME multi-part document, meaning the value in the lattermost configuration object will take precedence.
 
+--- 
+## Configuring node name
+
+We introduced an experiment feature in AL2023 AMI that allows you to name your node with EC2 instance name instead of EC2 private DNS name.
```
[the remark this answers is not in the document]
Issacwww: experimental title looks good! I think we can still mention this is for AL23, as some one may ask this feature for AL2 as well

nodeadm/internal/featuregates/featuregates.go
Issacwww: ack, removing it

nodeadm/doc/examples.md
```diff
@@ -61,6 +61,38 @@ spec:
 
 The configuration objects will be merged in the order they appear in the MIME multi-part document, meaning the value in the lattermost configuration object will take precedence.
 
+---
+## Configuring node name (experimental)
```
cartermckinnon, line 65: Clearer title since we aren't talking about setting the node name to an arbitrary value:
```suggestion
## Using instance ID as node name (experimental)
```

nodeadm/doc/examples.md
```diff
@@ -61,6 +61,38 @@ spec:
...
+
+We introduced the ability to name nodes using their EC2 instance id instead of EC2 private DNS name for AL2023.
```
cartermckinnon, line 67: I think we should talk about *why* this is a thing rather than just say it exists 😉 
```suggestion
When the `InstanceIdNodeName` feature gate is enabled, `nodeadm` will use the EC2 instance's ID (e.g. `i-abcdefg1234`) as the name of the `Node` object created by `kubelet`, instead of the EC2 instance's private DNS Name (e.g. `ip-192-168-1-1.ec2.internal`).

There are several benefits of doing this:
1. Your `Node` names are more meaningful in, for example, the output of `kubectl get nodes`.
2. The `Node` name, which is in the critical path of `kubelet` authentication, is non-volatile. While the private DNS name of an instance may change, its ID cannot.
3. The `ec2:DescribeInstances` permission can be removed from your node role's IAM policy; this is no longer necessary.
```

nodeadm/internal/featuregates/featuregates.go
```diff
@@ -0,0 +1,31 @@
+package featuregates
+
+type Feature string
```
cartermckinnon, line 3: This type should be in the `internal/api` package, since it mirrors an external API type. The `FeatureGate` struct and its helpers can stay here

cartermckinnon (commented):
Few small things

nodeadm/internal/featuregates/featuregates.go
Issacwww: `internal/api` has dependency on `FeatureGate`, move `Feature` to there will introduce the cycle dependency.

cartermckinnon: Makes sense to me!

nodeadm/internal/api/featuregates.go
```diff
@@ -0,0 +1,32 @@
+//go:build !ignore_autogenerated
```
Issacwww, line 1: Got this error form build after moving the feature gate to internal/api package.
```
/Volumes/workplace/githubForks/amazon-eks-ami/nodeadm/internal/api/featuregates.go:8:2: invalid field type: func(github.com/awslabs/amazon-eks-ami/nodeadm/internal/api.Feature, map[github.com/awslabs/amazon-eks-ami/nodeadm/internal/api.Feature]bool) bool
```
Thought this can be workaround by ignoring the featuregate file, by adding , but not sure if this is the correct way

cartermckinnon (approved):

PR #1732, opened 2024-03-19
fix context timeout associated with imds apis

ndbaker1:
**Issue #, if available:**

**Description of changes:**

the ec2imds apis fail fast in the `GetUserData` call, because of the context driven error:
```
operation error ec2imds: GetUserData, canceled, context deadline exceeded
```

notice that we only pass `context.TODO()` into this call, but the underlying middleware for the aws sdk call to imds adds its own deadline if one is missing or trivially set as 0.
https://github.com/aws/aws-sdk-go-v2/blob/main/feature/ec2/imds/request_middleware.go#L268-L280

> the timeout override that gets used is [5 seconds](https://github.com/aws/aws-sdk-go-v2/blob/main/feature/ec2/imds/request_middleware.go#L248C1-L248C43)

in this PR we are setting a context timeout greater than the retry period.

**Testing Done**

add e2e tests to help with this validation.


Issacwww (approved):

cartermckinnon:
This really sounds like a bug in the SDK to me: https://github.com/aws/aws-sdk-go-v2/issues/1247

IIUC the SDK is overriding the context we pass and as a result, making our retry options useless?

nodeadm/test/e2e/helpers.sh
```diff
@@ -102,9 +102,9 @@ function wait::dbus-ready() {
 
 function mock::aws() {
   local CONFIG_PATH=${1:-/etc/aemm-default-config.json}
-  imds-mock --config-file $CONFIG_PATH &
+  [ "${IMDS_MOCK_ONLY_CONFIGURE:-}" = "true" ] || imds-mock --config-file $CONFIG_PATH &
   export AWS_EC2_METADATA_SERVICE_ENDPOINT=http://localhost:1338
-  $HOME/.local/bin/moto_server -p5000 &
+  [ "${AWS_MOCK_ONLY_CONFIGURE:-}" = "true" ] || $HOME/.local/bin/moto_server -p5000 &
```
cartermckinnon, line 107: maybe just split this into 2 helpers, `mock::imds` and `mock::aws`? I combined them so that our standard set of mocks didn't get unruly, but it's not that big of a deal

ndbaker1:
> This really sounds like a bug in the SDK to me, we're not the first ones to run into it: https://github.com/aws/aws-sdk-go-v2/issues/1247
> IIUC the SDK is overriding the context we pass and as a result, making our retry options useless?

right, the underlying middleware for this client pretty much throws a context timeout onto your request, [xref](https://github.com/aws/aws-sdk-go-v2/blob/91a5e15e5218b768be724b71760e4523e32f1e2a/feature/ec2/imds/request_middleware.go#L251-L253). 
Not sure what they were thinking since its explicitly added [here](https://github.com/aws/aws-sdk-go-v2/blob/main/feature/ec2/imds/request_middleware.go#L57-L63), but it's not something i would expect the sdk to do

nodeadm/test/e2e/helpers.sh
ndbaker1: i was considering it, but I would probably still leave an `ONLY_CONFIGURE` option so that i can enable the service part way through the run

ndbaker1: there's not much pressure since this is a pretty unique test so lets not worry about it for now :)

ndbaker1:
updated the sdk to the versions with the fix, then also just moved the userdata helper into the imds wrapper

nodeadm/internal/aws/imds/imds.go
```diff
@@ -3,14 +3,22 @@ package imds
...
 
 var client *imds.Client
 
 func init() {
-	client = imds.New(imds.Options{})
+	client = imds.New(imds.Options{
+		DisableDefaultTimeout: true,
+		Retryer: retry.NewStandard(func(so *retry.StandardOptions) {
+			so.MaxAttempts = 15
+			so.MaxBackoff = 1 * time.Second
```
cartermckinnon, line 19: this means 1 second between each attempt, right?

ndbaker1: yea, the underlying exponential retryer just behaves linearly for 1 second backoff

cartermckinnon (approved):
LGTM

PR #1742, opened 2024-04-01
Fall back to previous K8s version if the latest Addons are missing

Cason33:
**Description of changes:**
If latest addons are not available, then we need to install the worker with the latest available.

I use the current K8s minor version to get the previous one, and after we get the Addons I check whether or not it actually got anything. If it didn't I got the prior Addons which will exist.

**Testing Done**

### DIFF
--- a/templates/al2/provisioners/install-worker.sh
+++ b/templates/al2/provisioners/install-worker.sh
@@ -392,7 +392,7 @@ if [[ "$CACHE_CONTAINER_IMAGES" == "true" ]] && ! [[ ${ISOLATED_REGIONS} =~ $BIN
   sudo systemctl enable containerd
 
   K8S_MINOR_VERSION=$(echo "${KUBERNETES_VERSION}" | cut -d'.' -f1-2)
-  PREVIOUS_K8S_MINOR_VERSION=$(echo ${KUBERNETES_MINOR_VERSION:0:1}."$((${KUBERNETES_MINOR_VERSION:2:3} - 1))")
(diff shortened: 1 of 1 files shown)


templates/al2/provisioners/install-worker.sh
```diff
@@ -392,9 +392,13 @@ if [[ "$CACHE_CONTAINER_IMAGES" == "true" ]] && ! [[ ${ISOLATED_REGIONS} =~ $BIN
   sudo systemctl enable containerd
 
   K8S_MINOR_VERSION=$(echo "${KUBERNETES_VERSION}" | cut -d'.' -f1-2)
+  PREVIOUS_K8S_MINOR_VERSION=$(echo ${KUBERNETES_MINOR_VERSION:0:1}."$((${KUBERNETES_MINOR_VERSION:2:3} - 1))")
```
cartermckinnon, line 395: I think we want to just skip image caching in this case; this will only happen for a brief period early in the version support cycle, before the version is GA on EKS

Cason33: Okay cool! So then the correct behavior should be:
If cached results is nil, then we skip everything up to the [SSM agent](https://github.com/awslabs/amazon-eks-ami/blob/main/templates/al2/provisioners/install-worker.sh#L469)

cartermckinnon: Yep! Instead of wrapping that whole block in an `if`, you could do something like move it into a `function` and then return-early if you find there's nothing to cache

Issue #1744, opened 2024-04-02
libnvidia-ml library not found in GPU AMIs


ugurgural:
**What happened**:
Since last week we can't run any GPU workload under the latest GPU AMI. Basically we realized something is wrong when we checked the nvidia-device-plugin pod, stating this error : "Detected non-NVML platform: could not load NVML library: libnvidia-ml.so.1: cannot open shared object file: No such file or directory". After debugging inside the node it simply says that lot of libraries are missing related with libnvidia. Detailed logs are below.

**What you expected to happen**:
Be able to run GPU workloads with default GPU AMI

**How to reproduce it (as minimally and precisely as possible)**:
 nvidia-device-plugin-deamonset output:

`I0402 12:47:47.247247       1 main.go:279] Retrieving plugins.
W0402 12:47:47.247300       1 factory.go:31] No valid resources detected, creating a null CDI handler
I0402 12:47:47.247350       1 factory.go:104] Detected non-NVML platform: could not load NVML library: libnvidia-ml.so.1: cannot open shared object file: No such file or directory
I0402 12:47:47.247391       1 factory.go:104] Detected non-Tegra platform: /sys/devices/soc0/family file not found
E0402 12:47:47.247405       1 factory.go:112] Incompatible platform detected
E0402 12:47:47.247409       1 factory.go:113] If this is a GPU node, did you configure the NVIDIA Container Toolkit?
E0402 12:47:47.247414       1 factory.go:114] You can check the prerequisites at: https://github.com/NVIDIA/k8s-device-plugin#prerequisites
E0402 12:47:47.247417       1 factory.go:115] You can learn how to set the runtime at: https://github.com/NVIDIA/k8s-device-plugin#quick-start
E0402 12:47:47.247422       1 factory.go:116] If this is not a GPU node, you should set up a toleration or nodeSelector to only deploy this plugin on GPU nodes
I0402 12:47:47.247432       1 main.go:308] No devices found. Waiting indefinitely.`

nvidia-smi output:

`sh-4.2$ sudo nvidia-smi
Tue Apr  2 12:52:53 2024
+---------------------------------------------------------------------------------------+
| NVIDIA-SMI 535.129.03             Driver Version: 535.129.03   CUDA Version: 12.2     |
|-----------------------------------------+----------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id        Disp.A | Volatile Uncorr. ECC |
[...]

PR #1754, opened 2024-04-05
Add bootstrap helper for GPU AMI

cartermckinnon:
**Issue #, if available:**

This is part of the fix for a race-condition occurring in the GPU, referenced in: #1744.

**Description of changes:**

We need a way to explicitly synchronize some background processes that configure GPU instances in the GPU AMI. This will run a helper script if it's present, which will perform GPU-AMI-specific actions. The helper script will only be present on GPU AMI's.

**As the code comment states, this is not intended to be a way for users to inject arbitrary logic into the bootstrap script! Don't do that.**


cartermckinnon:
/ci
+workflow:os_distros al2

Issacwww (approved):

PR #1758, opened 2024-04-09
Wait for network-online to configure clocksource

cartermckinnon:
**Description of changes:**

The script that configures the clocksource needs to query the hypervisor info from IMDS. We have retries in place, but this should add some more stability.

### DIFF
--- a/templates/shared/runtime/configure-clocksource.service
+++ b/templates/shared/runtime/configure-clocksource.service
@@ -1,5 +1,8 @@
 [Unit]
 Description=Configure kernel clocksource
+# the script needs to use IMDS, so wait for the network to be up
+Wants=network-online.target
+After=network-online.target
 
 [Service]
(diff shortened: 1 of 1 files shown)


templates/shared/runtime/configure-clocksource.service
```diff
@@ -1,5 +1,8 @@
 [Unit]
 Description=Configure kernel clocksource
+# the script needs to use IMDS, so wait for the network to be up
+Wants=network-online.target
+After=network-online.target
 
 [Service]
```
ndbaker1, line 7: not sure what behavior we've seen historically, but should we also set it to restart on failure?

ndbaker1 (approved):

cartermckinnon:
The 2 failures are unrelated, hit some limits in the CI account.

PR #1762, opened 2024-04-11
Decompress GZIP'd user data

### LINKED ISSUE #1734: nodeadm doesn't support userdata content-encoded with gzip
nodeadm fails with an error message about control characters in userdata when the userdata has content-encoded with gzip, while cloud-init uncompresses the userdata just fine.  Disabling gzip of userdata allowed nodeadm to work as expected.

cartermckinnon:
**Issue #, if available:**

Fixes #1734 

**Description of changes:**

Adds support for GZIP-compressed user data.

The following scenarios are supported:
1. User data that consists of a `NodeConfig` that is compressed with GZIP.
2. User data that is a multi-part MIME document that is compressed with GZIP.
3. User data that is a multi-part MIME document containing parts that are individually GZIP'd (and have the `Content-Encoding: gzip` header).
4. A combination of 2 and 3.

### DIFF
--- a/nodeadm/internal/configprovider/userdata.go
+++ b/nodeadm/internal/configprovider/userdata.go
@@ -2,6 +2,8 @@ package configprovider
 
 import (
 	"bytes"
+	"compress/gzip"
+	"encoding/base64"
 	"fmt"
 	"io"
 	"mime"
@@ -22,16 +24,38 @@ const (
 	nodeConfigMediaType        = "application/" + api.GroupName
 )
 
-type userDataConfigProvider struct{}
-
-func NewUserDataConfigProvider() ConfigProvider {
-	return &userDataConfigProvider{}
+type userDataProvider interface {
+	GetUserData() ([]byte, error)
 }
 
-func (ics *userDataConfigProvider) Provide() (*internalapi.NodeConfig, error) {
-	userData, err := imds.GetUserData()
+type imdsUserDataProvider struct{}
+
+func (p *imdsUserDataProvider) GetUserData() ([]byte, error) {
+	return imds.GetUserData()
+}
+
+type userDataConfigProvider struct {
+	userDataProvider userDataProvider
+}
+
+func NewUserDataConfigProvider() ConfigProvider {
+	return &userDataConfigProvider{
+		userDataProvider: &imdsUserDataProvider{},
+	}
+}
+
+func (p *userDataConfigProvider) Provide() (*internalapi.NodeConfig, error) {
+	userData, err := p.userDataProvider.GetUserData()
 	if err != nil {
 		return nil, err
+	}
+	userData, err = decodeIfBase64(userData)
+	if err != nil {
+		return nil, fmt.Errorf("failed to decode user data: %v", err)
+	}
+	userData, err = decompressIfGZIP(userData)
+	if err != nil {
+		return nil, fmt.Errorf("failed to decompress user data: %v", err)
 	}
 	// if the MIME data fails to parse as a multipart document, then fall back
 	// to parsing the entire userdata as the node config.
@@ -85,6 +109,14 @@ func parseMultipart(userDataReader *multipart.Reader) (*internalapi.NodeConfig, 
 				if err != nil {
 					return nil, err
 				}
+				nodeConfigPart, err = decodeIfBase64(nodeConfigPart)
+				if err != nil {
+					return nil, err
+				}
+				nodeConfigPart, err = decompressIfGZIP(nodeConfigPart)
+				if err != nil {
+					return nil, err
+				}
 				decodedConfig, err := apibridge.DecodeNodeConfig(nodeConfigPart)
 				if err != nil {
 					return nil, err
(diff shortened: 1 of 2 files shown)


ndbaker1 (commented):
the change LGTM. doesn't seem like we are going to adopt decompressing individual parts, so can we do a doc update before merging?

cartermckinnon:
>  doesn't seem like we are going to adopt decompressing individual parts

@ndbaker1 latest rev adds support for this, PTAL.

nodeadm/internal/configprovider/userdata_test.go
cartermckinnon, line 1: I rewrote the unit tests for this class; they're more verbose but I think being explicit about the input is more clear and lets us more easily add test cases in the future.

nodeadm/internal/configprovider/userdata.go
```diff
@@ -17,22 +18,41 @@ import (
...
+
+type imdsUserDataProvider struct{}
+
+func (p *imdsUserDataProvider) GetUserData() ([]byte, error) {
+	return imds.GetUserData()
+}
+
+type userDataConfigProvider struct {
+	userDataProvider userDataProvider
+}
```
mmerkes, line 39: Why do you need this wrapper?

nodeadm/internal/configprovider/userdata.go
```diff
@@ -17,22 +18,41 @@ import (
...
 
 func NewUserDataConfigProvider() ConfigProvider {
-	return &userDataConfigProvider{}
+	return &userDataConfigProvider{
+		userDataProvider: &imdsUserDataProvider{},
+	}
 }
 
 func (ics *userDataConfigProvider) Provide() (*internalapi.NodeConfig, error) {
```
mmerkes, line 47: Why is `ics` the name for `userDataConfigProvider`?

nodeadm/internal/configprovider/userdata.go
```diff
@@ -85,6 +105,13 @@ func parseMultipart(userDataReader *multipart.Reader) (*internalapi.NodeConfig,
 				if err != nil {
 					return nil, err
 				}
+				if part.Header.Get(contentEncodingHeader) == "gzip" {
+					decompressedNodeConfigPart, err := decompressIfGZIP(nodeConfigPart)
+					if err != nil {
+						return nil, err
+					}
+					nodeConfigPart = decompressedNodeConfigPart
+				}
```
mmerkes, line 114: Why do you need to make a check that it's `gzip` and then gracefully handle it not being `gzip`?

nodeadm/internal/configprovider/userdata_test.go
mmerkes: Did you do any integration testing here? Or is that covered by the existing integration tests?

nodeadm/internal/configprovider/userdata.go
ndbaker1: wouldn't just calling `decompressIfGZIP` in all cases work better since its fallible in constant time and safe either way?

To not hit this branch you'd have to set the nodeConfig mediaType but forget to add the gzip encoding header, which im sure would only happen on accident. maybe handling it explicitly is more "proper", but not sure if its necessary

nodeadm/internal/configprovider/userdata.go
ndbaker1: i think it's an artifact from when it was something like `imdsConfigProvider`? 🤣   
maybe we could just generally name the config provider references `cp`

nodeadm/internal/configprovider/userdata_test.go
ndbaker1: i like the change, thanks for skipping over the junk to setup mime documents and testing the actual provider interface

nodeadm/internal/configprovider/userdata.go
```diff
@@ -17,22 +18,41 @@ import (
...
+	contentEncodingHeader      = "Content-Encoding"
 	mimeBoundaryParam          = "boundary"
 	multipartContentTypePrefix = "multipart/"
 	nodeConfigMediaType        = "application/" + api.GroupName
 )
 
-type userDataConfigProvider struct{}
+type userDataProvider interface {
+	GetUserData() ([]byte, error)
+}
```
ndbaker1, line 29: nice way to make testing the provider interface easier 👍

nodeadm/internal/configprovider/userdata.go
cartermckinnon: tbh i like just calling it `this` but that's a Java-ism that gophers detest

nodeadm/internal/configprovider/userdata.go
cartermckinnon: This is for testing, it allows us to mock out the IMDS calls

nodeadm/internal/configprovider/userdata.go
cartermckinnon: MIME semantics dictate that if the section is GZIP'd, this header needs to be there. If a user (or library) is creating MIME documents that use GZIP but omit this header, that should fail because it's not conformant to the spec.

Generally, folks are only using GZIP compression via a library which should implement this properly -- folks aren't hand-writing and hand-compressing their MIME docs.

nodeadm/internal/configprovider/userdata.go
mmerkes: Haha. Let's change it to something that makes sense :P

cartermckinnon:
/ci
+workflow:os_distros al2023

cartermckinnon:
One flake:

> [Fail] [sig-api-machinery] CustomResourcePublishOpenAPI [Privileged:ClusterAdmin] [It] updates the published spec when one version gets renamed [Conformance] 

Re-running for sanity

nodeadm/internal/configprovider/userdata.go
cartermckinnon: Turns out cloudinit doesn't care what the spec says, and just tries to base64-decode and gzip-decompress pretty much anything it handles at any scope. It's more important for us to match the cloudinit behavior than to follow the spec, for better or worse. Updated in latest rev

cartermckinnon:
/ci
+workflow:os_distros al2023

nodeadm/internal/configprovider/userdata.go
```diff
@@ -85,6 +109,14 @@ func parseMultipart(userDataReader *multipart.Reader) (*internalapi.NodeConfig,
 				if err != nil {
 					return nil, err
 				}
+				nodeConfigPart, err = decodeIfBase64(nodeConfigPart)
+				if err != nil {
+					return nil, err
+				}
+				nodeConfigPart, err = decompressIfGZIP(nodeConfigPart)
+				if err != nil {
+					return nil, err
```
ndbaker1, line 118: nit: theres a custom error message around the calls above but not here :(

nodeadm/internal/configprovider/userdata.go
```diff
@@ -102,6 +134,39 @@ func parseMultipart(userDataReader *multipart.Reader) (*internalapi.NodeConfig,
...
+	}
+}
+
+func decodeIfBase64(data []byte) ([]byte, error) {
+	e := base64.StdEncoding
+	maxDecodedLen := e.DecodedLen(len(data))
+	decodedData := make([]byte, maxDecodedLen)
+	decodedLen, err := e.Decode(decodedData, data)
+	if err != nil {
+		return data, nil
```
ndbaker1, line 147: so none of the possible errors here end up mattering?

nodeadm/internal/configprovider/userdata.go
ndbaker1: actually the more i think on this, should we be more tolerant and `break` on these errors so that we can try the remaining parts?

nodeadm/internal/configprovider/userdata.go
ndbaker1: oh i guess it can only return [`CorruptInputError`](https://cs.opensource.google/go/go/+/refs/tags/go1.22.3:src/encoding/base64/base64.go;l=301-305) 🤔

ndbaker1 (approved):
lgtm as-is, one nit comment

Issue #1765, opened 2024-04-16
Hardened Source Image throws: `cannot stat ‘/tmp/worker/configure-clocksource.service’: No such file or directory` 


mebays:
**What happened**: 

When using `make 1.28 enable_fips=true source_ami_filter_name=<CIS STIG HARDENED Image> source_ami_owners=<OWNERID> aws_region=$AWS_REGION`  the packer build errors off on the install-worker.sh after the reboot for fips with the following error.  

```
2024-04-16T11:18:22-04:00:     amazon-ebs: mv: cannot stat ‘/tmp/worker/configure-clocksource.service’: No such file or directory
```

**What you expected to happen**: 

The image to build without any issues.

**Anything else we need to know?**: 

I have been watching other hardened issues and have not seen a reference to this specific issue. I will have a PR with a suggested fix. The assumed problem is around the ordering the provisioners. Specifically the order for creating the /tmp/worker and mounting of all the scripts before the reboot. While running the packer output in -debug and connecting to the image it was seen that /tmp/worker did not exist after reboot it did exist before reboot.  Reordering the creation of the /tmp/worker directory and the file provisioners after the reboot fixes this.

ndbaker1:
we stopped doing this as a service unit sometime back as well (https://github.com/awslabs/amazon-eks-ami/pull/2144). let us know if this still manifests in some way

PR #1766, opened 2024-04-16
Group working directory scripts after reboot

mebays:
**Issue #, if available:** #1765

**Description of changes:**

Group the file working directory file provisioning and scripts after reboot in case the files are not there after a restart. 

**Testing Done**

Building on a hardened image without this threw and error in the linked issue. After the changes the build process moved passed the issue. Then hits the issue with permissions of kubelet and awscli which has two pull requests #1513 and #1717.

### DIFF
--- a/templates/al2/template.json
+++ b/templates/al2/template.json
@@ -125,40 +125,9 @@
     {
       "type": "shell",
       "remote_folder": "{{ user `remote_folder`}}",
-      "inline": [
-        "mkdir -p {{user `working_dir`}}",
-        "mkdir -p {{user `working_dir`}}/log-collector-script"
-      ]
-    },
-    {
-      "type": "shell",
-      "remote_folder": "{{ user `remote_folder`}}",
       "script": "{{template_dir}}/provisioners/install-additional-repos.sh",
       "environment_vars": [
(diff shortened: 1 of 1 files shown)


mebays:
Actual cause after digging into this. The reason this occurred is al2 includes /tmp in the root file system.

```
$ mount | grep tmp
devtmpfs on /dev type devtmpfs (rw,nosuid,size=3960632k,nr_inodes=990158,mode=755)
tmpfs on /dev/shm type tmpfs (rw,nosuid,nodev)
tmpfs on /run type tmpfs (rw,nosuid,nodev,mode=755)
tmpfs on /sys/fs/cgroup type tmpfs (ro,nosuid,nodev,noexec,mode=755)
tmpfs on /run/user/1000 type tmpfs (rw,nosuid,nodev,relatime,size=793780k,mode=700,uid=1000,gid=1000)
tmpfs on /run/user/0 type tmpfs (rw,nosuid,nodev,relatime,size=793780k,mode=700)
```

On the instance where this failed (hardened instance),  `/tmp` is a tmpfs which gets thrown out on restart. 
```
tmpfs on /tmp type tmpfs (rw,nosuid,nodev,relatime,seclabel,size=5242880k)
```

This seems like a common scenario where /tmp is a tmpfs.  But I figured I would include what the actual cause was from.

PR #1773, opened 2024-04-26
Remove sudo from bootstrap commands

zhuoweilee:
**Issue #, if available:**

**Description of changes:**

`sudo` commands are removed from the bootstrap script.

This helps to resolve the issue outlined in [this article](https://access.redhat.com/solutions/5692661) which describes a permission issue with `sudo` commands in `cloud_init` scripts that results in a timeout when SELinux is enabled. This leads to extended bootstrap times (+25 sec per `sudo`). Before this pull request, customers that enable SELinux would have to either add an SELinux policy module or remove `sudo` commands from the bootstrap script to prevent timeout.


cartermckinnon:
I think we should add an assertion to the start of the script that it's being run as `root, something like:
```
if [ "$EUID" -ne 0 ]; then
  log "ERROR: must be run as root"
  exit 1
fi
```

cartermckinnon:
/ci
+workflow:os_distros al2

cartermckinnon:
This LGTM, I'm getting another set of eyes on it.

templates/al2/runtime/bootstrap.sh
```diff
@@ -46,6 +46,11 @@ function log {
 
 log "INFO: starting..."
 
+if [ "${EUID}" -ne 0 ]; then
+  log "ERROR: script must be run as root"
```
ndbaker1, line 50: for our usage this looks fine to me. In custom edge cases where this isn't satisfied i think it would just need a tweak to [`execute_command`](https://developer.hashicorp.com/packer/docs/provisioners/shell#execute-command-example)

ndbaker1 (approved):

PR #1794, opened 2024-05-10
al2023: Use base_runtime_spec to set default rlimits

### LINKED ISSUE #1746: AL2023: OOMKills caused by very high file descriptor limits
**What happened**:
Certain applications experience immediate OOM kills on AL2023 due to what I expect is a change in NOFILE's or Ulimit parameters on the AMI when compared to AL2
The apps to launch without issue
Launch https://github.com/DandyDeveloper/charts/blob/master/charts/redis-ha on an AL2023 node. It should immediately go into crash loop backoff
AL2 does not experience this issue
**Environment**:
- AWS Region: us-east-1
- Instance Type(s): any instance on AL2023
- EKS Platform version (use `aws eks describe-cluster --name <name> --query cluster.platformVersion`): eks.3
- Kubernetes version (use `aws eks describe-cluster --name <name> --query cluster.version`): 2.29
- AMI Version: AL2023
- Kernel (e.g. `uname -a`):

```
```

cartermckinnon:
**Issue #, if available:**

Fixes #1746 

**Description of changes:**

This uses the `base_runtime_spec` option of `containerd` to set `RLMIT_NOFILE` defaults that are much lower than 2^63-1, the default on AL2023. This approach lets us define the limits for containers without changing `containerd`'s own limits.

I've gone with `65536:1048576` with the understanding that programs using `select(2)` will need lower limits to be defined by the user to avoid undefined behavior. In practice, it's more common for our users to run programs that do not *raise* their limits appropriately--such as Envoy--so I think this is an acceptable default today. Bottlerocket is using the same values, so the consistency is beneficial: https://github.com/bottlerocket-os/bottlerocket/blob/fcf71a47c0ff005327ecde870d8a70877fc89196/sources/models/shared-defaults/oci-defaults-containerd-cri-resource-limits.toml

These defaults can be overridden by a new `NodeConfig` field, `Spec.Containerd.BaseRuntimeSpec`. This field behaves the same as `Spec.Kubelet.Config` -- it's an unstructured JSON/YAML document that will be merged on top of the default, allowing partial overrides.


cartermckinnon:
/ci
+workflow:os_distros al2023

nodeadm/internal/containerd/default-base-runtime-spec.json
```diff
@@ -0,0 +1,12 @@
+{
+  "ociVersion": "1.1.0",
+  "process": {
+    "rlimits": [
+      {
+        "type": "RLIMIT_NOFILE",
+        "soft": 65536,
+        "hard": 1048576
```
ndbaker1, line 8: appreciate making this configurable, do we have to set a default?  
asking based on the time we updated limits in https://github.com/awslabs/amazon-eks-ami/pull/1535 in al2, but I see the rationale on making this a breaking change to set the default and follow bottlerocket's example

cartermckinnon: Yeah, we have to set a lower default in order to fix #1746.

The issues caused by #1535 were mostly related to the `1024` effective soft limit (which is the "correct" config but is nevertheless incompatible with a lot of software written in the container age). That's why we use a much higher soft limit here, as well as a higher hard limit. I think this is fairly unlikely to be disruptive; and there's an easy mitigation if it does (overriding the limits with something appropriate for the specific workload).

cartermckinnon:
Argh, we have to pass a complete OCI spec instead of just `process.rlimits`. I'm going to switch up the API options here, because if someone wants to override the rlimits, they shouldn't have to pass all the other junk.

ps-jay:
Thanks for your work on this @cartermckinnon

cartermckinnon:
/ci
+workflow:os_distros al2023

dims:
@cartermckinnon is this the same limits used in bottlerocket?

cartermckinnon:
@dims yep!

cartermckinnon:
Need to fix something up with the capabilities block in the latest rev

> [sig-node] Security Context When creating a pod with privileged should run the container as unprivileged when false [LinuxOnly] [NodeConformance] [Conformance]

cartermckinnon:
/ci
+workflow:os_distros al2023

cartermckinnon:
/ci
+workflow:os_distros al2023

cartermckinnon:
Alright, there we go. Problem was `CAP_NET_ADMIN`.

nodeadm/internal/containerd/base_runtime_spec.go
```diff
@@ -0,0 +1,44 @@
...
+
+func init() {
+	if err := json.Unmarshal([]byte(baseRuntimeSpecData), &baseRuntimeSpecMap); err != nil {
+		panic(err)
+	}
+}
+
+func writeBaseRuntimeSpec(cfg *api.NodeConfig) error {
+	zap.L().Info("Writing containerd base runtime spec...", zap.String("path", containerdBaseRuntimeSpecFile))
+	baseRuntimeSpecData := baseRuntimeSpecData
```
ndbaker1, line 28: we end up pulling this from the global scope anyway so should you just move the init code inside here?

nodeadm/internal/containerd/base-runtime-spec.json
```diff
@@ -0,0 +1,174 @@
+{
```
Issacwww, line 1: out of curiosity, how this been generated? Is there any reference I can read to understand more in case we need to debug sth for this?

cartermckinnon: I derived this from a few sources:
1. The default base OCI spec given by `ctr oci spec`.
2. An actual spec for a running container grabbed with `ctr container info`.
3. The base spec used by Bottlerocket for this same purpose (providing RLIMIT overrides): https://github.com/bottlerocket-os/bottlerocket/blob/b3e476f382dbaf6bb6bea20d99d3bf143662cef6/packages/containerd/containerd-cri-base-json#L4

The spec itself is thoroughly documented here: https://github.com/opencontainers/runtime-spec

nodeadm/internal/containerd/base_runtime_spec.go
cartermckinnon: Sure, done

ndbaker1 (approved):
LGTM, thanks for tackling this 👏

Issacwww (approved):

bjhaid:
when will an AMI containing this be released?

bjhaid:
There was a pre-release 3 hours ago https://github.com/awslabs/amazon-eks-ami/releases/tag/v20240605 that does not contain this PR. @ndbaker1 @Issacwww what needs to be done to get this changes in the next AMI release?

cartermckinnon:
This will be included in an AMI next week.

msvechla:
Is there an update on when this will be released? We are waiting for this hotfix

bjhaid:
@cartermckinnon the week is almost over 😢 . We are actively blocked from completing our upgrade and this AMI will unblock us

cartermckinnon:
This is going out today in `v20240615`: https://github.com/awslabs/amazon-eks-ami/releases/tag/v20240615

PR #1808, opened 2024-05-16
[WIP] (al2023): Add default system reserved resources

### LINKED ISSUE #1490: Add a default systemReserved configuration to the kubelet-config.json

The default kubelet should configure dedicated systemReserved cpu and memory.

Something like the following should be added to the kubelet-config.json:
```json
"systemReserved": {
  "cpu": "50m",
  "memory": "128Mi"
}
```

**Why is this needed**:

We are experiencing nodes switching between Ready and NotReady when the node has high memory pressure caused by pods with unset memory limits.
Expectation is that the kubelet kills the pods if they over-allocate, or if other pods arrive with requests and the over-committed pods should then get killed.
Instead in these high memory pressure cases the entire node seems to die when using default EKS AMI configuration. Kubelet doesn't report back to the API server, and we also cannot connect to the nodes with SSM or Instance Connect.

Adding systemReserved configuration through `--kubelet-extra-args` might be an acceptable workaround, but it seems like something that should be configured by default on the nodes, so that even if the kubelet becomes unresponsive, services in the `system.slice` keep working so one can go and troubleshoot the node.

cartermckinnon:
**Issue #, if available:**

Fixes #1490 

**Description of changes:**

This PR's goal is to add a default CPU and memory reservation for the `/system` cgroup. We do not intend to enforce this reservation (i.e. throttle processes within this cgroup); this would merely subtract resources from the Node's allocatable totals, to provide more overhead in the event that pods burst beyond their limit(s).

This PR will only make this change for AL2023. It's unlikely at this point that we'll do this for AL2, especially on existing k8s versions, as it would decrease a cluster's total allocatable resources.

### DIFF
--- a/nodeadm/internal/kubelet/config.go
+++ b/nodeadm/internal/kubelet/config.go
@@ -55,6 +55,8 @@ func (k *kubelet) writeKubeletConfig(cfg *api.NodeConfig) error {
 // KubeletConfiguration types:
 // https://pkg.go.dev/k8s.io/kubelet/config/v1beta1#KubeletConfiguration
 type kubeletConfig struct {
+	metav1.TypeMeta `json:",inline"`
+
 	Address                  string                           `json:"address"`
 	Authentication           k8skubelet.KubeletAuthentication `json:"authentication"`
 	Authorization            k8skubelet.KubeletAuthorization  `json:"authorization"`
@@ -63,6 +65,7 @@ type kubeletConfig struct {
 	ClusterDNS               []string                         `json:"clusterDNS"`
 	ClusterDomain            string                           `json:"clusterDomain"`
 	ContainerRuntimeEndpoint string                           `json:"containerRuntimeEndpoint"`
+	EnforceNodeAllocatable   []string                         `json:"enforceNodeAllocatable"`
 	EvictionHard             map[string]string                `json:"evictionHard,omitempty"`
 	FeatureGates             map[string]bool                  `json:"featureGates"`
 	HairpinMode              string                           `json:"hairpinMode"`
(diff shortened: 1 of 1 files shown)


nodeadm/internal/kubelet/config.go
```diff
@@ -119,6 +122,7 @@ func defaultKubeletSubConfig() kubeletConfig {
 		CgroupRoot:               "/",
 		ClusterDomain:            "cluster.local",
 		ContainerRuntimeEndpoint: containerd.ContainerRuntimeEndpoint,
+		EnforceNodeAllocatable:   []string{"pods"},
```
cartermckinnon, line 125: This is the default -- just making it explicit here that we aren't enforcing either kube or system reserved.

cartermckinnon:
I'm starting off by measuring the CPU and memory usage at idle for the `m7i` family (general purpose Intel).

I launched a 1.29 cluster with the following instances:
| Instance type | vCPU | Memory (GB) |
|--------|--------|--------|
| m7i.large | 2 | 8 |
| m7i.xlarge | 4 | 16 |
| m7i.2xlarge | 8 | 32 |
| m7i.4xlarge | 16 | 64 |
| m7i.8xlarge | 32 | 128 |
| m7i.12xlarge | 48 | 192 |
| m7i.16xlarge | 64 | 256 | 
| m7i.24xlarge | 96 | 384 |

I grabbed the maximum CPU utilization (in vCPU's) for each instance and calculated a linear regression:

vCPU | Max usage at idle (vCPU)
-- | --
2 | 0.0458
4 | 0.0587
8 | 0.0729
16 | 0.097
32 | 0.1289
48 | 0.1574
96 | 0.2815

```
y = 0.0024 * x + 0.0497
```

Screenshot 2024-05-17 at 4 15 08 PM

tooptoop4:
will this apply to bottlerocket?

PR #1811, opened 2024-05-17
attempt to fix changelog formatting

ndbaker1:
**Issue #, if available:**

**Description of changes:**

fixes whitespace trimming in changelog PR auto-generation.

the js-flavor regex replace doesn't really work like `sed` or the normal line-based editors since the entire document is one string. Even using the multi-line and global modifiers `m`/`g` doesn't work as expected, and clobbers repeated empty lines when using a pattern with the rule `\s+$`. 

ultimately the intended behavior happens if you split the lines first to mimic a true line-based transform 🤷, which is safe because splitting on `\n` will leave carriage returns `\r` at the end of the line to be caught by `\s`.

tested by updating the local file using:
```js
fs.writeFileSync('CHANGELOG.md',
  fs.readFileSync('CHANGELOG.md', 'utf8')
    .split('\n')
    .map(s => s.replace(/\s+$/, ''))
    .join('\n')
)
```


cartermckinnon:
I think we should keep this, I tried to fix the issue but it didn't do the trick if you want to take a look at that: https://github.com/awslabs/amazon-eks-ami/blob/f59dfbdd9367f4713eb87fb35863c44956f77199/.github/workflows/update-changelog.yaml#L47-L48

ndbaker1:
> I think we should keep this, I tried to fix the issue but it didn't do the trick if you want to take a look at that:
> 
> https://github.com/awslabs/amazon-eks-ami/blob/f59dfbdd9367f4713eb87fb35863c44956f77199/.github/workflows/update-changelog.yaml#L47-L48

oh didn't realize 😲 i think the issue here...

.github/workflows/update-changelog.yaml
```diff
@@ -45,7 +45,7 @@ jobs:
             const newEntry = `# ${release.data.name}\n${release.data.body}`;
             let updatedChangelog = changelog.replace(placeholder, placeholder + '\n\n' + newEntry + '\n---\n');
             // if the release notes are modified in the GitHub web editor, trailing spaces can be added accidentally
-            updatedChangelog = updatedChangelog.replace(/\s+$/, '');
+            updatedChangelog = updatedChangelog.split('\n').map(s => s.replace(/\s+$/, '')).join('\n');
```
cartermckinnon, line 48: ahh nice! thanks

cartermckinnon (approved):

PR #1828, opened 2024-05-31
Tag AMI's with ena_support

cartermckinnon:
**Description of changes:**

Enables ENA by default on instances launched with the AMIs. We already have the `ena` kernel module installed, and the base Amazon Linux AMIs have this enabled -- we just haven't been propagating the attribute to our AMI's.


cartermckinnon:
This isn't actually needed, our AMI's already inherit this property. You just need to launch a Nitro instance and things work as expected.

PR #1850, opened 2024-06-17
Pin github action versions

cartermckinnon:
**Description of changes:**

Pins all of our external github action dependencies to their current latest versions. The `# 1.2.3` style comments are supported by dependabot.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1852, opened 2024-06-18
Use unique artifact names in CI workflow

cartermckinnon:
**Description of changes:**

#1850  updated us to 4.x of the upload-artifact action, which requires unique artifact names across an entire workflow.

Ex failure: https://github.com/awslabs/amazon-eks-ami/actions/runs/9568957937/job/26380369916

> Error: Failed to CreateArtifact: Received non-retryable error: Failed request: (409) Conflict: an artifact with this name already exists on the workflow run

### DIFF
--- a/.github/actions/ci/build/action.yaml
+++ b/.github/actions/ci/build/action.yaml
@@ -33,5 +33,5 @@ runs:
         echo "ami_id=$(jq -r .builds[0].artifact_id "${AMI_NAME}-manifest.json" | cut -d ':' -f 2)" >> $GITHUB_OUTPUT
     - uses: actions/upload-artifact@65462800fd760344b1a7b4382951275a0abb4808 # 4.3.3
       with:
-        name: version-info
+        name: version-info-${{ inputs.k8s_version }}-${{ inputs.os_distro }}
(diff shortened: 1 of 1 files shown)


.github/actions/ci/build/action.yaml
```diff
@@ -33,5 +33,5 @@ runs:
         echo "ami_id=$(jq -r .builds[0].artifact_id "${AMI_NAME}-manifest.json" | cut -d ':' -f 2)" >> $GITHUB_OUTPUT
     - uses: actions/upload-artifact@65462800fd760344b1a7b4382951275a0abb4808 # 4.3.3
       with:
-        name: version-info
+        name: version-info-${{ inputs.k8s_version }}-${{ inputs.os_distro }}
```
ndbaker1, line 36: lgtm, i just want to ask if this is enough or do people usually use a generated uuid or smth

ndbaker1 (approved):

.github/actions/ci/build/action.yaml
cartermckinnon: We could tack on the build_id, git_sha, etc. but this is sufficient for our workflows

PR #1854, opened 2024-06-18
bug(al2023): Do not bind containerd restarts to kubelet

### LINKED ISSUE #1846: EKS AMI shouldn't restart Kubelet when containerd is restarted
**What happened**:
If containerd is restarted, Kubelet also gets restarted. Other popular open source projects like Kops and kubeadm have not configured systemd configuration as such.

1. KOPS doesn't configure this behavior, Kops just ensures that containerd has started up before Kubelet - https://github.com/kubernetes/kops/blob/master/nodeup/pkg/model/kubelet.go#L389
2. Kubeadm doesn't configure either- https://github.com/kubernetes/release/blob/cd53840/cmd/krel/templates/latest/kubelet/kubelet.service
 

I would expect Kubelet to fail to communicate, but once containerd restarts it should automatically recover.

This is confirmed by my discussion on Slack with folks from sig-node: https://kubernetes.slack.com/archives/C0BP8PW9G/p1718063064902389

There are risks to restarting the Kubelet - 
1. Features like https://kubernetes.io/blog/2023/04/21/node-log-query-alpha/ , which don't depend on CRI should still work even if containerd is crashing.
2. In-case of a catastrophic failure mode where multiple Kubelets get stuck in a restart loop due to containerd crashing (ooms, panics, etc), it can cause availability issues for the control plane, as many Kubelets booting up together has historically resulted in higher load on the control plane. 

Launch an instance with the AMI
```
# Check Kubelet is running and check process start time
sudo systemctl status kubelet 
sudo systemctl status containerd
#Copy PID
sudo kill -9 <PID_OF_CONTAINERD>
# Check Kubelet restarted by checking the process start time
sudo systemctl status kubelet
```

cartermckinnon:
**Issue #, if available:**

Fixes #1846.

**Description of changes:**

This updates the `kubelet.service` on AL2023 to be loosely-coupled with `containerd`, so that restarts/kills of `containerd` do not impact `kubelet`.

We're not making this change on AL2 because folks may see different behavior if they're starting/stopping either of these units in their existing user data.

### DIFF
--- a/templates/al2023/runtime/rootfs/etc/systemd/system/kubelet.service
+++ b/templates/al2023/runtime/rootfs/etc/systemd/system/kubelet.service
@@ -2,7 +2,7 @@
 Description=Kubernetes Kubelet
 Documentation=https://github.com/kubernetes/kubernetes
 After=containerd.service
-Requires=containerd.service
+Wants=containerd.service
 
 [Service]
 Slice=runtime.slice


cartermckinnon:
The `1.28 / al2` failure was a network flake, not related.

suket22 (approved):

ndbaker1 (approved):

PR #1855, opened 2024-06-19
feat(al2023): Add Support for hardened Images

KrisJohnstone:
**Description of changes:**
When using hardened images (i.e. CIS or STIG) there are a few changes required to get everything running and have nodes join the cluster.

**Testing Done**
Here's a hardened image running in one of our test clusters:
image
image


templates/al2023/provisioners/hardened-images.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+set -o pipefail
+set -o nounset
+set -o errexit
+
+if [[ "$HARDENED_IMAGE" == "true" ]]; then
+  sudo chcon -t bin_t /usr/bin/nodeadm
+  sudo chcon -t bin_t /usr/bin/kubelet
+  sudo systemctl disable firewalld
```
cartermckinnon, line 10: can you add some context here? disabling the firewall != hardenening at first glance :)

KrisJohnstone: Judgement call I guess, a part of STIG and CIS (iirc) is that firewalld is enabled by default, however the management of its rule is manual. Figured there's already IPTables and if that's not enough Cilium has a host firewall option which is managed.

And with CIS and STIG they both have a minimum pass rate so the disabling of firewalld represents a small drop in compliance but a pass overall afaik.

PR #1871, opened 2024-07-04
Increase snapshot creation polling timeout

YngveMolnes:
**Issue #, if available:**

**Description of changes:**
Increase the snapshot creation polling timeout. When we build our AMIs we need to cache _a lot_ of container images on the AMIs. This, in turn, creates large snapshots. For this to be cheap and pragmatic we just use the defaults in the project and make a rather big volume for the images, without optimizing any IO-config. So - for us it might take up to 4 hours to wait for the snapshot creation. Either way; in order for the ami to get tagged properly and for the job to clean up - I had to increase this timeout.

**Testing Done**
Building an image with the params;

launch_block_device_mappings_volume_size=300 aws_region=eu-north-1 ssm_agent_version=3.3.380.0 make k8s=1.29 cache_container_images=true os_distro=al2 subnet_id=<my_private_subnet>

### DIFF
--- a/templates/al2/template.json
+++ b/templates/al2/template.json
@@ -85,7 +85,7 @@
       ],
       "aws_polling": {
         "delay_seconds": 30,
-        "max_attempts": 90
+        "max_attempts": 480
       },
       "ami_regions": "{{user `ami_regions`}}",
       "iam_instance_profile": "{{user `iam_instance_profile`}}",
--- a/templates/al2023/template.json
+++ b/templates/al2023/template.json
@@ -79,7 +79,7 @@
       ],
       "aws_polling": {
         "delay_seconds": 30,
-        "max_attempts": 90
+        "max_attempts": 480
       },
       "ami_regions": "{{user `ami_regions`}}",
       "ssh_username": "{{user `ssh_username`}}",


cartermckinnon:
Bumping the timeout values sounds fine to me, but I'm curious what kind of a benefit you see from packing so much into the AMI's snapshot. You avoid some roundtrips to ECR of course, but the actual data is still lazy-loaded at approx. S3 speeds.

YngveMolnes:
> but I'm curious what kind of a benefit you see from packing so much into the AMI's snapshot

We're a consultancy / service provider in the GIS industry. We serve a very specific solution with proprietary helm charts and proprietary docker images supplied through dockerhub. The images we serve are rather large and we need to serve several versions. Pulling from docker hub is rather slow when we have some ~15 unique images per version of the software and we serve up to three different versions at the time, whereas all of the images per version are rather large. It's a whole enterprise-y software suite. There is more complexity around it due to the implementation details of the software suite in question, but I'd rather be pragmatic about it than to openly critique our supplier publicly. (but it's a bit of a mess). We're using these eks images solely for the purpose of these deployments - and have different ec2nodeprovisoners with smaller amis in karpenter for all other workloads. 

The sole reason we even bake our own images is to solve for the slow pulling; it can take us hours to set up the suite without it. We do notice that the nodes take slightly more time to start, but it's not that bad, and we've tested with optimized IO as well. For us it's the lesser evil, if you will. We could also do something like https://docs.aws.amazon.com/AmazonECR/latest/userguide/pull-through-cache.html to solve the problem, but it's simply more pragmatic to just prepull the versions we need and be done with it. 

> Your instance launch times would take a hit as well.

I was under that assumption as well, but it ends up being fast enough. New nodes start in somewhere between 15 to 60 seconds. Haven't really measured things and benchmarked yet to get proper statistics. Still gathering experience here, but it's also just such an order of magnitude faster than our previous set up with kOps that to us it's still 'amazingly fast'.

cartermckinnon:
>  images supplied through dockerhub

ah, that makes sense, then. 😃 A pull-through cache on ECR (in-region) is worth taking a look at IMO but caching in EBS is a perfectly fine approach.

cartermckinnon (approved):

cartermckinnon:
CI issues, should be fixed now...

/ci build

cartermckinnon:
Sorry for the spam @YngveMolnes, using this PR to get the CI going.

/ci build

PR #1880, opened 2024-07-12
bug(nodeadm): Don't wrap kubelet systemd environment variable

ndbaker1:
**Issue #, if available:**

**Description of changes:**

referring to the docs on [`EnvironmentFile`](https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#EnvironmentFile=) wrapping `"` isn't needed (interior whitespace is preserved verbatum), and conflict with user data if providing quotes within their own flags

### DIFF
--- a/nodeadm/test/e2e/cases/kubelet-flags/config.yaml
+++ b/nodeadm/test/e2e/cases/kubelet-flags/config.yaml
@@ -14,3 +14,6 @@ spec:
       - --v=5
       - --node-labels=foo=bar,foo2=baz
       - --register-with-taints=foo=bar:NoSchedule
+      - |
+        --foo
+        --bar
--- a/nodeadm/test/e2e/cases/kubelet-flags/config.yaml
+++ b/nodeadm/test/e2e/cases/kubelet-flags/config.yaml
@@ -14,6 +14,3 @@ spec:
       - --v=5
       - --node-labels=foo=bar,foo2=baz
       - --register-with-taints=foo=bar:NoSchedule
-      - |
-        --foo
-        --bar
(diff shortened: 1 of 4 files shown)


Issacwww (approved):

nodeadm/test/e2e/cases/kubelet-flags/config.yaml
```diff
@@ -14,3 +14,6 @@ spec:
       - --v=5
       - --node-labels=foo=bar,foo2=baz
       - --register-with-taints=foo=bar:NoSchedule
+      - |
+        --foo
+        --bar
```
cartermckinnon, line 19: Not sure we want to allow this, looks pretty odd. What's the use case?

ndbaker1: included this for edge case coverage, not as an intended use case

cartermckinnon: The `EnvironmentFile` directive already has a bunch of documented handling of newlines, escape characters, etc. so I don't think we should be doing `santitize`, the underlying issue here was us wrapping the user's input in `"`, right? Isn't removing that sufficient?

> a "'"-quoted value after the "=" can span multiple lines and contain any character verbatim other than single quote
https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html

ndbaker1: got it, yea won't do any modification to the user's input. just going to fix the wrapped content issue

cartermckinnon (approved):

PR #1886, opened 2024-07-16
add retry to avoid ECR get auth token failed for some transient issues

wwvela:
**Issue #, if available:**
* ECR get Auth token api might failed if some transient issue happened like the short network disconnect. 
* Add retry here to prevent node failed in config stage for some transient failure

**Description of changes:**

**Testing Done**
make


nodeadm/internal/aws/ecr/ecr.go
```diff
@@ -19,12 +19,15 @@ func GetAuthorizationToken(awsRegion string) (string, error) {
 		return "", err
 	}
 	ecrClient := ecr.NewFromConfig(awsConfig)
-	token, err := ecrClient.GetAuthorizationToken(context.Background(), &ecr.GetAuthorizationTokenInput{})
-	if err != nil {
-		return "", err
+
+	for i := 0; i < 3; i++ {
```
ndbaker1, line 23: could use our internal `util.RetryExponentialBackoff` for this

wwvela: We couldn't use it directly here as the `util.RetryExponentialBackoff` function could accept method which will return error only. In this case, we need to return both token string and error, as the token is required to pull sandbox image.

I'm not sure if it is worthy to create another retry method like `func RetryExponentialBackoff(attempts int, initial time.Duration, f func() (string, error)) error ` which function in param could return both string and err.

ndbaker1: correct me if im wrong, but you can just use a closure for the token, something like 
```go
var token *ecr.GetAuthorizationTokenOutput
util.RetryExponentialBackoff(2, time.Second, func() error {
  token, err = ecrClient.GetAuthorizationToken(context.Background(), &ecr.GetAuthorizationTokenInput{})
  return err
})
```

ndbaker1 (commented):
trying to understand the issue more deeply. so there's a failure case when the ecr client [cannot resolve EC2 roles via IMDS](https://github.com/aws/aws-sdk-go-v2/blob/03768e0d0276b360a6abaa4d30318d4aedc44995/credentials/ec2rolecreds/provider.go#L181) and then this causes the ecr request to fail without even doing retries because this is part of the request finalizers and not the request itself.

then since we can't control how the imds client is configured within the ecr client, we're just using a retry around the entire call itself. Does that sounds right?

wwvela:
> trying to understand the issue more deeply.
> 
>     1. so there's a failure case when the ecr client [cannot resolve EC2 roles via IMDS](https://github.com/aws/aws-sdk-go-v2/blob/03768e0d0276b360a6abaa4d30318d4aedc44995/credentials/ec2rolecreds/provider.go#L181) because it fails to make the IMDSv2 call due to intermittent network issues and then falls back to IMDSv1 even when it is disabled for the instance.
> 
>     2. This then this causes the ecr request to fail without even doing retries because [this is part of the request finalizers](https://github.com/aws/aws-sdk-go-v2/blob/03768e0d0276b360a6abaa4d30318d4aedc44995/service/ecr/auth.go#L216-L231) and not the request itself.
> 
>     3. Finally, since we can't control how the imds client is configured within the ecr client, we're just using a retry around the entire call itself.
> 
> 
> Does that sound right?


Yes it is whole picture. And also attached the link that AL2023 has disabled IMDSv1 https://docs.aws.amazon.com/linux/al2023/ug/deprecated-al2023.html

nodeadm/internal/aws/ecr/ecr.go
wwvela: This might not work as it expected for following reasons. 
* First is because `func` passed to method `RetryExponentialBackoff` here is a function declare. And the token `var token *ecr.GetAuthorizationTokenOutput` declared outside of this func declare is not actually passed to this func.
```
func() error {
  token, err = ecrClient.GetAuthorizationToken(context.Background(), &ecr.GetAuthorizationTokenInput{})
  return err
}
```
* Where this func actually run is in`util.RetryExponentialBackoff`. So if we want to implement like passing a point param to this func(), 

  * The `token1` declared here need to be passed to `RetryExponentialBackoff` method. And also we need to define the func declare here to accept another `token2` in param
  ```
  var token1 *ecr.GetAuthorizationTokenOutput
  err := util.RetryExponentialBackoff(token1, 3, 2*time.Second, func(token2 *ecr.GetAuthorizationTokenOutput) error {
		token2, err = ecrClient.GetAuthorizationToken(context.Background(), &ecr.GetAuthorizationTokenInput{})
		return err
  })
  ```

  * And in `RetryExponentialBackoff` method, we can pass `token1` param to `f(token1)` where the func is actually run.
  ```
  func RetryExponentialBackoff(token1 *ecr.GetAuthorizationTokenOutput, attempts int, initial time.Duration, f func(*ecr.GetAuthorizationTokenOutput) error) error {
	var err error
	wait := initial
	for i := 0; i < attempts; i++ {
		if err = f(token1); err == nil {
			return nil
		}
		time.Sleep(wait)
		wait *= 2
	}
	return err
  }
  ```

But no matter how, it will require us to create another new `RetryExponentialBackoff` method in util.go

nodeadm/internal/aws/ecr/ecr.go
```diff
@@ -19,7 +21,14 @@ func GetAuthorizationToken(awsRegion string) (string, error) {
...
 	}
 	ecrClient := ecr.NewFromConfig(awsConfig)
-	token, err := ecrClient.GetAuthorizationToken(context.Background(), &ecr.GetAuthorizationTokenInput{})
+	var token *ecr.GetAuthorizationTokenOutput
+	err = util.RetryExponentialBackoff(3, 2*time.Second, func() error {
+		token, err = ecrClient.GetAuthorizationToken(context.Background(), &ecr.GetAuthorizationTokenInput{})
+		if err != nil {
+			return err
+		}
+		return nil
```
ndbaker1, line 30: nit: you can collapse these :) 
```suggestion
		return err
```

ndbaker1 (changes requested):

nodeadm/internal/aws/ecr/ecr.go
wwvela: Revised!

ndbaker1 (approved):

Issue #245, opened 2019-04-10
Can we implement EKS-AMI hardening?


khetanvallurupalli:
As per Our Infosec team, Every server should be using Hardened AMI according to there policies.
While we do the same for EKS AMI worker nodes are terminated before starting. Any suggestions?

rickard-von-essen:
Did you check why?

khetanvallurupalli:
As there AMI has passed the CIS benchmark test. with some agents like Splunk and TrendMicro are Baked into it that Hardened AMI.

micahhausler:
@khetanvallurupalli We have an issue (#99) for CIS benchmarks, would that cover your use case or are there additional changes?

pthrasher:
it may not be accurate... but I believe there's sort of 2 levels here. Linux ami hardening, and then kube hardening. Both with separate benchmarks from CIS.

burnertoday:
@pthrasher I believe you are correct:

Kubenetes hardening: CIS_Kubernetes_Benchmark_v1.4.1.pdf
AMI Hardening: CIS_Amazon_Linux_Benchmark_v2.1.0.pdf

There is no specific and official CIS hardened AMI for EKS that I can find. If anyone else can please point us.

hawkesnc:
Hi everyone,
Looking for some direction here. Is this on AWS' roadmap? I'm currently looking at trying to harden the AMI myself (specifically the linux AMI hardening) but if the work is already being done then that's great. Otherwise, I'm willing to open a PR but I would like some direction the preferred approach.

hawkesnc:
> Looking for aws hardened image, what are the options available currently.

Nothing official from AWS that I can find, but there are python/ansible scripts that you can search up that are open to use on Github

KYannick:
Does anyone has an idea what the status is of this issue? Are there plans to provide hardened images for EKS?

peteroruba:
I am surprised this receives so little attention.

Gangaram-Dewasi:
team, is there any update on this ? is AWS planing to provide a CIS AL2 hardened EKS AMI's which can be used as a part of the cluster node group ?

shazinahmed:
This [official repo](https://github.com/aws-samples/amazon-eks-custom-amis) has packer scripts to create custom hardened amis.

pierluigilenoci:
@mmerkes @abeer91 @heybronson is there any way to get AWS feedback on this?

mmerkes:
[...]

PR #1896, opened 2024-07-24
:sparkles: (feat): Add Support for Hardened Images

KrisJohnstone:
**Description of changes:**
When using hardened images (i.e. CIS or STIG) there are a few changes required to get everything running and have nodes join the cluster.

**Testing Done**
Here's a hardened image running in one of our test clusters:
image
image

Note: Accidentally closed old PR.


Issacwww:
Could you please share some context?
why only AL2023 need this change?

KrisJohnstone:
> Could you please share some context? why only AL2023 need this change?

AL2 EOL is just under a year away. For anyone whom going forth wants stateramp/fedramp accreditation, using AL2 as a basis is a bad idea (IMO). And from what issues I'd found during research it seemed like support was already there in AL2 (I've not tested personally).

templates/al2023/provisioners/hardened-images.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+set -o pipefail
+set -o nounset
+set -o errexit
+
+if [[ "$HARDENED_IMAGE" == "true" ]]; then
```
bryantbiggs, line 7: what are the downsides of setting these by default? where would this pose an issue?

cartermckinnon:
"Hardened" is very subjective 😄 and I don't think we want to attempt defining it in this repository at this time -- that will be a much larger effort. We'll continue tracking it in #245

KrisJohnstone:
@cartermckinnon what about if i changed the var to be SELINUX_ENFORCED or something like that?

templates/al2023/provisioners/hardened-images.sh
KrisJohnstone: Sorry my backlogs been rather large recently and missed this. Providing the binary exists by default then I can't foresee any.

PR #1524, opened 2023-11-18
[WIP] Check for ecr-fips endpoint availability

cartermckinnon:
**Description of changes:**

Adds a check that the `ecr-fips` endpoint actually exists falling back to the default ECR endpoint if not.

The FIPS endpoint only exists in a handful of regions: https://aws.amazon.com/compliance/fips/

And I'd like to avoid hard-coding that list here, so I've used a DNS lookup.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1897, opened 2024-07-24
Fall back to non fips endpoint per availability

Issacwww:
**Description of changes:**
apply same fix like https://github.com/awslabs/amazon-eks-ami/pull/1524 as FIPS endpoint only exists in a handful of regions: https://aws.amazon.com/compliance/fips/

### DIFF
--- a/nodeadm/internal/aws/ecr/ecr.go
+++ b/nodeadm/internal/aws/ecr/ecr.go
@@ -3,6 +3,7 @@ package ecr
 import (
 	"context"
 	"fmt"
+	"go.uber.org/zap"
 	"net"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


Issacwww:
/ci
+workflow:os_distros al2023
+workflow:k8s_versions 1.29

nodeadm/internal/aws/ecr/ecr.go
```diff
@@ -49,9 +49,8 @@ func GetEKSRegistry(region string) (ECRRegistry, error) {
 	}
 	if fipsInstalled && fipsEnabled {
 		fipsRegistry := getRegistry(account, "ecr-fips", region, servicesDomain)
-		if addresses, err := net.LookupHost(fipsRegistry); err != nil {
-			return "", err
-		} else if len(addresses) > 0 {
+		addresses, err := net.LookupHost(fipsRegistry)
+		if err == nil && len(addresses) > 0 {
```
ndbaker1, line 53: should we log whenever we skip this path? some other system log might catch this request, but i want it to be obvious when this path is not taken since i think `fipsEnabled` sets the expectation that the outcome will be fips

nodeadm/internal/aws/ecr/ecr.go
```diff
@@ -50,7 +51,7 @@ func GetEKSRegistry(region string) (ECRRegistry, error) {
 	if fipsInstalled && fipsEnabled {
 		fipsRegistry := getRegistry(account, "ecr-fips", region, servicesDomain)
 		if addresses, err := net.LookupHost(fipsRegistry); err != nil {
-			return "", err
+			zap.L().Info("Fail to look up Fips registry for requested region, fall back to default", zap.String("fipsRegistry", fipsRegistry))
 		} else if len(addresses) > 0 {
```
ndbaker1, line 55: think you still want to consider this for logging the fallback?

ndbaker1 (approved):
thanks!

Issue #1751, opened 2024-04-05
Slow startup of AL2023 AMI


stijndehaes:
After trying the AL2023 AMI I noticed startup is way slower compared to the AL2 image. 
After some help from @ndbaker1 I made the following analysis using `systemd-analyze`.

I added a plot using `systemd-analyze`, the nodeadm component appears to be in the hoth path.
With `nodeadm-config.service` taking 20s before it allows `cloud-init.service` (7.2s) to start, and then `nodeadm-run.service` takes 10.53s, I think it is waiting for containerd-service (8.7s) to start though. 

Looking at the logs of nodeadm-config, it appears to take a long time to start before the first log(5s), but also takes 13s between starting to configure kubelet and getting the kubelet version. 

Looking at the logs of nodeadm-run, most time apears to be spend looking up the sandbox image. I noticed in code this is done using `containerd config dump` and a regex, for some reason this takes a long time.

Not sure what can be done to optimize these steps, but I am willing to help :) I can write golang code and make custom AMI's with custom `nodeadm` if needed to test things out.

![plot](https://github.com/awslabs/amazon-eks-ami/assets/5019818/2cb3edb8-cd3d-4ffc-b017-818d2e97868a)

nodeadm-config logs, via `journalctl -u nodeadm-config`
```
Apr 04 08:56:49 localhost systemd[1]: Starting nodeadm-config.service - EKS Nodeadm Config...
Apr 04 08:56:54 localhost nodeadm[1693]: {"level":"info","ts":1712221014.1338603,"caller":"init/init.go:49","msg":"Checking user is root.."}
Apr 04 08:56:54 localhost nodeadm[1693]: {"level":"info","ts":1712221014.1339903,"caller":"init/init.go:57","msg":"Loading configuration..","configSource":"imds://user-data"}
Apr 04 08:56:54 localhost nodeadm[1693]: {"level":"info","ts":1712221014.2056038,"caller":"init/init.go:66","msg":"Loaded configuration","config":{"metadata":{"creationTimestamp":null},"spec":{"cluster":{"name":"datafy-dp-eks-dev-sdh","apiServerEndpoin>
Apr 04 08:56:54 localhost nodeadm[1693]: {"level":"info","ts":1712221014.2059212,"caller":"init/init.go:68","msg":"Enriching configuration.."}
[...]

PR #1910, opened 2024-08-05
Disable motd

stijndehaes:
Disabling motd results in a faster boot of the
AMI.

**Issue #, if available:** #1751

**Description of changes:** 

By disabling motd the AMI is able to boot up faster. The difference is about 20s. Analysis was done using systemd-analyze:

Before change:
Startup finished in 698ms (kernel) + 7.574s (initrd) + 1min 2.208s (userspace) = 1min 10.481s graphical.target reached after 1min 2.146s in userspace.

After change:
Startup finished in 454ms (kernel) + 3.842s (initrd) + 38.513s (userspace) = 42.810s graphical.target reached after 38.055s in userspace.

**Testing Done**: Launching a custom AMI made from one of the release AMI's

### DIFF
--- a/templates/al2023/template.json
+++ b/templates/al2023/template.json
@@ -140,6 +140,12 @@
     {
       "type": "shell",
       "inline": [
+        "sudo systemctl disable update-motd"
+      ]
+    },
+    {
+      "type": "shell",
+      "inline": [
         "sudo mkdir -p /etc/eks/log-collector-script/",
         "sudo cp -v {{user `working_dir`}}/log-collector-script/eks-log-collector.sh /etc/eks/log-collector-script/"
       ]


stewartsmith:
Is this observable with the stock AL2023 images? We did make sure the update-motd service didn't block the OS launching.

cartermckinnon:
This service doesn't block any of the important services (`nodeadm-run`, `containerd`, `kubelet`), so I don't see much benefit in disabling it. The "update is available" message written by `/etc/update-motd.d/70-available-updates` is helpful, IMO.

cartermckinnon:
Let me know if you disagree @stijndehaes

stijndehaes:
@cartermckinnon fair enough :)
I do notice kubelet launching 3s faster without motd, but that is probably within margin of error.
I would like to improve the hot code path, I am leaving on a 10 days of vacation tomorrow, but after that I would love to improve `nodeadm-config` and `nodeadm-run`. 
`nodeadm-config` uses kubelet making it slow, and `nodeadm-run` uses containerd making it slow.
Is there a better place than github issues to discuss solution designs? :)

ndbaker1:
appreciate the attention here, i also think it would benefit from a bit of caching and/or parallelism here so i might take a stab at it when i have time

PR #1916, opened 2024-08-09
bug(log-collector-script): look for kubelet config in the right place on AL2

cartermckinnon:
**Description of changes:**

Noticed our log bundles are missing the kubelet config in many cases, because on AL2 we put it at:
```
/etc/kubernetes/kubelet/kubelet-config.json
```
instead of:
```
/etc/kubernetes/kubelet/config.json
```

### DIFF
--- a/log-collector-script/linux/eks-log-collector.sh
+++ b/log-collector-script/linux/eks-log-collector.sh
@@ -474,6 +474,8 @@ get_k8s_info() {
 
       cp --force --recursive --dereference /etc/kubernetes/kubelet/config.json "${COLLECT_DIR}"/kubelet/config.json 2> /dev/null
       cp --force --recursive --dereference /etc/kubernetes/kubelet/config.json.d "${COLLECT_DIR}"/kubelet/config.json.d 2> /dev/null
+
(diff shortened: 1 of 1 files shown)


log-collector-script/linux/eks-log-collector.sh
```diff
@@ -474,6 +474,8 @@ get_k8s_info() {
 
       cp --force --recursive --dereference /etc/kubernetes/kubelet/config.json "${COLLECT_DIR}"/kubelet/config.json 2> /dev/null
       cp --force --recursive --dereference /etc/kubernetes/kubelet/config.json.d "${COLLECT_DIR}"/kubelet/config.json.d 2> /dev/null
+
+      cp --force --recursive --dereference /etc/kubernetes/kubelet/kubelet-config.json "${COLLECT_DIR}"/kubelet/config.json 2> /dev/null
```
dims, line 478: same destination file name as line 475?

cartermckinnon: Should be safe to assume only one of those source paths exist, but this is for debug after all :) fixing...

Issacwww (approved):

PR #1738, opened 2024-03-28
make 80-ec2.network match primary eni only

M00nF1sh:
**Issue #, if available:**
N/A

**Description of changes:**
Override the default `80-ec2.network` generated by amazon-ec2-net-utils to make systemd.network manage primary ENI only.
After this change, the sequence is as follows:
1. during initial boot, the `80-ec2.network` by amazon-ec2-net-utils will match all ENIs(only primary ENI at this moment), thus systemd.network will configure them properly with DHCP.
2. the `nodeadm-run` unit will execute, which will generate a `80-ec2.network` that matches the Mac of primary ENI(from ec2-metadata) only and reload the config.
3. after kubelet init and vpc-cni runs, vpc-cni will launch new secondary ENIs as pod scheduled, and those secondary ENIs will be unmanaged per `systemd.network`.

This is a temporary fix for EKS AL2023 AMI, where the `80-ec2.network` generated by amazon-ec2-net-utils will cause systemd.network managing all ENIs on host, and caused multiple issues including:
 1. systemd.network races against vpc-cni to configure secondary enis and might cause routing rules/routes from vpc-cni been flushed.
 6. routes for those secondary enis obtained from dhcp will appear in main route table, which is a drift from our AL2 behavior.

To address this issue temporarily, we override the `80-ec2.network` after boot to make it match against primary ENI only.
* the top-level `mac` from ec2 metadata is guaranteed to primary eni's mac and guaranteed to be available at boot according to [this code comment](https://github.com/amazonlinux/amazon-ec2-net-utils/blob/main/lib/lib.sh#L392) and internal code links.

TODOs after this PR: there are limitations on current solutions as well, and we should figure long term solution for this:
 1. the altNames for ENIs(a new feature in AL2023) were setup by amazon-ec2-net-utils via udev rules, but it's disabled by eks.
 2. the "unmanaged" ENI feature from vpc-cni is not supported, as those enis won't be setup by systemd.network.

Alternative solutions consider:
1. we considered to use a static file override to make `80-ec2.network` match `IFINDEX: 2`. However, after communications with ec2 networking, the primary ENI may not always be `IFINDEX: 2`.
[...]

PR #1924, opened 2024-08-14
Add build scripts for building Nvidia and Neuron AMIs based on AL2023

nkvetsinski:
**Issue #, if available:**

**Description of changes:**

Adding build scripts for building Nvidia and Neuron AMIs based on AL2023 OS.

**Testing Done**

***Neuron:***
1. Built the AMI.
2. Started an instance from the AMI.
3. Instance was able to join EKS cluster and after installing Neuron device plugin, Neuron devices were added as capacity to the node K8s object.

***Nvidia:***
1. Built 2 AMIs using the currently available drivers (555 and 560 versions) in the Nvidia [repo](https://developer.download.nvidia.com/compute/cuda/repos/amzn2023/).
2. Started instances from the AMIs.
3. Instances were able to join EKS cluster and after installing Nvidia device plugin, Nvidia GPUs were added as capacity to the nodes K8s objects.
4. Deployed Pod and executed `nvidia-smi` from it. 

Nodeadm E2E tests are also passing.


README.md
```diff
@@ -42,10 +42,22 @@ make k8s=1.29
...
 make k8s=1.29 os_distro=al2023
 
+# build Neuron AMI with a specific Kubernetes version
+make k8s=1.29 os_distro=al2023 accelerator_vendor=neuron
+
+# build Nvidia AMI with a specific Kubernetes version and default driver major version
+make k8s=1.29 os_distro=al2023 accelerator_vendor=nvidia
+
+# build Nvidia AMI with a specific Kubernetes version and a specific driver major version
+make k8s=1.29 os_distro=al2023 accelerator_vendor=nvidia nvidia_major_driver_version=550
```
Issacwww, line 52: Is 550 provided in the nvidia repo? I think we should add something around for how to check available versions

templates/al2023/variables-default.json
```diff
@@ -17,6 +18,7 @@
     "kms_key_id": "",
     "iam_instance_profile": "",
     "launch_block_device_mappings_volume_size": "20",
+    "nvidia_major_driver_version": "",
```
Issacwww, line 21: nit: can we setup default version here instead of Makefile?

README.md
nkvetsinski: My bad, 555 and 560 are available. Checking for available versions might be tricky, we need to check what is available in the repo.

templates/al2023/variables-default.json
nkvetsinski: yes, we can do that

nkvetsinski: Actually, I think we'll need to keep the default version in the Makefile, because we need it in the AMI_VARIANT variable. 

```
nvidia_major_driver_version := $(NVIDIA_MAJOR_DRIVER_VERSION_DEFAULT)
AMI_VARIANT := $(AMI_VARIANT)-$(nvidia_major_driver_version)
```

templates/al2023/provisioners/install-efa.sh
```diff
@@ -0,0 +1,51 @@
...
+if ! gpg --verify ./aws-efa-installer-${EFA_VERSION}.tar.gz.sig &> /dev/null ;then
+    echo "EFA Installer signature failed verification!"
+    exit 2
+fi
+
+##########################################################################################
+### Install and cleanup ##################################################################
+##########################################################################################
+tar -xf ${EFA_PACKAGE} && cd aws-efa-installer
+sudo ./efa_installer.sh -y
```
bryantbiggs, line 47: The user space libraries aren't required on the host:

```suggestion
sudo ./efa_installer.sh --minimal -y
```

templates/al2023/provisioners/install-neuron-driver.sh
```diff
@@ -0,0 +1,35 @@
...
+gpgcheck=1
+gpgkey=https://yum.repos.neuron.amazonaws.com/GPG-PUB-KEY-AMAZON-AWS-NEURON.PUB
+metadata_expire=0" |sudo tee /etc/yum.repos.d/neuron.repo
+
+################################################################################
+### Install packages ###########################################################
+################################################################################
+sudo dnf --setopt=install_weak_deps=False install -y \
+  aws-neuronx-dkms \
+  aws-neuronx-oci-hook \
```
bryantbiggs, line 34: we can drop the hooks, no longer needed - just the driver kernel mod - https://github.com/aws-neuron/aws-neuron-sdk/issues/794#issuecomment-1826938932

templates/al2023/provisioners/install-nvidia-driver.sh
```diff
@@ -0,0 +1,115 @@
...
+
+sudo mv ${WORKING_DIR}/gpu/nvidia-kmod-load.sh /etc/eks/
+sudo mv ${WORKING_DIR}/gpu/initialize_nvidia_clock.sh /etc/eks/
+sudo mv ${WORKING_DIR}/gpu/bootstrap_gpu.sh /etc/eks/
+sudo mv ${WORKING_DIR}/gpu/bootstrap_gpu.service /etc/systemd/system/bootstrap_gpu.service
+sudo systemctl daemon-reload
+sudo systemctl enable bootstrap_gpu
+
+################################################################################
+### Install Cuda ###############################################################
```
bryantbiggs, line 66: CUDA is massive and also a user space library - it should not be on the host, only the container toolkit

templates/al2023/runtime/gpu/initialize_nvidia_clock.sh
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env bash
```
bryantbiggs, line 1: this seems like something that would be better handled by nodeadm - for reference https://github.com/clowdhaus/eksnode/blob/main/eksnode/src/gpu.rs

templates/al2023/runtime/rootfs/etc/systemd/system/configure-multicard-interfaces.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
```
bryantbiggs, line 1: is this still required with AL2023 since we are now using ec2 net utils (AL2 we did not)?

templates/al2023/provisioners/install-efa.sh
nkvetsinski: sounds good

templates/al2023/provisioners/install-neuron-driver.sh
nkvetsinski: sounds good

templates/al2023/runtime/rootfs/etc/systemd/system/configure-multicard-interfaces.service
nkvetsinski: hm, let me double check

templates/al2023/provisioners/install-nvidia-driver.sh
bryantbiggs: correct - the cuda toolkit, cuda runtime, cuda libs, etc. all go into the container and not on the host. 

The host components are mostly kernel modules:
- NVIDIA driver
- GPUDirect RDMA (installed as part of EFA installer)
- FabricManager
- GDRCopy

And then the container toolkit of course

templates/al2023/runtime/gpu/initialize_nvidia_clock.sh
bryantbiggs: you could do it as a systemd unit (with bash script) - I'm just not that good with some gnarly bash-fu 😅 

also, not all cards will accept the max settings

templates/al2023/provisioners/install-nvidia-driver.sh
bryantbiggs: this should also save quite a bit in terms of consumed storage - a little over 9Gb I believe

templates/al2023/runtime/rootfs/etc/systemd/system/configure-multicard-interfaces.service
nkvetsinski: I'm going to drop this script for now. I'm discussing with VPC CNI people whether this is still relevant.

bryantbiggs: a few prior PRs for reference

- https://github.com/awslabs/amazon-eks-ami/pull/1528#discussion_r1404621827
- https://github.com/awslabs/amazon-eks-ami/pull/1539
- https://github.com/awslabs/amazon-eks-ami/pull/1738
- https://github.com/awslabs/amazon-eks-ami/blob/4f6fb3c563a078310e2fbd8fd1d60011c860254e/nodeadm/internal/system/networking.go#L19-L21

templates/al2023/runtime/gpu/bootstrap_gpu.sh
cartermckinnon, line 1: 1. let's use something other than "bootstrap" in naming, I don't want to cause confusion with the old AL2 script(s).
2. If all this needs to do is run the 2 other scripts with some error handling, it can be as simple as:
```
set -o errexit

/etc/eks/initialize_nvidia_clock.sh
/etc/eks/nvidia-kmod-load.sh
```

templates/shared/runtime/bin/gpu-ami-util
cartermckinnon, line 1: if these aren't actually used by the AL2 templates (yet), let's move them into the al2023/runtime directory so they don't end up in AL2 AMI's.

templates/al2023/provisioners/install-nvidia-driver.sh
```diff
@@ -0,0 +1,115 @@
...
+
+declare -A NVIDIA_DRIVER_CUDA_VERSION_MAP=(["555"]="12-5" ["560"]="12-6")
+CUDA_VERSION=${NVIDIA_DRIVER_CUDA_VERSION_MAP[${NVIDIA_MAJOR_DRIVER_VERSION}]}
+
+if [ -z $CUDA_VERSION ]; then
+  echo "undefined CUDA version for Nvidia driver version: ${NVIDIA_MAJOR_DRIVER_VERSION}"
+  exit 1
+fi
+
+sudo dnf --setopt=install_weak_deps=False -y install cuda-toolkit-${CUDA_VERSION} cuda-*-${CUDA_VERSION}-*
```
cartermckinnon: were you going to remove this? the toolkit can go, at least

templates/al2023/provisioners/install-nvidia-driver.sh
```diff
@@ -0,0 +1,115 @@
...
+echo "[Unit]
+Description=Display Nvidia driver EULA
+
+[Service]
+Type=oneshot
+ExecStart=/etc/eks/nvidia-eula
+
+[Install]
+WantedBy=basic.target
+"| sudo tee /etc/systemd/system/nvidia-eula.service
```
cartermckinnon, line 97: Let's just move this to a separate file

templates/al2023/provisioners/install-nvidia-driver.sh
```diff
@@ -0,0 +1,115 @@
...
+
+################################################################################
+### Install drivers ############################################################
+################################################################################
+sudo mkdir -p /etc/dkms
+echo "MAKE[0]=\"'make' -j$(grep -c processor /proc/cpuinfo) module\"" | sudo tee /etc/dkms/nvidia.conf
+sudo dnf -y install kernel-modules-extra.x86_64
+
+function archive-open-kmods(){
+  sudo dnf --setopt=install_weak_deps=False -y module install nvidia-driver:${NVIDIA_MAJOR_DRIVER_VERSION}-open
```
cartermckinnon, line 28: What is `--setopt=install_weak_deps=False` for?

templates/al2023/provisioners/install-neuron-driver.sh
```diff
@@ -0,0 +1,35 @@
...
+
+if [ "$ACCELERATOR_VENDOR" != "neuron" ]; then
+  exit 0
+fi 
+
+PARTITION=$(imds "/latest/meta-data/services/partition")
+
+if [ "$PARTITION" == "aws-iso" ] || [ "$PARTITION" == "aws-iso-b" ]; then
+  echo "Neuron repository not vailable in isolated regions"
+  exit 0
```
cartermckinnon, line 15: I think we want the build to fail in this case, so you don't end up with an AMI that looks like it contains the neuron bits, but doesn't.

Makefile
```diff
@@ -30,6 +31,16 @@ ifeq ($(enable_fips), true)
 	AMI_VARIANT := $(AMI_VARIANT)-fips
 endif
 
+ifeq ($(os_distro), al2023)
+	ifdef accelerator_vendor
+		AMI_VARIANT := $(AMI_VARIANT)-$(accelerator_vendor)
+		ifeq ($(accelerator_vendor), nvidia)
+			nvidia_major_driver_version := $(NVIDIA_MAJOR_DRIVER_VERSION_DEFAULT)
```
cartermckinnon, line 38: The user can't override this without `?=`. I'm not sure we want this in the top level Makefile anyway -- if the user really wants the driver version to appear in the AMI name, they can set the `ami_name` variable to whatever they want. The default value for the driver version var should be defined in the `variables-default.json`.

README.md
```diff
@@ -42,10 +42,22 @@ make k8s=1.29
...
 make k8s=1.29 os_distro=al2023
 
+# build Neuron AMI with a specific Kubernetes version
+make k8s=1.29 os_distro=al2023 accelerator_vendor=neuron
+
+# build Nvidia AMI with a specific Kubernetes version and default driver major version
+make k8s=1.29 os_distro=al2023 accelerator_vendor=nvidia
+
+# build Nvidia AMI with a specific Kubernetes version and a specific driver major version
+make k8s=1.29 os_distro=al2023 accelerator_vendor=nvidia nvidia_major_driver_version=555
```
cartermckinnon, line 52: I don't think we need this in the top-level README, it shouldn't be commonly used. Lets just document it in the template docs.

nodeadm/internal/containerd/oci_config.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+var execCommand CommandExecutor = RealCommandExecutor{}
+var fileWriter FileWriter = RealFileWriter{}
+var nvidiaCtkCommand = "/usr/bin/nvidia-ctk"
+var nvidiaCtkParams = []string{"--quiet", "runtime", "configure", "--runtime=containerd", "--set-as-default", "--dry-run"}
+
+func isAllowedCommand(command string, args ...string) bool {
+	return command == "/usr/bin/nvidia-ctk" && reflect.DeepEqual(args, nvidiaCtkParams)
+}
```
cartermckinnon, line 87: This is a bit odd. We don't need a generic `CommandExecutor` interface if it can only do one thing 😉

That interface (and it's impls) shouldn't be public either, AFAICT.

nodeadm/internal/containerd/oci_config.go
```diff
@@ -0,0 +1,148 @@
...
+
+	"github.com/awslabs/amazon-eks-ami/nodeadm/internal/api"
+	"github.com/awslabs/amazon-eks-ami/nodeadm/internal/util"
+	"github.com/pelletier/go-toml/v2"
+	"go.uber.org/zap"
+)
+
+type instanceTypeMixin struct {
+	instanceFamilies []string
+	apply            func(*[]byte) error
```
cartermckinnon, line 19: slices are passed by reference (for the most part), so a `*[]byte` can just be `[]byte` in this case

nodeadm/internal/containerd/oci_config.go
```diff
@@ -0,0 +1,148 @@
...
+		apply:            applyNeuron,
+	}
+
+	mixins = []instanceTypeMixin{
+		NvidiaInstanceTypeMixin,
+		NeuronInstanceTypeMixin,
+	}
+)
+
+// applyInstanceTypeMixins adds the needed OCI hook options to containerd config.toml
```
cartermckinnon, line 40: Comment here (and the name of the file) could be clearer. Is the intention of this interface to be a general-purpose way to do something special for certain instance types, or do we only want to handle alternative low-level runtimes?

nodeadm/internal/containerd/oci_config.go
```diff
@@ -0,0 +1,148 @@
...
+	return command == "/usr/bin/nvidia-ctk" && reflect.DeepEqual(args, nvidiaCtkParams)
+}
+
+// applyNvidia adds the needed Nvidia containerd options using nvidia container toolkit:
+// https://github.com/NVIDIA/nvidia-container-toolkit
+func applyNvidia(containerdConfig *[]byte) error {
+	zap.L().Info("Configuring Nvidia OCI hook..")
+	// before calling nvidia-ctk, we'll write the generated containerd config first
+	// nvidia-ctk will use it as a base
+	err := fileWriter.WriteFileWithDir(containerdConfigFile, *containerdConfig, containerdConfigPerm)
```
cartermckinnon, line 95: We need to avoid writing out the config in this file. The mixins should add to the containerdConfig blob and allow the result to be written to disk elsewhere (after all config routines are done).

If that means we stop using nvidia-ctk for this, I think that's okay. All we're doing is adding the same block of TOML with `neuron`/`nvidia` (and the BinaryName) plugged in.

templates/al2023/runtime/gpu/bootstrap_gpu.service
cartermckinnon, line 1: please use kebob-case for file names, to be consistent with what's already in this repo

templates/al2023/runtime/gpu/bootstrap_gpu.sh
nkvetsinski: will do

templates/shared/runtime/bin/gpu-ami-util
nkvetsinski: good point, I'll move them

templates/al2023/provisioners/install-nvidia-driver.sh
nkvetsinski: yes, I'm dropping CUDA.

templates/al2023/provisioners/install-nvidia-driver.sh
nkvetsinski: will do

nodeadm/internal/containerd/oci_config.go
nkvetsinski: I added the `CommandExecutor` so that I can unit test the code more easily(just passing mock command in the test).

nodeadm/internal/containerd/oci_config.go
nkvetsinski: okay, will change it

templates/al2023/provisioners/install-nvidia-driver.sh
cartermckinnon: Gotcha. Not a bad idea but it ultimately depends on the maintainer of the RPM spec to mark things correctly. We don't have much/any control over the spec, so unless this saves a lot of space I'd err on just installing the weak dependencies 🤷

nkvetsinski: I'll see how much we save, the AMI size dropped down after removing CUDA and the unneeded Neuron packages.

nodeadm/internal/containerd/oci_config.go
nkvetsinski: For now the intention is to only apply alternative low-level runtimes (if needed). I say if needed, because in the last commit I removed the special handling for Neuron, which Braynt commented was no longer needed. So for now it's just adding the nvidia runtime config.

nodeadm/internal/containerd/oci_config.go
nkvetsinski: I admit it's a bit weird, having to write the containerd file first so that we can generate the resulting conifg and write the final file somewhere else. If you're OK with not using `nvidia-ctk` command to generate the config I can do that. As you said, these are well known values.

Makefile
nkvetsinski: > The user can't override this without `?=`.

Not sure what you mean by that, one can override the `nvidia_major_driver_version` like so:

```
make k8s=1.28 os_distro=al2023 accelerator_vendor=nvidia nvidia_major_driver_version=560
```

But since you're the second reviewer to point this out and I totally missed the fact that `ami_name` can also be set, I'll put the version in `variables-default.json`.

README.md
nkvetsinski: okay

nodeadm/internal/containerd/oci_config.go
cartermckinnon: Yeah I think everything going on in this file would be simpler if we ditched `nvidia-ctk`, let's go that route

nodeadm/internal/containerd/oci_config.go
cartermckinnon: I think the instanceTypeMixin interface makes sense, just need a clearer file name

templates/al2023/runtime/gpu/initialize_nvidia_clock.sh
cartermckinnon: It's gonna take a lot more golang to accomplish the same thing as the bash script 😄 plus we don't need to block on it in nodeadm. A separate unit makes sense to me, downstream of `nvidia-persistenced`

nkvetsinski: Right now this file and loading the kernel modules are in the same unit. I can put them in their own.

nkvetsinski: Why I ended up doing is:
`nodeadm -> load-kmod -> nvidia-persistenced -> init-clocks`

README.md
```diff
@@ -58,4 +58,18 @@ For security issues or concerns, please do not open an issue or pull request on
...
-This sample code is made available under a modified MIT license. See the LICENSE file.
+This sample code is made available under a MIT-0 license. See the LICENSE file.
+
+Although this repository is released under the MIT license, when using Nvidia accelerated AMIs you agree to the NVIDIA Cloud End User License Agreement: https://s3.amazonaws.com/EULA/NVidiaEULAforAWS.pdf.
+
+Although this repository is released under the MIT license, Nvidia accelerated AMIs  
+use the third party [open-gpu-kernel-modules](https://github.com/NVIDIA/open-gpu-kernel-modules). The open-gpu-kernel-modules project's licensing includes the dual MIT/GPLv2 license.
+
+Although this repository is released under the MIT license, Nvidia accelerated AMIs  
+use the third party [nvidia-container-toolkit](https://github.com/NVIDIA/nvidia-container-toolkit). The nvidia-container-toolkit project's licensing includes the Apache-2.0 license.
```
cartermckinnon, line 69: style nit: 
```suggestion
Although this repository is released under the MIT license, when using NVIDIA accelerated AMIs you agree to the NVIDIA Cloud End User License Agreement: https://s3.amazonaws.com/EULA/NVidiaEULAforAWS.pdf.

Although this repository is released under the MIT license, NVIDIA accelerated AMIs  
use the third party [open-gpu-kernel-modules](https://github.com/NVIDIA/open-gpu-kernel-modules). The open-gpu-kernel-modules project's licensing includes the dual MIT/GPLv2 license.

Although this repository is released under the MIT license, NVIDIA accelerated AMIs  
use the third party [nvidia-container-toolkit](https://github.com/NVIDIA/nvidia-container-toolkit). The nvidia-container-toolkit project's licensing includes the Apache-2.0 license.
```

doc/usage/al2023.md
```diff
@@ -28,6 +29,7 @@
 | `kubernetes_build_date` |  |
 | `kubernetes_version` |  |
 | `launch_block_device_mappings_volume_size` |  |
+| `nvidia_major_driver_version` | Driver version to install, depends on what is available in Nvidia upstream yum repository. |
```
cartermckinnon, line 32:
```suggestion
| `nvidia_driver_major_version` | Driver version to install, depends on what is available in Nvidia upstream yum repository. |
```

templates/al2023/runtime/gpu/initialize-nvidia-clock.sh
```diff
@@ -0,0 +1,45 @@
...
+if ! gpu-ami-util has-nvidia-devices; then
+  echo >&2 "no NVIDIA devices are present, nothing to do!"
+  exit 0
+fi
+
+# gpu boost clock
+if command -v nvidia-smi &> /dev/null; then
+  echo >&2 "INFO: nvidia-smi found"
+
+  nvidia-smi -q > /tmp/nvidia-smi-check
```
ndbaker1, line 16: is this existence check needed anymore? if it doesn't exist then the unit fails and that doesn't hurt anything does it?

templates/al2023/runtime/gpu/nvidia-kmod-load.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Loading Nvidia kernel modules
+After=nodeadm-run.service
```
ndbaker1, line 3: Can you drop a blurb here about what we're waiting for nodeadm to do before starting?

templates/al2023/runtime/gpu/gpu-ami-util
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env bash
+
+function is-isolated-partition() {
```
ndbaker1, line 3: this is unused, think its because you needed the partition name anyway in the other scripts

templates/al2023/provisioners/configure-nvidia-eula.sh
```diff
@@ -0,0 +1,37 @@
...
+echo "[Unit]
+Description=Display Nvidia driver EULA
+
+[Service]
+Type=oneshot
+ExecStart=/etc/eks/nvidia-eula
+
+[Install]
+WantedBy=basic.target
+"| sudo tee /etc/systemd/system/nvidia-eula.service
```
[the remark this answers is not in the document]
nkvetsinski: Adding it to `al2023/runtime/rootfs/` will make it available to plain AL2023, I thought that might be a problem.

nodeadm/internal/containerd/runtime_config.go
```diff
@@ -0,0 +1,78 @@
...
+[plugins.'io.containerd.grpc.v1.cri'.containerd]
+default_runtime_name = 'nvidia'
+discard_unpacked_layers = true
+[plugins."io.containerd.grpc.v1.cri".containerd.runtimes.nvidia]
+base_runtime_spec = "/etc/containerd/base-runtime-spec.json"
+runtime_type = "io.containerd.runc.v2"
+[plugins."io.containerd.grpc.v1.cri".containerd.runtimes.nvidia.options]
+BinaryName = "/usr/bin/nvidia-container-runtime"
+SystemdCgroup = true
+`
```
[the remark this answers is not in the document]
nkvetsinski: 1. Moved out of the function and declared as constant.
2. I'm not sure how swapping will help. IIUC `[plugins."io.containerd.grpc.v1.cri".containerd.runtimes.nvidia]` and `[plugins."io.containerd.grpc.v1.cri".containerd.runtimes.nvidia.options]` are not present in `config.template.toml`, so if we want them to present in the resulting config, we'll have pass everything we need as is (SystemdCgroup, BinaryName, etc).

nodeadm/internal/containerd/runtime_config.go
```diff
@@ -0,0 +1,78 @@
...
+
+	if containerdConfig != nil {
```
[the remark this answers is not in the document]
nkvetsinski: You're right, it won't be.

templates/al2023/provisioners/install-efa.sh
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env bash
+set -o pipefail
+set -o nounset
+set -o errexit
+
+if [ "$ACCELERATOR_VENDOR" != "neuron" ] && [ "$ACCELERATOR_VENDOR" != "nvidia" ]; then
+  exit 0
+fi
```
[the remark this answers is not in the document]
nkvetsinski: make sense

nodeadm/internal/containerd/runtime_config.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+func (m *instanceTypeMixin) matches(cfg *api.NodeConfig) bool {
+	instanceFamily := strings.Split(cfg.Status.Instance.Type, ".")[0]
+	return slices.Contains(m.instanceFamilies, instanceFamily)
+}
+
+var (
+	// TODO: fetch this list dynamically
+	nvidiaInstances         = []string{"p3", "p3dn", "p4d", "p4de", "p5", "g4", "g4dn", "g5", "g6"}
```
bryantbiggs, line 25:
```suggestion
	nvidiaInstances         = []string{"p3", "p3dn", "p4d", "p4de", "p5", "g4", "g4dn", "g5", "g6", "g6e"}
```

Hopefully pulling this from the EC2 API is a fast follow 😅

bryantbiggs: I would actually vote we add this now, or proactively add the instance types we know are launching fairly soon.

templates/al2023/provisioners/install-neuron-driver.sh
```diff
@@ -29,4 +29,4 @@ metadata_expire=0" |sudo tee /etc/yum.repos.d/neuron.repo
 ################################################################################
 ### Install packages ###########################################################
 ################################################################################
-sudo dnf --setopt=install_weak_deps=False install -y aws-neuronx-dkms
\ No newline at end of file
+sudo dnf install -y aws-neuronx-dkms
```
bryantbiggs, line 32: I think adding the `aws-neuronx-tools ` would be a worthy 140mb - the Neuron AMI should be pretty lightweight as is. You can think of it as `nvidia-smi` for Neuron somewhat

templates/al2023/template.json
```diff
@@ -170,6 +179,14 @@
         "ENABLE_FIPS={{user `enable_fips`}}"
       ]
     },
+    {
+      "type": "shell",
+      "expect_disconnect": true,
+      "script": "{{template_dir}}/runtime/bin/limit-c-states.sh",
```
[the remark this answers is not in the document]
nkvetsinski: you're right

templates/al2023/runtime/gpu/initialize-nvidia-clock.sh
nkvetsinski: yes, that's some leftovers, `nvidia-smi` should always be present on nvidia ami

templates/al2023/runtime/bin/limit-c-states.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+set -o pipefail
+set -o nounset
+set -o errexit
+
+if [ "$ACCELERATOR_VENDOR" == "neuron" ] || [ "$ACCELERATOR_VENDOR" == "nvidia" ]; then
```
bryantbiggs, line 6: I think this will go along with [Carter's comment above](https://github.com/awslabs/amazon-eks-ami/pull/1924/files#r1727776198) for the use case where EFA is used, regardless of whether or not accelerators or GPUs are utilized/present

templates/al2023/runtime/gpu/initialize-nvidia-clock.sh
```diff
@@ -0,0 +1,45 @@
...
+    elif [[ $GPUNAME == *"V100"* ]]; then
+      nvidia-smi -ac 877,1530
+    elif [[ $GPUNAME == *"K80"* ]]; then
+      nvidia-smi -ac 2505,875
+    elif [[ $GPUNAME == *"T4"* ]]; then
+      nvidia-smi -ac 5001,1590
+    elif [[ $GPUNAME == *"M60"* ]]; then
+      nvidia-smi -ac 2505,1177
+    elif [[ $GPUNAME == *"H100"* ]]; then
+      nvidia-smi -ac 2619,1980
```
bryantbiggs, line 36: if we aren't going to dynamically pull these at boot, we should proactively add all known cards - `L40S` (g6e) was just launched - `B200`, `GB200`, etc.?

templates/al2023/runtime/gpu/gpu-ami-util
nkvetsinski: you're right, I don't think we need this now

templates/al2023/runtime/gpu/initialize-nvidia-clock.sh
bryantbiggs: also, do we need to set `nvidia-smi -lgc (--lock-gpu-clocks) xx` and `nvidia-smi -lmc (--lock-memory-clocks) yy`? do we test and validate that the clock speeds are maintained across reboots?

templates/al2023/runtime/gpu/nvidia-kmod-load.service
nkvetsinski: Hm, thinking about it again, we don't really need to wait for nodeadm to setup the node. Loading the modules and nodeadm doing its thing can be done at the same time right?

nodeadm/internal/containerd/runtime_config.go
nkvetsinski: I'd rather do it on a separate PR, this one is big enough. In the meantime any other instance type beside `g6e` that's worth adding?

templates/al2023/provisioners/install-neuron-driver.sh
nkvetsinski: sounds good, will add it

templates/al2023/provisioners/install-nvidia-driver.sh
```diff
@@ -52,51 +55,21 @@ archive-open-kmods
...
 ################################################################################
-### Prepare for nvidia bootstrap ###############################################
+### Prepare for nvidia init ####################################################
 ################################################################################
 
 sudo mv ${WORKING_DIR}/gpu/nvidia-kmod-load.sh /etc/eks/
-sudo mv ${WORKING_DIR}/gpu/initialize_nvidia_clock.sh /etc/eks/
-sudo mv ${WORKING_DIR}/gpu/bootstrap_gpu.sh /etc/eks/
-sudo mv ${WORKING_DIR}/gpu/bootstrap_gpu.service /etc/systemd/system/bootstrap_gpu.service
+sudo mv ${WORKING_DIR}/gpu/initialize-nvidia-clock.sh /etc/eks/
```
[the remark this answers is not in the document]
nkvetsinski: `configure-nvidia-clocks`, naming sucks?

nodeadm/internal/containerd/runtime_config.go
bryantbiggs: makes sense - no, just `g6e` is missing from the current list of supported instances. we should also have a "louder" failure for unsupported instances (i.e. - `p2`) if possible

templates/al2023/provisioners/install-nvidia-driver.sh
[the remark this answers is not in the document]
bryantbiggs: > naming sucks?

it does!

`set-nvidia-clocks.sh`? 🤷🏽

templates/al2023/provisioners/install-nvidia-driver.sh
```diff
@@ -52,51 +55,21 @@ archive-open-kmods
...
-sudo systemctl enable bootstrap_gpu
-
+sudo systemctl enable nvidia-kmod-load.service
+sudo systemctl enable initialize-nvidia-clock.service
 
 ################################################################################
 ### Install other dependencies #################################################
 ################################################################################
-sudo dnf --setopt=install_weak_deps=False -y install nvidia-fabric-manager nvidia-container-toolkit
+sudo dnf -y install nvidia-fabric-manager nvidia-container-toolkit
```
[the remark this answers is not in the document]
nkvetsinski: I'm inclined to leave it off for now based on the [caveats](https://developer.nvidia.com/gdrcopy) and the CUDA [requirement](https://github.com/NVIDIA/gdrcopy/blob/master/README.md#requirements). @cartermckinnon do you know if we had customer requests for this?

templates/al2023/runtime/gpu/initialize-nvidia-clock.sh
nkvetsinski: > if we aren't going to dynamically pull these at boot, we should proactively add all known cards - `L40S` (g6e) was just launched - `B200`, `GB200`, etc.?

I'll go over the existing instances and make sure we have all cards.




> also, do we need to set `nvidia-smi -lgc (--lock-gpu-clocks) xx` and `nvidia-smi -lmc (--lock-memory-clocks) yy`? do we test and validate that the clock speeds are maintained across reboots?

IIUC from `nvidia-smi` help, the flag that we currently set ( `--applications-clocks`) 
specifies the clocks when there is something running on the GPU. While the flags you mention are setting gpu and mem clocks at all times (with or without load). We don't use lock-gpu and lock-mem for the other GPU AMI that we support, so if you think it's beneficial, we can set them.

I'll check the reboot persistence, because this is executed as systemd unit, it should persist across reboots.

templates/al2023/runtime/bin/limit-c-states.sh
nkvetsinski: sounds good

templates/al2023/provisioners/install-nvidia-driver.sh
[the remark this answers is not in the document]
cartermckinnon: NVIDIA doesn't provide an RPM for GDRCopy for AL2023 right now: https://developer.download.nvidia.com/compute/redist/gdrcopy/

This has come up in the past but we determined the utility of it wasn't clear. My understanding is it's mostly beneficial on aging instance types, it doesn't help much on p5's, for example.

nodeadm/internal/containerd/runtime_config.go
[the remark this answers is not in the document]
cartermckinnon: I see what you mean. That's really not ideal, we don't want to have to keep this blob in sync with what the rest of the config flow does to these nested structs. Does the nvidia-ctk clone your existing runtime config and just change the name? How does this work there?

templates/al2023/provisioners/configure-nvidia-eula.sh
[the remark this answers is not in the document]
cartermckinnon: chatted offline, we'll try adding this to motd instead of a service unit

nodeadm/internal/containerd/runtime_config.go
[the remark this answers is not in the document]
nkvetsinski: `nvidia-ctk` uses `/etc/containerd/config.toml` as base and does an in-place change (adding whatever it needs to the base). This is why initially I was writing the config file first, calling nvidia-ctk, passing the resulting config back. Let me see what will happen if we call it on empty file and merge with our template.

doc/usage/al2023.md
```diff
@@ -23,11 +24,13 @@
 | `enable_fips` | Install openssl and enable fips related kernel parameters |
 | `encrypted` |  |
 | `iam_instance_profile` | The name of an IAM instance profile to launch the EC2 instance with. |
+| `install_efa` | Valid options are ```true``` or ```false```. Wheather or not to install the software needed to use AWS Elastic Fabric Adapter (EFA) network interfaces. |
```
cartermckinnon, line 27: `enable_efa` would align better with existing vars, I think

doc/usage/al2023.md
```diff
@@ -5,6 +5,7 @@
 <!-- template-variable-table-boundary -->
 | Variable | Description |
 | - | - |
+| `accelerator_vendor` | Vendor that provides the GPU or accelerator hardware. Currently we support Neuron and NVIDIA. |
```
cartermckinnon, line 8: not loving this name 😂 maybe `enable_accelerator`? what else have you considered?

also should mention in the doc for `nvidia_driver_major_version` that it's only used when this var == `nvidia`

cartermckinnon (commented):
Some final nits 👍

nodeadm/internal/containerd/runtime_config.go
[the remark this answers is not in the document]
cartermckinnon: I'm wondering about how `nvidia-ctk` handles options like `SystemdCgroup`, `base_runtime_spec`, etc. It must be copying our existing `runc` runtime config and just plugging in `nvidia` and `BinaryName`.

Could you just update our existing containerd config template such that you can swap in `nvidia`? https://github.com/awslabs/amazon-eks-ami/blob/main/nodeadm/internal/containerd/config.template.toml

doc/usage/al2023.md
nkvetsinski: >what else have you considered?

you don't wanna know :)

I'm good with `enable_accelerator`

Issacwww:
please fix the lint, should be easy via `make fmt && make lint` under the root path

Issacwww:
/ci
+workflow:os_distros al2023

nodeadm/internal/containerd/runtime_config.go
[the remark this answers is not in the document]
nkvetsinski: I think I addressed that, check the latest code.

templates/al2023/runtime/gpu/initialize-nvidia-clock.sh
nkvetsinski: Clock settings persist reboot, I left out the lgc and lmc flags.

nkvetsinski:
> please fix the lint, should be easy via `make fmt && make lint` under the root path

done

templates/al2023/runtime/gpu/nvidia-kmod-load.sh
```diff
@@ -0,0 +1,75 @@
...
+        return 1
+      fi
+    done
+  done
+  return 0
+}
+
+# load the nvidia kernel module appropriate for the attached devices
+MODULE_NAME=""
+if devices-support-open; then
```
bryantbiggs, line 47: since the older cards that require the proprietary driver are a known quantity, and new cards that will be released in the future are unknown to this script (and therefore the PCI class codes are unknown here) - should we flip this logic to default to using the open GPU kernel module unless we know the proprietary driver is absolutely required?

templates/al2023/runtime/gpu/set-nvidia-clocks.sh
```diff
@@ -0,0 +1,36 @@
...
+set -o xtrace
+
+if ! gpu-ami-util has-nvidia-devices; then
+  echo >&2 "no NVIDIA devices are present, nothing to do!"
+  exit 0
+fi
+
+# gpu boost clock
+sudo nvidia-smi -pm 1 # set persistence mode
+sudo nvidia-smi --auto-boost-default=0
```
bryantbiggs, line 14: in the final else case below, we should not disable auto boost if the card is unknown to this script

if a new instance is launched, we don't want to hamper its performance in any form if we skip updating this script - let it fall back to the default behavior of the card

templates/al2023/runtime/gpu/initialize-nvidia-clock.sh
bryantbiggs: sounds good to me!

bryantbiggs:
looks great to me - two comments that can be followed up in later PRs to ensure we fail open to the appropriate path
:shipit:

templates/al2023/runtime/gpu/nvidia-kmod-load.sh
nkvetsinski: Let's check with @Issacwww about that. I'm not the original writer of this.

Issacwww: I think this is a fair call out, I will follow up with next PR as current one is big enough :)

nkvetsinski: Sounds good, I'm also in favor of changing this in different PR.

templates/al2023/runtime/gpu/set-nvidia-clocks.sh
nkvetsinski: @bryantbiggs do you mind if we address this in another PR?

bryantbiggs: absolutely - as stated below, the two added comments here are great candidates for follow PRs if we think the changes are worth. But I think this turkey is cooked as is so - :shipit:

bryantbiggs (approved):

templates/al2023/runtime/gpu/nvidia-kmod-load.sh
cartermckinnon: > new cards that will be released in the future are unknown to this script (and therefore the PCI class codes are unknown here)

The PCI class codes don't change per device, these have been the same for a few decades now. :)

> should we flip this logic to default to using the open GPU kernel module unless we know the proprietary driver is absolutely required

I didn't go this route because it's way messier. Compiling an exhaustive list of every PCI device ID used by those EC2 instance types is quite a chore. With the current approach, we generate the supported device file from the driver archive and do a simple lookup.

Long term, yes we want to default to the open source kmod. NVIDIA is going to do that in the 560 driver branch, and we can revisit/refactor accordingly then, IMO.

cartermckinnon (approved):

PR #1810, opened 2024-05-17
fix shell variable expansion for cache images

ndbaker1:
**Issue #, if available:**

**Description of changes:**

fixes an issue where missing addons causes the build to fail when caching images.

this was originally fixed in https://github.com/awslabs/amazon-eks-ami/pull/1133, but since removing the unconditional `PAUSE_CONTAINER` image, it's now possible for the entire `CACHE_IMGS` variable to be empty, resulting in

```
line 486: CACHE_IMGS[@]: unbound variable
```

added empty string as a fallback to empty or unset values of `CACHE_IMGS` and `PULLED_IMGS`, and shortened the alternative syntax `+${...}` to `:-` which avoids rewriting the same expression.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

**Testing Done**

*[See this guide for recommended testing for PRs.](../doc/CONTRIBUTING.md#testing-changes) Some tests may not apply. Completing tests and providing additional validation steps are not required, but it is recommended and may reduce review time and time to merge.*

PR #1926, opened 2024-08-16
bug(al2): skip empty container images

ndbaker1:
**Issue #, if available:**

**Description of changes:**

follow up for https://github.com/awslabs/amazon-eks-ami/pull/1810. when there are no images to cache during worker installation, the blank variable `img` will be passed to the `pull-image.sh` script.

the reason for the issue is the fallback `"${VAR[@]:-}"` still results in a single loop iteration over the empty string `""`, but would not do so with `${VAR[@]:-}`

this PR short circuits when the loop variable is blank, which fixes this case and other cases where the image is actually blank, and also adds logging to help diagnose issues with the images in the future.

**Testing Done**

build with the fix alleviates the issue.


ndbaker1:
/ci build

ndbaker1:
had to redo the approach since the linter feel strongly about quoting array expansions

/ci build

Issacwww (approved):

Issacwww:
Thanks for the quick fix! 🚀

cartermckinnon (approved):

PR #1927, opened 2024-08-19
Added LLM-based issue summarizer.

trm109:
**Issue #, if available:**
N/A
**Description of changes:**
Added a GitHub Action that will summarize issues via: 
- `/summarize`
  - *Will summarize the issue in which the command is being called*
- `/summarize 123`
  - *Will summarize issue number 123*
- `/summarize bob dotfiles 123`
  - *Will summarize issue number 123 of the repository: bob/dotfiles*

**Example Output:**
> /summarize awslabs amazon-eks-ami 1751

>Here's a summary of the GitHub Issue reply chain:
>
>Issue: Startup is significantly slower on the AL2023 AMI compared to the AL2 AMI for Amazon EKS.
>
>Key points:
>1. The issue creator (stijndehaes) noticed that startup time increased from about 40 seconds on AL2 to over 100 seconds on AL2023.
>
>2. Analysis using systemd-analyze showed that nodeadm components were in the critical path, with nodeadm-config.service and nodeadm-run.service taking significant time.
>
>3. Investigations revealed that getting the kubelet version was a major bottleneck, taking about 13 seconds.
>
>4. Enabling EBS fast-restore improved startup time to around 50 seconds, but it was still slower than AL2.
>
>5. Experiments were conducted to optimize the startup process:
>   - Caching the kubelet version in an environment variable reduced nodeadm-config.service time from 21s to 5.12s.
>   - Disabling update-motd.service further improved startup time.
>
>6. Despite optimizations, AL2023 startup remained slower than AL2 (about 40s vs. 22s).
>
>Conclusions/Proposed Solutions:
>1. A pull request was created to disable update-motd.service for performance improvement.
>2. Caching the kubelet version during build time was suggested to avoid slow executable access during startup.
>3. The team is considering various optimizations to reduce startup time, including changing how the kubelet version is obtained and stored.
>
>The issue remains open, with ongoing discussions about further optimizations to bring AL2023 startup time closer to that of AL2.

The bot utilizes AWS Bedrock to trigger a `InvokeModelCommand` to Claude 3.5 Sonnet. In order to achieve this, a few things will have to be configured on a AWS Account:
- [An OIDC Identity Provider for GitHub](https://docs.github.com/en/actions/security-for-github-actions/security-hardening-your-deployments/configuring-openid-connect-in-amazon-web-services)
- (An IAM Role with access to Bedrock)[https://docs.aws.amazon.com/bedrock/latest/userguide/security_iam_id-based-policy-examples.html]
  - This role must have a trust policy allowing the GitHub OIDC Identity Provider to `sts:AssumeRoleWithWebIdentity`

**⚠️ This Repository must also provide the ARN of this role in Secrets as BEDROCK_ACTION_ROLE_ARN**

**Testing Done**
- Tested on forked repository.
- Tested on awslabs/amazon-eks-ami, awslabs/aws-shell, hashicorp/packer and verified robustness of summarization.
- Tested auth (confirmed that only owner/members can invoke command)

### DIFF
--- a/.github/actions/summarizer/index.js
+++ b/.github/actions/summarizer/index.js
@@ -6,6 +6,14 @@ const context = github.context;
 const { BedrockRuntimeClient, InvokeModelCommand } = require("@aws-sdk/client-bedrock-runtime");
 
 (async () => {
+
+  const author = payload.comment.user.login;
+  const authorized = ["OWNER", "MEMBER"].includes(payload.comment.author_association);
+  if (!authorized) {
+    console.log(`Comment author is not authorized: ${author}`);
+    return;
+  }
+  console.log(`Comment author is authorized: ${author}`);
   // Split the command into parts
   const parts = process.env.COMMENT_BODY.trim().split(' ');
 
--- a/.github/actions/summarizer/index.js
+++ b/.github/actions/summarizer/index.js
@@ -2,11 +2,11 @@ const github  = require('@actions/github');
 const core    = require('@actions/core');
 const token   = process.env.GITHUB_TOKEN;
 const octokit = new github.getOctokit(token);
-const context = github.context;
 const { BedrockRuntimeClient, InvokeModelCommand } = require("@aws-sdk/client-bedrock-runtime");
 
 (async () => {
-
+  const context = github.context;
+  const payload = context.payload;
   const author = payload.comment.user.login;
   const authorized = ["OWNER", "MEMBER"].includes(payload.comment.author_association);
   if (!authorized) {
@@ -60,7 +60,7 @@ const { BedrockRuntimeClient, InvokeModelCommand } = require("@aws-sdk/client-be
       Assistant:
     `
   });
-  const payload = {
+  const modelInput = {
     anthropic_version: "bedrock-2023-05-31",
     max_tokens: 16384, // Adjust this if issue comment chain is long.
     messages: messages
@@ -68,11 +68,11 @@ const { BedrockRuntimeClient, InvokeModelCommand } = require("@aws-sdk/client-be
 
   const command = new InvokeModelCommand({
     contentType: "application/json",
-    body: JSON.stringify(payload),
+    body: JSON.stringify(modelInput),
     modelId: process.env.MODEL_ID,
   });
 
-  console.log("Prompting LLM with:\n" + JSON.stringify(payload));
+  console.log("Prompting LLM with:\n" + JSON.stringify(modelInput));
 
   try {
     const response = await client.send(command);
--- a/.github/actions/summarizer/index.js
+++ b/.github/actions/summarizer/index.js
@@ -9,15 +9,20 @@ const { BedrockRuntimeClient, InvokeModelCommand } = require("@aws-sdk/client-be
   const payload = context.payload;
   const author = payload.comment.user.login;
   const authorized = ["OWNER", "MEMBER"].includes(payload.comment.author_association);
+  // Ensure that invoker is either a Owner or member of awslabs / amazon-eks-ami
   if (!authorized) {
     console.log(`Comment author is not authorized: ${author}`);
     return;
   }
   console.log(`Comment author is authorized: ${author}`);
+
   // Split the command into parts
   const parts = process.env.COMMENT_BODY.trim().split(' ');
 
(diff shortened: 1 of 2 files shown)


.github/workflows/summarize.yml
```diff
@@ -0,0 +1,44 @@
...
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          role-to-assume: ${{ secrets.BEDROCK_ACTION_ROLE_ARN }}
+          aws-region: ${{ env.AWS_REGION }}
+          role-session-name: GITHUB_ACTION
+        # invoke Bedrock to summarize the issue
+      - name: Summarize issues
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          MODEL_ID    : "anthropic.claude-3-5-sonnet-20240620-v1:0"
```
cartermckinnon, line 42: If this is going to be hard-coded, it can just go in your Action js script

.github/actions/summarizer/index.js
```diff
@@ -0,0 +1,106 @@
...
+
+  const command = new InvokeModelCommand({
+    contentType: "application/json",
+    body: JSON.stringify(modelInput),
+    modelId: process.env.MODEL_ID,
+  });
+
+  console.log("Prompting LLM with:\n" + JSON.stringify(modelInput));
+
+  try {
```
cartermckinnon, line 85: 1. too much going on in a single try/catch
2. if you just catch the error, log it, and rethrow -- you don't need the try/catch

.github/actions/summarizer/index.js
```diff
@@ -0,0 +1,106 @@
...
+    }
+  ];
+
+  messages[0].content.push({
+    type: "text",
+    text: `
+      Human: ${prompt}
+      Assistant:
+    `
+  });
```
cartermckinnon, line 69: Why don't you move this into the declaration of `messages` above?

.github/actions/summarizer/index.js
```diff
@@ -0,0 +1,106 @@
...
+  commentLog += `${issue.user.login} created the issue:\n "${issue.body}"\n`
+
+  for (const comment of comments) {
+    if(
+      (comment.user.login != "github-actions[bot]") && 
+      (!comment.body.startsWith("/"))
+    ){
+      commentLog += `${comment.user.login} says:\n"${comment.body}"\n`
+    }
+  }
```
cartermckinnon, line 49: Building the string like this is a bit clunky, maybe try something like:
```
const commentLog = comments.filter(c => c.user.login != "github-actions[bot]")
    .map(c => `${c.user.login} says: ...`)
    .join('\n');
```

.github/actions/summarizer/index.js
```diff
@@ -0,0 +1,106 @@
...
+  const context = github.context;
+  const payload = context.payload;
+  const author = payload.comment.user.login;
+  const authorized = ["OWNER", "MEMBER"].includes(payload.comment.author_association);
+  // Ensure that invoker is either a Owner or member of awslabs / amazon-eks-ami
+  if (!authorized) {
+    console.log(`Comment author is not authorized: ${author}`);
+    return;
+  }
+  console.log(`Comment author is authorized: ${author}`);
```
cartermckinnon, line 17: We need to extend the existing bot vs reimplementing the authz here

.github/actions/summarizer/index.js
```diff
@@ -0,0 +1,106 @@
...
+
+  // Split the command into parts
+  const parts = process.env.COMMENT_BODY.trim().split(' ');
```
cartermckinnon, line 20: What if the comment has more than just the command in it? what if it has multiple commands?

.github/workflows/summarize.yml
trm109: True, will migrate in next revision.

.github/actions/summarizer/index.js
trm109: Good catch, revising in next revision.

.github/actions/summarizer/index.js
trm109: Currently, it requires that you only type a single command. Working on moving the summarizer over to the existing bot.

PR #1952, opened 2024-09-04
Configure multicard interfaces

nkvetsinski:
**Issue #, if available:**

**Description of changes:**
In the AMI we force `systemd-networkd` to only manage the primary interface. In this commit I'm adding a script that will configure ip addresses and routes for interfaces that are not managed by `systemd-networkd`.

**Testing Done**

I verified that:

* Pod networking is working as expected (pods on local and remote machines can communicate)
* Non zero card interfaces on remote machines can communicate. I tested this with `iperf`, starting a server on one pod and binding it to non zero card interface. On another pod (running on different machine), using `iperf` client (bound to non zero card interface). Running `tcpdump` on both machines and verifying traffic is flowing thru the correct interfaces.

### DIFF
--- a/templates/al2023/runtime/rootfs/usr/bin/configure-multicard-interfaces.sh
+++ b/templates/al2023/runtime/rootfs/usr/bin/configure-multicard-interfaces.sh
@@ -1,0 +1,69 @@
+#!/usr/bin/env bash
+
+# EKS already forces amazon-ec2-net-utils to manage the primary ENI only (0/0):
+# https://github.com/awslabs/amazon-eks-ami/pull/1738
+#
+# When VPC CNI is installed it creates another ENI from network card with index 0 (0/1).
+# VPC CNI will continue adding additional ENIs until max number of interfaces for the intance
+# is reached (0/2, 0/3, etc).
+#
+# This script configures IP and routing for Elastic Network Interfaces that are part of
+# non-zero indexed EC2 network cards (1/0, 1/1, 2/1, 3/1, etc). The way we find out whether
+# we need to configure the interface is by checking IMDS:
+# /latest/meta-data/network/interfaces/macs/${mac_address}/network-card/
+# This script will skip any interfaces that are part of the 0 indexed card.
+
+cni_managed_card_index=0
+
+instance_type=$(imds "/latest/meta-data/instance-type")
+echo "instance type is $instance_type"
+
+macs_array=($(imds '/latest/meta-data/network/interfaces/macs'))
+
+if [[ "${#macs_array[@]}" -le 1 ]]; then
+  echo "this instance does not have multiple network cards, skip configuration"
+  exit 0
+fi
+
+table_id=1001
+pref_id=32765
+
+for mac in "${macs_array[@]}"; do
+  trimmed_mac=$(echo $mac | sed 's:/*$::')
+  if_name=$(ip -o link show | grep -F "link/ether $trimmed_mac" | awk -F'[ :]+' '{print $2}')
+  ec2_card_index=$(imds "/latest/meta-data/network/interfaces/macs/${trimmed_mac}/network-card/")
+
+  if [ "$ec2_card_index" -eq "$cni_managed_card_index" ]; then
+    echo "skipping cni managed interface ${if_name} - ${trimmed_mac}"
+  else
+    echo "handling interface $if_name"
+
+    if_ip_addr=$(imds "/latest/meta-data/network/interfaces/macs/$trimmed_mac/local-ipv4s" | head -1)
+    vpc_subnet_cidr=$(imds "/latest/meta-data/network/interfaces/macs/$trimmed_mac/subnet-ipv4-cidr-block")
+
+    network=$(echo ${vpc_subnet_cidr} | awk -F'/' '{print $1}')
+    netmask=$(echo ${vpc_subnet_cidr} | awk -F'/' '{print $2}')
+    router=$(($(echo ${network} | cut -d. -f4) + 1))
+    default_gw_ip="$(echo ${network} | cut -d. -f1-3).${router}"
+
+    echo "configuring IP addr: ${if_ip_addr}/${netmask} for ${if_name} ..."
+    ip link set $if_name down
+    ip addr add $if_ip_addr/$netmask metric 1024 dev $if_name
+    ip link set dev $if_name mtu 9001
+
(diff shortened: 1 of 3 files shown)


templates/al2023/runtime/rootfs/usr/bin/configure-multicard-interfaces.sh
```diff
@@ -0,0 +1,69 @@
...
+#
+# When VPC CNI is installed it creates another ENI from network card with index 0 (0/1).
+# VPC CNI will continue adding additional ENIs until max number of interfaces for the intance
+# is reached (0/2, 0/3, etc).
+#
+# This script configures IP and routing for Elastic Network Interfaces that are part of
+# non-zero indexed EC2 network cards (1/0, 1/1, 2/1, 3/1, etc). The way we find out whether
+# we need to configure the interface is by checking IMDS:
+# /latest/meta-data/network/interfaces/macs/${mac_address}/network-card/
+# This script will skip any interfaces that are part of the 0 indexed card.
```
cartermckinnon, line 14: A+++++ comment, thx.

I'm wondering if we can leverage systemd-networkd here instead of setting all the routes up ourselves. We modify the default networkd config so that it only targets the primary ENI (by using `PermanentMACAddress`). Can we add more MAC's there for the other interfaces on other cards? Or create additional networkd configs for those cards?

nkvetsinski: If we add the mac addresses to the current drop-in file, the default routes for them will end up in the main route table, which I think it's not a good idea.

If we want to leverage `systemd-networkd` we'll have to create separate `.network` file for the non zero cards populate the mac addresses and configure the policy routing. I was thinking about that too, but figured it is probably easier/clear to setup everything with `iproute` package commands.  I'm fine with changing direction, also asked somebody more familiar with networking to take a look, let's see what their take will be too :)

cartermckinnon: > it is probably easier/clear to setup everything with iproute package commands

I would much rather defer to networkd 😂  we have to phone a friend every time we touch this script because it's pretty esoteric stuff

suket22: +1 to seeing what we can do to avoid adding this into our AMIs for AL2023. I think we need to figure out how amazon-ec2-net-utils can support management of just the primary ENI and let others (like VPC CNI) handle all secondary ENI management.

We're also going through this much effort to bring up network interfaces that aren't on network card 0, but those interfaces aren't actually used by K8s pods in any case since we only assign IPs from network card 0 right? This is something we've got to chase with the VPC CNI to make it network card aware, but until then I wonder how much value this PR brings us other than consistency with our AL2 AMI.

nkvetsinski: >+1 to seeing what we can do to avoid adding this into our AMIs for AL2023. I think we need to figure out how amazon-ec2-net-utils can support management of just the primary ENI and let others (like VPC CNI) handle all secondary ENI manageme

We already "tell" ec2-net-utils to manage the primary ENI, by using [drop-in](https://github.com/awslabs/amazon-eks-ami/pull/1738). When checking with VPC CNI people I was told that there is backlog item about handling non primary interfaces, but I didn't get anything more than that. I can ask further. 

>We're also going through this much effort to bring up network interfaces that aren't on network card 0, but those interfaces aren't actually used by K8s pods in any case since we only assign IPs from network card 0 right?

Potential use case is if someone runs the pod in the host network namespace and specifically binds their app to one of non zero card interfaces.

>I wonder how much value this PR brings us other than consistency with our AL2 AMI.

My main goal was parity with AL2 AMI.

Pavani-Panakanti: > +1 to seeing what we can do to avoid adding this into our AMIs for AL2023. I think we need to figure out how amazon-ec2-net-utils can support management of just the primary ENI and let others (like VPC CNI) handle all secondary ENI management

Making VPC CNI aware of network cards other than card 0 is going to be a much bigger change and currently not prioritized by the networking team. We went ahead with same approach as AL2 AMI to not block the AL2023 AMI on this.
cc: @jayanthvn on setting up routes via iproute vs using networkd by creating separate .network files

jayanthvn: The CNI plugin will not manage the ENI setup and routes for the primary ENIs on non-zero network cards, even if the CNI plugin manages the secondary ENIs on the non-zero network cards.

M00nF1sh:
1. I prefer to use drop-ins to configure those non-zero network cards whenever possible.  As I think it's better to align with EC2's original behavior of AL2023 for those ENIs not managed by vpc-cni as much as possible instead of align with original EKS AL2 behavior.
2. do we need a support flag in node-adm to conditionally enable/disable this behavior? as it's a drift from previous AL2023 behav

nkvetsinski:
> 1. I prefer to use drop-ins to configure those non-zero network cards whenever possible.  As I think it's better to align with EC2's original behavior of AL2023 for those ENIs not managed by vpc-cni as much as possible instead of align with original EKS AL2 behavior.
> 
>     2. do we need a support flag in node-adm to conditionally enable/disable this behavior? as it's a drift from current EKS AL2023 behavior

@M00nF1sh few cla

nkvetsinski:
> 1. I prefer to use drop-ins to configure those non-zero network cards whenever possible.  As I think it's better to align with EC2's original behavior of AL2023 for those ENIs not managed by vpc-cni as much as possible instead of align with original EKS AL2 behavior.
> 
>     2. do we need a support flag in node-adm to conditionally enable/disable this behavior? as it's a drift from current EKS AL2023 behavior

We synced on the side, Yang is in favor of creating separate `.network` files for the interfaces. I want to get an alignment with @cartermckinnon whether we need this in `nodeadm`. My take is that it will be easier to do it as shell script and since it's not a super critical part of node bootstrap I'd rather not add it `nodeadm` workflow.

cartermckinnon:
We already modify the existing networkd config in `nodeadm`, so I think this should go there as well (plus it's more easily unit tested vs a shell script)

nkvetsinski:
I'm going to open a new PR for this guys, since we're changing the approach.

Issue #1868, opened 2024-06-29
[k8s>=1.30] switch off NodeSwap feature gate for AL2 based images


dims:
As of https://github.com/kubernetes/kubernetes/pull/122745, `NodeSwap` has been switched to `true` by default. Since we know that `NodeSwap` works only with cgroupv2 and AL2 defaults to cgroupv1, can we switch off the feature gate?

See https://github.com/kubernetes/test-infra/pull/32876 for context where we had to do this to ensure CI jobs are stabilized with AL2.

cc @cartermckinnon @njtran

cartermckinnon:
We'd need to add some logic for older (and future) k8s versions where the feature flag doesn't exist, and since using swap requires you to also disable `failOnSwap` *and* configure the swap volumes, I'm not sure it makes sense to have this additional guardrail/callout. @ndbaker1 what do you think here?

ndbaker1:
> add some logic for older (and future) k8s versions where the feature flag doesn't exist

we don't need to do anything for older version right?  

If we already had logic in place to enforce `NodeSwap=false` it would make a lot of sense, but it sounds like the default makes sense from the PR since it won't be configures as `NoSwap`

I'm also not eager to track more featureGates besides the ones we currently set (which are all related to credentials afaict)

cartermckinnon:
> we don't need to do anything for older version right?

the feature gate was added in 1.22, so we can't specify it for 1.21 or kubelet will complain. When the feature gate is removed in the future, we'll have to come back and add a check for that boundary as well

> If we already had logic in place to enforce NodeSwap=false it would make sense, but sounds like the default from the PR is NoSwap which works as expected?

yeah that's my thinking, the feature gate default changing was a no-op anyway, since you have to opt-in with several other config options

cartermckinnon:
I think we'll leave things as-is for now

PR #1958, opened 2024-09-08
disable node swap on AL2

Issacwww:
**Issue #, if available:**
https://github.com/awslabs/amazon-eks-ami/issues/1868
There is some instance have swap turned on, `c1.medium` and when it run with AL2 + 1.30+ AMIs, node failed to join the cluster

**Description of changes:**
Disable NodeSwap feature flag and not fail when swap on to unblock node join cluster

**Testing Done**
- repo: launch node with latest AMI on c1.medium, node fail to join
- fix: node able to join with fx

### DIFF
--- a/.github/workflows/dependency-review.yaml
+++ b/.github/workflows/dependency-review.yaml
@@ -19,7 +19,7 @@ jobs:
         with:
           go-version-file: 'nodeadm/go.mod'
       - run: go install github.com/securego/gosec/v2/cmd/gosec@latest
-      - run: gosec -exclude-generated ./...
+      - run: gosec -exclude-generated -exclude=G115 ./...
         working-directory: nodeadm
   govulncheck:
     runs-on: ubuntu-latest
--- a/templates/al2/runtime/kubelet-config.json
+++ b/templates/al2/runtime/kubelet-config.json
@@ -26,8 +26,10 @@
   "readOnlyPort": 0,
   "cgroupDriver": "cgroupfs",
   "cgroupRoot": "/",
+  "failSwapOn": false,
   "featureGates": {
-    "RotateKubeletServerCertificate": true
+    "RotateKubeletServerCertificate": true,
+    "NodeSwap": false
   },
   "protectKernelDefaults": true,
   "serializeImagePulls": false,


templates/al2/runtime/bootstrap.sh
```diff
@@ -548,6 +548,10 @@ fi
 
 KUBELET_ARGS="$KUBELET_ARGS --cloud-provider=$KUBELET_CLOUD_PROVIDER"
 
+if vercmp "$KUBELET_VERSION" gteq "1.30.0"; then
+  KUBELET_ARGS="$KUBELET_ARGS --fail-swap-on=false --feature-gates=NodeSwap=false"
```
cartermckinnon, line 552: We should do this in the config file instead of with flags, IMO. We also shouldn't change the default for `failSwapOn`; if the feature gate is disabled that param shouldn't matter?

cartermckinnon: > default failSwapOn is true. so this flag does matter in terms of node join

That would only happen if the feature gate is enabled, right?

Issacwww: I think we need it per this [doc](https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/)
> Enable the kubelet to allocate swap memory for Kubernetes workloads on a node. Must be used with KubeletConfiguration.failSwapOn set to false. For more details, please see [swap memory](https://kubernetes.io/docs/concepts/architecture/nodes/#swap-memory)

cartermckinnon: https://github.com/kubernetes/kubernetes/commit/32c265f60c175cedbd685f30e389c95fb1e66ee3

looks like this option has been around since at least 1.12, and it's never been tied to the `NodeSwap` feature gate 🤨.

cartermckinnon: I guess we can set it to `false`, it just seems odd to set this option in such a way to support swap -- when we are explicitly *not* allowing swap

Issacwww: yes, I feel this 2 level of feature flag control is odd.. 
above test I did was add 
```
  "featureGates": {
    "RotateKubeletServerCertificate": true,
    "NodeSwap": false
  },
```
but still run into the same issue.
running on t3 for another test

dims: > looks like this option has been around since at least 1.12, and it's never been tied to the `NodeSwap` feature gate 🤨.

experimental in 1.8 🙃 https://github.com/kubernetes/kubernetes/pull/47181/commits/71e8c8eba43a0fade6e4edfc739b331ba3cc658a

cartermckinnon: lol touché!

I think it's safe to say that `c1.medium` and `m1.small` have never worked on EKS with AL2. I'm leaning "lets just document this and move on" vs touching the swap settings, which could have downstream impact on custom AMIs or other users of boostrap.sh (like canonical for Ubuntu)

Issue #1984, opened 2024-10-01
Check for ecr-fips endpoint availability - VPC Endpoint


jeremymcgee73:
**What happened**:
There is a bug when you have enabled FIPS on the image, in a region with FIPS endpoints, and have VPC endpoints enabled. The issue is that the check implemented in #1524 , checks to see if the FIPS endpoint resolves. In an isolated environment, the endpoint does resolve. But, there is not a FIPS enabled ECR VPC endpoint available. 

Error:
```
[   68.459399] cloud-init[1077]: E1001 16:45:44.492873    1141 remote_image.go:135] PullImage "013241004608.dkr.ecr-fips.us-gov-west-1.amazonaws.com/eks/pause:3.5" from image service failed: rpc error: code = DeadlineExceeded desc = failed to pull and unpack image "013241004608.dkr.ecr-fips.us-gov-west-1.amazonaws.com/eks/pause:3.5": failed to resolve reference "013241004608.dkr.ecr-fips.us-gov-west-1.amazonaws.com/eks/pause:3.5": failed to do request: Head "https://013241004608.dkr.ecr-fips.us-gov-west-1.amazonaws.com/v2/eks/pause/manifests/3.5": dial tcp 52.222.42.110:443: i/o timeout
```

**What you expected to happen**:
Instead of checking for if the FIPS endpoint resolves, check for connectivity.

https://github.com/Issacwww/amazon-eks-ami/blob/9ef1b17cc4b250496096d89fd8ec3c1b129943bd/nodeadm/internal/aws/ecr/ecr.go#L53

**How to reproduce it (as minimally and precisely as possible)**:

**Environment**:
- AWS Region: us-gov-west-
- Instance Type(s):
- Cluster Kubernetes version:
- Node Kubernetes version:
- AMI Version:

jeremymcgee73:
This issue was only happening on RHEL. You can close if this isn't applicable.

cartermckinnon:
Sounds legit to me, feel free to open a PR 👍

ndbaker1:
merged https://github.com/awslabs/amazon-eks-ami/pull/2000 which removed this code path by caching the image at build time

PR #1990, opened 2024-10-03
Added HTTP Check for FIPS endpoint

jeremymcgee73:
**Issue #, if available:**
#1984
**Description of changes:**

There is a bug when you have enabled FIPS on the image, in a region with FIPS endpoints, and have VPC endpoints enabled. The issue is that the check implemented in https://github.com/awslabs/amazon-eks-ami/pull/1524 , checks to see if the FIPS endpoint resolves. In an isolated environment, the endpoint does resolve. But, there is not a FIPS enabled ECR VPC endpoint available. I switched the check from being a DNS request, to a HTTP call.

**Testing Done**
I used packer to build an Al2023 node with FIPS enabled to test. I also used the same code in a RHEL build downstream.

### DIFF
--- a/nodeadm/internal/aws/ecr/ecr.go
+++ b/nodeadm/internal/aws/ecr/ecr.go
@@ -4,7 +4,7 @@ import (
 	"context"
 	"fmt"
 	"go.uber.org/zap"
-	"net"
+	"net/http"
 	"strings"
 
 	"github.com/aws/aws-sdk-go-v2/config"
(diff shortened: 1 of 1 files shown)


nodeadm/internal/aws/ecr/ecr.go
```diff
@@ -49,8 +49,8 @@ func GetEKSRegistry(region string) (ECRRegistry, error) {
 	}
 	if fipsInstalled && fipsEnabled {
 		fipsRegistry := getRegistry(account, "ecr-fips", region, servicesDomain)
-		addresses, err := net.LookupHost(fipsRegistry)
-		if err == nil && len(addresses) > 0 {
+		resp, err := http.Get(fipsRegistry)
```
cartermckinnon, line 52: `fipsRegistry` isn't a URL, right? It's just the registry's hostname. You would need to add a scheme for this to work

jeremymcgee73:
Yeah, that could be the case. I need to test to make sure its not throwing an error, and is always using the non FIPS endpoint.

nodeadm/internal/aws/ecr/ecr.go
jeremymcgee73: You were right, it was failing everytime. My latest commit added the scheme, and I checked on AL2023 and RHEL. I was also able to hit both paths, with a VPC endpoint and without.

cartermckinnon:
This is going to be obsolete after #2000, which removes the dependency on ECR entirely.

PR #2000, opened 2024-10-08
feat(al2023): cache and retag pause container

ndbaker1:
**Issue #, if available:**

**Description of changes:**

In the past we've added/removed pause container image caching due to various issues with image GC and regional images. here I'm bringing back image caching for AL2023 and retagging the image to remove the association with the image's registry

the new field `pause_container_image` will be necessary when building AMIs in an environment that cannot hit the configured default ecr uri (in `us-west-2`)

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

**Testing Done**

*[See this guide for recommended testing for PRs.](../doc/CONTRIBUTING.md#testing-changes) Some tests may not apply. Completing tests and providing additional validation steps are not required, but it is recommended and may reduce review time and time to merge.*

PR #1991, opened 2024-10-04
feat(nodeadm): use cached kubelet version to reduce config time

cartermckinnon:
**Issue #, if available:**

Improves #1751.

**Description of changes:**

This PR aims to reduce `nodeadm-config` time by determining the `kubelet` version using a text file created during the AMI build process, instead of executing `kubelet --version`.

Using `kubelet --version` involves EBS lazy-loading time plus the overhead of the child process, and doing this during `nodeadm-config` delays execution of the user's user data scripts and `nodeadm-run`.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #2011, opened 2024-10-16
feat(al2023): initialize critical binaries from EBS filesystem

cartermckinnon:
**Issue #, if available:**

Improves #1751.

**Description of changes:**

This PR adds a systemd "template unit"  (`ebs-initialize-bin@.service`) that is used to "initialize" various critical binaries which are stored on the root EBS filesystem (`nodeadm`, `kubelet`, `containerd`).

EBS blocks are lazily loaded, meaning the first read of one of these binaries usually involves a delay of a handful of seconds. By accessing these files as early in the boot process as possible, this delay is not felt in hot path of `nodeadm-config` and `nodeadm-run`.

### DIFF
--- a/templates/al2023/runtime/rootfs/etc/systemd/system/ebs-initialize-bin@.service
+++ b/templates/al2023/runtime/rootfs/etc/systemd/system/ebs-initialize-bin@.service
@@ -1,0 +1,10 @@
+[Unit]
+Description=Initialize critical binary (%i) from an EBS filesystem
+After=basic.target
+
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/ebs-initialize /usr/bin/%i
+
+[Install]
+WantedBy=multi-user.target
--- a/templates/al2023/provisioners/install-worker.sh
+++ b/templates/al2023/provisioners/install-worker.sh
@@ -133,6 +133,8 @@ fi
 sudo dnf install -y runc-${RUNC_VERSION}
 sudo dnf install -y containerd-${CONTAINERD_VERSION}
 
+sudo systemctl enable ebs-initialize-bin@containerd
+
 ###############################################################################
 ### Nerdctl setup #############################################################
 ###############################################################################
@@ -189,6 +191,8 @@ sudo rm ./*.sha256
 
 kubelet --version > "${WORKING_DIR}/kubelet-version.txt"
 sudo mv "${WORKING_DIR}/kubelet-version.txt" /etc/eks/kubelet-version.txt
+
+sudo systemctl enable ebs-initialize-bin@kubelet
 
 ################################################################################
 ### ECR Credential Provider Binary #############################################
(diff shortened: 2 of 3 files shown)


cartermckinnon:
Looks like we got a comparable speedup in `nodeadm-run` as we did in #1991 for `nodeadm-config`. 🥳 

I've opened #2012 to get some better numbers in the log bundle.

cartermckinnon:
Scoping down to AL2023 only

/ci
+workflow:os_distros al2023

cartermckinnon:
### 1.24

<details>
<summary><i>Expand</i></summary>

systemd-analyze
</details>

<details>
<summary><i>Expand</i></summary>

systemd-analyze
</details>

<details>
<summary><i>Expand</i></summary>

systemd-analyze
</details>

### 1.31
<details>
<summary><i>Expand</i></summary>

systemd-analyze
</details>
<details>
<summary><i>Expand</i></summary>

systemd-analyze
</details>
<details>
<summary><i>Expand</i></summary>

systemd-analyze
</details>

cartermckinnon:
`containerd` benefits most from this change. Before, it took 2.5-5 seconds to become active. After, it takes about 500ms, with much less variation in timing.

This allows `kubelet` to become active 2-4 seconds earlier after this change. I expect #2000 to shave another 1-2 seconds off this, and make the timing more consistent.

There is no noticeable benefit of `ebs-initialize@nodeadm`, because `nodeadm-config` becomes runnable at the same point in the lifecycle. I'll remove that one.

cartermckinnon:
Getting some more samples...

/ci
+workflow:os_distros al2023

templates/al2023/runtime/rootfs/etc/systemd/system/ebs-initialize-bin@.service
```diff
@@ -0,0 +1,10 @@
+[Unit]
+Description=Initialize critical binary (%i) from an EBS filesystem
+After=basic.target
+
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/ebs-initialize /usr/bin/%i
```
ndbaker1, line 7: seems like a pretty niche script 😂 thoughts on just inlining it into the service?
```suggestion
ExecStart=/usr/bin/bash -c 'dd if="/usr/bin/%i" of=/dev/null bs=1M'
```

Issacwww (approved):

templates/al2023/runtime/rootfs/etc/systemd/system/ebs-initialize-bin@.service
cartermckinnon: I wanted it to be testable/benchmarkable without the systemd dependency. The hardcoded “/usr/bin/“ part in the unit is because passing file paths as the instance name for template units is janky, and I didn’t want to move that down into the script

ndbaker1 (approved):

PR #2022, opened 2024-10-24
feat(al2023): support NVIDIA variant in isolated partitions

whoix:
**Issue #, if available:**

**Description of changes:**

Pull request to start having AL2023 GPU amis support in isolated partitions. Added some local helpers and the AL2023 repo. No functional changes and should not affect prod.

**Testing Done**

### DIFF
--- a/templates/al2023/provisioners/install-nvidia-driver.sh
+++ b/templates/al2023/provisioners/install-nvidia-driver.sh
@@ -7,6 +7,45 @@ set -o errexit
 if [ "$ENABLE_ACCELERATOR" != "nvidia" ]; then
   exit 0
 fi
+
+################################################################################
+### Add AL2023 Nvidia repository ###############################################
+################################################################################
+echo "[amzn2023-nvidia]
+name=Amazon Linux 2023 Nvidia repository
+mirrorlist=https://al2023-repos-$awsregion-de612dc2.s3.$awsregion.$awsdomain/nvidia/mirrors/$releasever/$basearch/mirror.list
+priority=20
+enabled=1
+repo_gpgcheck=0
+type=rpm
+gpgcheck=0
+gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-amazon-linux-2023" | sudo tee /etc/yum.repos.d/amzn2023-nvidia.repo
+
+# utility function for pulling rpms from an S3 bucket
+function rpm_install() {
+  local RPMS=($@)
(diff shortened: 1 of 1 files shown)


templates/al2023/provisioners/install-nvidia-driver.sh
```diff
@@ -7,6 +7,45 @@ set -o errexit
 if [ "$ENABLE_ACCELERATOR" != "nvidia" ]; then
   exit 0
 fi
+
+################################################################################
+### Add AL2023 Nvidia repository ###############################################
+################################################################################
+echo "[amzn2023-nvidia]
```
cartermckinnon, line 14: this should only be done `if is-isolated-partition`, right?

templates/al2023/provisioners/install-nvidia-driver.sh
```diff
@@ -7,22 +7,58 @@ set -o errexit
...
 ################################################################################
 ### Add repository #############################################################
 ################################################################################
 # Determine the domain based on the region
-if [[ $AWS_REGION == cn-* ]]; then
-  DOMAIN="nvidia.cn"
+if is-isolated-partition; then
+  echo "[amzn2023-nvidia]
+  name=Amazon Linux 2023 Nvidia repository
+  mirrorlist=https://al2023-repos-$awsregion-de612dc2.s3.$awsregion.$awsdomain/nvidia/mirrors/$releasever/$basearch/mirror.list
```
ndbaker1, line 42: need to present these from being substituted using single quotes?

templates/al2023/provisioners/install-nvidia-driver.sh
```diff
@@ -7,22 +7,58 @@ set -o errexit
 if [ "$ENABLE_ACCELERATOR" != "nvidia" ]; then
   exit 0
 fi
+
+# utility function for pulling rpms from an S3 bucket
+function rpm_install() {
```
ndbaker1, line 12: function not used right?

templates/al2023/provisioners/install-nvidia-driver.sh
ndbaker1: yea i think using single instead of double quotes to encase this blob should suffice, the neuron excerpt doesn't have any `$VAR` items in it that might get replaced with a shell variable like this `mirrorlist` value

templates/al2023/provisioners/install-nvidia-driver.sh
ndbaker1: we did yea, but i think its cleaner to introduce that when it becomes necessary

ndbaker1 (approved):
lgtm

Issacwww (approved):

PR #2056, opened 2024-11-15
feat(al2023): warn about bootstrap.sh usage

cartermckinnon:
**Description of changes:**

Adds more helpful logging to `/var/log/cloud-init-output.log` when AL2 user data is used on an AL2023 instance, using a fake `bootstrap.sh`.

### DIFF
--- a/templates/al2023/runtime/rootfs/etc/eks/bootstrap.sh
+++ b/templates/al2023/runtime/rootfs/etc/eks/bootstrap.sh
@@ -1,0 +1,15 @@
+#!/usr/bin/env bash
+
+echo >&2 '
+!!!!!!!!!!
+!!!!!!!!!! ERROR: bootstrap.sh has been removed from AL2023-based EKS AMIs.
+!!!!!!!!!!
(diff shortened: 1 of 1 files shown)


templates/al2023/runtime/rootfs/usr/bin/bootstrap.sh
ndbaker1, line 1: we actually want this on the same path as al2 right? i.e. `/etc/eks/bootstrap.sh`?

cartermckinnon: ah yep, forgot we put this there. makes so much sense for a script to be in `etc`, after all 😭

ndbaker1 (approved):

s7an-it:
@cartermckinnon, what does it mean, does it menn bootstrap.sh can't be used and in the terraform eks module that option should be turned off, how can I pass custom script, using the cloudinit and multipart shell + nodeconfig?

bryantbiggs:
What are you passing to the shell script?

PR #2072, opened 2024-11-22
Fix kernel tools version mismatch

hotspoons:
This PR adds the kernel-tools package to the version-locked packages in the AL2 AMI. Currently, kernel-tools can be updated to versions independent of the installed kernel version when applying other updates when we run `yum update -y` to patch the OS. This gets flagged as a "high" vulnerability by our Tenable scans, because Tenable thinks that there is a pending kernel upgrade that we need to reboot our EC2s for to apply, even though there is no pending kernel upgrade:

image

By version locking kernel tools to the kernel version, this should no longer be an issue, and we can coordinate kernel-tools upgrades with kernel upgrades by manually unlocking the packages when we are ready to apply updates.

******************************************

**Testing Done**

### DIFF
--- a/templates/al2/provisioners/install-worker.sh
+++ b/templates/al2/provisioners/install-worker.sh
@@ -56,7 +56,7 @@ sudo yum install -y \
   yum-plugin-versionlock
 
 # lock the version of the kernel and associated packages before we yum update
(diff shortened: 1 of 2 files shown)


hotspoons:
@cartermckinnon -I checked with our internal service team, and they are building the EKS AMI on top of their STIG baseline AL2 AMI, which does pull in kernel-tools. Sorry for the confusion, I will work with them to address this in their build process.

PR #2074, opened 2024-11-23
feat(al2023): ensure the sandbox image is present periodically

cartermckinnon:
**Description of changes:**

When the sandbox image goes missing, manual repairs are often necessary. This PR adds a service unit with an associated timer that will (re)import the local sandbox image tarball if necessary. The unit runs once every 5 minutes (beginning 1 minute after boot) and is a no-op if the image is present.


cartermckinnon:
We haven't found this to be a problem in practice (outside of a few weird cases), not moving forward.

Issue #2092, opened 2024-12-16
bug: /usr/bin/imds fails with printf format in user-data


vflaux:
**What happened**:
I use `/usr/bin/imds latest/user-data` to get the instance user-data and get the error ``/usr/bin/imds: line 40: printf: `D': invalid format character`` at some point.
My user-data contains an URL with the sequence `%3D` ("=") which get interpreted by the "printf" command in the imds script and then fails.

**What you expected to happen**:

No error and the full user-data returned.

**How to reproduce it (as minimally and precisely as possible)**:

1. Create an instance with user-data containing "%D"  (I use karpenter)
2. On the instance, run `imds latest/user-data`
3. See error

**Environment**:
- AWS Region: N/A
- Instance Type(s): N/A
- Cluster Kubernetes version: N/A
- Node Kubernetes version: 1.30
- AMI Version: Amazon Linux 2023.6.20241111

I do not think that calling `printf "$(cat $OUTPUT_FILE)\n"` with no additional arguments is needed. A simple `cat $OUTPUT_FILE; echo` should be sufficient.
https://github.com/awslabs/amazon-eks-ami/blob/e297621e85ad0d6545a8d0d82db688f123976a32/templates/shared/runtime/bin/imds#L40

This is not blocking as I can read the user-data from the `cloud-init` command or file directly.

cartermckinnon:
@vflaux thanks for the report! Should be fixed 👍

PR #2093, opened 2024-12-16
bug: remove printf from imds helper

ndbaker1:
**Issue #, if available:**

fix https://github.com/awslabs/amazon-eks-ami/issues/2092

**Description of changes:**

the imds helper uses `printf` when printing output, which can be confused by format specifiers inside of the returned data (like userdata)

**Testing Done**

```shell
$ echo foo > foo
$ printf "$(cat foo)\n"
foo
$ cat foo
foo
```

### DIFF
--- a/templates/shared/runtime/bin/imds
+++ b/templates/shared/runtime/bin/imds
@@ -37,7 +37,7 @@ function imdscurl() {
     cat >&2 $OUTPUT_FILE
     return $CODE
   fi
-  printf "$(cat $OUTPUT_FILE)\n"
+  cat $OUTPUT_FILE
   rm $OUTPUT_FILE
 }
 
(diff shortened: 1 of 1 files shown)


templates/shared/runtime/bin/imds
```diff
@@ -37,7 +37,7 @@ function imdscurl() {
     cat >&2 $OUTPUT_FILE
     return $CODE
   fi
-  printf "$(cat $OUTPUT_FILE)\n"
+  cat $OUTPUT_FILE
```
cartermckinnon, line 40: Maintain the trailing newline:
```suggestion
  cat $OUTPUT_FILE
  printf “\n”
```

ndbaker1: oh hah i think the `\n` is already omit when shell subbing `$(cat ..)`
adding this line was breaking the tests

cartermckinnon (approved):

Issue #803, opened 2021-11-11
Add retry logic on `aws eks wait cluster-active` to `/etc/eks/bootstrap.sh`


cucxabong:
**What would you like to be added**:
I get error `Waiter ClusterActive failed: Too Many Requests` when launch cluster with 400 Node, only 256 Node registered with the master and no more

**Why is this needed**:
Adding logic to retry on `aws eks wait cluster-active` the same way we did with `aws eks describe-cluster`

suket22:
As a possible workaround, you could pass the API Server endpoint and the Cluster certificate data to the bootstrap script and avoid the EKS API calls

l1n:
Any updates on this? Would be pretty useful.

cartermckinnon:
The best solution here is to avoid the `eks:DescribeCluster` call entirely by passing the cluster CA and endpoint to the bootstrap script.

PR #2096, opened 2024-12-18
fix: check return code of aws eks wait call in bootstrap script

msarfaty:
**Issue #, if available:** None Active (closed: https://github.com/awslabs/amazon-eks-ami/issues/803)

**Description of changes:**

this checks the return code for `aws eks wait cluster-active` in the bootstrap script.

we've noticed that if a lot of nodes are booting at the same time, this can get backed off via AWS:
```
Waiter ClusterActive failed: An error occurred (TooManyRequestsException): Too Many Requests
Exited with error on line 286
```
since `errexit` is set, this automatically fails the bootstrap script. With this change, I'm intending for this loop to check _both_ returns from `wait` and `describe cluster` to avoid throttling/errors from both.

**Testing Done**

- I created an AMI via packer in our account via `make k8s=1.29`
- I used that AMI for a test node and the nodes joined successfully

testing a failure scenario would take a little more time and the code change is minimal, but I can try to replicate it if desired (probably easiest by creating an AMI where I set `$rc` to `286` manually on the first attempt between the wait / describe calls)

### DIFF
--- a/templates/al2/runtime/bootstrap.sh
+++ b/templates/al2/runtime/bootstrap.sh
@@ -362,9 +362,9 @@ if [[ -z "${B64_CLUSTER_CA}" ]] || [[ -z "${APISERVER_ENDPOINT}" ]]; then
 
     aws eks wait cluster-active \
       --region=${AWS_DEFAULT_REGION} \
-      --name=${CLUSTER_NAME}
+      --name=${CLUSTER_NAME} || rc=$?
 
-    aws eks describe-cluster \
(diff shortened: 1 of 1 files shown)


cartermckinnon:
Change looks reasonable to me. Worth noting that you can avoid this category of issue if you pass your cluster details to `bootstrap.sh`, instead of relying on the `eks:DescribeCluster` fallback 👍 

/ci

cartermckinnon (approved):

PR #2099, opened 2024-12-26
Bump pause image to 3.10

dims:
We are on an old pause image, let's switch to newer ones.

example https://github.com/containerd/containerd/pull/10258

### DIFF
--- a/templates/al2/variables-default.json
+++ b/templates/al2/variables-default.json
@@ -22,7 +22,7 @@
     "kms_key_id": "",
     "iam_instance_profile": "",
     "launch_block_device_mappings_volume_size": "4",
-    "pause_container_version": "3.5",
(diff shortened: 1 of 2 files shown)


cartermckinnon:
All the changes since 3.5 are for Windows, AFAICT. The `3.10` tag seems to be available in our standard ECR repos. Update seems fine, but is there something motivating this that I'm missing @dims ?

dims:
@cartermckinnon nothing specific other than keeping things up-to-date.

cartermckinnon:
Roger that! I noticed that the template var for AL2 isn't actually used, since we removed caching of this container image; so no concern there. Making sure things are good on AL2023...

/ci
+workflow:os_distros al2023

cartermckinnon (approved):

PR #2100, opened 2024-12-26
Enable CDI in containerd configuration

dims:
`enable_cdi = true` is set to true in containerd 2.0:
https://github.com/containerd/containerd/commit/c8e8a093ce3dd9231173042e9828a679972f5d6b

Some DRA drivers are using CDI:
https://github.com/search?q=repo%3ANVIDIA%2Fk8s-dra-driver%20cdi&type=code

A bunch of CI jobs in upstream are now enabling CDI:
https://cs.k8s.io/?q=enable_cdi&i=nope&files=&excludeFiles=&repos=

So let's switch it on in our images as well please.

### DIFF
--- a/nodeadm/internal/api/merge_test.go
+++ b/nodeadm/internal/api/merge_test.go
@@ -119,6 +119,7 @@ discard_unpacked_layers = true
 
 [plugins."io.containerd.grpc.v1.cri"]
 sandbox_image = "{{.SandboxImage}}"
+enable_cdi = true
 
 [plugins."io.containerd.grpc.v1.cri".registry]
 config_path = "/etc/containerd/certs.d:/etc/docker/certs.d"
@@ -195,6 +196,7 @@ discard_unpacked_layers = false
 
 [plugins."io.containerd.grpc.v1.cri"]
 sandbox_image = "{{.SandboxImage}}"
+enable_cdi = true
 
 [plugins."io.containerd.grpc.v1.cri".registry]
 config_path = "/etc/containerd/certs.d:/etc/docker/certs.d"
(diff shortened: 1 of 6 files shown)


dims:
@cartermckinnon does this look ok to you?

cartermckinnon:
@dims sorry, catching up on this one. I don't know in detail what this option enables, and I'm hesitant to switch it on in 1.7.x until the early adopters bake it a bit. IIUC this won't be functional for DRA on EKS anyway until that feature is default on (in 1.33, TBD)?

dims:
@cartermckinnon you can read about it here - https://github.com/cncf-tags/container-device-interface

yes 1.33 is targetted for DRA/Beta to be enabled by default.

cartermckinnon:
Ah thanks! I'm mostly concerned about what is toggled in the containerd internals, e.g. what if the `cdi_spec_dirs` don't exist/can't be created or can't be watched? https://github.com/containerd/containerd/blob/2ab62acd6379cda100b5af7e3e0b65877359c8b1/internal/cri/server/service_linux.go#L107

It looks like those errors are not currently surfaced here, but that's an impl detail: https://github.com/cncf-tags/container-device-interface/blob/4149a8a524dfe96686d97789833cc78f61eb6631/pkg/cdi/cache.go#L102

This part looks safe, if the CRI spec doesn't have CDI devices or annotations, it's a no-op: https://github.com/containerd/containerd/blob/2ab62acd6379cda100b5af7e3e0b65877359c8b1/internal/cri/server/container_create_linux.go#L124

I'm on board with enabling it for 1.32 for now, since we'll need it in 1.33 for DRA

dims:
> This part looks safe, if the CRI spec doesn't have CDI devices or annotations, it's a no-op: 

yep!

@austinvazquez sounds about right?

austinvazquez (approved):
@dims, yep this looks kosher to me. Double checked sandbox server is having the same implementation as CRI server here as well.

dims:
thanks @austinvazquez 
@cartermckinnon we can merge this when you feel ready!

dims:
@cartermckinnon do we want to do this now?

cartermckinnon:
Continuing this in #2173...

PR #2112, opened 2025-01-08
fix: increase IMDS retries

cartermckinnon:
**Description of changes:**

This increases the number of retry attempts for IMDS requests from 15 to 60. We use exponential backoff with a max delay of 1 second, so this results in about 1 minute of retries. We've seen rare failures where the network is not up until the previous retries (15) were exhausted.


ndbaker1 (approved):
thanks!

ndbaker1:
any value in exercising a longer downtime in the e2e test?
https://github.com/awslabs/amazon-eks-ami/blob/b5c6c83b526c1ce664a782a5d9473a00908f5bfe/nodeadm/test/e2e/cases/imds-timeouts/run.sh#L12-L23

PR #2135, opened 2025-02-04
refactor(nodeadm): rework containerd template mixin interface

ndbaker1:
**Issue #, if available:**

**Description of changes:**

In order to recognize when the running instance has GPU devices, we currently use a hard coded list of instance type families. to keep this list updated or remove it we can do one of the following:
1. use workflows to generate PRs for the changes. this isn't so bad but introduces a build time dependency
2. make api calls to `ec2.DescribeInstanceTypes` to get whether the current instance has a GPU device
3. read local data on the instance to determine if there is an nvidia device attached

this PR implements (3) from above by looking for the `nvidia` vendor name in `/proc/bus/pcie/device`. the original logic is left in and PCIe devices are used as a fallback.

**Testing Done**


nodeadm/internal/containerd/config.go
```diff
@@ -30,11 +30,15 @@ type containerdTemplateVars struct {
 	RuntimeBinaryName string
 }
 
-func writeContainerdConfig(cfg *api.NodeConfig) error {
-	if err := writeBaseRuntimeSpec(cfg); err != nil {
-		return err
-	}
+type configurator interface {
```
cartermckinnon, line 33: Name is a little clunky, I liked `mixin` in the prev impl 😉 or maybe `modifier`?

nodeadm/internal/containerd/daemon.go
```diff
@@ -20,6 +20,9 @@ func NewContainerdDaemon(daemonManager daemon.DaemonManager) daemon.Daemon {
 }
 
 func (cd *containerd) Configure(c *api.NodeConfig) error {
+	if err := writeBaseRuntimeSpec(c); err != nil {
+		return err
+	}
```
cartermckinnon, line 25: What's this about? Doesn't seem related to the other changes

nodeadm/internal/containerd/config.go
```diff
@@ -59,12 +63,25 @@ func writeContainerdConfig(cfg *api.NodeConfig) error {
...
+		if configurator.Matches(cfg) {
+			configurators = append(configurators, configurator)
+		}
+	}
+	switch len(configurators) {
+	case 0:
+	case 1:
+		configurators[0].Transform(&configVars)
+	default:
+		zap.L().Error("Skipping configuration because more than one configurator was matched", zap.Any("configurators", configurators))
```
cartermckinnon, line 84: I don't think this should be an error; `configurator`-s should be composable or there's not much benefit to the abstraction.

nodeadm/internal/containerd/nvidia.go
```diff
@@ -0,0 +1,48 @@
...
+	return slices.Contains(nvidiaInstanceFamilies, family)
+}
+
+func matchesPCIeVendor() bool {
+	devices, err := os.ReadFile(pcieDevicesPath)
+	if err != nil {
+		zap.L().Error("Failed to read PCIe devices", zap.Error(err))
+		return false
+	}
+	return strings.Contains(string(devices), "nvidia")
```
cartermckinnon, line 47: Can you add a code comment with an example of what `devices` should look like?

nodeadm/internal/containerd/nvidia_test.go
```diff
@@ -0,0 +1,48 @@
...
+		configurator.Transform(&template)
+		assert.Equal(t, "nvidia", template.RuntimeName)
+		assert.Equal(t, "/usr/bin/nvidia-container-runtime", template.RuntimeBinaryName)
+	})
+
+	t.Run("IsNvidiaUsingPCIe", func(t *testing.T) {
+		originalPCIeDevicesPath := pcieDevicesPath
+		pcieDevicesPath = filepath.Join(t.TempDir(), "pcie-devices")
+		defer func() { pcieDevicesPath = originalPCIeDevicesPath }()
+		os.WriteFile(pcieDevicesPath, []byte("nvidia"), 0777)
```
cartermckinnon, line 27: Could `nvidiaConfigurator` have a field for `pcieDevicesPath` that could be overridden? setting the package `var` to a value specific to this test case is a little unsavory

cartermckinnon (commented):
Bit too much refactoring mixed in with the logical changes, IMO; I think this PR can be split

nodeadm/internal/containerd/config.go
ndbaker1: I agree this looks funky, mainly was a byproduct of trying to continue using the abstraction without a good reason. It doesnt work like a mixin, it just populates a set of parameters.

I'd ideally just break on the first match and swallow the thought that another item *couldve* been matched but it won't because we validate that ahead if time

nodeadm/internal/containerd/nvidia_test.go
ndbaker1: Good call :)

nodeadm/internal/containerd/config.go
ndbaker1: ill try going with `modifier`

ndbaker1:
https://github.com/awslabs/amazon-eks-ami/pull/2146 replaces this PR for PCIe detection functionality, while this will be turned into a refactor for the `mixin` interface,

ndbaker1:
superceded by https://github.com/awslabs/amazon-eks-ami/pull/2227

PR #2146, opened 2025-02-08
[WIP] feat(nodeadm): PCIe detection for nvidia GPU instances

ndbaker1:
**Issue #, if available:**

**Description of changes:**

In order to decide when to configure containerd using the `nvidia-container-runtime`, we currently rely on checking the ec2 instanceType family of the node. Because of this, we have a hard coded list of instance type families. 

1. use ci workflows to generate PRs for the update. this isn't so bad but introduces a build time dependency
2. make api calls to `ec2.DescribeInstanceTypes` to get whether the current instance has a GPU device. this introduces more api load and can suffer from thundering herd issues.
3. read data on the instance to determine if there is an nvidia PCIe device attached
4. rely on the components baked into the AMI.

this PR implements (3) from above by looking for the `nvidia` driver name in `/proc/bus/pcie/device`. see https://elixir.bootlin.com/linux/v6.12.19/source/drivers/pci/proc.c#L367-L404 for reference. the original logic is left in and PCIe device info is used as a fallback.

**Testing Done**

### DIFF
--- a/nodeadm/internal/containerd/runtime_config.go
+++ b/nodeadm/internal/containerd/runtime_config.go
@@ -1,6 +1,7 @@
 package containerd
 
 import (
+	"os"
 	"slices"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


ndbaker1:
We need to reconsider the ordering of services for this to work.

because the `nvidia-kmod-load.service` has a dependency chain of `network-online.target` -> `cloud-init.service`, the containerd config generated in `nodeadm-config.service` will not be able to take into account the `nvidia` pcie devices driver as they haven't yet been loaded.

ndbaker1:
rather than taking a dependency on the kmods being loaded, ill change the logic to call `lspci`. this'll be easier to e2e test than using the `c` headers to call the same libs that `lspci` uses to parse vendor names from the device.

ndbaker1:
closing in favor of https://github.com/awslabs/amazon-eks-ami/pull/2227

PR #2185, opened 2025-03-14
[WIP] feat: set OOMScoreAdjust=-999 for system.slice

### LINKED ISSUE #2175: bug: Adjust the OOMScore of the critical system services lower than Kubernetes guaranteed pod OOMScore
Adjust the OOMScore of critical systemd services (in general, the units in the system.slice are the critical services) so that the their OOMScore is less than -997. This can be configured by setting `OOMScoreAdjust` to -998 or -999 in the systemd units.

**Why is this needed**:
[In kubernetes, containers with a guaranteed quality of service are given an oom_score_adj of -997](https://kubernetes.io/docs/concepts/scheduling-eviction/node-pressure-eviction/#node-out-of-memory-behavior). This is problematic because this is a lower OOMScore than the systemd services that are required to provide service to a node. The linux oomkiller kills processes in the affected cgroup with the highest oomscore + memory usage. When linux memory pressure is due to a guaranteed pod, the linux oomkiller can not kill the guaranteed pod first due to the low oomscore, and so it must kill other processes to free memory. This may be unrelated (burstable) user workloads, or if none exist, then the node's systemd services.

In practice, we have seen this triggered by straightforward user behavior: force deleting a pod. A user that was debugging a "stuck" or busy container decided to force-delete a guaranteed pod that was part of a statefulset, without overriding the grace period and the pod had a long termination grace period. This resulted in a replacement being scheduled to the same node, and kubelet started it while the old container was still terminating, creating memory pressure from the duplicated container. When this happened the oomkiller decided to kill critical services like ssm, sshd, postfix, etc. The node in kubernetes showed it's ready status as "Unknown" and never recovered. We've since reproduced this behavior by running a guaranteed with "stress" and a long prestop hook, and force-deleting it to cause kubelet to race and have both the terminating and replacement pods running on the node.

As a cluster admin, I believe it should not be possible for a user to crash a node and make it unready by deleting or scheduling a non-privileged guaranteed pod to a node. So in order to protect against this, AL should harden the OOMScore of critical systemd services.

Note: this behavior occurred in AL2 and we will be testing AL2023 soon, but don't expect it to behave differently.

cartermckinnon:
**Issue #, if available:**

Resolves #2175.

**Description of changes:**

Adjusts the OOM score of processes in `system.slice` (`sshd`, `amazon-ssm-agent`, etc.) below that of Kubernetes `Guaranteed` QoS pods. This ensures that system services are not killed before pods.


cartermckinnon:
I'm no longer at EKS, so I'm going to close this PR. If there's interest from the current maintainers -- it worked fine in my testing. The only consideration is some units in the `system.slice` (like `sshd`) have an `oom_score_adj` of `-1000`, which you may not want to override.

PR #2195, opened 2025-03-26
feat(test): build nodeadm on host

cartermckinnon:
**Description of changes:**

This speeds up repeated runs of `make test-e2e` by using the `nodeadm` executable built by the `build` target (on the host) instead of in a multi-stage build of the test infra container image. Lots of time is wasted re-downloading go dependencies with that approach.


ndbaker1 (approved):
cool, seems like it saved ~1 min on average?

cartermckinnon:
YMMV but especially on slower connections it'll help.

PR #2209, opened 2025-04-09
fix(build): correct build variable order of precedence

mselim00:
**Issue #, if available:**

**Description of changes:**

At the moment, variables in the "defaults" are actually forced at a higher precedence than any other variable files. This is due to the way packer handles precedence. Specifically, variables are resolved in descending order of precedence by command-line flags, variable files, environment variables, and variable defaults [ref](https://developer.hashicorp.com/packer/tutorials/aws-get-started/aws-get-started-variables#build-image-with-variables). Packer uses a 'first defined, first used' heuristic for variables in general [ref](https://developer.hashicorp.com/packer/tutorials/aws-get-started/aws-get-started-variables#build-image-with-variables), so by specifying the default variable file first, we were disabling any other files from overriding values set there. 

**Testing Done**


mselim00:
Nevermind - the last time a variable is set in a variable file seems to take precedence over earlier files

PR #2136, opened 2025-02-04
build(nvidia): install nvidia-imex driver

ndbaker1:
**Issue #, if available:**

**Description of changes:**

Installs the `nvidia-imex` package that support imex channels and domains.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

**Testing Done**

built an ami with 
```
make k8s=1.31 os_distro=al2023 enable_accelerator=nvidia
```

*[See this guide for recommended testing for PRs.](../doc/CONTRIBUTING.md#testing-changes) Some tests may not apply. Completing tests and providing additional validation steps are not required, but it is recommended and may reduce review time and time to merge.*

PR #2239, opened 2025-05-03
feat(al2023/nvidia): install nvlsm

mselim00:
**Issue #, if available:**

**Description of changes:**

Nvidia systems using fourth generation NVSwitches, like DGX B200, NVIDIA HGX B200 8-GPU, and NVIDIA HGX B100 8-GPU, have an additional dependency on NVLink Subnet Manager (NVLSM). [ref](https://docs.nvidia.com/datacenter/tesla/fabric-manager-user-guide/index.html#systems-using-fourth-generation-nvswitches) 

**Testing Done**

This change is adding documented dependencies - I did spin up `p4d.24xlarge` instances to verify backwards compatibility though.

### DIFF
--- a/templates/al2023/provisioners/install-nvidia-driver.sh
+++ b/templates/al2023/provisioners/install-nvidia-driver.sh
@@ -119,6 +119,16 @@ archive-open-kmods
 archive-proprietary-kmod
 
 ################################################################################
+### Install NVLSM ##############################################################
+################################################################################
+# https://docs.nvidia.com/datacenter/tesla/fabric-manager-user-guide/index.html#systems-using-fourth-generation-nvswitches
+
+echo "ib_umad" | sudo tee -a /etc/modules-load.d/ib-umad.conf
+sudo dnf -y install libibumad
+sudo dnf -y install infiniband-diags
+sudo dnf -y install nvlsm
+
+################################################################################
 ### Prepare for nvidia init ####################################################
 ################################################################################
 
--- a/templates/al2023/provisioners/install-nvidia-driver.sh
+++ b/templates/al2023/provisioners/install-nvidia-driver.sh
@@ -123,11 +123,14 @@ archive-proprietary-kmod
 ################################################################################
 # https://docs.nvidia.com/datacenter/tesla/fabric-manager-user-guide/index.html#systems-using-fourth-generation-nvswitches
 
-echo "ib_umad" | sudo tee -a /etc/modules-load.d/ib-umad.conf
-sudo dnf -y install \
-  libibumad \
-  infiniband-diags \
-  nvlsm
+# TODO: install unconditionally once availability is guaranteed
+if ! is-isolated-partition; then
+  echo "ib_umad" | sudo tee -a /etc/modules-load.d/ib-umad.conf
+  sudo dnf -y install \
+    libibumad \
+    infiniband-diags \
+    nvlsm
+fi
 
 ################################################################################
 ### Prepare for nvidia init ####################################################
--- a/templates/al2023/provisioners/install-nvidia-driver.sh
+++ b/templates/al2023/provisioners/install-nvidia-driver.sh
@@ -124,9 +124,10 @@ archive-proprietary-kmod
 # https://docs.nvidia.com/datacenter/tesla/fabric-manager-user-guide/index.html#systems-using-fourth-generation-nvswitches
 
 echo "ib_umad" | sudo tee -a /etc/modules-load.d/ib-umad.conf
-sudo dnf -y install libibumad
-sudo dnf -y install infiniband-diags
-sudo dnf -y install nvlsm
+sudo dnf -y install \
+  libibumad \
+  infiniband-diags \
+  nvlsm
 
 ################################################################################
 ### Prepare for nvidia init ####################################################


templates/al2023/provisioners/install-nvidia-driver.sh
```diff
@@ -118,6 +118,16 @@ function archive-proprietary-kmod() {
...
 archive-proprietary-kmod
 
+################################################################################
+### Install NVLSM ##############################################################
+################################################################################
+# https://docs.nvidia.com/datacenter/tesla/fabric-manager-user-guide/index.html#systems-using-fourth-generation-nvswitches
+
+echo "ib_umad" | sudo tee -a /etc/modules-load.d/ib-umad.conf
+sudo dnf -y install libibumad
+sudo dnf -y install infiniband-diags
```
bryantbiggs, line 128: I don't believe this should be added (given EFA). Are there plans to add support for IMEX though?

mselim00: Thanks for the catch! I just did some poking around the efa-installer, I see it does indeed include these through the installation of rdma-core. It does also check to see if the packages have already been installed, and skips them if so. Since these packages are required for nvlsm, I think an unconditional install here is probably still reasonable so that EFA installation isn't required for support

For IMEX - recent AMIs already include it, we just haven't started/enabled the service ref: #2136

mselim00:
/ci
+workflow:k8s_versions 1.32
+test --instance-types=g5.8xlarge

mselim00:
Looking at the entrypoint for the Fabric Manager service, it only launches NVLSM if it detects InfiniBand devices when searching under `/sys/class/Infiniband`, and checking for the keyword `SW_MNG` in it's device/vpd file. 

snippet ref from `/usr/bin/nvidia-fabricmanager-start.sh` on version `570.133.20-1` of nvidia-fm 
```
    134 # Loop through each Infiniband device directory
    135 for dir in /sys/class/infiniband/*/device; do
    136     # Define the path to the VPD file
    137     vpd_file="$dir/vpd"
    138
    139     # Check if the VPD file exists
    140     if [ -f "$vpd_file" ]; then
    141         # Search for 'SW_MNG' in the VPD file
    142         if grep -q "SW_MNG" "$vpd_file"; then
    143             echo "Detected NVL5+ system"
    144             # Extract the Infiniband device name using parameter expansion
    145             device_name="${dir%/device}"  # Removes '/device' from the end of $dir
    146             device_name="${device_name##*/}"  # Extracts the part after the last '/'
    147             echo "Checking $device_name..."
    148             launch_ibstat $device_name
    149             if [ $? -ne 0 ]; then
    150                 remove_cruft_files
    151                 exit 1
    152             fi
    153             capture_portguids
    154             remove_cruft_files
    155             post_nvl5=true
    156             # if port guid was captured, return since we are only grabbing first GUID
    157             if [ "$capture_flag" = "true" ]; then
    158                 echo "Using device $device_name, port $prefix_port_guid"
    159                 break
    160             fi
    161         fi
    162     fi
    163 done
    164
    165 if [ "$post_nvl5" = "true" ]; then
    166     check_required_module
    167     launch_nvlsm_proc
    168     # SM takes few mins to start the GRPC service, hence lets wait before starting FM
    169     sleep 5
    170     launch_fm_proc
    171 else
    172     echo "Detected Pre-NVL5 system"
    173     launch_fm_proc
    174 fi
```

So this should be generally safe to do, we should be able to rely on the fabric manager to handle the determination for whether or not nvlsm is needed.

templates/al2023/provisioners/install-nvidia-driver.sh
```diff
@@ -118,6 +118,16 @@ function archive-proprietary-kmod() {
 archive-open-kmods
 archive-proprietary-kmod
 
+################################################################################
+### Install NVLSM ##############################################################
+################################################################################
+# https://docs.nvidia.com/datacenter/tesla/fabric-manager-user-guide/index.html#systems-using-fourth-generation-nvswitches
+
```
ndbaker1, line 125: we should probably exclude isolated regions since we don't know whether nvlsm will be present in AL repos

ndbaker1 (commented):
nice deep dive, my current understanding from the [nvidia docs](https://docs.nvidia.com/datacenter/tesla/fabric-manager-user-guide/index.html#systems-using-fourth-generation-nvswitches) is that they've moved from directly connecting NVSwitch bridges to using higher level CX7 bridges that are exposed/unified via the the [OFI stack](https://github.com/ofiwg/libfabric). I'm assuming that's why the CX7's are surfaced as Infiniband devices, since that's where the most common RDMA apis are globbed under.

templates/al2023/provisioners/install-nvidia-driver.sh
```diff
@@ -118,6 +118,16 @@ function archive-proprietary-kmod() {
...
+echo "ib_umad" | sudo tee -a /etc/modules-load.d/ib-umad.conf
+sudo dnf -y install libibumad
+sudo dnf -y install infiniband-diags
+sudo dnf -y install nvlsm
```
cartermckinnon, line 129:
```suggestion
sudo dnf -y install \
  libibumad \
  infiniband-diags \
  nvlsm
```

templates/al2023/provisioners/install-nvidia-driver.sh
mselim00: Good catch! I excluded the kernel module and infiniband packages as well to avoid unnecessary bloat where they're not needed

raiderss:
The performance optimizations here are quite clever. Nice engineering work!

cartermckinnon:
Giving this one more go before we merge...

/ci
+workflow:k8s_versions 1.32
+test --instance-types=g5.8xlarge

cartermckinnon (approved):

templates/al2023/provisioners/install-nvidia-driver.sh
```diff
@@ -118,6 +118,20 @@ function archive-proprietary-kmod() {
...
 archive-proprietary-kmod
 
+################################################################################
+### Install NVLSM ##############################################################
+################################################################################
+# https://docs.nvidia.com/datacenter/tesla/fabric-manager-user-guide/index.html#systems-using-fourth-generation-nvswitches
+
+# TODO: install unconditionally once availability is guaranteed
+if ! is-isolated-partition; then
+  echo "ib_umad" | sudo tee -a /etc/modules-load.d/ib-umad.conf
```
ndbaker1, line 128: asking to make double sure the need here since `rdma-core` includes a [module conf that will load `ib_umad`](https://github.com/linux-rdma/rdma-core/blob/master/kernel-boot/modules/infiniband.conf#L6).

cartermckinnon: chatted offline, no harm in keeping it explicit.

ndbaker1 (approved):

PR #2244, opened 2025-05-09
feat(ci): add a workflow to verify al2 ecr match list

mselim00:
**Issue #, if available:**

**Description of changes:**
Add a manual CI to ensure no region is missing from the ECR account match function for AL2

### DIFF
--- a/.github/actions/ci/build/action.yaml
+++ b/.github/actions/ci/build/action.yaml
@@ -27,6 +27,19 @@ runs:
     - id: build
       shell: bash
       run: |
+        exit_code=0
+        ECR_SCRIPT_PATH=templates/al2/runtime/get-ecr-uri.sh
+        for region in $(aws account list-regions| jq '.Regions[].RegionName' -r); do
+            if ! grep -q $region $ECR_SCRIPT_PATH; then
+            echo "missing ECR account for $region in $ECR_SCRIPT_PATH"
+            exit_code=1
+            fi
+        done
+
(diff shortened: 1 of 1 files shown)


.github/actions/ci/build/action.yaml
```diff
@@ -27,6 +27,19 @@ runs:
...
+        for region in $(aws account list-regions| jq '.Regions[].RegionName' -r); do
+            if ! grep -q $region $ECR_SCRIPT_PATH; then
+            echo "missing ECR account for $region in $ECR_SCRIPT_PATH"
+            exit_code=1
+            fi
+        done
+
+        if (($exit_code)); then
+            exit $exit_code;
+        fi
```
cartermckinnon, line 41: less error prone than bash arithmetic:
```suggestion
        ECR_MISSING_REGION="false"
        ECR_SCRIPT_PATH=templates/al2/runtime/get-ecr-uri.sh
        for REGION in $(aws account list-regions | jq -r '.Regions[].RegionName'); do
          if ! grep -q "${REGION}" "${ECR_SCRIPT_PATH}"; then
            echo >&2 "ERROR: missing account for ${REGION} in ${ECR_SCRIPT_PATH}"
            ECR_MISSING_REGION="true"
          fi
        done
        if [ "$ECR_MISSING_REGION" = "true" ]; then
          exit 1
        fi
```

cartermckinnon (approved):

PR #2256, opened 2025-05-15
fix(log-collector,io-throttling): only exclude procs with 0 cs of delay

### LINKED ISSUE #2255: IO Throttling collector misses values that end with the zero digit
https://github.com/awslabs/amazon-eks-ami/blob/77a969c3788caa6562bda1ca30321f9d5af798a4/log-collector-script/linux/eks-log-collector.sh#L867

The `grep -v 0$` there throws out any results that have values ending with zero. (Rather than any values of zero.) It should be `grep -v " 0$"`

scole66:
**Issue #, if available:**

Fixes #2255.

**Description of changes:**

In `get_io_throttled_processes`, the generated output is filtered to remove those process that have no throttling. It used a pattern that matched any throttling value the ended with the digit zero, rather than a match of the value zero.

This change repairs that. A process with `1000000` centiseconds of delay will now show up in the output.

**Testing Done**


mselim00 (approved):
Thanks for the fix!

PR #2258, opened 2025-05-16
fix(al2023): remove duplicate entries in nvidia-open-supported-devices files

HusainZafar:
**Issue #, if available:**

**Description of changes:**
Remove duplicate entries in nvidia-open-supported-devices files

**Testing Done**

### DIFF
--- a/templates/al2023/runtime/gpu/nvidia-open-supported-devices-555.txt
+++ b/templates/al2023/runtime/gpu/nvidia-open-supported-devices-555.txt
@@ -23,11 +23,9 @@
 0x1E07 NVIDIA GeForce RTX 2080 Ti
 0x1E30 Quadro RTX 6000
 0x1E30 Quadro RTX 8000
-0x1E30 Quadro RTX 8000
-0x1E30 Quadro RTX 8000
 0x1E36 Quadro RTX 6000
-0x1E78 Quadro RTX 8000
 0x1E78 Quadro RTX 6000
+0x1E78 Quadro RTX 8000
 0x1E81 NVIDIA GeForce RTX 2080 SUPER
 0x1E82 NVIDIA GeForce RTX 2080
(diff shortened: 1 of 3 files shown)


templates/al2023/runtime/gpu/nvidia-open-supported-devices-555.txt
```diff
@@ -339,203 +83,63 @@
...
-0x1FF0 NVIDIA T1000 8GB
-0x1FF0 NVIDIA T1000 8GB
-0x1FF2 NVIDIA T400 4GB
 0x1FF2 NVIDIA T400 4GB
 0x1FF2 NVIDIA T400E
 0x1FF2 NVIDIA T400 4GB
-0x1FF2 NVIDIA T400 4GB
 0x1FF2 NVIDIA T400E
 0x1FF2 NVIDIA T400 4GB
 0x1FF2 NVIDIA T400E
```
ndbaker1, line 120: might want to make the filter `sort | uniq`?

another thing to note is the device IDs are the same across these variations, but only device ID matters at the moment. any cases where it might?
https://github.com/awslabs/amazon-eks-ami/blob/ec2d96ef4e6b0ac9c3c865c2ccc68ac955bb6dee/templates/al2023/runtime/gpu/nvidia-kmod-load.sh#L33-L38

HusainZafar: > might want to make the filter `sort | uniq`?

Updated to `sort | uniq`

> another thing to note is the device IDs are the same across these variations, but only device ID matters at the moment

Should we address that in a different PR and just do the unique part in this one? `NVIDIA T400 4GB` and `NVIDIA T400` and other variations could be merged as part of that change

cartermckinnon (approved):

PR #2265, opened 2025-05-23
Add GRID driver support

HusainZafar:
**Issue #, if available:**

**Description of changes:**

**Testing Done**

### DIFF
--- a/templates/al2023/provisioners/install-nvidia-driver.sh
+++ b/templates/al2023/provisioners/install-nvidia-driver.sh
@@ -79,6 +79,7 @@ sudo dnf -y install \
   kernel-modules-extra-common-$(uname -r)
 
 function archive-open-kmods() {
+  echo "Archiving open kmods"
   if is-isolated-partition; then
     sudo dnf -y install "kmod-nvidia-open-dkms-${NVIDIA_DRIVER_MAJOR_VERSION}.*"
   else
@@ -106,6 +107,7 @@ function archive-open-kmods() {
 }
 
 function archive-proprietary-kmod() {
+  echo "Archiving proprietary kmods"
   if is-isolated-partition; then
     sudo dnf -y install "kmod-nvidia-latest-dkms-${NVIDIA_DRIVER_MAJOR_VERSION}.*"
   else
@@ -115,8 +117,37 @@ function archive-proprietary-kmod() {
   sudo kmod-util remove nvidia
 }
 
+function archive-grid-kmod() {
+  echo "Archiving GRID kmods"
+  local workdir=NVIDIA-Linux
+  sudo mkdir -p ${workdir}
+  pushd ${workdir}
+
+  sudo aws s3 cp s3://ec2-linux-nvidia-drivers/grid-18.1/NVIDIA-Linux-x86_64-570.133.20-grid-aws.run .
+  local package=$(ls NVIDIA-Linux-x86_64*.run)
+  sudo sh ${package} --extract-only
+  sudo rm -rf ${package}
+
+  cd $(ls)
+
+  sudo sed -i 's/PACKAGE_NAME="nvidia"/PACKAGE_NAME="nvidia-grid"/g' kernel/dkms.conf
+
+  sudo dnf -y install dkms
+  sudo ./nvidia-installer --kernel-module-type proprietary --dkms --silent
+
+  sudo kmod-util archive nvidia-grid
+  sudo kmod-util remove nvidia-grid
+
+  sudo ./nvidia-installer --uninstall --silent
+  sudo ./nvidia-installer --no-kernel-modules --silent
+
+  popd
+  sudo rm -rf ${workdir}
+}
+
 archive-open-kmods
 archive-proprietary-kmod
+archive-grid-kmod
 
 ################################################################################
 ### Install NVLSM ##############################################################
--- a/templates/al2023/runtime/gpu/nvidia-kmod-load.sh
+++ b/templates/al2023/runtime/gpu/nvidia-kmod-load.sh
@@ -14,6 +14,11 @@ PCI_CLASS_CODES=(
   "0300" # VGA controller; instance types like g3, g4
   "0302" # 3D controller; instance types like p4, p5
 )
+NVIDIA_GRID_SUBDEVICES=(
+  "27b8:1733" # L4:L4-3Q
+  "27b8:1735" # L4:L4-6Q
+  "27b8:1737" # L4:L4-12Q
+)
 
 INSTANCE_TYPE=$(imds /latest/meta-data/instance-type)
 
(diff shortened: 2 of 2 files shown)


templates/al2023/provisioners/install-nvidia-driver.sh
```diff
@@ -115,8 +117,37 @@ function archive-proprietary-kmod() {
   sudo kmod-util remove nvidia
 }
 
+function archive-grid-kmod() {
+  echo "Archiving GRID kmods"
+  local workdir=NVIDIA-Linux
+  sudo mkdir -p ${workdir}
+  pushd ${workdir}
+
+  sudo aws s3 cp s3://ec2-linux-nvidia-drivers/grid-18.1/NVIDIA-Linux-x86_64-570.133.20-grid-aws.run .
```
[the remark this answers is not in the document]
ndbaker1: also what is the `grid-18.1` part for?

templates/al2023/provisioners/install-nvidia-driver.sh
```diff
@@ -115,8 +117,37 @@ function archive-proprietary-kmod() {
...
+  local package=$(ls NVIDIA-Linux-x86_64*.run)
+  sudo sh ${package} --extract-only
+  sudo rm -rf ${package}
+
+  cd $(ls)
+
+  sudo sed -i 's/PACKAGE_NAME="nvidia"/PACKAGE_NAME="nvidia-grid"/g' kernel/dkms.conf
+
+  sudo dnf -y install dkms
+  sudo ./nvidia-installer --kernel-module-type proprietary --dkms --silent
```
ndbaker1, line 136: so grid only works with `--kernel-module-type proprietary`?

templates/al2023/runtime/gpu/nvidia-kmod-load.sh
```diff
@@ -42,9 +47,23 @@ function devices-support-open() {
   return 0
 }
 
+function device-supports-grid() {
+  for NVIDIA_GRID_SUBDEVICE in "${NVIDIA_GRID_SUBDEVICES[@]}"; do
+    for NVIDIA_SUBDEVICE in $(lspci -n -mm -d 10de:  | tr -d '"' | awk '{print $4":"$7}'); do
```
ndbaker1, line 52: consistency with the existing method: https://github.com/awslabs/amazon-eks-ami/blob/fd505f6c036c8ed1729a5cd779ff4c89c980c32d/templates/al2023/runtime/gpu/nvidia-kmod-load.sh#L41

```suggestion
    for NVIDIA_SUBDEVICE in $(lspci -n -mm -d "${NVIDIA_VENDOR_ID}:" | awk '{print $4":"$7}' | tr -d '"'); do
```

templates/al2023/provisioners/install-nvidia-driver.sh
```diff
@@ -115,8 +117,37 @@ function archive-proprietary-kmod() {
...
+  sudo ./nvidia-installer --uninstall --silent
+  sudo ./nvidia-installer --no-kernel-modules --silent
+
+  popd
+  sudo rm -rf ${workdir}
+}
+
 archive-open-kmods
 archive-proprietary-kmod
+archive-grid-kmod
```
ndbaker1, line 150: i don't love that our final driver installation is with the runfile, i think the existing ordering implicitly enforces that the proprietary package installation is last as its packages are preserved. (the dkms modules just get removed)

https://github.com/awslabs/amazon-eks-ami/blob/38c24acb099d6823433cfce30ebf3736a041befe/templates/al2023/provisioners/install-nvidia-driver.sh#L108-L119

I think you can remove the last installer line and then use the ordering (1) grid (2) open (3) proprietary?

templates/al2023/provisioners/install-nvidia-driver.sh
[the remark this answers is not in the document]
HusainZafar: grid-18.1 is the NVIDIA vGPU software version that includes the GRID drivers. Alternatively we could use latest instead of grid-18.1 if that sounds better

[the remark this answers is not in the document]
HusainZafar: This is the only run file available in the grid-18.1 folder - NVIDIA-Linux-x86_64-570.133.20-grid-aws.run. We can short circuit arm until that is available but that logic would be in the runtime load file

templates/al2023/provisioners/install-nvidia-driver.sh
HusainZafar: Also works with open but need to make more changes for that to work. Specifically while using open, I wasn't able to use a custom name with dkms for grid. Will look into switching to open

templates/al2023/provisioners/install-nvidia-driver.sh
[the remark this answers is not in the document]
ndbaker1: where do we track releases of this and related issues with a particular version? if there's breakage lets pin it but what are the downsides to using latest

templates/al2023/provisioners/install-nvidia-driver.sh
ndbaker1: if we're going to switch open, the path might be something different like `kernel-open/dkms.conf`. i can't recall right now

templates/al2023/provisioners/install-nvidia-driver.sh
HusainZafar: Yes that sounds good

templates/al2023/provisioners/install-nvidia-driver.sh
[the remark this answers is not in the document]
HusainZafar: The driver versions can be found using `aws s3 ls s3://ec2-linux-nvidia-drivers/`
Different grid driver versions correspond to different nvidia driver major versions
For eg. grid-17.6 points to 550.163.01 whereas the latest(grid-18.1) points to 70.133.20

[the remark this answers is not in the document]
HusainZafar: Updated the install script to short circuit non x86_64

templates/al2023/provisioners/install-nvidia-driver.sh
HusainZafar: `kernel-open/dkms.conf` -> This was helpful thanks

templates/al2023/provisioners/install-nvidia-driver.sh
```diff
@@ -78,7 +78,41 @@ sudo dnf -y install \
   kernel-headers-$(uname -r) \
   kernel-modules-extra-common-$(uname -r)
 
+function archive-grid-kmod() {
+  local machine=x86_64
```
ndbaker1: think you wanted something more like this?
```suggestion
  local arch=$(uname -m)
```

templates/al2023/provisioners/install-nvidia-driver.sh
```diff
@@ -78,7 +78,41 @@ sudo dnf -y install \
...
+  if [ "$machine" != "x86_64" ]; then
+    return
+  fi
+  echo "Archiving GRID kmods"
+  local grid_driver_version=grid-18.1
+  local workdir=NVIDIA-Linux
+  sudo mkdir -p ${workdir}
+  pushd ${workdir}
+
+  sudo aws s3 cp "s3://ec2-linux-nvidia-drivers/${grid_driver_version}/NVIDIA-Linux-${machine}-${NVIDIA_DRIVER_MAJOR_VERSION}.133.20-grid-aws.run" .
```
ndbaker1: maintaining this still looks awkward, can we add a check that all archived dkms tarballs use the same driver version? but this introduces a dependency on the source for all 3 versions that we need to make sure is not blocking our releases

templates/al2023/provisioners/install-nvidia-driver.sh
HusainZafar: Oh yea fixed

templates/al2023/provisioners/install-nvidia-driver.sh
HusainZafar: We don't have the same driver versions today for open and proprietary

cartermckinnon:
/ci build

HusainZafar:
/ci build

cartermckinnon (approved):

PR #2272, opened 2025-06-04
feat(nodeadm): pass nvidia gpu startup labels to kubelet

ndbaker1:
**Issue #, if available:**

**Description of changes:**

**Testing Done**

```
make test test-e2e
```


bryantbiggs:
Could we do the same for Neuron and possibly EFA


( on mobile so I can't direct link but see lines 83-84 https://github.com/aws-ia/terraform-aws-eks-blueprints/blob/main/patterns/aws-neuron-efa/eks.tf )

ndbaker1:
> Could we do the same for Neuron and possibly EFA
> 
> ( on mobile so I can't direct link but see lines 83-84 https://github.com/aws-ia/terraform-aws-eks-blueprints/blob/main/patterns/aws-neuron-efa/eks.tf )

thanks for the pointer @bryantbiggs, i wasn't aware these had a presence 😅. are there any manifests/charts that already use these as a standard the way that NVIDIA does, or is it mainly for customers to schedule? (which i guess is a method of scheduling independent of requesting resources from the device plugin)

the concept is the same though, so I'm not opposed to adding those in a follow up

suket22:
+1 to @bryantbiggs. I think it helps things like the Neuron Device Plugin refer to that single label rather than [auto-generate a list of Neuron instances](https://github.com/aws-neuron/neuron-helm-charts/blob/53a6b9fa8c15ef80987c6adb87b710920c42cc14/charts/neuron-helm-chart/templates/_helpers.tpl#L132C34-L132C49) and feed it into the helm chart. I think this might also let us reference this label in the EKS NodeMonitoringAgent's helm chart too instead of the current `node.kubernetes.io/instance-type` affinity? Might be hard with backwards compatibility (need multiple nodeSelectorTerms?) but it's definitely a better direction forward.

nodeadm/test/e2e/cases/nvidia-gpu-labels/run.sh
```diff
@@ -0,0 +1,24 @@
...
+mock::kubelet 1.32.0
+wait::dbus-ready
+
+# since we cannot modify the kernel sysfs paths, we bind mount our mock
+# directory on top of an existing pci device. the only requirement for the check
+# is that the vendor ID file is correct.
+pci_mock_dst=/sys/bus/pci/devices/$(ls /sys/bus/pci/devices/ | head -n 1)
+pci_mock_src=$(mktemp -d)
+echo "0x10de" > $pci_mock_src/vendor
+mount --bind $pci_mock_src $pci_mock_dst
```
suket22: Do you need to `umount` after the test?

suket22 (approved):

nodeadm/test/e2e/cases/nvidia-gpu-labels/run.sh
ndbaker1: nothing's host mounted or persistent in the test harness, so I'm just letting the container clean all that up as long as it's not affecting the test case

nodeadm/internal/kubelet/config.go
```diff
@@ -197,6 +197,20 @@ func (ksc *kubeletConfig) withOutpostSetup(cfg *api.NodeConfig) error {
 	return nil
 }
 
+func (ksc *kubeletConfig) withNvidiaLabels(flags map[string]string) {
```
cartermckinnon: Sounds like we could make use of this for other things, so I think it's worth restructuring so that additional labels are easy to add.

bryantbiggs:
> +1 to @bryantbiggs. I think it helps things like the Neuron Device Plugin refer to that single label rather than [auto-generate a list of Neuron instances](https://github.com/aws-neuron/neuron-helm-charts/blob/53a6b9fa8c15ef80987c6adb87b710920c42cc14/charts/neuron-helm-chart/templates/_helpers.tpl#L132C34-L132C49) and feed it into the helm chart. I think this might also let us reference this label in the EKS NodeMonitoringAgent's helm chart too instead of the current `node.kubernetes.io/instance-type` affinity? Might be hard with backwards compatibility (need multiple nodeSelectorTerms?) but it's definitely a better direction forward.

ya, the two big ones are the [Neuron device plugin values](https://github.com/aws-neuron/neuron-helm-charts/blob/53a6b9fa8c15ef80987c6adb87b710920c42cc14/charts/neuron-helm-chart/values.yaml#L1) that Suket has pointed out, and similarly in the [EFA device plugin values](https://github.com/aws/eks-charts/blob/2a91e36816290c68b96c7581b55a856b4fe796aa/stable/aws-efa-k8s-device-plugin/values.yaml#L13-L168)

The only precedence currently is the use of `nvidia.com/gpu.present=true` but it seems like a simple and straight forward approach to follow on other devices to avoid the static instance lists

nodeadm/internal/kubelet/config.go
ndbaker1: updated the implementation PTAL!

nodeadm/internal/kubelet/config.go
```diff
@@ -298,13 +316,15 @@ func (k *kubelet) GenerateKubeletConfig(cfg *api.NodeConfig) (*kubeletConfig, er
...
-		return nil, err
-	}
 
+	kubeletConfig.withPodInfraContainerImage(cfg, k.flags)
 	kubeletConfig.withVersionToggles(cfg, k.flags)
 	kubeletConfig.withCloudProvider(cfg, k.flags)
 	kubeletConfig.withDefaultReservedResources(cfg)
+	kubeletConfig.withNodeLabels(k.flags, map[string]LabelValueFunc{
+		// see: https://github.com/NVIDIA/gpu-operator/commit/e25291b86cf4542ac62d8635cda4bd653c4face3
+		"nvidia.com/gpu.present": withNvidiaGPULabel(),
```
bryantbiggs: We should strive to only set this when true. It doesn't make a lot of sense to add this if the value is false and just adds to the node object size (which is usually quite large to begin with)

ndbaker1: agreed, it only sets the label given no error and and `ok` which i never set unless we've got devices. just keeping the api flexible

ndbaker1:
/ci
+workflow:k8s_versions 1.33

ndbaker1:
i pulled up the node logs from the ci and confirmed the labels
```bash
# nodeadm logs
Jul 07 19:16:23 localhost nodeadm[1961]: {"level":"info","ts":1751915783.9248972,"caller":"kubelet/config.go:212","msg":"Adding node label","label":"nvidia.com/gpu.present=true"}

# kubelet logs
Jul 07 19:16:32 ip-192-168-151-135.us-west-2.compute.internal kubelet[2876]: I0707 19:16:32.864865    2876 flags.go:64] FLAG: --node-labels="nvidia.com/gpu.present=true"
```

nodeadm/internal/util/devices.go
cartermckinnon, line 1: Can we move this into the `system` package? We have some utility-style methods there to get things like available CPU/mem, etc.

nodeadm/test/e2e/cases/nvidia-gpu-labels/run.sh
```diff
@@ -0,0 +1,29 @@
...
+assert::file-not-contains /etc/eks/kubelet/environment '--node-labels=nvidia.com/gpu.present'
+assert::file-contains /etc/eks/kubelet/environment '--node-labels=foo=bar'
+
+# since we cannot modify the kernel sysfs paths, we bind mount our mock
+# directory on top of an existing pci device. the only requirement for the check
+# is that the vendor ID file is correct.
+pci_mock_dst=/sys/bus/pci/devices/$(ls /sys/bus/pci/devices/ | head -n 1)
+pci_mock_src=$(mktemp -d)
+echo "0x10de" > $pci_mock_src/vendor
+mount --bind $pci_mock_src $pci_mock_dst
```
cartermckinnon: Seems useful to move this into `helpers.sh` and have a `mock::nvidia-device` or something to re-use

ndbaker1: ill prob do smth like `mock::pci-device` so that we can use it for neuron and friends as well

ndbaker1:
@cartermckinnon scoped the version to start once we launch 1.34+ and updated docs, see the preview in https://ndbaker1.github.io/amazon-eks-ami/nodeadm/doc/api-concepts/

fletcherw:
closing in favor of #2607

PR #2279, opened 2025-06-05
fix: make kmod-util module-version comptaible with all dkms versions

HusainZafar:
**Issue #, if available:**

**Description of changes:**
Some versions of dkms have a different format like so `nvidia, 550.163.01: added`. This change looks at the dkms module path directly to determine the module version

**Testing Done**

### DIFF
--- a/templates/al2023/runtime/gpu/kmod-util
+++ b/templates/al2023/runtime/gpu/kmod-util
@@ -14,7 +14,7 @@ function log() {
 
 # get the version of a registered kernel module
 function module-version() {
-  ${DKMS} status -m "${1}" | tr -s ',' '/' | tr -d ' ' | awk -F '/' '{print $2}'
+  ls /var/lib/dkms/${1}/ | head -n 1
 }
 
 # load a kernel module from the archives


templates/al2023/runtime/gpu/kmod-util
```diff
@@ -14,7 +14,7 @@ function log() {
 
 # get the version of a registered kernel module
 function module-version() {
-  ${DKMS} status -m "${1}" | tr -s ',' '/' | tr -d ' ' | awk -F '/' '{print $2}'
+  ${DKMS} status -m "${1}" | tr -s ',:' '/' | tr -d ' ' | awk -F '/' '{print $2}'
```
ndbaker1: Parsing this feels hacky at this point, could we just read the dkms paths?
```suggestion
  ls /var/lib/dkms/${1}/ | head -n 1
```

HusainZafar: Yes this works well although it relies on an assumption about the directory setup which may change and break things in the future?

HusainZafar:
/ci build

HusainZafar:
Canceling PR in favor of https://github.com/awslabs/amazon-eks-ami/pull/2284

PR #2280, opened 2025-06-06
fix: circular dependency issue in nodeadm-run.service

### LINKED ISSUE #2269: Circular dependency issue
This is the error in my boot log:

```
sudo journalctl -b -p 3
<date> <time> <host-name> systemd[1]: multi-user.target: Job nodeadm-run.service/start deleted to break ordering cycle starting with multi-user.target/start
```

After the investigation, I found that there is a circular dependency issue:

https://github.com/awslabs/amazon-eks-ami/blob/79a046046cacc4c64e33e8e7e8bcfa5e1bd74687/templates/al2023/runtime/rootfs/etc/systemd/system/nodeadm-run.service#L6

https://github.com/awslabs/amazon-eks-ami/blob/eb6757e2157c4c6f30a83ed7cddbaf01d223cd7d/templates/al2023/runtime/rootfs/etc/systemd/system/nodeadm-run.service#L14

https://github.com/canonical/cloud-init/blob/c3848d5747f0d1342ea703b6838e1c568c776d9a/systemd/cloud-final.service#L5
```js
After=multi-user.target
```

`multi-user.target` wants `nodeadm-run.service` but:
- `nodeadm-run.service` must run after `cloud-final.service`.
- `cloud-final.service` must run after `multi-user.target`.

ansidev:
**Issue #, if available:**
Fixes #2269 

**Description of changes:**
nodeadm-run.service will be triggered by cloud-init.target instead of multi-user.target

**Testing Done**

I updated the systemd service nodeadm-run.service and rebooted the EC2 instance.
Then I ran the command `sudo journalctl -b -p 3`.

Result: the error boot log did not contain any error related to noeadm-run.service anymore (expected).

### DIFF
--- a/templates/al2023/runtime/rootfs/etc/systemd/system/nodeadm-run.service
+++ b/templates/al2023/runtime/rootfs/etc/systemd/system/nodeadm-run.service
@@ -11,4 +11,4 @@ Type=oneshot
 ExecStart=/usr/bin/nodeadm init --skip config
 
 [Install]
-WantedBy=multi-user.target
+WantedBy=cloud-init.target


fletcherw:
Closing this since we could not reproduce the issue on the EKS AMIs.

PR #2285, opened 2025-06-09
feat(nodeadm): enable DRA for 1.33+

cartermckinnon:
**Description of changes:**

This opens the `DynamicResourceAllocation` feature gate for Kubernetes 1.33 and above.

### DIFF
--- a/nodeadm/internal/kubelet/config.go
+++ b/nodeadm/internal/kubelet/config.go
@@ -229,6 +229,11 @@ func (ksc *kubeletConfig) withVersionToggles(cfg *api.NodeConfig, flags map[stri
 		ksc.KubeAPIQPS = ptr.Int(10)
 		ksc.KubeAPIBurst = ptr.Int(20)
 	}
+
+	// EKS enables DRA on 1.33+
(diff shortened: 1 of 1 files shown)


nodeadm/internal/kubelet/config.go
```diff
@@ -229,6 +229,11 @@ func (ksc *kubeletConfig) withVersionToggles(cfg *api.NodeConfig, flags map[stri
 		ksc.KubeAPIQPS = ptr.Int(10)
 		ksc.KubeAPIBurst = ptr.Int(20)
 	}
+
+	// EKS enables DRA on 1.33+
```
HusainZafar, line 233: Why does 1.33 need a separate logic? Can this not be combined with the other versions, whatever enables it upto 1.32?

HusainZafar (approved):

nodeadm/internal/kubelet/config.go
cartermckinnon: The feature gate only exists in 1.30 and above, so this was just the simplest way to do it. When the feature gate is removed in the future, it's easy to tack on `<= $FINAL_VERSION` here

PR #2293, opened 2025-06-12
[POC] feat(nodeadm): reconcile max pods with ipamd

mselim00:
**Issue #, if available:**

**Description of changes:**
This adds a new nodeadm phase (after run) that will poll IPAMD for some pre-configured timeout, and then re-configure the reserved resources for kubelet after receiving a response. It is intended as a one-off, so it will short-circuit if there's an indication a previous configuration ran. It also requires enabling a new feature gate, and running a supported version of kubelet. This is intended to allow nodeadm to automatically configure kubelet based on the features of the VPC CNI actively in-use, gauged via the locally running instance of ipamd.

This utilizes a feature branch of the VPC CNI for interrogating the allocatable IPs: https://github.com/aws/amazon-vpc-cni-k8s/compare/master...mselim00:amazon-vpc-cni-k8s:expose-ip-count

**Testing Done**


jaydeokar:
High level question - Do we have to vendor the CNI repository here? Can't we add as a go.mod dependency ?

mselim00:
> Do we have to vendor the CNI repository here? Can't we add as a go.mod dependency ?

We vendor all dependencies for nodeadm to allow for consistent and network-less builds. I'll look into the size difference to make sure we're not getting too much unnecessary bloat though

ref: 
https://github.com/awslabs/amazon-eks-ami/blob/6137919b86dad11cd8d847732ae91f91b8932049/templates/al2023/provisioners/install-nodeadm.sh#L19-L26

nodeadm/cmd/nodeadm-aux/main.go
cartermckinnon, line 1: Maybe we should name this `nodeadm-internal` to make the support/intent as clear as possible?

nodeadm/cmd/nodeadm-aux/nodeadm-reconcile.go
cartermckinnon, line 1: this should go in a directory + file with the same name as the command (`reconcile/reconcile.go`), to fit with our structure elsewhere.

also prob needs a more specific name, `reconcile-cni-config` or something?

nodeadm/internal/cli/config.go
```diff
@@ -0,0 +1,50 @@
...
+	"go.uber.org/zap"
+
+	"github.com/awslabs/amazon-eks-ami/nodeadm/internal/api"
+	"github.com/awslabs/amazon-eks-ami/nodeadm/internal/aws/imds"
+	"github.com/awslabs/amazon-eks-ami/nodeadm/internal/kubelet"
+)
+
+// Various initializations and verifications of the NodeConfig and
+// perform in-place updates when allowed by the user
+func EnrichConfig(log *zap.Logger, cfg *api.NodeConfig) error {
```
cartermckinnon: kind of a weird place for this IMO, the `cli` package is intended to be a framework for building the command-line interface, not really the business logic of commands. What about a new `config` or `runtime` package?

nodeadm/internal/ipamd/grpc.go
```diff
@@ -0,0 +1,40 @@
...
+	conn, err := grpcClient.Dial(cniBackendServerAddress, grpc.WithTransportCredentials(insecure.NewCredentials()))
+	if err != nil {
+		return 0, fmt.Errorf("failed to connect to ipamd server: %v", err)
+	}
+	defer conn.Close()
+
+	cniBackendClient := rpcwrapper.New().NewCNIBackendClient(conn)
+	resp, err := cniBackendClient.GetAllocatableValues(ctx, &emptypb.Empty{})
+	if err != nil {
+		return 0, err
```
cartermckinnon:
```suggestion
		return -1, err
```

nodeadm/internal/ipamd/grpc.go
```diff
@@ -0,0 +1,40 @@
...
+	}
+	if resp.MaxAllocatableIPs <= 0 {
+		return 0, fmt.Errorf("received a non-positive value for max IPs: %d", resp.MaxAllocatableIPs)
```
cartermckinnon:
```suggestion
		return -1, fmt.Errorf("received a non-positive value for max IPs: %d", resp.MaxAllocatableIPs)
```

nodeadm/internal/ipamd/grpc.go
```diff
@@ -0,0 +1,40 @@
...
+	"google.golang.org/protobuf/types/known/emptypb"
+)
+
+const cniBackendServerAddress = "127.0.0.1:50051"
+
+func GetMaxAllocatableIPs(ctx context.Context) (int32, error) {
+	grpcClient := grpcwrapper.New()
+	conn, err := grpcClient.Dial(cniBackendServerAddress, grpc.WithTransportCredentials(insecure.NewCredentials()))
+	if err != nil {
+		return 0, fmt.Errorf("failed to connect to ipamd server: %v", err)
```
cartermckinnon:
```suggestion
		return -1, fmt.Errorf("failed to connect to ipamd server: %v", err)
```

nodeadm/cmd/nodeadm-aux/nodeadm-reconcile.go
```diff
@@ -0,0 +1,124 @@
...
+	if err := cli.EnrichConfig(log, nodeConfig); err != nil {
+		return err
+	}
+
+	zap.L().Info("Validating configuration..")
+	if err := api.ValidateNodeConfig(nodeConfig); err != nil {
+		return err
+	}
+
+	if !api.IsFeatureEnabled(api.CNIReconcile, nodeConfig.Spec.FeatureGates) {
```
cartermckinnon: Could you check this before calling `EnrichConfig`?

nodeadm/internal/kubelet/eni_max_pods.go
```diff
@@ -63,5 +95,59 @@ func CalcMaxPods(awsRegion string, instanceType string) int32 {
...
+
+// getMaxPodsCeiling provides a basic ceiling to ensure that reserved max pods does not
+// overconsume reserved memory, creating a diminishing effect on the number of schedulable pods.
+// Limits the value so that the max pods use no more than 25% of memory, to a hard maximum of 737 pods.
+func getMaxPodsCeiling(totalMemoryMB int) int32 {
+	memoryMaxPodsCeiling := float64(projectMaxPodsFromMemoryMBReserved(float64(totalMemoryMB) * maxPercentMemoryReserved))
+	return int32(math.Min(memoryMaxPodsCeiling, maxPodsHardCeiling))
+}
+
+func PollInterrogateMaxPods(ctx context.Context, maxDuration time.Duration, backoff time.Duration) (int32, error) {
```
cartermckinnon: The request-level logic for timeout/retries/etc. should be moved into the `ipamd` package. I think what we need here is a function that takes an ENI count and returns a maxPods value after applying the various heuristics

nodeadm/internal/system/resources.go
```diff
@@ -81,6 +82,15 @@ func GetMilliNumCores() (int, error) {
 
 }
 
+func GetTotalMemoryMB() uint64 {
+	var sysinfo syscall.Sysinfo_t
+	err := syscall.Sysinfo(&sysinfo)
+	if err != nil {
+		return 0
+	}
+	return (uint64(sysinfo.Totalram) * uint64(sysinfo.Unit)) / 1000000
```
cartermckinnon: does this always give us the total RAM for the entire system, or does it differ if you're within a cgroup?

cartermckinnon (commented):
Gave this a pass for structure and a few lower-level things. I think we need to spell out how this will change reserved resources in various scenarios, it's difficult to review what's going on in `eni_max_pods.go` without real examples.

nodeadm/internal/kubelet/config.go
cartermckinnon, line 1: Can you get away with adding a targeted func here for writing a maxPods dropin, instead of having to expose _all_ the kubelet config details?

nodeadm/internal/system/resources.go
mselim00: I'm not sure, I'll have to double check this one

nodeadm/internal/kubelet/eni_max_pods.go
mselim00: makes sense moving the retry and stuff to imds, I'll work on changing that. the one thing about this is we I don't think we want to try setting defaults or anything in this dynamic config file, e.g. in case an EC2/IMDS intermittent error is hit on that run but wasn't in the initial nodeadm. so for the dynamic configuration, it should have more granularity of specifying what goes into the response it gets. I'm thinking maybe a variadic function with function inputs for those different heuristics

nodeadm/cmd/nodeadm-aux/nodeadm-reconcile.go
mselim00: makes sense, I'll revise to do that

nodeadm/internal/kubelet/config.go
mselim00: that would work too, but I was thinking this might be more extensible in case we end up wanting to manage more mini-configs like this one. the kubelet package is also module-internal so it's really only exposing it to other parts of nodeadm

mselim00:
Working on splitting off the limiting aspect for the max pods value in another PR: https://github.com/awslabs/amazon-eks-ami/pull/2325

nodeadm/internal/kubelet/config.go
mselim00: switched it to a dedicated function to generate the config, but it's a little awkward with the default config, which is generated via a generic Configure function to the kubelet daemon. this one's generated by the command with a Generate function directly

nodeadm/internal/system/resources.go
mselim00: tested this out in https://github.com/awslabs/amazon-eks-ami/pull/2325#discussion_r2211932553
and moved this logic out to that PR, so closing this thread in favor of that one

nodeadm/internal/cli/config.go
mselim00: added it into a `config` package

nodeadm/internal/kubelet/eni_max_pods.go
mselim00: still sort of working on the details for what that interface should look like. there's more to this in https://github.com/awslabs/amazon-eks-ami/pull/2325, where I also split off the max pods logic into another package

mselim00:
Superseded by #2446

PR #2277, opened 2025-06-05
chore(nodeadm): add retry attempt error logs for ec2.DescribeInstances

kummatty:
**Description of changes:**

Enhanced error visibility and optimized retry handling for EC2 DescribeInstance calls in nodeadm.
Added detailed error logs for each retry attempt during ec2.DescribeInstances calls with specific error codes. Additionally, disabled AWS SDK's built-in retry configuration to eliminate redundant retry logic, ensuring the custom instance waiter is the single source of retry behavior.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

**Testing Done**

* Created a cluster via packer (make os_distro=al2023) in personal account
* Ensured node can join the cluster.
* Simulated error scenario by disabling permissions for ec2.DescribeInstances in the node IAM role

*[See this guide for recommended testing for PRs.](../doc/CONTRIBUTING.md#testing-changes) Some tests may not apply. Completing tests and providing additional validation steps are not required, but it is recommended and may reduce review time and time to merge.*

PR #2299, opened 2025-06-20
fix(nodeadm): use aws retry and timeout definitions in ec2 waiter

ndbaker1:
**Issue #, if available:**

**Description of changes:**

in https://github.com/awslabs/amazon-eks-ami/pull/2277 we've updated the functionality of the waiter to be a replacement for the default ec2 client retry mechanism, however it does not handle other generic cases such as network timeouts. 

This PR takes the default timeout and retry definitions from the AWS go SDK and uses them to decide whether the error is retryable by the waiter wrapper.

**Testing Done**

### DIFF
--- a/nodeadm/internal/aws/ec2/instance_waiter.go
+++ b/nodeadm/internal/aws/ec2/instance_waiter.go
@@ -2,13 +2,12 @@ package ec2
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"time"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/aws/retry"
 	"github.com/aws/aws-sdk-go-v2/service/ec2"
-	"github.com/aws/smithy-go"
 	"github.com/aws/smithy-go/middleware"
 	smithytime "github.com/aws/smithy-go/time"
 	smithywaiter "github.com/aws/smithy-go/waiter"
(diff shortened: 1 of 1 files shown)


nodeadm/internal/aws/ec2/instance_waiter.go
```diff
@@ -175,13 +171,12 @@ func (w *InstanceConditionWaiter) WaitForOutput(ctx context.Context, params *ec2
...
-		var apiErr smithy.APIError
-		ok := errors.As(err, &apiErr)
-		if !ok {
-			return false, fmt.Errorf("expected err to be of type smithy.APIError, got %w", err)
+		if retry.IsErrorRetryables(retry.DefaultRetryables).IsErrorRetryable(err).Bool() {
+			zap.L().Warn("retryable error encountered", zap.Error(err))
+			return true, nil
 		}
-
-		if "InvalidInstanceID.NotFound" == apiErr.ErrorCode() {
```
mselim00, line 184: I'm not really seeing where this is covered atm? Is this still needed b/c of the eventual consistency nature of some APIs?

also seems like we're mostly mirroring the standard retryer, is there a notable feature of it we're avoiding?

ndbaker1: This is just the most miminal way to use the same retryer logic, i can link the code for this retry in a bit for clarification

cartermckinnon: We need to keep the `InvalidInstanceID.NotFound` check, otherwise the changes look fine to me.

ndbaker1: oops, added that back to the custom status code map

ndbaker1: all essentially pulled from https://github.com/aws/aws-sdk-go-v2/blob/bba018b7b8200fecfe0fab51e38470ff612c9689/aws/retry/standard.go#L171-L201, and the only difference with the adaptive retryer is using [DefaultThrottleErrors](https://github.com/aws/aws-sdk-go-v2/blob/bba018b7b8200fecfe0fab51e38470ff612c9689/aws/retry/standard.go#L90) in two places

PR #2304, opened 2025-06-24
feat(log-collector): collect systemd-network configs and unattended-upgrades

Issacwww:
**Issue #, if available:**
https://github.com/aws/amazon-vpc-cni-k8s/issues/2103

**Description of changes:**
Add this log collection to  better verify if [the known **systemd-udev** issue](https://github.com/aws/amazon-vpc-cni-k8s/blob/f6ed39b50840e7f6985f33113feeb8e7572bfb31/docs/troubleshooting.md#known-issues) occur

**Test**
The collected log looks like below
```
# /usr/lib/systemd/network/99-default.link
# SPDX-License-Identifier: MIT-0
#
# This config file is installed as part of systemd.
# It may be freely copied and edited (following the MIT No Attribution license).
#
# To make local modifications, one of the following methods may be used:
# 1. add a drop-in file that extends this file by creating the
#    /etc/systemd/network/99-default.link.d/ directory and creating a
#    new .conf file there.
# 2. copy this file into /etc/systemd/network or one of the other paths checked
#    by systemd-udevd and edit it there.
# This file should not be edited in place, because it'll be overwritten on upgrades.

[Match]
OriginalName=*

[Link]
NamePolicy=keep kernel database onboard slot path
AlternativeNamesPolicy=database onboard slot path
MACAddressPolicy=persistent

# /etc/systemd/network/99-default.link.d/99-no-policy.conf
# Ensure MACAddressPolicy=none, reinstalling systemd-udev writes /usr/lib/systemd/network/99-default.link
# with value set to persistent
# https://github.com/aws/amazon-vpc-cni-k8s/issues/2103#issuecomment-1321698870
[Link]
MACAddressPolicy=none
```

### DIFF
--- a/log-collector-script/linux/eks-log-collector.sh
+++ b/log-collector-script/linux/eks-log-collector.sh
@@ -71,6 +71,7 @@ COMMON_DIRECTORIES=(
 COMMON_LOGS=(
   syslog
   messages
+  unattended-upgrades
   aws-routed-eni # eks
   containers     # eks
   pods           # eks
@@ -668,7 +669,32 @@ get_networking_info() {
     ethtool -S ${ifc} >> "${COLLECT_DIR}"/networking/ethtool.txt 2>&1
     echo -e "\n" >> "${COLLECT_DIR}"/networking/ethtool.txt
   done
+
+  get_systemd_network_config
   ok
+}
+
+get_systemd_network_config() {
+  # The possible locations of systemd-networkd configuration files
+  # https://www.freedesktop.org/software/systemd/man/latest/systemd.network.html
+  folders=(
+    "/usr/lib/systemd/network"
+    "/etc/systemd/network"
+    "/run/systemd/network"
+    "/usr/local/lib/systemd/network"
+  )
+
+  mkdir -p "${COLLECT_DIR}"/networking/systemd-network
+
+  # Process each folder
+  for folder in "${folders[@]}"; do
+    [ ! -d "$folder" ] && continue
+    [ -z "$(ls -A "$folder")" ] && continue
+    for unit in "$folder"/*.*; do
+      [ ! -f "$unit" ] && continue
+      systemd-analyze cat-config "${unit}" > "${COLLECT_DIR}"/networking/systemd-network/$(basename "$unit") 2> /dev/null
+    done
+  done
 }
 
 get_cni_config() {


log-collector-script/linux/eks-log-collector.sh
```diff
@@ -71,6 +71,7 @@ COMMON_DIRECTORIES=(
 COMMON_LOGS=(
   syslog
   messages
+  unattended-upgrades
```
wwvela, line 74: never notice this log file before, does the `unattended-upgrades` a dir under `/var/log/` and file `unattended-upgrades.log` is under unattended-upgrades dir? Saw you mentioned unattended-upgrades.log in title

Issacwww: This log is generated from [Ubuntu AutomaticSecurityUpdates](https://help.ubuntu.com/community/AutomaticSecurityUpdates)
Which EKS optimized AMI disabled so you never see it. But we have seen use case that for ubuntu AMI, and this auto-update modify MacAddressPolicy, and end up with broken pod network.

wwvela: I see. Thanks for the explaining. LGTM

wwvela (approved):

log-collector-script/linux/eks-log-collector.sh
```diff
@@ -668,6 +669,11 @@ get_networking_info() {
     ethtool -S ${ifc} >> "${COLLECT_DIR}"/networking/ethtool.txt 2>&1
     echo -e "\n" >> "${COLLECT_DIR}"/networking/ethtool.txt
   done
+
+  # collect systemd network configuration
```
ndbaker1: do we also want to try `networkctl status -a`, it lists all of the links and also descriptive logs like:
```
Jun 18 00:16:06 localhost systemd-networkd[2040]: ens5: Configuring with /usr/lib/systemd/network/80-ec2.network.
```

log-collector-script/linux/eks-log-collector.sh
```diff
@@ -668,6 +669,11 @@ get_networking_info() {
...
+  cp --force --recursive --dereference /usr/lib/systemd/network/ "${COLLECT_DIR}"/networking/usr-lib-systemd-network/ 2> /dev/null
+  cp --force --recursive --dereference /etc/systemd/network/ "${COLLECT_DIR}"/networking/etc-systemd-network/ 2> /dev/null
```
ndbaker1: this sort of relies on us having all of the drop-in locations from https://www.freedesktop.org/software/systemd/man/latest/systemd.network.html, otherwise we *could* miss something.

thinking maybe we could list all final systemd configs based on this directory with something like:
```bash
for unit in /usr/lib/systemd/network/*.*; do
  systemd-analyze cat-config $unit > "${COLLECT_DIR}"/networking/systemd-network/$unit
done
```

Issacwww: Updated with this approach to collect all possible localtions

ndbaker1: Do you end up collecting duplicate files with the way its done right now? `cat-config` will output with the drop-ins accounted for so i wonder if you need a filter on `.conf` or something. If there's no duplicates then it should be fine

ndbaker1: Ah nvm i see the filter on directories 👍

log-collector-script/linux/eks-log-collector.sh
```diff
@@ -668,9 +669,35 @@ get_networking_info() {
...
+
+  mkdir -p "${COLLECT_DIR}"/networking/systemd-network
+
+  # Process each folder
+  for folder in "${folders[@]}"; do
+    [ ! -d "$folder" ] && continue
+    [ -z "$(ls -A "$folder")" ] && continue
+    for unit in "$folder"/*.*; do
+      [ ! -f "$unit" ] && continue
+      basename=$(basename "$unit")
```
ndbaker1: nit: using variable names that shadow other binaries/functions is awkward IMO, could just inline this

PR #2305, opened 2025-06-25
feat(nodeadm): add SOCI snapshotter support

HusainZafar:
**Issue #, if available:**

**Description of changes:**
Configures kubelet and containerd to use SOCI snapshotter if the `FastContainerImagePull` feature flag is set in nodeadm

**Testing Done**

### DIFF
--- a/nodeadm/api/v1alpha1/nodeconfig_types.go
+++ b/nodeadm/api/v1alpha1/nodeconfig_types.go
@@ -136,6 +136,6 @@ const (
 	// InstanceIdNodeName will use EC2 instance ID as node name
 	InstanceIdNodeName Feature = "InstanceIdNodeName"
 
-	// Allows more granular control for faster image pull
+	// FastContainerImagePull reduces image pull time
 	FastContainerImagePull Feature = "FastContainerImagePull"
 )
--- a/nodeadm/api/v1alpha1/nodeconfig_types.go
+++ b/nodeadm/api/v1alpha1/nodeconfig_types.go
@@ -129,10 +129,13 @@ const (
 )
 
 // Feature specifies which feature gate should be toggled
-// +kubebuilder:validation:Enum={InstanceIdNodeName}
+// +kubebuilder:validation:Enum={InstanceIdNodeName,FastContainerImagePull}
 type Feature string
 
 const (
 	// InstanceIdNodeName will use EC2 instance ID as node name
 	InstanceIdNodeName Feature = "InstanceIdNodeName"
+
+	// Allows more granular control for faster image pull
+	FastContainerImagePull Feature = "FastContainerImagePull"
 )
--- a/nodeadm/internal/api/featuregates.go
+++ b/nodeadm/internal/api/featuregates.go
@@ -14,6 +14,9 @@ var featureVerifiers = map[Feature]func(Feature, map[Feature]bool) bool{
 	// InstanceIdNodeNameGate controls whether to use instance ID as the node's name.
 	// By default, this feature is disabled, and the private DNS Name will be used.
 	InstanceIdNodeName: DefaultFalse,
+
+	// Allows more granular control for faster image pull. Disabled by default
+	FastContainerImagePull: DefaultFalse,
 }
 
 func IsFeatureEnabled(feature Feature, featureGates map[Feature]bool) bool {
--- a/nodeadm/internal/api/featuregates.go
+++ b/nodeadm/internal/api/featuregates.go
@@ -15,7 +15,7 @@ var featureVerifiers = map[Feature]func(Feature, map[Feature]bool) bool{
 	// By default, this feature is disabled, and the private DNS Name will be used.
 	InstanceIdNodeName: DefaultFalse,
 
-	// Allows more granular control for faster image pull. Disabled by default
+	// FastContainerImagePull reduces image pull time. Disabled by default
 	FastContainerImagePull: DefaultFalse,
 }
 
--- a/nodeadm/internal/api/types.go
+++ b/nodeadm/internal/api/types.go
@@ -122,4 +122,7 @@ type Feature string
 const (
 	// InstanceIdNodeName will use EC2 instance ID as node name
 	InstanceIdNodeName Feature = "InstanceIdNodeName"
+
+	// Allows more granular control for faster image pull
+	FastContainerImagePull Feature = "FastContainerImagePull"
 )
--- a/nodeadm/internal/containerd/config.go
+++ b/nodeadm/internal/containerd/config.go
@@ -26,7 +26,7 @@ var (
 	containerdConfigTemplate     = template.Must(template.New(containerdConfigFile).Parse(containerdConfigTemplateData))
 
 	//go:embed snapshotter/soci-snapshotter.config.toml
-	sociSnapshotterTemplateData string
+	sociSnapshotterTemplateData []byte
 )
 
 type containerdTemplateVars struct {
@@ -49,7 +49,7 @@ func writeContainerdConfig(cfg *api.NodeConfig) error {
 
 func writeSnapshotterConfig(cfg *api.NodeConfig) error {
 	if api.IsFeatureEnabled(api.FastContainerImagePull, cfg.Spec.FeatureGates) {
-		return util.WriteFileWithDir(sociSnapshotterConfigFile, []byte(sociSnapshotterTemplateData), configPerm)
+		return util.WriteFileWithDir(sociSnapshotterConfigFile, sociSnapshotterTemplateData, configPerm)
 	}
 	return nil
 }
--- a/nodeadm/internal/containerd/config.go
+++ b/nodeadm/internal/containerd/config.go
@@ -15,55 +15,79 @@ import (
 const ContainerRuntimeEndpoint = "unix:///run/containerd/containerd.sock"
 
 const (
-	containerdConfigFile = "/etc/containerd/config.toml"
-	containerdConfigPerm = 0644
+	containerdConfigFile      = "/etc/containerd/config.toml"
+	sociSnapshotterConfigFile = "/etc/soci-snapshotter-grpc/config.toml"
+	configPerm                = 0644
 )
 
 var (
 	//go:embed config.template.toml
 	containerdConfigTemplateData string
 	containerdConfigTemplate     = template.Must(template.New(containerdConfigFile).Parse(containerdConfigTemplateData))
+
+	//go:embed snapshotter/soci-snapshotter.config.toml
+	sociSnapshotterTemplateData string
 )
 
 type containerdTemplateVars struct {
-	EnableCDI         bool
-	SandboxImage      string
-	RuntimeName       string
-	RuntimeBinaryName string
+	EnableCDI          bool
+	SandboxImage       string
+	RuntimeName        string
+	RuntimeBinaryName  string
+	UseSOCISnapshotter bool
 }
 
 func writeContainerdConfig(cfg *api.NodeConfig) error {
-	containerdConfig, err := generateContainerdConfig(cfg)
+	containerdConfig, err := combineContainerdConfigs(cfg)
 	if err != nil {
 		return err
+	}
+
+	zap.L().Info("Writing containerd config to file..", zap.String("path", containerdConfigFile))
+	return util.WriteFileWithDir(containerdConfigFile, containerdConfig, configPerm)
+}
+
+func writeSnapshotterConfig(cfg *api.NodeConfig) error {
+	if api.IsFeatureEnabled(api.FastContainerImagePull, cfg.Spec.FeatureGates) {
+		return util.WriteFileWithDir(sociSnapshotterConfigFile, []byte(sociSnapshotterTemplateData), configPerm)
+	}
+	return nil
+}
+
+func combineContainerdConfigs(cfg *api.NodeConfig) ([]byte, error) {
+	containerdConfig, err := generateContainerdConfig(cfg)
+	if err != nil {
+		return nil, err
 	}
 
 	// because the logic in containerd's import merge decides to completely
 	// overwrite entire sections, we want to implement this merging ourselves.
 	// see: https://github.com/containerd/containerd/blob/a91b05d99ceac46329be06eb43f7ae10b89aad45/cmd/containerd/server/config/config.go#L407-L431
+
+	// Merge user-provided config if present
 	if len(cfg.Spec.Containerd.Config) > 0 {
 		containerdConfigMap, err := util.Merge(containerdConfig, []byte(cfg.Spec.Containerd.Config), toml.Marshal, toml.Unmarshal)
 		if err != nil {
-			return err
+			return nil, err
 		}
 		containerdConfig, err = toml.Marshal(containerdConfigMap)
 		if err != nil {
-			return err
+			return nil, err
 		}
 	}
 
-	zap.L().Info("Writing containerd config to file..", zap.String("path", containerdConfigFile))
-	return util.WriteFileWithDir(containerdConfigFile, containerdConfig, containerdConfigPerm)
+	return containerdConfig, nil
 }
 
 func generateContainerdConfig(cfg *api.NodeConfig) ([]byte, error) {
 	runtimeOptions := getRuntimeOptions(cfg)
 
 	configVars := containerdTemplateVars{
-		SandboxImage:      cfg.Status.Defaults.SandboxImage,
-		RuntimeBinaryName: runtimeOptions.RuntimeBinaryPath,
-		RuntimeName:       runtimeOptions.RuntimeName,
-		EnableCDI:         semver.Compare(cfg.Status.KubeletVersion, "v1.32.0") >= 0,
+		SandboxImage:       cfg.Status.Defaults.SandboxImage,
+		RuntimeBinaryName:  runtimeOptions.RuntimeBinaryPath,
+		RuntimeName:        runtimeOptions.RuntimeName,
+		EnableCDI:          semver.Compare(cfg.Status.KubeletVersion, "v1.32.0") >= 0,
+		UseSOCISnapshotter: api.IsFeatureEnabled(api.FastContainerImagePull, cfg.Spec.FeatureGates),
 	}
 	var buf bytes.Buffer
 	if err := containerdConfigTemplate.Execute(&buf, configVars); err != nil {
--- a/nodeadm/internal/containerd/containerd_config_test.go
+++ b/nodeadm/internal/containerd/containerd_config_test.go
@@ -1,0 +1,112 @@
+package containerd
+
+import (
+	"testing"
+
+	"github.com/awslabs/amazon-eks-ami/nodeadm/internal/api"
+	"github.com/pelletier/go-toml/v2"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestContainerdConfigWithUserConfigAndFastContainerImagePullFeature(t *testing.T) {
+	cfg := &api.NodeConfig{
+		Spec: api.NodeConfigSpec{
+			FeatureGates: map[api.Feature]bool{
+				api.FastContainerImagePull: true,
+			},
+			Containerd: api.ContainerdOptions{
+				Config: api.ContainerdConfig(`
+[plugins.'io.containerd.grpc.v1.cri'.registry.mirrors]
+"docker.io" = ["https://my-custom-mirror.example.com"]
+`),
+			},
+		},
+	}
+	containerdConfig, err := combineContainerdConfigs(cfg)
+	assert.NoError(t, err)
+
+	// Parse the containerdConfig
+	var configMap map[string]interface{}
+	err = toml.Unmarshal(containerdConfig, &configMap)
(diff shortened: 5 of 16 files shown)


nodeadm/internal/api/featuregates.go
```diff
@@ -14,6 +14,10 @@ var featureVerifiers = map[Feature]func(Feature, map[Feature]bool) bool{
 	// InstanceIdNodeNameGate controls whether to use instance ID as the node's name.
 	// By default, this feature is disabled, and the private DNS Name will be used.
 	InstanceIdNodeName: DefaultFalse,
+
+	// InstanceIdNodeNameGate controls whether containerd should be setup to use SOCI Snaphotter
```
fletcherw: nit: copy-paste error in comment, and typo for 'snapshotter'

nodeadm/internal/containerd/snapshotter_test.go
```diff
@@ -0,0 +1,68 @@
...
+
+	// Verify containerd snapshotter
+	plugins, ok := configMap["plugins"].(map[string]interface{})
+	assert.True(t, ok)
+	criPlugin, ok := plugins["io.containerd.grpc.v1.cri"].(map[string]interface{})
+	assert.True(t, ok)
+	containerdSettings, ok := criPlugin["containerd"].(map[string]interface{})
+	assert.True(t, ok)
+	assert.Equal(t, "soci", containerdSettings["snapshotter"], "incorrect snapshotter configuration")
+}
```
fletcherw: Can you add a test with both user-specific containerd config in `Conf.Spec.Containerd.Config` and the feature enabled to ensure that the three-way config merge works properly as well?

nodeadm/internal/api/types.go
```diff
@@ -122,4 +122,7 @@ type Feature string
 const (
 	// InstanceIdNodeName will use EC2 instance ID as node name
 	InstanceIdNodeName Feature = "InstanceIdNodeName"
+
+	// Allows more granular control for faster image pull
+	FastContainerImagePull Feature = "FastContainerImagePull"
```
fletcherw, line 127: If we migrate to containerd v2.1, will this flag continue to enable SOCI or is the plan to have it switch over to using the built-in parallel fetch?

nodeadm/internal/kubelet/config.go
```diff
@@ -294,6 +295,10 @@ func (ksc *kubeletConfig) withPodInfraContainerImage(cfg *api.NodeConfig, flags
 func (k *kubelet) GenerateKubeletConfig(cfg *api.NodeConfig) (*kubeletConfig, error) {
 	kubeletConfig := defaultKubeletSubConfig()
 
+	if api.IsFeatureEnabled(api.FastContainerImagePull, cfg.Spec.FeatureGates) {
```
fletcherw, line 296: nit: can you put this at the end of the function (line 318) rather than at the beginning? Since this is conditional config, it's distracting to have it first.

nodeadm/internal/api/featuregates.go
fletcherw: stale comment, please ignore.

nodeadm/internal/api/types.go
HusainZafar: This flag will continue to enable SOCI. Once all features are available in containerd, we have the option to deprecate this feature flag

nodeadm/internal/kubelet/config.go
HusainZafar: Fixed this to match the existing pattern

nodeadm/internal/containerd/snapshotter_test.go
HusainZafar: Added another test that verifies the three-way config merge

HusainZafar:
> Have you tested enabling the feature on an ami that includes the soci binary and confirmed that the pull works correctly? I know you had your previous test ami but I wasn't sure if that used this logic to generate the config.

Yes the AMI was generated with the latest changes that included the SOCI binary. I retested again with the latest changes. There are more configuration changes specifically for the SOCI settings, will update the PR with those changes today

nodeadm/internal/containerd/containerd_config_test.go
```diff
@@ -0,0 +1,112 @@
...
+	// Verify user config
+	registry, ok := criPlugin["registry"].(map[string]interface{})
+	assert.True(t, ok)
+
+	mirrors, ok := registry["mirrors"].(map[string]interface{})
+	assert.True(t, ok)
+
+	dockerMirrors, ok := mirrors["docker.io"].([]interface{})
+	assert.True(t, ok)
+	assert.Equal(t, "https://my-custom-mirror.example.com", dockerMirrors[0], "User config was not merged correctly with the containerd config")
```
fletcherw: nit: `assert.ElementsMatch([]{"https://my-custom-mirror.example.com"}, dockerMirrors, [...])`

(code untested but hopefully you get the idea)

nodeadm/internal/containerd/snapshotter/soci-snapshotter.config.toml
```diff
@@ -0,0 +1,26 @@
+debug = true
```
fletcherw: Is this the debug from https://github.com/awslabs/soci-snapshotter/blob/main/docs/config.md? 

If so this should probably be disabled: "Enables debugging for go-fuse in logs. This often emits sensitive data, so this should be false in production. Default: false."

nodeadm/internal/containerd/daemon.go
```diff
@@ -19,18 +19,27 @@ func NewContainerdDaemon(daemonManager daemon.DaemonManager) daemon.Daemon {
...
 }
 
-func (cd *containerd) EnsureRunning() error {
-	return cd.daemonManager.StartDaemon(ContainerdDaemonName)
+func (cd *containerd) EnsureRunning(cfg *api.NodeConfig) error {
+	if err := cd.daemonManager.StartDaemon(ContainerdDaemonName); err != nil {
+		return err
+	}
+	if api.IsFeatureEnabled(api.FastContainerImagePull, cfg.Spec.FeatureGates) {
+		return cd.daemonManager.StartDaemon("soci-snapshotter")
```
fletcherw: It looks like soci supports using a systemd socket file to start the snapshotter on demand (see https://github.com/awslabs/soci-snapshotter/pull/1378). Can we use that here to avoid running the snapshotter constantly in the background? You might need https://www.freedesktop.org/software/systemd/man/latest/systemd-socket-proxyd.html in order to have the snapshotter shut down when idle.

cartermckinnon: +1, image pulls can be pretty infrequent. If this feature will apply to all instance types, should we adjust our kube-reserved values when it's enabled? The snapshotter runs in the same cgroup as kubelet and containerd IIUC

nodeadm/internal/containerd/config.go
```diff
@@ -15,14 +15,21 @@ import (
...
+	configPerm                = 0644
 )
 
 var (
 	//go:embed config.template.toml
 	containerdConfigTemplateData string
 	containerdConfigTemplate     = template.Must(template.New(containerdConfigFile).Parse(containerdConfigTemplateData))
+
+	//go:embed snapshotter/soci-snapshotter.containerd-config.toml
+	containerdConfigSociSnapshotterData string
```
cartermckinnon: You only use this as a `[]byte` AFAICT, so you could use that type here instead of `string`.

I'm not sure declaring this in a separate file and merging it in is optimal. You could add this to the existing `config.template.toml` and wrap it with `{{if .UseSOCISnapshotter}} ... {{end}}`. The merging operation isn't necessarily cheap, so it'd be nice to avoid it

templates/al2023/runtime/rootfs/etc/systemd/system/soci-snapshotter.service
```diff
@@ -0,0 +1,14 @@
+[Unit]
+Description=soci snapshotter containerd plugin
+Documentation=https://github.com/awslabs/soci-snapshotter
+After=network.target
+Before=containerd.service
+
+[Service]
+Type=notify
+ExecStart=/usr/local/bin/soci-snapshotter-grpc --address fd://
```
cartermckinnon: What does `--address fd://` mean?

nodeadm/internal/containerd/containerd_config_test.go
HusainZafar: Updated to use ElementsMatch

nodeadm/internal/containerd/snapshotter/soci-snapshotter.config.toml
HusainZafar: Good callout. Switched to false

nodeadm/internal/containerd/config.go
HusainZafar: Makes sense. Updated the exiting config template to include the SOCI changes conditionally

templates/al2023/runtime/rootfs/etc/systemd/system/soci-snapshotter.service
HusainZafar: `soci-snapshotter-grpc` accepts a flag `--address` for the snapshotter socket. `fd://` indicates the service is using the file descriptor passed by systemd socket activation

nodeadm/internal/containerd/daemon.go
HusainZafar: Thanks for the callout. Updated to take advantage of socket activation

nodeadm/internal/containerd/daemon.go
```diff
@@ -19,18 +20,27 @@ func NewContainerdDaemon(daemonManager daemon.DaemonManager) daemon.Daemon {
...
 }
 
-func (cd *containerd) EnsureRunning() error {
-	return cd.daemonManager.StartDaemon(ContainerdDaemonName)
+func (cd *containerd) EnsureRunning(cfg *api.NodeConfig) error {
+	if err := cd.daemonManager.StartDaemon(ContainerdDaemonName); err != nil {
+		return err
+	}
+	if api.IsFeatureEnabled(api.FastContainerImagePull, cfg.Spec.FeatureGates) {
+		return cd.daemonManager.StartDaemon(SociSnapshotterSocketName)
```
fletcherw: do you need to start the soci daemon if you have the `.socket` systemd config set up?

templates/al2023/runtime/rootfs/etc/systemd/system/soci-snapshotter.service
fletcherw: does this shut down automatically when idle?

HusainZafar: Added a new line to indicate that restarts are handled by the socket
```
Restart=no
```

Also confirmed that socket activation is working:
```
[root@host bin]# sudo systemctl stop soci-snapshotter.service
Warning: Stopping soci-snapshotter.service, but it can still be activated by:
  soci-snapshotter.socket
```

Based on my checks, it appears that the service will always have active connections from containerd so it will never go idle

nodeadm/internal/containerd/daemon.go
HusainZafar: Based on the current approach, the daemons are started via explicit daemon start calls in our code. I would like to match the same with the soci snapshotter so that any issues with it appear in the nodeadm-run logs

HusainZafar: Updated to let systemd handle the soci snapshotter daemon start

nodeadm/api/v1alpha1/nodeconfig_types.go
```diff
@@ -129,10 +129,13 @@ const (
...
 // Feature specifies which feature gate should be toggled
-// +kubebuilder:validation:Enum={InstanceIdNodeName}
+// +kubebuilder:validation:Enum={InstanceIdNodeName,FastContainerImagePull}
 type Feature string
 
 const (
 	// InstanceIdNodeName will use EC2 instance ID as node name
 	InstanceIdNodeName Feature = "InstanceIdNodeName"
+
+	// Allows more granular control for faster image pull
```
cartermckinnon:
```suggestion
	// FastContainerImagePull reduces image pull time.
	// Note that this consumes additional CPU and memory, and may not be suitable for all workloads.
```

nodeadm/internal/containerd/snapshotter/soci-snapshotter.config.toml
```diff
@@ -0,0 +1,21 @@
...
+[content_store]
+type = 'containerd'
+
+[cri_keychain]
+enable_keychain = true
+image_service_path = '/run/containerd/containerd.sock'
+
+[pull_modes.parallel_pull_unpack]
+    enable = true
+    concurrent_download_chunk_size = ""
```
cartermckinnon, line 12: Why is this blank?

nodeadm/internal/daemon/systemd.go
```diff
@@ -105,6 +106,10 @@ func (m *systemdDaemonManager) Close() {
 }
 
 func getServiceUnitName(name string) string {
+	if strings.HasSuffix(name, ".service") ||
+		strings.HasSuffix(name, ".socket") {
+		return name
+	}
```
cartermckinnon: This doesn't seem necessary, and conflicts with the name/purpose of the function

templates/al2023/runtime/rootfs/etc/systemd/system/kubelet.service
```diff
@@ -1,8 +1,8 @@
 [Unit]
 Description=Kubernetes Kubelet
 Documentation=https://github.com/kubernetes/kubernetes
-After=containerd.service
-Wants=containerd.service
+After=containerd.service soci-snapshotter.socket
+Wants=containerd.service soci-snapshotter.socket
```
cartermckinnon: This will cause the snapshotter to be started even if the feature isn't enabled, right?

templates/al2023/runtime/rootfs/etc/systemd/system/soci-snapshotter.socket
cartermckinnon, line 1: what does this unit actually do?

templates/al2023/provisioners/install-worker.sh
```diff
@@ -140,6 +140,33 @@ sudo dnf versionlock containerd-*
 
 sudo systemctl enable ebs-initialize-bin@containerd
 
+###############################################################################
+### SOCI Snapshotter setup ##########################################################
+###############################################################################
+function install-soci-snapshotter() {
+  PARTITION=$(imds /latest/meta-data/services/partition)
+  SUPPORTED_PARTITIONS=("aws" "aws-us-gov")
```
cartermckinnon: if we can only install this in these two partitions, what happens if you open the feature gate in an unsupported partition?

templates/al2023/runtime/rootfs/etc/systemd/system/soci-snapshotter.socket
fletcherw: My recollection is that we discovered containerd was going to maintain a persistent connection to the snapshotter socket anyways, so using .socket activation didn't provide much value. I think we can remove this unit, which lets us also remove the change to `systemd.go`

EDIT: after further discussion this is still good to keep around so that we can implicitly start soci and don't block the node from joining the cluster if soci fails to start.

templates/al2023/provisioners/install-worker.sh
fletcherw: also wondering this -- can we force-disable the flag in other environments? seems like if you enabled it outside of those envs, you'd get a broken configuration since your containerd config would still be trying to pull from soci.

nodeadm/internal/containerd/config.go
```diff
@@ -15,55 +15,79 @@ import (
...
+	configPerm                = 0644
 )
 
 var (
 	//go:embed config.template.toml
 	containerdConfigTemplateData string
 	containerdConfigTemplate     = template.Must(template.New(containerdConfigFile).Parse(containerdConfigTemplateData))
+
+	//go:embed snapshotter/soci-snapshotter.config.toml
+	sociSnapshotterTemplateData string
```
ndbaker1:
```suggestion
	sociSnapshotterTemplateData []byte
```

templates/al2023/runtime/rootfs/etc/systemd/system/kubelet.service
fletcherw: This would just start the socket unit which just creates the socket file, doesn't actually start soci until something connects to the socket.

templates/al2023/provisioners/install-worker.sh
HusainZafar: In other environments where the soci snapshotter is not setup, the node fails to join. This is expected because the kubelet is set to use soci-snapshotter as the image service

Note: The partition specific logic will be removed in the next revision

nodeadm/internal/containerd/snapshotter/soci-snapshotter.config.toml
HusainZafar: This field is of the type string and the default value is blank "". This indicates unbounded chunk size

cartermckinnon: What does "unbounded" mean in this case?

nodeadm/internal/daemon/systemd.go
cartermckinnon: I think you can remove this? What's it for?

templates/al2023/runtime/rootfs/etc/systemd/system/kubelet.service
cartermckinnon: Why is this set on `kubelet`, though?

templates/al2023/runtime/rootfs/etc/systemd/system/soci-snapshotter.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=soci snapshotter containerd plugin
+Documentation=https://github.com/awslabs/soci-snapshotter
+Before=containerd.service kubelet.service
```
cartermckinnon: what is this dependency actually doing? if we're relying on socket activation, do we need to declare that this should start before kubelet/containerd?

nodeadm/internal/containerd/snapshotter/soci-snapshotter.config.toml
HusainZafar: [ref](https://github.com/awslabs/soci-snapshotter/blob/790a445854406da5b174b267104be43a01bd37ec/config/defaults.go#L128)

> When unbounded, full layer is downloaded in one request

This is the default that matches the default containerd behavior

nodeadm/internal/daemon/systemd.go
HusainZafar: I will remove it in the next revision, can create another function in future for getting the socket name, if required

templates/al2023/runtime/rootfs/etc/systemd/system/soci-snapshotter.service
HusainZafar: Removed the dependency. Right, it is no longer needed with socket activation

nodeadm/internal/containerd/snapshotter/soci-snapshotter.config.toml
cartermckinnon: Don't we want to increase that? The goal is to leverage the snapshotter to download the layer in chunks, in parallel, right?

templates/al2023/runtime/rootfs/etc/systemd/system/soci-snapshotter.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=soci snapshotter containerd plugin
+Documentation=https://github.com/awslabs/soci-snapshotter
+
+[Service]
+Type=notify
+ExecStart=/usr/bin/soci-snapshotter-grpc --address fd://
```
cartermckinnon, line 7: For my future self: the `--address fd://` option here tells the snapshotter to look for file descriptors passed to the process by `systemd` as part of the socket activation flow: https://github.com/coreos/go-systemd/blob/7d375ecc2b092916968b5601f74cca28a8de45dd/activation/files_unix.go#L39

`systemd` will own the "listening" socket, and will create a "proxy" fd to route traffic to the backend service after it's started.

nodeadm/internal/containerd/snapshotter/soci-snapshotter.config.toml
HusainZafar: We wanted to keep the defaults as containerd defaults. It can be overridden via userdata scripts to reasonable values such as "8MB", etc that works best for the instance type, image size and number of image layers. Changing from unbounded to a smaller chunk size will help when `max_concurrent_downloads_per_image` is set appropriately

wwvela:
/ci
+build binary_bucket_name=amazon-wesley-beta-ami containerd_version=1.7.27 install_containerd_from_s3=true arch=x86_64 enable_accelerator=nvidia

cartermckinnon:
Superseded by #2352

PR #2312, opened 2025-06-30
feat(al2023): install kernel headers by default

cartermckinnon:
**Description of changes:**

It's handy to have these installed on all variants, they're needed by some things like security and telemetry agents (Falco, DataDog, etc.)

### DIFF
--- a/templates/al2023/provisioners/install-worker.sh
+++ b/templates/al2023/provisioners/install-worker.sh
@@ -46,9 +46,6 @@ fi
 ### Packages ###################################################################
 ################################################################################
 
-# Update the OS to begin with to catch up to the latest packages.
-sudo dnf update -y
-
 # Install necessary packages
 sudo dnf install -y \
   aws-cfn-bootstrap \
@@ -65,6 +62,17 @@ sudo dnf install -y \
   mdadm \
   pigz \
   python3-dnf-plugin-versionlock
+
+# we need to specify the fully-formed version of these packages
+# because the 6.1 and 6.12 branches share a namespace
+sudo dnf install -y \
+  kernel-headers-$(uname -r) \
+  kernel-devel-$(uname -r)
+# versionlock them for the same reason
+sudo dnf versionlock kernel-headers kernel-devel
+# versionlock the kernel to make sure it matches
+# pattern has to match "kernel-6.1..." and "kernel6.12-6.12..."
+sudo dnf versionlock kernel*-$(uname -r)
 
 ################################################################################
 ### Networking #################################################################


cartermckinnon:
Versions look good to me, except that the base AL2023 AMI seems to be including the "wrong" versions of `kernel-libbpf` and `kernel-tools` (it has the 6.12 version of these packages). I'll reach out to Amazon Linux about that.
```
        "kernel": "6.1.141-155.222.amzn2023",
        "kernel-devel": "6.1.141-155.222.amzn2023",
        "kernel-headers": "6.1.141-155.222.amzn2023",
        "kernel-libbpf": "6.12.31-35.92.amzn2023",
        "kernel-livepatch-repo-s3": "2023.7.20250623-0.amzn2023",
        "kernel-srpm-macros": "1.0-14.amzn2023.0.3",
```

wwvela (approved):

templates/al2023/provisioners/install-worker.sh
```diff
@@ -66,6 +63,14 @@ sudo dnf install -y \
   pigz \
   python3-dnf-plugin-versionlock
 
+# we need to specify the fully-formed version of these packages
+# because the 6.1 and 6.12 branches share a namespace
+sudo dnf install -y \
+  kernel-headers-$(uname -r) \
+  kernel-devel-$(uname -r)
+# versionlock them for the same reason
+sudo dnf versionlock kernel-headers kernel-devel
```
HusainZafar, line 72: Does this version lock need a suffix or does this work as is?

mselim00, line 72: should we version lock the kernel too to ensure that consistency

cartermckinnon: it works as you'd expect, the currently installed version of the package will end up in the versionlock list.

cartermckinnon: yeah I'll add this

templates/al2023/provisioners/install-worker.sh
```diff
@@ -66,6 +63,14 @@ sudo dnf install -y \
   pigz \
   python3-dnf-plugin-versionlock
 
+# we need to specify the fully-formed version of these packages
+# because the 6.1 and 6.12 branches share a namespace
+sudo dnf install -y \
+  kernel-headers-$(uname -r) \
+  kernel-devel-$(uname -r)
+# versionlock them for the same reason, along with the kernel to ensure things match
+sudo dnf versionlock kernel kernel-headers kernel-devel
```
mselim00: the package for the 6.12 kernel is actually called kernel6.12, so I think this would versionlock the 6.1 kernel

might have to do something creative like
```
dnf versionlock kernel*-$(uname -r)
```

mselim00: just hijacking on this to also add that we could remove the installs from install-(neuron|nvidia)-driver.sh, but I think dnf will just exit 0 with nothing to do atm so it shouldn't cause an issue

cartermckinnon: yeah had the same though re: 6.12 after pushing, I'm working on a rev

kalavt:
this might result and relevant to calico network CNI
eBPF map creation fails with “Invalid argument (-22)”
https://github.com/projectcalico/calico/issues/10734

ndbaker1:
superceded by https://github.com/awslabs/amazon-eks-ami/pull/2345

Issue #2282, opened 2025-06-07
Node log collector script - add logs for requests to ECR private endpoints.


david-a-aws:
**What would you like to be added**:
For EKS node log collector script, add a test to make a network request to ECR private.

e.g. `curl -v https://<account_id>.dkr.ecr.<region>.amazonaws.com` and `curl -v https://api.ecr.<region>.amazonaws.com`

**Why is this needed**:
Recently had an issue where this was the root cause for node failing to join cluster, but none of the other logs mentioned timeout with reaching ECR private to get the sandbox-image/pause image for kubelet to start.

david-a-aws:
I can create a PR for this if the idea is approved

mselim00:
Sounds helpful, curious on your thoughts for the specific implementation though. I think the main challenge with this is figuring out which registry to use for the check, since only access to the correct regional EKS container registry is expected. Could perhaps parse it out from https://docs.aws.amazon.com/eks/latest/userguide/add-ons-images.html but seems less than ideal

PR #2319, opened 2025-07-06
feat(log-collector): log ec2 regional endpoint reachability

mselim00:
**Issue #, if available:**

**Description of changes:**

Reachability to the EC2 regional endpoint is essential for various core tasks that run on the instance, e.g. for IPAMD to manage ENIs or for nodeadm/bootstrap.sh to find out the instance's private DNS name. This adds a check to confirm that the endpoint can be reached at log collection time. This gives some insight on whether an earlier error encountered reaching the EC2 endpoint was a intermittent/one-off or sustained, since this would reflect the connectivity at log collection time.

**Testing Done**


log-collector-script/linux/eks-log-collector.sh
```diff
@@ -650,6 +663,9 @@ get_networking_info() {
   # test some network connectivity
   timeout 75 ping -A -c 10 amazon.com > "${COLLECT_DIR}"/networking/ping_amazon.com.txt
   timeout 75 ping -A -c 10 public.ecr.aws > "${COLLECT_DIR}"/networking/ping_public.ecr.aws.txt
+  if [[ -n "${REGION:-}" ]]; then
+    timeout 75 ping -A -c 10 "ec2.${REGION}.amazonaws.com" > "${COLLECT_DIR}/networking/ping_ec2.${REGION}.amazonaws.com.txt"
```
ndbaker1, line 667: any other endpoints we may as well gather connectivity on as well?

mselim00: I think a private ECR endpoint as suggested in #2282 would be nice and the only other core endpoint I can think of, it's just a bit trickier to figure out what registry to use for the test

PR #2321, opened 2025-07-07
feat(nodeadm): support for nodeadm version display

guessi:
**Issue #, if available:**

**Description of changes:**

By current design, there have no version support for `nodeadm`, so it would inherit the setup from `integrii/flaggy`

```golang
// defaultVersion is applied to parsers when they are created
const defaultVersion = "0.0.0"
```

* https://github.com/integrii/flaggy/blob/v1.5.2/flaggy.go#L24-L25

With this PR change, not only make it possible to identify which version it is about, but also help customer to identify which `nodeadm` it is running with.

Cross ref: https://github.com/aws/eks-hybrid/issues/639 and https://github.com/aws/eks-hybrid/pull/640

**Testing Done**

```bash
$ make build
```

```bash
$ ./_bin/nodeadm --version
Version: v20250627-7-ged80709a
```


nodeadm/Makefile
```diff
@@ -5,6 +5,10 @@ else
 GOBIN=$(shell go env GOBIN)
 endif
 
+# Version from git tags
+GITTAGVERSION  := $(shell git describe --tags --abbrev=8)
+LDFLAGS := -X 'main.gitTagVersion=$(GITTAGVERSION)'
```
ndbaker1:
```suggestion
# Version from git tags
VERSION ?= $(shell git describe --tags --abbrev=8 --always)
LDFLAGS ?= -X 'main.gitTagVersion=$(VERSION)'
```

ndbaker1 (commented):
that's a nice idea, ill see if its plumbed through correctly today

guessi:
Notice merge conflicts, let me fix it.

nodeadm/cmd/nodeadm/main.go
```diff
@@ -9,9 +9,14 @@ import (
 	"github.com/awslabs/amazon-eks-ami/nodeadm/internal/cli"
 )
 
+var (
+	gitTagVersion = "0.0.0"
```
cartermckinnon:
```suggestion
	version = "0.0.0"
```

nodeadm/Makefile
```diff
@@ -5,6 +5,10 @@ else
 GOBIN=$(shell go env GOBIN)
 endif
 
+# Version from git tags
+VERSION ?= $(shell git describe --tags --abbrev=8 --always)
```
cartermckinnon: Unfortunately, this won't work with our current setup. The `nodeadm` build happens in a container as a part of the AMI build, and there is no `.git` directory in that workspace. Additionally, the `git` tag for an release does not actually exist when the AMI build happens. We could wire something up, but it would have to happen in our internal release tooling.

We could do this as a pre-req without changing behavior, and then set the version in a future release:
```suggestion
VERSION ?= 0.0.0
```

guessi:
# Test after rebased

### With no tag

```bash
% make build
GOOS= go build -ldflags="-X 'main.gitTagVersion=v20250715-6-g6534ca01'" -o <redacted>/_bin/ ./cmd/...
```

```bash
% ./_bin/nodeadm --version
Version: v20250715-6-g6534ca01
```

### With tag (for test only, not actually a release)

```bash
% git tag v20250717
% git describe --exact-match --tags
v20250717
```

```bash
% make build
GOOS= go build -ldflags="-X 'main.gitTagVersion=v20250717'" -o <redacted>/_bin/ ./cmd/...
```

```bash
% ./_bin/nodeadm --version
Version: v20250717
```

nodeadm/Makefile
```diff
@@ -5,6 +5,10 @@ else
...
+LDFLAGS ?= -X 'main.gitTagVersion=$(VERSION)'
```
cartermckinnon:
```suggestion
LDFLAGS ?= -X 'main.version=$(VERSION)'
```

nodeadm/cmd/nodeadm/main.go
```diff
@@ -9,9 +9,14 @@ import (
...
+)
+
 func main() {
 	flaggy.SetName("nodeadm")
 	flaggy.SetDescription("From zero to Node faster than you can say Elastic Kubernetes Service")
+	flaggy.SetVersion(gitTagVersion)
```
cartermckinnon:
```suggestion
	flaggy.SetVersion(version)
```

nodeadm/Makefile
guessi: Umm... interesting, ~sounds like we need somethings like sed/awk/...~

Never mind, I just made updates, and I can see could be done by...

```bash
% VERSION=1.2.3 make build
GOOS= go build -ldflags="-X 'main.version=1.2.3'" -o <redacted>/_bin/ ./cmd/...`
```

```bash
% ./_bin/nodeadm --version
Version: 1.2.3
```

guessi:
Just made an update to support both `nodeadm` and `nodeadm-internal`.

```bash
% VERSION=v1.99.88 make build
```
```bash
% ./_bin/nodeadm --version
Version: v1.99.88
```
```bash
% ./_bin/nodeadm-internal --version
Version: v1.99.88
```

nodeadm/Makefile
cartermckinnon: Yeah exactly, we can pass the correct version with our internal tooling

PR #2322, opened 2025-07-10
fix(nodeadm): increase retries for AWS API calls

cartermckinnon:
**Description of changes:**

The default retryer only makes 3 attempts, with a max delay of 20 seconds between attempts (per the exponential delay policy). Connections will time out after 30 seconds. This PR increases the retries significantly, because a failure in these API calls leads to a terminal node.

We do not override the entire `Retryer`, and instead only override the max attempt count. This allows a user to override the value using the standard environment variable `AWS_MAX_ATTEMPTS`, if push comes to shove.

### DIFF
--- a/nodeadm/cmd/nodeadm/init/init.go
+++ b/nodeadm/cmd/nodeadm/init/init.go
@@ -168,6 +168,15 @@ func enrichConfig(log *zap.Logger, cfg *api.NodeConfig) error {
 	if err != nil {
 		return err
 	}
+	if awsConfig.RetryMaxAttempts == 0 {
+		// use a very generous retry policy to accomodate delays in network readiness
+		// we only specify the max attempts if it is unset by the user
+		// so it's possible to override with the AWS_MAX_ATTEMPTS environment variable.
+		// NOTE that this is the number of attempts that will be made in a blocking fashion
+		// i.e. an SDK client.ExampleAPICall() will not return until these attempts are exhausted
+		// we'll give up after approximately 10 minutes
+		awsConfig.RetryMaxAttempts = 30
+	}
 	instanceDetails, err := api.GetInstanceDetails(context.TODO(), cfg.Spec.FeatureGates, ec2.NewFromConfig(awsConfig))
 	if err != nil {
 		return err
--- a/nodeadm/test/e2e/helpers.sh
+++ b/nodeadm/test/e2e/helpers.sh
@@ -100,17 +100,64 @@ function wait::dbus-ready() {
   wait::path-exists /run/systemd/private
 }
 
-function mock::aws() {
-  local CONFIG_PATH=${1:-/etc/aemm-default-config.json}
-  [ "${IMDS_MOCK_ONLY_CONFIGURE:-}" = "true" ] || imds-mock --config-file $CONFIG_PATH &
-  export AWS_EC2_METADATA_SERVICE_ENDPOINT=http://localhost:1338
-  [ "${AWS_MOCK_ONLY_CONFIGURE:-}" = "true" ] || $HOME/.local/bin/moto_server -p5000 &
-  export AWS_ACCESS_KEY_ID='testing'
-  export AWS_SECRET_ACCESS_KEY='testing'
-  export AWS_SECURITY_TOKEN='testing'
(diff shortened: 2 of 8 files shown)


ndbaker1 (approved):
sgtm, though we basically stop allowing users to provide an override with this right? not that providing an override is exactly ergonomic/idiomatic with nodeadm, but it's locking users into our preferences a little tighter.

cartermckinnon:
@ndbaker1 updated to preserve the override path.

nodeadm/test/e2e/helpers.sh
```diff
@@ -100,17 +100,60 @@ function wait::dbus-ready() {
...
-  export AWS_SESSION_TOKEN='testing'
-  export AWS_REGION=us-east-1
-  export AWS_ENDPOINT_URL=http://localhost:5000
+  wait::server-responding localhost 5000 10
   # ensure that our instance exists in the API
   aws ec2 run-instances
 }
+
+function mock::connection-timeout-server() {
+  if [ "$#" -ne 1 ]; then
```
ndbaker1, line 144: ig that's the accurate end effect but should we just call it something related to `mock::packet-drops` instead of connection timeout?

nodeadm/test/e2e/helpers.sh
```diff
@@ -100,17 +100,60 @@ function wait::dbus-ready() {
...
+function mock::imds() {
   local CONFIG_PATH=${1:-/etc/aemm-default-config.json}
-  [ "${IMDS_MOCK_ONLY_CONFIGURE:-}" = "true" ] || imds-mock --config-file $CONFIG_PATH &
-  export AWS_EC2_METADATA_SERVICE_ENDPOINT=http://localhost:1338
+  imds-mock --config-file $CONFIG_PATH &
+  wait::server-responding localhost 1338 10
+}
+
+function mock::aws() {
+  [ "${IMDS_MOCK_ONLY_CONFIGURE:-}" = "true" ] || mock::imds ${1:-}
```
ndbaker1: updating all the tests to explicitly call `mock::imds` and `mock::aws` might not be so bad

cartermckinnon: 😛

nodeadm/test/e2e/helpers.sh
cartermckinnon: eh I don't really love either

nodeadm/cmd/nodeadm/init/init.go
```diff
@@ -168,6 +168,15 @@ func enrichConfig(log *zap.Logger, cfg *api.NodeConfig) error {
...
 		return err
 	}
+	if awsConfig.RetryMaxAttempts == 0 {
+		// use a very generous retry policy to accomodate delays in network readiness
+		// we only specify the max attempts if it is unset by the user
+		// so it's possible to override with the AWS_MAX_ATTEMPTS environment variable.
+		// NOTE that this is the number of attempts that will be made in a blocking fashion
+		// i.e. an SDK client.ExampleAPICall() will not return until these attempts are exhausted
+		// we'll give up after approximately 10 minutes
+		awsConfig.RetryMaxAttempts = 30
```
cartermckinnon, line 178: The math for this is:
1. Max delay between attempts is 20 seconds.
2. Delay is exponential, starting at 2; but there's jitter applied. So, the actual delay is 0-100% of the quantity.

The sequence looks like:
| Attempt | Delay (s) |
|--------|--------|
| 1 | 2 |
| 2 | 4 |
| 3 | 8 |
| 4 | 16 |
| 5 | 20 | 

Any attempt after the fourth attempt will have a delay of 20 seconds. Jitter is not applied once the max delay has been reached. We'll just treat each attempt as taking 20 seconds, the first 4 attempts don't change much.

So, 10 minutes (600 seconds) / 20 seconds = 30 attempts.

cartermckinnon: Note that if the connection times out, the delay between each attempt will be 30 seconds longer. That's where our custom waiter's timeout comes in -- we will still give up after about 10 minutes, even if we don't make all 30 attempts.

tzneal (approved):
lgtm

PR #2325, opened 2025-07-11
feat(nodeadm): add feature gate to limit max pods value on memory

mselim00:
**Issue #, if available:**

**Description of changes:**

Adds a new feature gate called `LimitMaxPods`, which, if enabled, will limit the max pods value by the amount of memory it will require in kubeReserved. Implicitly globally limits the max pods value to 169 because of the considered memory range on the system.

Also adds a new binary, `nodeadm-docs`, as a helper for generating doc references using nodeadm internals. One command is added int here to generate a table to compare max pods values across configurations. There is also a new CI workflow in this change to execute that binary weekly to make sure the list is in-sync for new instance types.

**Testing Done**
TBD - will add unit testing here


mselim00:
Quick link for the generated table showing max pods limits and values: https://github.com/mselim00/amazon-eks-ami/blob/limit-max-pods/nodeadm/doc/feature-gates/references/maxpods.md

nodeadm/internal/system/resources.go
```diff
@@ -81,6 +82,16 @@ func GetMilliNumCores() (int, error) {
...
 
+// TODO before merging: double check this always returns total memory for the entire system
+func GetTotalMemoryMiB() uint64 {
+	var sysinfo syscall.Sysinfo_t
+	err := syscall.Sysinfo(&sysinfo)
+	if err != nil {
+		return 0
+	}
+	return (uint64(sysinfo.Totalram) * uint64(sysinfo.Unit)) / 1048576
+}
```
mselim00: I tested this out with
```
yum install libcgroup-tools
sudo cgcreate -g memory:testgroup
sudo cgset -r memory.limit_in_bytes=500M testgroup
sudo cgexec -g memory:testgroup mem
```
where `mem` was a binary that just prints the output of this function. On a `t3.medium` that showed `3827` which is the same value as `MemTotal` in `/proc/meminfo` (times 1024, since meminfo is in KiB). So it should always be fine.

One caveat though is the discrepancy between that value and the 4096 MiB expected on a `t3.medium` (e.g. per DescribeInstanceTypes), which is due to reserved memory for the kernel:
>  MemTotal — Total amount of usable RAM, in kibibytes, which is physical RAM minus a number of reserved bits and the kernel binary code. 

https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/6/html/deployment_guide/s2-proc-meminfo

The current max pods values are static for an instance type, so it might a bit of a peculiar change in behavior for that value to decrease if the kernel binary grows. The table in the docs will also always show higher values because of this

nodeadm/internal/kubelet/maxpods/maxpods.go
```diff
@@ -0,0 +1,98 @@
...
+
+// GetMaxPodsLimitForMemoryMiB returns a max pods limit value based on memory reserved
+func GetMaxPodsLimitForMemoryMiB(memoryMiB uint64) int32 {
+	percentsPerRange := []int{25, 20, 10, 6}
+	memoryBands := []int{512, 4096, 8192, 16384, 131072}
+	// #nosec G115 // value is limited by pre-defined bands and percents, so we know this cannot overflow
+	memoryFromBands := int32(util.GetAmountFromPercentageBands(percentsPerRange, memoryBands, memoryMiB))
+	maxReservable := 299 + memoryFromBands
+	return projectMaxPodsFromMemoryMiBReserved(maxReservable)
+}
```
mselim00: This is the same formula GKE uses for their reservations https://cloud.google.com/kubernetes-engine/docs/concepts/plan-node-sizes. There's two minor tweaks here:
1. The lower bound is bumped from `255` for nodes with less than `1 GiB` to `299` for nodes with less than `512 MiB`, which guarantees at least 4 pods on any viable instance type 
2. The upper bound of memory considered is capped at `128 GiB`, whereas GKE would continue reserving 2% of any leftover memory after this band.

These tweaks are mainly because of a difference in approaches we handle reservations for the max pods and how GKE does. GKE discourages changes from the upstream recommended default of `110`, and does not adjust memory reservations based on max pods. 

Several years back, we made an optimization to reduce our memory reservations by dynamically setting them as a function of max pods instead of memory (ref: https://github.com/awslabs/amazon-eks-ami/pull/367). This has historically just worked because of an implicit relationship that exists between allocatable IPs and memory available on a given instance type. The one gap though, which this PR tries to address, is that disproportionate increases in allocatable IPs (e.g. by using prefix delegation) for an instance can cause unreasonably large max pods values without available memory to back it.

Importantly, this function is only used as a cap on the max pods value we will pass to kubelet, never directly as the value. The value will be set as a minimum of the max pods value calculated from allocatable IPs and this limit, since we still need to ensure each pod can get an IP. This means that we could also continue considering memory beyond the `128 GiB` with minimal immediate impact on the reserved memory. I left this out because for instance types like `u7in-32tb.224xlarge` with a lot of memory, that would mean we can 'support' up to `30,643` pods (the applied value would be set to around ~6,000 if adjusted for prefix delegation), which unnecessary increases our memory reservations without practical gain. This change creates an effective hard ceiling at `861`, which is higher than the current allocatable value on almost every instance type.

mselim00:
Superseded by #2446

PR #2336, opened 2025-07-17
feat(nodeadm): add custom image credential provider support

adrien-f:
**Issue #, if available:** N/A

**Description of changes:**

In a process where extra credential providers binaries are added, this enable users, instead of overriding the `image-credential-provider-config` flag themselves, to pass a custom CredentialProviderConfig directly.

Let me know what you think,

**Testing Done**

Adding extra tests to verify what gets written as well as an extra e2e test.

### DIFF
--- a/nodeadm/api/v1alpha1/nodeconfig_types.go
+++ b/nodeadm/api/v1alpha1/nodeconfig_types.go
@@ -68,6 +68,12 @@ type KubeletOptions struct {
 	// Flags are [command-line `kubelet` arguments](https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/).
 	// that will be appended to the defaults.
 	Flags []string `json:"flags,omitempty"`
+
+	// ImageCredentialProviderConfig is a custom Go template to configure
+	// the image credential provider. If not set, a default template is used.
+	// The template is rendered with the ConfigApiVersion, EcrProviderName
+	// and ProviderApiVersion variables.
+	ImageCredentialProviderConfig string `json:"imageCredentialProviderConfig,omitempty"`
 }
 
 // ContainerdOptions are additional parameters passed to `containerd`.
--- a/nodeadm/doc/examples.md
+++ b/nodeadm/doc/examples.md
@@ -132,3 +132,48 @@ spec:
             soft: 1024
             hard: 1024
 ```
+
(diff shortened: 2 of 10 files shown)


nodeadm/api/v1alpha1/nodeconfig_types.go
```diff
@@ -68,6 +68,12 @@ type KubeletOptions struct {
 	// Flags are [command-line `kubelet` arguments](https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/).
 	// that will be appended to the defaults.
 	Flags []string `json:"flags,omitempty"`
+
+	// ImageCredentialProviderConfig is a custom Go template to configure
+	// the image credential provider. If not set, a default template is used.
+	// The template is rendered with the ConfigApiVersion, EcrProviderName
+	// and ProviderApiVersion variables.
+	ImageCredentialProviderConfig string `json:"imageCredentialProviderConfig,omitempty"`
```
cartermckinnon, line 76: I think we'll want to use stronger typing than a go template, could we import `CredentialProvider`: https://kubernetes.io/docs/reference/config-api/kubelet-config.v1/#kubelet-config-k8s-io-v1-CredentialProvider

Or define an equivalent type if we need to handle multiple `apiVersion`-s of `CredentialProvider`?

nodeadm/doc/examples.md
```diff
@@ -132,3 +132,48 @@ spec:
...
+              "*.dkr.ecr-fips.*.amazonaws.com",
+              "*.dkr-ecr-fips.*.on.aws",
+              "*.dkr.ecr.*.c2s.ic.gov",
+              "*.dkr.ecr.*.sc2s.sgov.gov",
+              "*.dkr.ecr.*.cloud.adc-e.uk",
+              "*.dkr.ecr.*.csp.hci.ic.gov"
+            ],
+            "defaultCacheDuration": "12h",
+            "apiVersion": "{{.ProviderApiVersion}}"
+          },
```
cartermckinnon, line 170: I think we'll want to merge this into the config so users don't have to replicate it themselves; is there a reason you wouldn't want this to be injected?

cartermckinnon (commented):
Great idea, thanks for putting this together!

PR #2339, opened 2025-07-23
fix: install containerd 1.7 with CVE-2025-22874 and CVE-2025-4673 patches from s3

wwvela:
**Issue #, if available:**
* install containerd 1.7 with CVE-2025-22874 and CVE-2025-4673 patches from s3

### DIFF
--- a/templates/al2/provisioners/install-worker.sh
+++ b/templates/al2/provisioners/install-worker.sh
@@ -149,7 +149,20 @@ sudo yum install -y runc-${RUNC_VERSION}
 sudo yum versionlock runc-*
 
 # install containerd and lock version
-sudo yum install -y containerd-${CONTAINERD_VERSION}
+function rpm_install() {
+  local RPMS=($@)
+  echo "Pulling and installing local rpms from s3 bucket"
+  for RPM in "${RPMS[@]}"; do
+    aws s3 cp --region ${BINARY_BUCKET_REGION} s3://${BINARY_BUCKET_NAME}/rpms/${RPM} ${WORKING_DIR}/${RPM}
+    sudo yum localinstall -y ${WORKING_DIR}/${RPM}
+  done
+}
+
+if [[ "$CONTAINERD_VERSION" == 1.* ]]; then
+  rpm_install "containerd-1.7.27-1.amzn2.0.4.$(uname -m).rpm"
+else
+  sudo yum install -y containerd-${CONTAINERD_VERSION}
+fi
 sudo yum versionlock containerd-*
 
 # install cri-tools for crictl, needed to interact with containerd's CRI server
--- a/templates/al2023/provisioners/install-worker.sh
+++ b/templates/al2023/provisioners/install-worker.sh
@@ -133,9 +133,21 @@ fi
 ###############################################################################
 ### Containerd setup ##########################################################
 ###############################################################################
+function rpm_install() {
+  local RPMS=($@)
+  echo "Pulling and installing local rpms from s3 bucket"
+  for RPM in "${RPMS[@]}"; do
+    aws s3 cp --region ${BINARY_BUCKET_REGION} s3://${BINARY_BUCKET_NAME}/rpms/${RPM} ${WORKING_DIR}/${RPM}
+    sudo dnf localinstall -y ${WORKING_DIR}/${RPM}
+  done
+}
 
 sudo dnf install -y runc-${RUNC_VERSION}
-sudo dnf install -y containerd-${CONTAINERD_VERSION}
+if [[ "$CONTAINERD_VERSION" == 1.* ]]; then
+  rpm_install "containerd-1.7.27-1.amzn2023.0.4.$(uname -m).rpm"
+else
+  sudo dnf install -y containerd-${CONTAINERD_VERSION}
+fi
 sudo dnf versionlock containerd-*
 
 sudo systemctl enable ebs-initialize-bin@containerd


templates/al2/provisioners/install-worker.sh
```diff
@@ -149,7 +149,20 @@ sudo yum install -y runc-${RUNC_VERSION}
 sudo yum versionlock runc-*
 
 # install containerd and lock version
-sudo yum install -y containerd-${CONTAINERD_VERSION}
+function rpm_install() {
```
HusainZafar: Can we extract this into a common helper, similar to the imds helper? Would be useful elsewhere too

templates/al2/provisioners/install-worker.sh
```diff
@@ -149,7 +149,20 @@ sudo yum install -y runc-${RUNC_VERSION}
...
+  local RPMS=($@)
+  echo "Pulling and installing local rpms from s3 bucket"
+  for RPM in "${RPMS[@]}"; do
+    aws s3 cp --region ${BINARY_BUCKET_REGION} s3://${BINARY_BUCKET_NAME}/rpms/${RPM} ${WORKING_DIR}/${RPM}
+    sudo yum localinstall -y ${WORKING_DIR}/${RPM}
+  done
+}
+
+if [[ "$CONTAINERD_VERSION" == 1.* ]]; then
+  rpm_install "containerd-1.7.27-1.amzn2.0.4.$(uname -m).rpm"
```
HusainZafar: Nit: Can extract 1.7.27-1 as a variable `CONTAINERD_PATCH_VERSION`

HusainZafar: `$(uname -m)` already exists as $MACHINE

templates/al2/provisioners/install-worker.sh
mselim00: +1, we actually already have this in another place for AL2023 https://github.com/awslabs/amazon-eks-ami/blob/f59beffefeecb784d1d771137ab6abbb864d79d7/templates/al2023/provisioners/install-nvidia-driver.sh#L23-L30

templates/al2/provisioners/install-worker.sh
```diff
@@ -149,7 +149,20 @@ sudo yum install -y runc-${RUNC_VERSION}
...
+function rpm_install() {
+  local RPMS=($@)
+  echo "Pulling and installing local rpms from s3 bucket"
+  for RPM in "${RPMS[@]}"; do
+    aws s3 cp --region ${BINARY_BUCKET_REGION} s3://${BINARY_BUCKET_NAME}/rpms/${RPM} ${WORKING_DIR}/${RPM}
+    sudo yum localinstall -y ${WORKING_DIR}/${RPM}
+  done
+}
+
+if [[ "$CONTAINERD_VERSION" == 1.* ]]; then
```
mselim00: should this be gated on something more targeted?

templates/al2/provisioners/install-worker.sh
wwvela: * After discussion with Carter, we won't install contianerd from s3 bucket directly this way and this also will be an issue for custom AMI builder.
* We will provide `containerd_rpm_url` param which can be used by us as well as custom AMI builder if they want to install containerd rpm from outside url. 
* And for us, we will generate the containerd presigned s3 url for containerd 1.7 with patches before the ami build and pass it though this param.

wwvela:
/ci
+workflow:k8s_versions 1.32
+build binary_bucket_name=amazon-wesley-beta-ami

wwvela:
/ci
+workflow:k8s_versions 1.32
+build binary_bucket_name=amazon-wesley-beta-ami install_containerd_from_s3=true

templates/al2/provisioners/install-worker.sh
```diff
@@ -149,7 +149,27 @@ sudo yum install -y runc-${RUNC_VERSION}
...
+  echo "Pulling and installing local containerd binary from s3 bucket"
+  for bianry in "${CONTAINERD_BINARIES[@]}"; do
+    aws s3 cp --region ${BINARY_BUCKET_REGION} s3://${BINARY_BUCKET_NAME}/containerd/${bianry} .
+    sudo chmod +x $binary
+    sudo mv $binary /usr/bin/
+  done
+  sudo mkdir -p /var/lib/containerd
+  sudo mv $WORKING_DIR/containerd.service /etc/systemd/system/containerd.service
+  sudo chown root:root /etc/systemd/system/containerd.service
+else
```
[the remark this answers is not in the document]
wwvela: currently only install binary from s3 when containerd version required is `1.7.*` and `install_containerd_from_s3` is `true`

templates/al2023/provisioners/install-worker.sh
```diff
@@ -133,9 +133,26 @@ fi
 ###############################################################################
 ### Containerd setup ##########################################################
 ###############################################################################
-
 sudo dnf install -y runc-${RUNC_VERSION}
-sudo dnf install -y containerd-${CONTAINERD_VERSION}
+# TO-DO: this is a temp way to install binary from s3, need to change once we figure out how to get and store containerd binaries in long term
+if [[ "$CONTAINERD_VERSION" == "1.7.*" && "$INSTALL_CONTAINERD_FROM_S3" == "true" ]]; then
```
ndbaker1: this should have no attachment to containerd version right

templates/al2023/provisioners/install-worker.sh
```diff
@@ -133,9 +133,26 @@ fi
 ###############################################################################
 ### Containerd setup ##########################################################
 ###############################################################################
-
 sudo dnf install -y runc-${RUNC_VERSION}
-sudo dnf install -y containerd-${CONTAINERD_VERSION}
+# TO-DO: this is a temp way to install binary from s3, need to change once we figure out how to get and store containerd binaries in long term
```
ndbaker1: if this is temporary can we do any other way that doesn't add new variable config?

templates/al2/provisioners/install-worker.sh
```diff
@@ -149,7 +149,27 @@ sudo yum install -y runc-${RUNC_VERSION}
 sudo yum versionlock runc-*
 
 # install containerd and lock version
-sudo yum install -y containerd-${CONTAINERD_VERSION}
+# TO-DO: this is a temp way to install binary from s3, need to change once we figure out how to get and store containerd binaries in long term
+if [[ "$CONTAINERD_VERSION" == "1.7.*" && "$INSTALL_CONTAINERD_FROM_S3" == "true" ]]; then
```
cartermckinnon: Why scope it to `1.7.*`? Seems like this should override that option entirely

templates/al2/provisioners/install-worker.sh
```diff
@@ -149,7 +149,27 @@ sudo yum install -y runc-${RUNC_VERSION}
...
+  CONTAINERD_BINARIES=(
+    containerd
+    containerd-shim
+    containerd-shim-runc-v1
+    containerd-shim-runc-v2
+    ctr
+  )
+  echo "Pulling and installing local containerd binary from s3 bucket"
+  for bianry in "${CONTAINERD_BINARIES[@]}"; do
+    aws s3 cp --region ${BINARY_BUCKET_REGION} s3://${BINARY_BUCKET_NAME}/containerd/${bianry} .
```
cartermckinnon:
```suggestion
  for binary in "${CONTAINERD_BINARIES[@]}"; do
    aws s3 cp --region ${BINARY_BUCKET_REGION} s3://${BINARY_BUCKET_NAME}/containerd/${binary} .
```

cartermckinnon: I think we need a version in the S3 key. If the `INSTALL_CONTAINERD_FROM_S3` option is used, we should expect a fully-formed version string in `CONTAINERD_VERSION`.

templates/al2023/provisioners/install-worker.sh
wwvela: I was thinking if we can use the s3 bucket as the marker as internal or external and discussing this with Carter before. We don't want to have the internal bucket name showed in code and cx will use s3 bucket variable to use their own s3 bucket sometime, so I create a new variable config is still needed in this case.

templates/al2/provisioners/install-worker.sh
wwvela: 1.34 version build starts and we will merge containerd v2 migration PR soon. I'm wondering if we want to extract binary from containerd 2.0.5 rpms now and make this process also compatible with containerd v2 migration for 1.34. Or should we keep containerd v2 install from AL repo before we finally figure out how we will get and store those binaries

wwvela: synced offline, removed the version check

templates/al2023/provisioners/install-worker.sh
wwvela: removed the version check

wwvela:
/ci
+workflow:k8s_versions 1.32
+build binary_bucket_name=amazon-wesley-beta-ami install_containerd_from_s3=true

templates/al2/provisioners/install-worker.sh
wwvela: According to latest discussion, we will install containerd bianries from s3 if `install_containerd_from_s3=true`

wwvela:
/ci
+workflow:k8s_versions 1.32
+build binary_bucket_name=amazon-wesley-beta-ami install_containerd_from_s3=true

templates/al2/variables-default.json
```diff
@@ -13,7 +13,8 @@
     "binary_bucket_region": "us-west-2",
     "cache_container_images": "false",
     "cni_plugin_version": "v1.2.0",
-    "containerd_version": "1.7.*",
+    "containerd_version": "1.7.27",
```
ndbaker1, line 16: is this pin permanent or temporary? it seems like the decision could be up to whoever runs the build to provide a fully versioned string

wwvela: This is permanent. Others can still pass `1.7.*` without `install_containerd_from_s3` set to true. This would still allow them to install the latest 1.7.* from AL repo. And for internal, we would set a fully-formed version string in CONTAINERD_VERSION and also have the fully formed version in s3 key

templates/al2/provisioners/install-worker.sh
```diff
@@ -149,7 +149,25 @@ sudo yum install -y runc-${RUNC_VERSION}
 sudo yum versionlock runc-*
 
 # install containerd and lock version
-sudo yum install -y containerd-${CONTAINERD_VERSION}
+# TO-DO: this is a temp way to install binary from s3, need to change once we figure out how to get and store containerd binaries in long term
```
cartermckinnon:
```suggestion
```

templates/al2/provisioners/install-worker.sh
```diff
@@ -149,7 +149,25 @@ sudo yum install -y runc-${RUNC_VERSION}
...
+if [[ "$INSTALL_CONTAINERD_FROM_S3" == "true" ]]; then
+  CONTAINERD_BINARIES=(
+    containerd
+    containerd-shim-runc-v2
+    ctr
+  )
+  echo "Pulling and installing local containerd binary from s3 bucket"
```
cartermckinnon:
```suggestion
  echo "Installing containerd from S3..."
```

templates/al2/provisioners/install-worker.sh
```diff
@@ -149,7 +149,25 @@ sudo yum install -y runc-${RUNC_VERSION}
...
+  for binary in "${CONTAINERD_BINARIES[@]}"; do
+    aws s3 cp --region ${BINARY_BUCKET_REGION} s3://${BINARY_BUCKET_NAME}/containerd/${CONTAINERD_VERSION}/${binary} .
+    sudo chmod +x $binary
+    sudo mv $binary /usr/bin/
+  done
+  sudo mkdir -p /var/lib/containerd
+  sudo mv $WORKING_DIR/containerd.service /etc/systemd/system/containerd.service
+  sudo chown root:root /etc/systemd/system/containerd.service
```
cartermckinnon, line 166: Do you need to `daemon-reload` and enable the service?

templates/al2023/provisioners/install-worker.sh
```diff
@@ -133,9 +133,24 @@ fi
...
```
cartermckinnon:
```suggestion
```

templates/al2/provisioners/install-worker.sh
wwvela: The containerd will be enabled by default with these setting, seems no need to do daemon-reload or enable it explicitly here.

templates/al2/variables-default.json
ndbaker1: id like if we could use the existing wildcard convention and just have this details matter when installing from direct sources like s3

templates/al2/provisioners/install-worker.sh
```diff
@@ -149,7 +149,24 @@ sudo yum install -y runc-${RUNC_VERSION}
...
+else
+  sudo yum install -y containerd-${CONTAINERD_VERSION}
+fi
 sudo yum versionlock containerd-*
```
ndbaker1: this just no-ops when the package is not installed?

templates/al2023/provisioners/install-worker.sh
```diff
@@ -133,9 +133,23 @@ fi
...
 sudo dnf install -y runc-${RUNC_VERSION}
-sudo dnf install -y containerd-${CONTAINERD_VERSION}
+if [[ "$INSTALL_CONTAINERD_FROM_S3" == "true" ]]; then
+  CONTAINERD_BINARIES=(
+    containerd
+    containerd-shim-runc-v2
+    ctr
+  )
+  for binary in "${CONTAINERD_BINARIES[@]}"; do
+    echo "Pulling and installing local containerd binary ${binary} from s3 bucket"
```
ndbaker1: repeat of a previous suggestion
```suggestion
    echo "Installing containerd from S3..."
```

templates/al2/variables-default.json
wwvela: Oh you mean. we pass a fully-formated version when version only when install_containerd_from_s3 is true?

wwvela: revised, we would need to add detailed version in internal build

templates/al2/provisioners/install-worker.sh
ndbaker1: we also need to find a decide on solution for preventing folks from installing containerd from the AL repo when its installed via the binary method above. 

spitballing maybe we can add `excludepkgs=containerd` to the default dnf/yum conf to prevent folks from being able to overwrite the binaries

wwvela: * For al2, we need to exclude both `exclude=containerd*,docker*` as it will still trying to install docker when running `sudo yum install containerd` and throw out docker related errors.
* For al2023, `exclude=containerd*` is enough to prevent

below are outputs after this change
```
# For al2
> sudo yum install containerd
amzn2-core                                                                                                                                                           | 3.6 kB  00:00:00
amzn2extra-docker                                                                                                                                                    | 2.9 kB  00:00:00
amzn2extra-kernel-5.10                                                                                                                                               | 3.0 kB  00:00:00
Nothing to do

# for al2023
> sudo yum install containerd
Last metadata expiration check: 15:34:00 ago on Tue Jul 29 04:19:35 2025.
All matches were filtered out by exclude filtering for argument: containerd
Error: Unable to find a match: containerd
```

ndbaker1: does dnf/yum have any kind of drop-in config support, just so that its as easy as possible to drop this guard rail if user's want to? it's probably viable to say use `--disableexcludes` since there are no defaults, but want to make sure that's the last resort

wwvela: Find this in [dnf configuration doc](https://dnf.readthedocs.io/en/latest/conf_ref.html). Seems the `--disableexcludes` is the only way except manually remove the exclude setting in dnf/yum.conf
```
Exclude packages of this repository, specified by a name or a glob and separated by a comma, from all operations. Can be disabled using --disableexcludes command line switch. Defaults to [].
```

wwvela:
/ci
+workflow:k8s_versions 1.32
+build binary_bucket_name=amazon-wesley-beta-ami install_containerd_from_s3=true arch=arm64

wwvela:
/ci
+workflow:k8s_versions 1.32
+build binary_bucket_name=amazon-wesley-beta-ami install_containerd_from_s3=true

templates/al2/provisioners/install-worker.sh
```diff
@@ -149,8 +149,27 @@ sudo yum install -y runc-${RUNC_VERSION}
...
-sudo yum versionlock containerd-*
+if [[ "$INSTALL_CONTAINERD_FROM_S3" == "true" ]]; then
+  CONTAINERD_BINARIES=(
+    containerd
+    containerd-shim-runc-v2
+    ctr
+  )
+  echo "Installing containerd from S3..."
+  for binary in "${CONTAINERD_BINARIES[@]}"; do
+    aws s3 cp --region ${BINARY_BUCKET_REGION} s3://${BINARY_BUCKET_NAME}/containerd/${CONTAINERD_VERSION}/${binary} .
```
cartermckinnon: don't you need the S3 key to include the architecture? e.g.
```
s3://mybucket/bin/containerd/1.7.27-1/x86_64/containerd
```

templates/al2/provisioners/install-worker.sh
cartermckinnon: Are you sure? `systemctl enable` creates a symbolic link, I don't think that would happen implicitly just by putting something in this directory. Are we enabling it elsewhere somewhere? check `systemctl status containerd`

templates/al2/provisioners/install-worker.sh
wwvela: Oh I only update the al2023 one. Update for al2 in latest change

templates/al2/provisioners/install-worker.sh
wwvela: * In AL2, saw we run systemctl daemon-reload and systemctl enable for containerd here https://github.com/awslabs/amazon-eks-ami/blob/main/templates/al2/provisioners/install-worker.sh#L403-L405

* In al2023, nodeAdm will enable the containerd https://github.com/awslabs/amazon-eks-ami/blob/99eff32dbfdf9a1492086fc4d081f2115d40e3b8/templates/al2023/provisioners/install-nodeadm.sh#L7

wwvela:
/ci
+workflow:k8s_versions 1.32
+build binary_bucket_name=amazon-wesley-beta-ami install_containerd_from_s3=true arch=arm64

wwvela:
/ci
+workflow:k8s_versions 1.32
+build binary_bucket_name=amazon-wesley-beta-ami install_containerd_from_s3=true arch=x86_64

wwvela:
/ci
+workflow:k8s_versions 1.32
+build binary_bucket_name=amazon-wesley-beta-ami install_containerd_from_s3=true arch=x86_64

wwvela:
/ci
+workflow:k8s_versions 1.32
+build binary_bucket_name=amazon-wesley-beta-ami containerd_version=1.7.27 install_containerd_from_s3=true arch=x86_64

wwvela:
previous ci failed because of containerd_version, need to specify fully format containerd_version like 1.7.27 when build

wwvela:
I think we should still maintain fully formated `containerd_version` like `1.7.27` in variables-xx.json file instead of pass this info in internal build. 
* As it required us to maintain some containerd version check according to k8s version in internal build once containerd 2.0 used in 1.34+. 
* And it also make the `containerd_version` pin in variable file meaningless as we will have to maintain get version logic in ami build

PR #2347, opened 2025-08-01
fix(ci): remove vpc-cni deployer flags

cartermckinnon:
**Description of changes:**

There's a bug in the latest VPC CNI version (v1.20.0) that is causing our CI to fail. We shouldn't need to bump the CNI version in our AL2023 jobs anymore, so let's just drop the flags.


cartermckinnon:
<cancelled CI run, this branch's workflow wouldn't be used anyway>

calvinbui:
i've opened an issue on the vpc-cni repo https://github.com/aws/amazon-vpc-cni-k8s/issues/3394

PR #2350, opened 2025-08-05
fix(build/al2023): start with latest release version

mselim00:
**Issue #, if available:**

**Description of changes:**
AL2023, by default, locks packages visible to a given AMI release by associating all `dnf` operations with a namesake repository ([ref](https://docs.aws.amazon.com/linux/al2023/ug/deterministic-upgrades-usage.html#using-a-deterministic-upgraded-system)). This is done by a `releasever` variable in the repo URL, e.g. `https://al2023-repos-$awsregion-de612dc2.s3$dualstack.$awsregion.$awsdomain/core/mirrors/$releasever/debuginfo/$basearch/$mirrorlist`. If `/etc/dnf/vars/releasever` is empty (as it is by default), `dnf` attempts to resolve the value by checking the installed version of one of several known marker packages, in this case `system-release` ([1](https://github.com/rpm-software-management/dnf/blob/b7eb2e399c42d10444f7d045689fd50f91cf72db/dnf/const.py.in#L25) -> [2](https://github.com/rpm-software-management/dnf/blob/b7eb2e399c42d10444f7d045689fd50f91cf72db/dnf/rpm/__init__.py#L44)). 

AL2023 AMIs release versions include that package with the same version, and each repository includes that package at the same version as in the repository URL. Using `releasever=latest` indicates to use the lastest package repository, which includes the latest version of `system-release`. A generic `dnf upgrade` on that package repo closes the loop by installing a newer version of `system-release`, causing all future `dnf` operations to be bound to that new version by default.

**Testing Done**

### DIFF
--- a/templates/al2023/provisioners/install-worker.sh
+++ b/templates/al2023/provisioners/install-worker.sh
@@ -47,7 +47,7 @@ fi
 ################################################################################
 
 # Update the OS to begin with to catch up to the latest packages.
-sudo dnf update -y
+sudo dnf upgrade -y --releasever=latest
 
 # Install necessary packages
 sudo dnf install -y \


mselim00:
AL2023's release version concept is much more granular than AL2's. In AL2, the release version is just `2` for AL**2**, whereas for AL2023 it's a specific release (e.g. `2023.8.20250721`). That in turn means any `yum` operation on AL2 will always use the latest packages for AL2 by default

PR #2352, opened 2025-08-05
feat(nodeadm) add SOCI snapshotter support

fletcherw:
Add a feature flag to NodeConfig that allows enabling the parallel SOCI snapshotter when pulling images. This may result in faster image pulls on large instances, at the cost of higher CPU and memory usage.

See also #2305 

**Testing Done**

* Unit testing
* Build and run AMI in cluster, pull large image

Co-authored-by: husaiz <husaiz@amazon.com>

### DIFF
--- a/nodeadm/api/v1alpha1/nodeconfig_types.go
+++ b/nodeadm/api/v1alpha1/nodeconfig_types.go
@@ -129,10 +129,16 @@ const (
 )
 
 // Feature specifies which feature gate should be toggled
-// +kubebuilder:validation:Enum={InstanceIdNodeName}
+// +kubebuilder:validation:Enum={InstanceIdNodeName,AggressiveImagePull}
 type Feature string
 
 const (
 	// InstanceIdNodeName will use EC2 instance ID as node name
 	InstanceIdNodeName Feature = "InstanceIdNodeName"
+
+	// AggressiveImagePull enables a parallel image pull for container
+	// images. This will use more instance CPU and Memory during image pull, but
+	// may result in faster image pull times. This flag will be ignored on
+	// instances with memory and vCPU below a certain threshold.
+	AggressiveImagePull Feature = "AggressiveImagePull"
 )
--- a/nodeadm/doc/examples.md
+++ b/nodeadm/doc/examples.md
@@ -102,7 +102,7 @@ This has the benefit of potentially decreasing image pull time, at the cost of i
 
 ### To enable this feature:
 1. Ensure your instance type is a larger instance type. Currently we recommend a 2xlarge instance or larger, but that value may change.
-2. Make sure your workloads can tolerate the increased CPU and memory usage during image pull.
+2. Make sure your workloads can tolerate the increased CPU and memory usage during image pull. This makes the most sense when you need to pull a very large container image early in a node's lifecycle, before other workloads are running.
 3. Enable the feature gate in your user data:
 ```
 ---
--- a/nodeadm/doc/examples.md
+++ b/nodeadm/doc/examples.md
@@ -92,6 +92,28 @@ spec:
 ```
 
 ---
+## Enabling aggressive image pull (experimental)
+
+When the `AggressiveImagePull` feature gate is enabled, `nodeadm` will configure the container runtime to pull and unpack container images in parallel.
+
+This has the benefit of potentially decreasing image pull time, at the cost of increased CPU and memory usage during image pull.
+
+⚠️ **Note**: This flag will be ignored on instance sizes below a certain vCPU and memory threshold.
+
+### To enable this feature:
+1. Ensure your instance type is a larger instance type. Currently we recommend a 2xlarge instance or larger, but that value may change.
+2. Make sure your workloads can tolerate the increased CPU and memory usage during image pull.
+3. Enable the feature gate in your user data:
+```
+---
+apiVersion: node.eks.aws/v1alpha1
+kind: NodeConfig
+spec:
+  featureGates:
+    AggressiveImagePull: true
+```
+
+---
 
 ## Configuring `containerd`
 
--- a/nodeadm/internal/api/featuregates.go
+++ b/nodeadm/internal/api/featuregates.go
@@ -14,6 +14,12 @@ var featureVerifiers = map[Feature]func(Feature, map[Feature]bool) bool{
 	// InstanceIdNodeNameGate controls whether to use instance ID as the node's name.
 	// By default, this feature is disabled, and the private DNS Name will be used.
 	InstanceIdNodeName: DefaultFalse,
+
+	// AggressiveImagePull enables a parallel image pull for container
+	// images. This will use more instance CPU and Memory during image pull, but
+	// may result in faster image pull times. This flag will be ignored on
+	// instances with memory and vCPU below a certain threshold.
+	AggressiveImagePull: DefaultFalse,
 }
 
 func IsFeatureEnabled(feature Feature, featureGates map[Feature]bool) bool {
--- a/nodeadm/internal/api/types.go
+++ b/nodeadm/internal/api/types.go
@@ -122,4 +122,6 @@ type Feature string
 const (
 	// InstanceIdNodeName will use EC2 instance ID as node name
 	InstanceIdNodeName Feature = "InstanceIdNodeName"
+
+	AggressiveImagePull Feature = "AggressiveImagePull"
 )
--- a/nodeadm/internal/containerd/config.go
+++ b/nodeadm/internal/containerd/config.go
@@ -9,6 +9,7 @@ import (
 	"text/template"
 
 	"github.com/awslabs/amazon-eks-ami/nodeadm/internal/api"
+	"github.com/awslabs/amazon-eks-ami/nodeadm/internal/system"
 	"github.com/awslabs/amazon-eks-ami/nodeadm/internal/util"
 	"github.com/pelletier/go-toml/v2"
 	"go.uber.org/zap"
@@ -18,8 +19,9 @@ import (
 const ContainerRuntimeEndpoint = "unix:///run/containerd/containerd.sock"
 
 const (
-	containerdConfigFile = "/etc/containerd/config.toml"
-	containerdConfigPerm = 0644
+	containerdConfigFile      = "/etc/containerd/config.toml"
+	sociSnapshotterConfigFile = "/etc/soci-snapshotter-grpc/config.toml"
+	configPerm                = 0644
 )
 
 var (
@@ -27,6 +29,9 @@ var (
 	containerdConfigTemplateData string
 	//go:embed config2.template.toml
 	containerdConfigTemplateData2 string
+
+	//go:embed snapshotter/soci-snapshotter.config.toml
+	sociSnapshotterTemplateData []byte
 )
 
 type ConfigSchema string
@@ -34,6 +39,8 @@ type ConfigSchema string
 const (
 	ConfigSchemaV2 ConfigSchema = "2"
 	ConfigSchemaV3 ConfigSchema = "3"
+
+	gibibyte int64 = 1024 * 1024 * 1024
 )
 
 var containerdTemplateVersionMap = map[ConfigSchema]string{
@@ -42,13 +49,14 @@ var containerdTemplateVersionMap = map[ConfigSchema]string{
 }
 
 type containerdTemplateVars struct {
-	EnableCDI         bool
-	SandboxImage      string
-	RuntimeName       string
-	RuntimeBinaryName string
+	EnableCDI          bool
+	SandboxImage       string
+	RuntimeName        string
+	RuntimeBinaryName  string
+	UseSOCISnapshotter bool
 }
 
-func writeContainerdConfig(cfg *api.NodeConfig) error {
+func writeContainerdConfig(cfg *api.NodeConfig, resources system.Resources) error {
 	isContainerdV2, err := isContainerdV2()
 	if err != nil {
 		return err
@@ -57,27 +65,17 @@ func writeContainerdConfig(cfg *api.NodeConfig) error {
 	if err != nil {
 		return err
 	}
-	containerdConfig, err := generateContainerdConfig(cfg, templateVersion)
+	containerdConfig, err := generateContainerdConfig(cfg, resources, templateVersion)
+	if err != nil {
+		return err
+	}
+	containerdConfig, err = combineContainerdConfigs(containerdConfig, cfg.Spec.Containerd.Config)
 	if err != nil {
 		return err
 	}
 
-	// because the logic in containerd's import merge decides to completely
-	// overwrite entire sections, we want to implement this merging ourselves.
-	// see: https://github.com/containerd/containerd/blob/a91b05d99ceac46329be06eb43f7ae10b89aad45/cmd/containerd/server/config/config.go#L407-L431
-	if len(cfg.Spec.Containerd.Config) > 0 {
-		containerdConfigMap, err := util.Merge(containerdConfig, []byte(cfg.Spec.Containerd.Config), toml.Marshal, toml.Unmarshal)
-		if err != nil {
-			return err
-		}
-		containerdConfig, err = toml.Marshal(containerdConfigMap)
-		if err != nil {
-			return err
-		}
-	}
-
 	zap.L().Info("Writing containerd config to file..", zap.String("path", containerdConfigFile))
-	err = util.WriteFileWithDir(containerdConfigFile, containerdConfig, containerdConfigPerm)
+	err = util.WriteFileWithDir(containerdConfigFile, containerdConfig, configPerm)
 	if err != nil {
 		return err
 	}
@@ -90,14 +88,32 @@ func writeContainerdConfig(cfg *api.NodeConfig) error {
 	return nil
 }
 
-func generateContainerdConfig(cfg *api.NodeConfig, templateVersion ConfigSchema) ([]byte, error) {
+func combineContainerdConfigs(configA []byte, configB api.ContainerdConfig) ([]byte, error) {
+	// because the logic in containerd's import merge decides to completely
+	// overwrite entire sections, we want to implement this merging ourselves.
+	// see: https://github.com/containerd/containerd/blob/a91b05d99ceac46329be06eb43f7ae10b89aad45/cmd/containerd/server/config/config.go#L407-L431
+	if len(configB) <= 0 {
+		return configA, nil
+	}
+
+	containerdConfigMap, err := util.Merge(configA, []byte(configB), toml.Marshal, toml.Unmarshal)
+	if err != nil {
+		return nil, err
+	}
+
+	return toml.Marshal(containerdConfigMap)
+
+}
+
+func generateContainerdConfig(cfg *api.NodeConfig, resources system.Resources, templateVersion ConfigSchema) ([]byte, error) {
 	runtimeOptions := getRuntimeOptions(cfg)
 
 	configVars := containerdTemplateVars{
-		SandboxImage:      cfg.Status.Defaults.SandboxImage,
-		RuntimeBinaryName: runtimeOptions.RuntimeBinaryPath,
-		RuntimeName:       runtimeOptions.RuntimeName,
-		EnableCDI:         semver.Compare(cfg.Status.KubeletVersion, "v1.32.0") >= 0,
+		SandboxImage:       cfg.Status.Defaults.SandboxImage,
+		RuntimeBinaryName:  runtimeOptions.RuntimeBinaryPath,
+		RuntimeName:        runtimeOptions.RuntimeName,
+		EnableCDI:          semver.Compare(cfg.Status.KubeletVersion, "v1.32.0") >= 0,
+		UseSOCISnapshotter: UseSOCISnapshotter(cfg, resources),
 	}
 	var buf bytes.Buffer
 	containerdConfigTemplate := template.Must(template.New(containerdConfigFile).Parse(containerdTemplateVersionMap[templateVersion]))
@@ -137,5 +153,34 @@ func migrateConfig() error {
 	if err != nil {
 		return err
 	}
-	return util.WriteFileWithDir(containerdConfigFile, migratedConfig, containerdConfigPerm)
+	return util.WriteFileWithDir(containerdConfigFile, migratedConfig, configPerm)
+}
+
+func writeSnapshotterConfig(cfg *api.NodeConfig, resources system.Resources) error {
+	if UseSOCISnapshotter(cfg, resources) {
+		return util.WriteFileWithDir(sociSnapshotterConfigFile, sociSnapshotterTemplateData, configPerm)
+	}
+
+	return nil
+}
+
+func UseSOCISnapshotter(cfg *api.NodeConfig, resources system.Resources) bool {
+	if !api.IsFeatureEnabled(api.AggressiveImagePull, cfg.Spec.FeatureGates) {
+		return false
+	}
+
+	totalCPUMillicores, err := resources.GetMilliNumCores()
+	if err != nil {
+		zap.L().Error("Error getting total CPU millicores", zap.Error(err))
+		return false
+	}
+
+	totalMemory, err := resources.GetOnlineMemory()
+	if err != nil {
+		zap.L().Error("Error getting total memory", zap.Error(err))
+		return false
+	}
+
+	return totalMemory >= 8*gibibyte && totalCPUMillicores >= 8000
+
 }
--- a/nodeadm/internal/containerd/config.template.toml
+++ b/nodeadm/internal/containerd/config.template.toml
@@ -8,6 +8,10 @@ address = "/run/containerd/containerd.sock"
 [plugins."io.containerd.grpc.v1.cri".containerd]
 default_runtime_name = "{{.RuntimeName}}"
 discard_unpacked_layers = true
+{{- if .UseSOCISnapshotter}}
+snapshotter = "soci"
+disable_snapshot_annotations = false
+{{- end}}
 
 [plugins."io.containerd.grpc.v1.cri"]
 sandbox_image = "{{.SandboxImage}}"
@@ -27,3 +31,13 @@ SystemdCgroup = true
 [plugins."io.containerd.grpc.v1.cri".cni]
 bin_dir = "/opt/cni/bin"
 conf_dir = "/etc/cni/net.d"
+
+{{if .UseSOCISnapshotter}}
+[proxy_plugins.soci]
+type = "snapshot"
+address = "/run/soci-snapshotter-grpc/soci-snapshotter-grpc.sock"
+
+[proxy_plugins.soci.exports]
+root = "/var/lib/soci-snapshotter-grpc"
+
+{{- end}}
--- a/nodeadm/internal/containerd/snapshotter/soci-snapshotter.config.toml
+++ b/nodeadm/internal/containerd/snapshotter/soci-snapshotter.config.toml
@@ -1,0 +1,21 @@
+debug = false
+
+[content_store]
+type = 'containerd'
+
+[cri_keychain]
+enable_keychain = true
+image_service_path = '/run/containerd/containerd.sock'
+
+[pull_modes.parallel_pull_unpack]
+enable = true
+concurrent_download_chunk_size = "16mb"
+discard_unpacked_layers = true
+max_concurrent_downloads = -1
+max_concurrent_downloads_per_image = 32
+max_concurrent_unpacks = -1
+max_concurrent_unpacks_per_image = 12
+
+[pull_modes.parallel_pull_unpack.decompress_streams."gzip"]
+path = '/usr/bin/unpigz'
+args = ['-d', '-c']
--- a/nodeadm/internal/system/resources.go
+++ b/nodeadm/internal/system/resources.go
@@ -17,6 +17,14 @@ var (
 	cpusPath     = "/sys/devices/system/cpu"
 )
 
+type Resources interface {
+	GetMilliNumCores() (int, error)
+	GetOnlineMemory() (int64, error)
+}
+
+// Real implementation of Resources interface that reads values from sysfs.
+type SysfsResources struct{}
+
 const (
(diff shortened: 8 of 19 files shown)


nodeadm/internal/containerd/config.go
```diff
@@ -15,55 +16,104 @@ import (
...
+
+	totalCPUMillicores, err := resources.GetMilliNumCores()
+	if err != nil {
+		zap.L().Error("Error getting total CPU millicores", zap.Error(err))
+		return false
+	}
+
+	if totalCPUMillicores < 8000 {
+		return false
+	}
```
wwvela: nit: This and move to the final return

nodeadm/internal/containerd/config.go
```diff
@@ -15,55 +16,104 @@ import (
...
+
+	totalMemory, err := resources.GetOnlineMemory()
+	if err != nil {
+		zap.L().Error("Error getting total memory", zap.Error(err))
+		return false
+	}
+
+	return totalMemory >= 8*1024*1024*1024
```
wwvela: This can be `return totalMemory >= 8*1024*1024*1024 && totalCPUMillicores >= 8000`

nodeadm/internal/containerd/config.template.toml
```diff
@@ -8,6 +8,10 @@ address = "/run/containerd/containerd.sock"
 [plugins."io.containerd.grpc.v1.cri".containerd]
 default_runtime_name = "{{.RuntimeName}}"
 discard_unpacked_layers = true
+{{- if .UseSOCISnapshotter}}
```
wwvela, line 11: We merged the containerd v2 changes, this also needed in config2.template.toml
In containerd v2 config, `snapshotter` and `disable_snapshot_annotations` moved under `[plugins.'io.containerd.cri.v1.images']`. 
* here is their full config: https://github.com/containerd/containerd/blob/main/docs/cri/config.md#full-configuration

nodeadm/internal/containerd/config.go
```diff
@@ -15,55 +16,104 @@ import (
...
+	EnableCDI          bool
+	SandboxImage       string
+	RuntimeName        string
+	RuntimeBinaryName  string
+	UseSOCISnapshotter bool
 }
 
-func writeContainerdConfig(cfg *api.NodeConfig) error {
-	containerdConfig, err := generateContainerdConfig(cfg)
+func writeContainerdConfig(cfg *api.NodeConfig, resources system.Resources) error {
```
wwvela, line 59: The writecontainerdconfig and generateContainerdConfig method are changed a bit for containerd 2.0 migration which adding the containerd version check and choose which config version need to be used logic here.

nodeadm/internal/api/featuregates.go
```diff
@@ -14,6 +14,12 @@ var featureVerifiers = map[Feature]func(Feature, map[Feature]bool) bool{
 	// InstanceIdNodeNameGate controls whether to use instance ID as the node's name.
 	// By default, this feature is disabled, and the private DNS Name will be used.
 	InstanceIdNodeName: DefaultFalse,
+
+	// AggressiveImagePullTradeoff enables a parallel image pull for container
+	// images. This will use more instance CPU and Memory during image pull, but
+	// may result in faster image pull times. This flag cannot be enabled on
+	// instances with memory and vCPU below a certain threshold.
```
mselim00:
```suggestion
	// may result in faster image pull times. This flag will be ignored on
	// instances with memory and vCPU below a certain threshold.
```
**cannot** makes it sound like it will error, when in reality we're ignoring the flag in these cases. The distinction's important because avoiding an error would driver users to be invested in what the unspecified threshold is, but it seems like we want to keep that as an implementation detail

nodeadm/internal/containerd/config.go
mselim00: I think storing this as a `gibibyte` constant for would make this a little bit more intuitive (maybe even worth also storing the intervening KiB and MiB and building up to it)

fletcherw: Done.

nodeadm/internal/containerd/config.template.toml
fletcherw: Updated the v2 templates as well, can you double check me that I'm putting the config in the right place?

nodeadm/internal/containerd/config.go
fletcherw: Fixed.

nodeadm/internal/containerd/config.go
```diff
@@ -137,5 +148,34 @@ func migrateConfig() error {
...
+		return false
+	}
+
+	totalMemory, err := resources.GetOnlineMemory()
+	if err != nil {
+		zap.L().Error("Error getting total memory", zap.Error(err))
+		return false
+	}
+
+	return totalMemory >= 8*1024*1024*1024 && totalCPUMillicores >= 8000
```
mselim00: I think it's worth creating a constant for a gibibyte to reference here, maybe even KiB and MiB to build up to it

nodeadm/internal/containerd/config.template.toml
wwvela: Looks correct to me

wwvela:
/ci
+build binary_bucket_name=amazon-wesley-beta-ami containerd_version=1.7.27 install_containerd_from_s3=true arch=x86_64

wwvela (commented):
changes LGTM let's wait the ci passed before merge

nodeadm/internal/system/resources.go
```diff
@@ -270,3 +279,57 @@ func isCpuOnline(path string, cpuID uint64) (bool, error) {
...
+			return 0, fmt.Errorf("failed to read online path for memory '%s': %w", onlinePath, err)
+		}
+
+		if strings.TrimSpace(string(onlineContents)) == "1" {
+			onlineMemory += 1
+		}
+	}
+
+	return blockSize * onlineMemory, nil
+}
```
mselim00: I tried comparing this heuristic in shell to some other sources on a `g6f.2xlarge`:
```
$ ONLINE_BLOCKS=0; BLOCK_SIZE_BYTES=$(cat /sys/devices/system/memory/block_size_bytes); for MEMORY_BLOCK_FILE in /sys/devices/system/memory/memory*; do ONLINE_BLOCKS=$((ONLINE_BLOCKS+1)); done && echo "$((ONLINE_BLOCKS*BLOCK_SIZE_BYTES))"
1976000000
$ SYS_MEM_KIB=$(cat /proc/meminfo | grep 'MemTotal' | awk '{ print $2 }') && SYS_MEM_BYTES=$((SYS_MEM_KIB*1024)) && echo $SYS_MEM_BYTES
32373919744
$ EC2_MEM_MIB=$(aws ec2 describe-instance-types --instance-types=$(cat /sys/devices/virtual/dmi/id/product_name)  --query 'InstanceTypes[0].MemoryInfo.SizeInMiB') && EC2_MEM_BYTES=$((EC2_MEM_MIB*1024*1024)) && echo $EC2_MEM_BYTES
34359738368
```
not sure if I missed something in my translation but this is showing way less than expected

nodeadm/internal/containerd/snapshotter/soci-snapshotter.config.toml
```diff
@@ -0,0 +1,21 @@
+debug = false
```
mselim00, line 1: since we enable soci at build time, is it worth also unconditionally writing this config at the same time? or are we intentionally leaving that open as an avenue for users to configure it completely independently

nodeadm/internal/system/resources.go
```diff
@@ -17,13 +17,22 @@ var (
 	cpusPath     = "/sys/devices/system/cpu"
 )
 
+type Resources interface {
+	GetMilliNumCores() (int, error)
+	GetOnlineMemory() (int64, error)
+}
+
+type ProdResources struct{}
```
mselim00: nit: what does this name mean?

templates/al2023/provisioners/install-worker.sh
```diff
@@ -250,6 +250,12 @@ sudo chmod +x $ECR_CREDENTIAL_PROVIDER_BINARY
 sudo mkdir -p /etc/eks/image-credential-provider
 sudo mv $ECR_CREDENTIAL_PROVIDER_BINARY /etc/eks/image-credential-provider/
 
+###############################################################################
+### SOCI Snapshotter setup ##########################################################
+###############################################################################
+sudo dnf install -y soci-snapshotter
+sudo systemctl enable soci-snapshotter.socket
+
```
wwvela: The image build failed with error 
```
2025-08-06T23:29:16Z: ==> amazon-ebs: Error:
2025-08-06T23:29:16Z: ==> amazon-ebs:  Problem: conflicting requests
2025-08-06T23:29:16Z: ==> amazon-ebs:   - package soci-snapshotter-0.11.1-1.amzn2023.0.1.aarch64 from amazonlinux requires containerd, but none of the providers can be installed
2025-08-06T23:29:16Z: ==> amazon-ebs:   - package soci-snapshotter-0.7.0-1.amzn2023.0.1.aarch64 from amazonlinux requires containerd, but none of the providers can be installed
2025-08-06T23:29:16Z: ==> amazon-ebs:   - package soci-snapshotter-0.7.0-1.amzn2023.0.2.aarch64 from amazonlinux requires containerd, but none of the providers can be installed
2025-08-06T23:29:16Z: ==> amazon-ebs:   - package soci-snapshotter-0.8.0-1.amzn2023.0.1.aarch64 from amazonlinux requires containerd, but none of the providers can be installed
2025-08-06T23:29:16Z: ==> amazon-ebs:   - package soci-snapshotter-0.9.0-1.amzn2023.0.1.aarch64 from amazonlinux requires containerd, but none of the providers can be installed
2025-08-06T23:29:16Z: ==> amazon-ebs:   - package soci-snapshotter-0.9.0-1.amzn2023.0.3.aarch64 from amazonlinux requires containerd, but none of the providers can be installed
2025-08-06T23:29:16Z: ==> amazon-ebs:   - package soci-snapshotter-0.9.0-1.amzn2023.0.4.aarch64 from amazonlinux requires containerd, but none of the providers can be installed
2025-08-06T23:29:16Z: ==> amazon-ebs:   - package containerd-1.6.19-1.amzn2023.0.1.aarch64 from amazonlinux is filtered out by exclude filtering
2025-08-06T23:29:16Z: ==> amazon-ebs:   - package containerd-1.6.8-2.amzn2023.0.3.aarch64 from amazonlinux is filtered out by exclude filtering
2025-08-06T23:29:16Z: ==> amazon-ebs:   - package containerd-1.6.8-2.amzn2023.0.4.aarch64 from amazonlinux is filtered out by exclude filtering
2025-08-06T23:29:16Z: ==> amazon-ebs:   - package containerd-1.7.11-1.amzn2023.0.1.aarch64 from amazonlinux is filtered out by exclude filtering
2025-08-06T23:29:16Z: ==> amazon-ebs:   - package containerd-1.7.2-1.amzn2023.0.1.aarch64 from amazonlinux is filtered out by exclude filtering
2025-08-06T23:29:16Z: ==> amazon-ebs:   - package containerd-1.7.2-1.amzn2023.0.2.aarch64 from amazonlinux is filtered out by exclude filtering
2025-08-06T23:29:16Z: ==> amazon-ebs:   - package containerd-1.7.2-1.amzn2023.0.3.aarch64 from amazonlinux is filtered out by exclude filtering
2025-08-06T23:29:16Z: ==> amazon-ebs:   - package containerd-1.7.2-1.amzn2023.0.4.aarch64 from amazonlinux is filtered out by exclude filtering
2025-08-06T23:29:16Z: ==> amazon-ebs:   - package containerd-1.7.20-1.amzn2023.0.1.aarch64 from amazonlinux is filtered out by exclude filtering
2025-08-06T23:29:16Z: ==> amazon-ebs:   - package containerd-1.7.22-1.amzn2023.0.1.aarch64 from amazonlinux is filtered out by exclude filtering
2025-08-06T23:29:16Z: ==> amazon-ebs:   - package containerd-1.7.22-1.amzn2023.0.2.aarch64 from amazonlinux is filtered out by exclude filtering
2025-08-06T23:29:16Z: ==> amazon-ebs:   - package containerd-1.7.23-1.amzn2023.0.1.aarch64 from amazonlinux is filtered out by exclude filtering
2025-08-06T23:29:16Z: ==> amazon-ebs:   - package containerd-1.7.25-1.amzn2023.0.1.aarch64 from amazonlinux is filtered out by exclude filtering
2025-08-06T23:29:16Z: ==> amazon-ebs:   - package containerd-1.7.27-1.amzn2023.0.1.aarch64 from amazonlinux is filtered out by exclude filtering
2025-08-06T23:29:16Z: ==> amazon-ebs:   - package containerd-1.7.27-1.amzn2023.0.2.aarch64 from amazonlinux is filtered out by exclude filtering
2025-08-06T23:29:16Z: ==> amazon-ebs:   - package containerd-1.7.27-1.amzn2023.0.3.aarch64 from amazonlinux is filtered out by exclude filtering
2025-08-06T23:29:16Z: ==> amazon-ebs:   - package containerd-2.0.5-1.amzn2023.0.1.aarch64 from amazonlinux is filtered out by exclude filtering
2025-08-06T23:29:16Z: ==> amazon-ebs:   - package containerd-2.0.5-1.amzn2023.0.2.aarch64 from amazonlinux is filtered out by exclude filtering
```

We disabled the containerd install from AL repo and use the binary installed from internal s3 recently as containerd 1.7.* no longer supported in AL repo. here is the change https://github.com/awslabs/amazon-eks-ami/pull/2339
We need to install soci-snapshotter after containerd setup

nodeadm/internal/containerd/config.go
mselim00: oops duplicate, the code changed

templates/al2023/provisioners/install-worker.sh
wwvela: If move soci install after containerd still failed because of trying to install containerd from AL repo, we probably need to install the soci rpm without it's dependency. I verified these are required dependency for soci. containerd and runc and libseccomp should be installed after containerd setup phase.
```
bash-5.2# dnf install soci-snapshotter
Amazon Linux 2023 repository                     29 MB/s |  42 MB     00:01
Last metadata expiration check: 0:00:10 ago on Wed Aug  6 23:38:29 2025.
Dependencies resolved.
========================================================================================================================
 Package                       Architecture        Version                               Repository                Size
========================================================================================================================
Installing:
 soci-snapshotter              x86_64              0.11.1-1.amzn2023.0.1                 amazonlinux               23 M
Installing dependencies:
 containerd                    x86_64              2.0.5-1.amzn2023.0.2                  amazonlinux               26 M
 libseccomp                    x86_64              2.5.3-1.amzn2023.0.2                  amazonlinux               71 k
 runc                          x86_64              1.2.6-1.amzn2023.0.1                  amazonlinux              3.7 M

```

nodeadm/internal/system/resources.go
fletcherw: /sys/devices/system/memory/block_size_bytes is a hex value, hence the ParseInt 16. Thank you for checking me though.

nodeadm/internal/api/featuregates.go
fletcherw: Good point, done.

nodeadm/internal/system/resources.go
fletcherw: I added a comment and updated the name to SysfsResources.

nodeadm/internal/system/resources.go
mselim00: ah so I did miss something
```
ONLINE_BLOCKS=0; BLOCK_SIZE_BYTES=$(("0x$(cat /sys/devices/system/memory/block_size_bytes)")); for MEMORY_BLOCK_FILE in /sys/devices/system/memory/memory*; do ONLINE_BLOCKS=$((ONLINE_BLOCKS+1)); done && echo "$((ONLINE_BLOCKS*BLOCK_SIZE_BYTES))"
33151778816
```
very cool, this is much closer to the total. do you know what the gap is between this and total system memory though?

nodeadm/internal/containerd/config.go
fletcherw: made a constant.

nodeadm/internal/containerd/snapshotter/soci-snapshotter.config.toml
fletcherw: I just wanted to keep the config file a part of the feature flag for now.

nodeadm/internal/system/resources.go
fletcherw: not sure, presumably the kernel is keeping some memory for itself.

mselim00: that should be the difference between `MemTotal` and system memory though, which appears to be larger in this case, ref https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/6/html/deployment_guide/s2-proc-meminfo

not critical for this PR btw, it's a trivial difference for this use case. I'm just trying to understand the behavior because I have a use case in [another PR](https://github.com/awslabs/amazon-eks-ami/pull/2325) for this information where that difference does matter

templates/al2023/provisioners/install-worker.sh
```diff
@@ -250,6 +250,15 @@ sudo chmod +x $ECR_CREDENTIAL_PROVIDER_BINARY
 sudo mkdir -p /etc/eks/image-credential-provider
 sudo mv $ECR_CREDENTIAL_PROVIDER_BINARY /etc/eks/image-credential-provider/
 
+###############################################################################
+### SOCI Snapshotter setup ##########################################################
+###############################################################################
+
+# Use skip-broken as this won't recognize our locally-installed containerd1.x
+# TODO remove this flag when we've migrated off containerd1.x
```
wwvela: We won't remove this as we plan to use pure containerd binaries for both 1.* and 2.* eventually and will no longer reply on containerd in AL repo
```suggestion
# Use skip-broken as dnf will still try to install containerd from AL as a denpendency and it will break because we exclude containerd from AL repo to versionlock it. dnf can't recognize the containerd binaries we already setup.
```

wwvela:
/ci
+build binary_bucket_name=amazon-wesley-beta-ami containerd_version=1.7.27 arch=arm64

wwvela:
just initial another ci without `install_containerd_from_s3=true` to make sure it won't break external custom AMI builder

templates/al2023/provisioners/install-worker.sh
fletcherw: Updated comment, thanks.

templates/al2023/provisioners/install-worker.sh
```diff
@@ -250,6 +250,19 @@ sudo chmod +x $ECR_CREDENTIAL_PROVIDER_BINARY
...
+###############################################################################
+
+# Use skip-broken as dnf will still try to install containerd from Amazon Linux
+# as a dependency, and it will break because we explicitly exclude it in favor
+# of our manual version in order to version lock it. dnf doesn't recognize the
+# locally installed version.
+# TODO consider installing a stub package that provides containerd so that we can
+# remove this flag.
+sudo dnf install -y --skip-broken soci-snapshotter
+sudo systemctl enable soci-snapshotter.socket
```
mselim00: This would just skip installing soci when `containerd` is excluded. `dnf` is very opinionated about what it will install, it'll absolutely refuse to do anything that would break the specified chain of dependencies and offers no workaround.

mselim00 (commented):
Generally lgtm, I think we just need to drill down on how this plays with https://github.com/awslabs/amazon-eks-ami/pull/2339. 

In essence, we need a way to install `soci-snapshotter` even if `containerd` is excluded. The `rpm` CLI would let us do this with the `--nodeps`, but IMO we should only consider implementations that disregard the specific `containerd` deprecation since we know that one is satisfied. Though it seems to currently be the only dependency `soci` has, I'm not sure we want to own tracking the dependency chain to correct this logic when that change

nodeadm/internal/containerd/snapshotter/soci-snapshotter.config.toml
```diff
@@ -0,0 +1,21 @@
...
+concurrent_download_chunk_size = "16mb"
+discard_unpacked_layers = true
+max_concurrent_downloads = -1
+max_concurrent_downloads_per_image = 32
+max_concurrent_unpacks = -1
+max_concurrent_unpacks_per_image = 12
+
+[pull_modes.parallel_pull_unpack.decompress_streams."gzip"]
+path = '/usr/bin/unpigz'
+args = ['-d', '-c']
```
mselim00, line 21: Also, do you mind sharing some of the results of the benchmarking done for posterity?

Am I understanding correctly that the defaults that would be applied are these: https://github.com/awslabs/soci-snapshotter/blob/c8bc2dcc8b596075be5769c3800e3c2ecdd86166/docs/parallel-mode.md?plain=1#L29

If so, can you also talk a little bit about the changes?
```
enable: false -> true
concurrent_download_chunk_size: size_of_layer -> 16mb
discard_unpacked_layers: false -> true (in line with previous containerd config)
max_concurrent_downloads_per_image: 3 -> 32 (how do we interpret this? and any reason for this vs # of cores or something?)
max_concurrent_unpacks_per_image: 1 -> 12 (same as above)
```

templates/al2023/provisioners/install-worker.sh
wwvela: Yeah. Double checked in the succeeded ci build. soci is skipped
```
2025-08-07T00:26:45Z: ==> amazon-ebs: ================================================================================
2025-08-07T00:26:45Z: ==> amazon-ebs:  Package             Arch      Version                     Repository      Size
2025-08-07T00:26:45Z: ==> amazon-ebs: ================================================================================
2025-08-07T00:26:45Z: ==> amazon-ebs: Skipping packages with broken dependencies:
2025-08-07T00:26:45Z: ==> amazon-ebs:  soci-snapshotter    x86_64    0.11.1-1.amzn2023.0.1       amazonlinux     23 M
2025-08-07T00:26:45Z: ==> amazon-ebs:
2025-08-07T00:26:45Z: ==> amazon-ebs: Transaction Summary
2025-08-07T00:26:45Z: ==> amazon-ebs: ================================================================================
2025-08-07T00:26:45Z: ==> amazon-ebs: Skip  1 Package
```

Another work around is to install pure rmp file pkg using `--downloadonly` flag, it will install pkg and all its dependencies rpm. And remove the containerd one. And deploy these pure rpms using `rpmcpio` which won't trigger dnf 
```
bash-5.2# dnf install --downloadonly --downloaddir=/tmp/soci soci-snapshotter 
bash-5.2# cd /tmp/soci
bash-5.2# ls
containerd-2.0.5-1.amzn2023.0.2.x86_64.rpm
libseccomp-2.5.3-1.amzn2023.0.2.x86_64.rpm
runc-1.2.6-1.amzn2023.0.1.x86_64.rpm
soci-snapshotter-0.11.1-1.amzn2023.0.1.x86_64.rpm
```
If install it after runc install, it will only install containerd and soci rpms
```
bash-5.2# dnf install --downloadonly --downloaddir=/tmp/soci soci-snapshotter 
bash-5.2# cd /tmp/soci
bash-5.2# ls
containerd-2.0.5-1.amzn2023.0.2.x86_64.rpm
soci-snapshotter-0.11.1-1.amzn2023.0.1.x86_64.rpm
```

But yeah, it is a hacky way, and we might need to handle all pkg which rely on containerd this way

fletcherw:
/ci
+build binary_bucket_name=amazon-wesley-beta-ami containerd_version=1.7.27 install_containerd_from_s3=true arch=arm64

fletcherw:
/ci
+build binary_bucket_name=amazon-wesley-beta-ami containerd_version=1.7.27 install_containerd_from_s3=true arch=arm64

wwvela:
/ci
+workflow:k8s_versions 1.33
+build binary_bucket_name=amazon-wesley-beta-ami containerd_version=1.7.27 install_containerd_from_s3=true arch=x86_64 enable_accelerator=nvidia

mselim00:
trying this out with the feature enabled

/ci
+test --nodeadm-feature-gates=AggressiveImagePullTradeoff=true

nodeadm/api/v1alpha1/nodeconfig_types.go
```diff
@@ -129,10 +129,16 @@ const (
...
+// +kubebuilder:validation:Enum={InstanceIdNodeName,AggressiveImagePullTradeoff}
 type Feature string
 
 const (
 	// InstanceIdNodeName will use EC2 instance ID as node name
 	InstanceIdNodeName Feature = "InstanceIdNodeName"
+
+	// AggressiveImagePullTradeoff enables a parallel image pull for container
+	// images. This will use more instance CPU and Memory during image pull, but
+	// may result in faster image pull times. This flag cannot be enabled on
```
mselim00:
```suggestion
	// may result in faster image pull times. This flag will be ignored on
```

nodeadm/api/v1alpha1/nodeconfig_types.go
```diff
@@ -129,10 +129,16 @@ const (
 )
 
 // Feature specifies which feature gate should be toggled
-// +kubebuilder:validation:Enum={InstanceIdNodeName}
+// +kubebuilder:validation:Enum={InstanceIdNodeName,AggressiveImagePullTradeoff}
```
cartermckinnon: Can we call this `AggressiveImagePull`? Not a hill I need to die on if you think `Tradeoff` is adding something notable

nodeadm/api/v1alpha1/nodeconfig_types.go
```diff
@@ -129,10 +129,16 @@ const (
...
 type Feature string
 
 const (
 	// InstanceIdNodeName will use EC2 instance ID as node name
 	InstanceIdNodeName Feature = "InstanceIdNodeName"
+
+	// AggressiveImagePullTradeoff enables a parallel image pull for container
+	// images. This will use more instance CPU and Memory during image pull, but
+	// may result in faster image pull times. This flag cannot be enabled on
+	// instances with memory and vCPU below a certain threshold.
```
cartermckinnon: Because our doc gen doesn't use this (bummer), can you add a block to the docs that describe the feature? Something in examples, like what we have here: https://github.com/awslabs/amazon-eks-ami/blob/main/nodeadm/doc/examples.md#using-instance-id-as-node-name-experimental

nodeadm/internal/api/featuregates.go
```diff
@@ -14,6 +14,12 @@ var featureVerifiers = map[Feature]func(Feature, map[Feature]bool) bool{
 	// InstanceIdNodeNameGate controls whether to use instance ID as the node's name.
 	// By default, this feature is disabled, and the private DNS Name will be used.
 	InstanceIdNodeName: DefaultFalse,
+
+	// AggressiveImagePullTradeoff enables a parallel image pull for container
+	// images. This will use more instance CPU and Memory during image pull, but
+	// may result in faster image pull times. This flag will be ignored on
+	// instances with memory and vCPU below a certain threshold.
+	AggressiveImagePullTradeoff: DefaultFalse,
```
cartermckinnon: (not blocking) I think we need to extend `Feature` to support additional, arbitrary requirements (like the system resources in this case). The `containerd` package doesn't seem like the best place for it IMO, but it'll work for now.

nodeadm/internal/api/types.go
```diff
@@ -122,4 +122,10 @@ type Feature string
 const (
 	// InstanceIdNodeName will use EC2 instance ID as node name
 	InstanceIdNodeName Feature = "InstanceIdNodeName"
+
+	// AggressiveImagePullTradeoff enables a parallel image pull for container
+	// images. This will use more instance CPU and Memory during image pull, but
+	// may result in faster image pull times. This flag will be ignored on
+	// instances with memory and vCPU below a certain threshold.
```
cartermckinnon: for future ref, this `internal/api` package is sealed so no need to replicate the doc comments

nodeadm/api/v1alpha1/nodeconfig_types.go
fletcherw: Done, no strong preference.

nodeadm/api/v1alpha1/nodeconfig_types.go
fletcherw: Shoot fixed, thanks for catching this one as well.

nodeadm/api/v1alpha1/nodeconfig_types.go
fletcherw: Done.

nodeadm/internal/api/featuregates.go
fletcherw: That's fair, the way this is implemented now also repeatedly checks CPU/Memory every time we check the flag value. Having a notion of config separate from the raw config file would probably work.

nodeadm/internal/api/types.go
fletcherw: thanks, removed.

nodeadm/doc/examples.md
```diff
@@ -91,6 +91,28 @@ spec:
...
+
+When the `AggressiveImagePull` feature gate is enabled, `nodeadm` will configure the container runtime to pull and unpack container images in parallel.
+
+This has the benefit of potentially decreasing image pull time, at the cost of increased CPU and memory usage during image pull.
+
+⚠️ **Note**: This flag will be ignored on instance sizes below a certain vCPU and memory threshold.
+
+### To enable this feature:
+1. Ensure your instance type is a larger instance type. Currently we recommend a 2xlarge instance or larger, but that value may change.
+2. Make sure your workloads can tolerate the increased CPU and memory usage during image pull.
```
cartermckinnon:
```suggestion
2. Make sure your workloads can tolerate the increased CPU and memory usage during image pull. This makes the most sense when you need to pull a very large container image early in a node's lifecycle, before other workloads are running.
```

nodeadm/internal/api/featuregates.go
cartermckinnon: Yeah that was one thing that caught my eye. We could do some magic with `sync.Once` for the additional checks, either on the `Feature` or in `system` for this specific check. We can iterate in later PR's

wwvela (approved):

Issue #2354, opened 2025-08-06
bug(AL2023): iptable_nat module not loaded, breaking Istio and other iptables-legacy users in v20250801


nnductuan:
**What happened**:

After upgrading from AMI `amazon-eks-node-al2023-x86_64-standard-1.33-v20250704` to `amazon-eks-node-al2023-x86_64-standard-1.33-v20250801`, Istio init containers fail with CrashLoopBackOff. 

The `istio-init` container cannot initialize iptables NAT rules because the `iptable_nat` kernel module is not loaded (It was loaded in the old AMI). The module exists on the system (`modinfo iptable_nat` shows it as "iptables legacy nat table") but is not auto-loaded when containers attempt to use iptables commands.

Error from istio-init:
```
{"level":"error","time":"2025-08-05T07:08:22.818942Z","msg":"Command error: xtables parameter problem: iptables-restore: unable to initialize table 'nat'\nError occurred at line: 1"}
{"level":"error","time":"2025-08-05T07:08:22.821469Z","msg":"exit status 2\niptables-restore v1.8.10 (legacy): iptables-restore: unable to initialize table 'nat'"}
```

Verification shows:
- `lsmod | grep iptable_nat` returns nothing (module not loaded)
- After `sudo modprobe iptable_nat`, istio-init containers work correctly

**What you expected to happen**:
- I'm reporting this issue for awareness if anyone wants to upgrade the AMI - they should know about this change 
- Perhaps a script could be added to the AMI to load the legacy iptables modules (iptable_nat, etc.) during node startup if possible, to make the AMI backward compatible.

**Environment**:
- Cluster Kubernetes version: 1.33
- Node Kubernetes version: 1.33
- AMI Version: `amazon-eks-node-al2023-x86_64-standard-1.33-v20250801`
- Previous working AMI: `amazon-eks-node-al2023-x86_64-standard-1.33-v20250704`
- Istio version: 1.25

cartermckinnon:
I don’t see any reason this would have changed between these AMI releases. Do you see a difference in the /etc/modprobe.d/ config files? Have you tried the 20250715 release, to narrow things down?

ayushj0909:
Im facing the same issues with new AMIs. I think this started after the July 10th release. Never using @latest in my EC2NodeClass again.

shvbsle:
[...]

PR #2355, opened 2025-08-07
WIP: load ip_tables and iptable_nat for AMIs with kernel 6.12

shvbsle:
**Issue #, if available:**
https://github.com/awslabs/amazon-eks-ami/issues/2354

**Description of changes:**
The dependency for `iptable_nat` module changes across kernel versions (kernel 6.1 to kernel 6.12). On kernel 6.12, it depends on `ip_tables` and `nf_nat` modules but on kernel 6.1 its only dependency is `nf_nat`. In both kernels `nf_nat` is loaded but newer dependent kernel module are not.

**Testing Done**

### DIFF
--- a/templates/al2023/provisioners/install-worker.sh
+++ b/templates/al2023/provisioners/install-worker.sh
@@ -111,6 +111,15 @@ cat << EOF | sudo tee /etc/systemd/network/99-default.link.d/99-no-policy.conf
 MACAddressPolicy=none
 EOF
 
+if [[ "$(uname -r)" == 6.12.* ]]; then
+  # Create module loading configuration for iptable_nat on kernel 6.12
+  cat << EOF | sudo tee -a /etc/modules-load.d/iptables.conf
+ip_tables
+iptable_nat
+EOF
+fi
+
+
 ################################################################################
 ### SSH ########################################################################
 ################################################################################


templates/al2023/provisioners/install-worker.sh
```diff
@@ -111,6 +111,14 @@ cat << EOF | sudo tee /etc/systemd/network/99-default.link.d/99-no-policy.conf
 MACAddressPolicy=none
 EOF
 
+if [[ "$(uname -r)" == 6.12.* ]]; then
```
fletcherw: do we want a >= here or are we assuming that this will be fixed in newer kernel versions?

cartermckinnon: I think we can just create this as a static file in the `rootfs` tree, any downside to it on 6.1, etc.?

shvbsle: > any downside to it on 6.1, etc.?

No obvious downsides but given that we don't load this module for AMIs on 6.1 kernel by default, and such AMIs are in a stable state, I'm thinking we only stick to AMIs with 6.12+ kernels since they are relatively new and reduced risk of unintentionally breaking something for already stable AMIs

shvbsle: I just confirmed that on 6.1 kernels `ip_tables` (the other dependency of `iptable_nat`) is a built-in kernel module. So I think it should be safe to add it for all 6* kernels. I'll send a revision

fletcherw (approved):

cartermckinnon (changes requested):
We're talking to some Amazon Linux folks about this before making this change.

cartermckinnon:
We discussed a few options with Amazon Linux, and this workaround makes the most sense. I do think the burden of compatibility is ultimately on Istio -- I'm not sure if the Istio component is using the wrong `iptables` binary, or an old version, but this problem is likely to occur on other 6.12+ kernel builds and should be handled by Istio.

PR #2361, opened 2025-08-11
feat(ci): handle kubetest2 deployer and tester args

mselim00:
**Issue #, if available:**

**Description of changes:**
Splits up the arguments passable to the CI into test and build arguments. The two goals are still `build` and `test`, but the arguments that are parsed in the `test` goal are more granular. Arguments that were previously passed to `test` will need to be passed to `deploy`, `test` will now be reserved for specific test suites (e.g. `+test --instance-types g6f.2xlarge` -> `+deploy --instance-types g6f.2xlarge`

**Testing Done**


.github/actions/ci/kubetest2/action.yaml
```diff
@@ -35,25 +38,26 @@ runs:
...
             ;;
           *)
             echo >&2 "unknown os_distro: ${{ inputs.os_distro }}"
             exit 1
           ;;
         esac
 
-        if [ ! "${{ inputs.additional_arguments }}" = "" ]; then
-          KUBETEST2_ARGS="${KUBETEST2_ARGS} ${{ inputs.additional_arguments }}"
+        if [ ! "${{ inputs.deployer_arguments }}" = "" ]; then
```
fletcherw: `deployer` vs `deploy` ?

.github/actions/ci/kubetest2/action.yaml
```diff
@@ -18,7 +18,10 @@ inputs:
   log_bucket:
     required: true
     type: string
-  additional_arguments:
```
fletcherw, line 21: do any existing actions use this param?

mselim00: AFAICT this action's only called by workflows/ci.yaml, which is changed here. https://github.com/search?q=repo%3Aawslabs%2Famazon-eks-ami%20additional_arguments&type=code

.github/actions/ci/kubetest2/action.yaml
mselim00: yeah I wasn't sure about this one either tbh. kubetest2 is split up into `deployer` and `tester` phases, which I wanted to honor here. our CI has traditionally used `build` and `test`arguments though, which I was thinking makes sense to keep as an imperative. not sure where in the internals the transition makes the most sense

cartermckinnon: Verb makes more sense to me for the user-facing stuff, don’t think it matters much for the workflow internals

fletcherw: sorry to be clear I was flagging that your if condition said `deployer_arguments` and your if body used `deploy_arguments`.

Either is fine to me.

mselim00: ah lol, I was looking at this in the conversations page so didn't see that in the snippet. fixed, thanks

.github/actions/ci/kubetest2/action.yaml
```diff
@@ -35,25 +38,26 @@ runs:
...
-          KUBETEST2_ARGS="${KUBETEST2_ARGS} ${{ inputs.additional_arguments }}"
+        if [ ! "${{ inputs.deploy_arguments }}" = "" ]; then
+          DEPLOYER_ARGS="${DEPLOYER_ARGS} ${{ inputs.deploy_arguments }}"
         fi
 
-        TESTER_NAME=ginkgo
-        # k8s_version below 1.26 use the older ginkgo tester
-        if [ "${{ inputs.k8s_version }}" = "$(echo -e "${{ inputs.k8s_version }}\n1.26" | sort -V | head -n1)" ]; then
-          TESTER_NAME=ginkgo-v1
+        # run k8s conformance suite if no tests specified
```
cartermckinnon, line 56: I think we probably want to run the conformance tests in all cases, and allow additional suites to be enabled with wordy args or a more convenient CI goal (`test-nvidia`, `test-gpu`, whatever)

cartermckinnon:
This will be nice for super specific runs, but I think we probably want some kind of shorthand for nvidia/gpu tests (if that's the broader motivation for the PR)

.github/actions/ci/kubetest2/action.yaml
mselim00: Agreed, but there's unfortunately a small intersect of the current GPU tests in there that are expected to succeed on all `nvidia` instance types (currently just the [single-node](https://github.com/aws/aws-k8s-tester/blob/331138c5faf89c942c4beadcc3550e97954b458a/test/cases/nvidia/mpi_test.go#L144) and [bert-inference](https://github.com/aws/aws-k8s-tester/blob/main/test/cases/nvidia-inference/bert_inference_test.go#L37) tests AFAICT), so I think configuration would still end up being needed in many cases

I've been wanting to implement a plugin for `aws-k8s-tester` that'll just dump out supported tests for a given instance type, maybe that's worth implementing first and then using here

cartermckinnon: If something like that is possible to write, then we should be able to push that logic into the testers, I think? Ideally we just run all the tests that are applicable for the deployed cluster, instead of making the user spell everything out explicitly

mselim00: yeah I think so too. to be directly part of the tester, I think would require some re-structuring so it's all one executable that both identifies the needed tests and runs them. would still need a way to handle necessary configurations like the image URI to use for the test pods but maybe these can all be achieved with reasonable defaults

PR #2369, opened 2025-08-13
feat: run shellcheck on all shell files

fletcherw:
* Update existing shellcheck workflow to find all shell files, not just those ending in .sh
* Fix existing shellcheck errors.
* Split lint step into two separate steps since the lint-docs step is slow.

### DIFF
--- a/.github/workflows/ci-auto.yaml
+++ b/.github/workflows/ci-auto.yaml
@@ -6,13 +6,17 @@ on:
       - reopened
       - synchronize
 jobs:
-  lint:
+  lint-docs:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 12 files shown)


cartermckinnon:
We actually already do this, as part of the `lint` check: https://github.com/awslabs/amazon-eks-ami/blob/69b114045d41dc3ea16b52cddefe8f4bbb9d2475/Makefile#L62

The filter for `*.sh` filenames is probably why you're seeing a non-empty diff. I don't remember why that was added, so you could try removing it

fletcherw:
Added a script for finding shell files in the repo that also looks at shebangs.

cartermckinnon (approved):

PR #2371, opened 2025-08-13
chore: drop shellcheck level to warning

fletcherw:
* update shellcheck presubmit to report warnings and above
* use shfmt's logic for finding shell files instead of my hacky script
* fix most existing warnings

**Testing Done**


templates/shared/runtime/bin/imds
```diff
@@ -23,8 +23,10 @@ function log() {
 }
 
 function imdscurl() {
-  local OUTPUT_FILE=$(mktemp)
-  local CODE=$(curl \
+  local OUTPUT_FILE
+  OUTPUT_FILE=$(mktemp)
+  local CODE
+  CODE=$(curl \
```
cartermckinnon, line 29: I'm lukewarm on bumping the shellcheck level, because the findings get increasingly subjective (like this one). I chose `error` initially because it gave us enough of a guardrail without getting nit picky / being overly-defensive. Did you find a notable bug somewhere in here that justifies going to `warning`?

fletcherw: I see where you're coming from. This split definition and assignment is not great. It's probably only really an issue if we have scripts that set errexit and then don't set nounset, since whatever we assigned to will error out later as soon as it's used.

If you think it's too much noise, I'm happy to silence that check and just ensure that our scripts all `set -euo pipefail`.

There were no huge errors that I caught, but a lot of unused variables, setting of AWS command options that had no effect, and some code that would break if any variables had spaces in them.

I think a lot of the array warnings can be kind of confusing and would be open to disabling those as well.

mselim00: FWIW this actually is a functional difference (and the suggestion is better) &mdash; the exit code of the local variable allocation supersedes that of the assignment. So in the existing code, we could see a zero exit code even if the temporary file fails to create. Spent a bundle of time stuck on an issue arising from this in the past 😅

fletcherw: I agree it's a functional difference, in practice though won't the failure be caught when you attempt to use the variable assuming you've set `nounset`?

ndbaker1: I'm on the side that this isn't just subjective, per the case that mazen mentioned, but the impact is definitely conditional based on the other options like `errexit`.

rather than worrying about interactions on a case by case basis i think it makes sense to just follow whatever the best practices are. although I'm also assuming that `warning` doesn't additionally contain pedantic stuff that's actually unnecessary 😉

cartermckinnon: It's not a hill I need to die on, I just don't think it's adding a ton of value. We can always just turn it off if it's annoying.

cartermckinnon (approved):

cartermckinnon:
@fletcherw can you open a PR to add this commit to https://github.com/awslabs/amazon-eks-ami/blob/main/.git-blame-ignore-revs?

PR #2372, opened 2025-08-14
feat(nodeadm): auto-collect diagnostics on failure

mselim00:
**Issue #, if available:**

**Description of changes:**
WIP

**Testing Done**


nodeadm/internal/logshim/logs.go
```diff
@@ -0,0 +1,51 @@
...
+import (
+	"fmt"
+	"os/exec"
+	"path/filepath"
+	"strings"
+)
+
+const logCollectorScript = "/etc/eks/log-collector-script/eks-log-collector.sh"
+const baseOutputDirectory = "/etc/eks/nodeadm/"
+
```
ndbaker1, line 12: just noticing that this really reduces down to a post-fail-hook,
if we gave customers the ability to provide failure scripts, they could verify easily do something like this:
```yaml
diagnosticOptions:
  failureScript: |
    /etc/eks/log-collector-script/eks-log-collector.sh --output XXX.tgz
    aws s3 cp XXX.tgz s3://mybucket/mydiagnostics/
```

or maybe they just write a systemd unit that watches for the failure of the `nodeadm-run` service.

not that we'd want to do this, and I like the struct that this PR brings and how it would make investigation more standardized if it were enabled by default.

nodeadm/cmd/nodeadm/init/init.go
```diff
@@ -45,28 +48,52 @@ func (c *initCmd) Flaggy() *flaggy.Subcommand {
...
-
-	log.Info("Checking user is root..")
-	root, err := cli.IsRunningAsRoot()
-	if err != nil {
-		return err
-	} else if !root {
-		return cli.ErrMustRunAsRoot
+func (c *initCmd) collectDiagnostics(ctx context.Context, log *zap.Logger, nodeConfig *api.NodeConfig) error {
+	if !api.IsFeatureEnabled(api.AutoCollectDiagnostics, nodeConfig.Spec.FeatureGates) {
+		return nil
```
ndbaker1, line 53: i think we just might want this as new, toggle-able behavior, not so much a featureGate? i.e. just a `diagnosticOptions` "enabled" flag that has a default?

log-collector-script/linux/eks-log-collector.sh
```diff
@@ -321,14 +324,14 @@ collect() {
 pack() {
   try "archive gathered information"
 
-  tar --create --verbose --gzip --file "${LOG_DIR}"/eks_"${INSTANCE_ID}"_"${CURRENT_TIME}"_"${PROGRAM_VERSION}".tar.gz --directory="${COLLECT_DIR}" . > /dev/null 2>&1
+  tar --create --verbose --gzip --file "${OUTPUT_DIR}"/eks_"${INSTANCE_ID}"_"${CURRENT_TIME}"_"${PROGRAM_VERSION}".tar.gz --directory="${COLLECT_DIR}" . > /dev/null 2>&1
```
ndbaker1, line 327: would output file just be more powerful than output dir?

nodeadm/api/v1alpha1/nodeconfig_types.go
```diff
@@ -29,10 +29,11 @@ type NodeConfigList struct {
 }
 
 type NodeConfigSpec struct {
-	Cluster    ClusterDetails    `json:"cluster,omitempty"`
-	Containerd ContainerdOptions `json:"containerd,omitempty"`
-	Instance   InstanceOptions   `json:"instance,omitempty"`
-	Kubelet    KubeletOptions    `json:"kubelet,omitempty"`
+	Cluster            ClusterDetails     `json:"cluster,omitempty"`
+	Containerd         ContainerdOptions  `json:"containerd,omitempty"`
+	DiagnosticsOptions DiagnosticsOptions `json:"diagnosticsOptions,omitempty"`
```
ndbaker1, line 34:
```suggestion
	Diagnostics DiagnosticsOptions `json:"diagnostics,omitempty"`
```

nodeadm/internal/logshim/logs.go
mselim00: this is very true, and the simplicity of doing this in user data does make the complexity of this PR seem not worth it. the one thing that user data doesn't cover (that this PR didn't address either) is running a diagnostic **during** nodeadm-config. that currently requires creating a custom AMI with a service built-in since `nodeadm-config` precedes `cloud-init`,  whereas `nodeadm-config` could potentially execute a `prestartHook` to carry out that diagnostic. the use case for this is probably super niche though 

diagnostics **after** `nodeadm-config` can be trivially collected in user data because user data scripts execute after `nodeadm-config`, e.g.
```
MIME-Version: 1.0
Content-Type: multipart/mixed; boundary="BOUNDARY"

--BOUNDARY
Content-Type: text/x-shellscript; charset="us-ascii"

#!/bin/bash
/etc/eks/log-collector-script/eks-log-collector.sh

--BOUNDARY
Content-Type: application/node.eks.aws

---
apiVersion: node.eks.aws/v1alpha1
kind: NodeConfig
spec: ...

--BOUNDARY--
```

diagnostics **after** `nodeadm-run` or `kubelet` are a bit more complex because they can't natively use `systemd` ordering directives to accomplish it through a service (the `start` would block indefinitely since `nodeadm-run` needs to start after `cloud-init` which would be executing the `start`). as you suggested, users would need to instead directly watch for the status of the service and then execute their post script. in any case though, I think there's more cases where running something after `nodeadm-config` is helpful rather than `nodeadm-run`

mselim00:
closing based on the discussion in https://github.com/awslabs/amazon-eks-ami/pull/2372#discussion_r2279586019

PR #2378, opened 2025-08-18
chore(nodeadm): tweak image pull flag config and docs

fletcherw:
Update the max concurrent value based on some testing.

Add recommendations about EBS throughput to docs.

### DIFF
--- a/nodeadm/internal/containerd/snapshotter/soci-snapshotter.config.toml
+++ b/nodeadm/internal/containerd/snapshotter/soci-snapshotter.config.toml
@@ -12,7 +12,7 @@ enable = true
 concurrent_download_chunk_size = "16mb"
 discard_unpacked_layers = true
 max_concurrent_downloads = -1
-max_concurrent_downloads_per_image = 32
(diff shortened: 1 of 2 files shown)


nodeadm/internal/containerd/snapshotter/soci-snapshotter.config.toml
```diff
@@ -12,7 +12,7 @@ enable = true
 concurrent_download_chunk_size = "16mb"
 discard_unpacked_layers = true
 max_concurrent_downloads = -1
-max_concurrent_downloads_per_image = 32
+max_concurrent_downloads_per_image = 20
```
ndbaker1, line 15: are there any testing anecdotes to share? was it % faster with this config for example?

fletcherw: from tests, it was sometimes a little bit faster with 32 vs 20, but it was more likely to hit throttling on the API. 20 gets the majority of the speed up.

Issue #2128, opened 2025-01-28
feat(nodeadm): support for HTTP proxy


MageshSrinivasulu:
In EKS 1.29, we were using Amazon Linux 2 (AL2) with a user data script that configured proxy environment settings before executing the bootstrap.sh script, and everything was functioning properly.

However, after upgrading to EKS 1.30, we encountered some breaking changes, primarily with nodeadm, which initializes as the node boots up. As a result, I'm noticing a few things that are not working as expected. I suspect this behaviour is mainly related to timing or the order of execution.

**Issue: nodeadm failing with some requests getting timed out because of the proxy details are not getting set**

I believe nodeadm executes several tasks in parallel, causing some to run before the user_data script. Modifications done by user_data script are crucial as they configure the proxy details in the environment rest of the services depend on that for making the API call. 

Note: The execution of the user_data script completes despite the error, as I can observe the intended changes in the node.

Below is the error that we got when running `journalctl -u nodeadm-config`

```
Jan 28 10:04:51 localhost systemd[1]: Starting nodeadm-config.service - EKS Nodeadm Config...
 
Jan 28 10:04:53 localhost nodeadm[2833]: {"level":"info","ts":1738058693.179373,"caller":"init/init.go:51","msg":"Checking user is root.."}
 
Jan 28 10:04:53 localhost nodeadm[2833]: {"level":"info","ts":1738058693.179425,"caller":"init/init.go:59","msg":"Loading configuration..","configSource":"imds://user-data"}

Jan 28 10:04:53 localhost nodeadm[2833]: {"level":"info","ts":1738058693.333687,"caller":"init/init.go:70","msg":"Enriching configuration.."}
 
Jan 28 10:04:53 localhost nodeadm[2833]: {"level":"info","ts":1738058693.3336952,"caller":"init/init.go:152","msg":"Fetching instance details.."}
 
Jan 28 10:04:53 localhost nodeadm[2833]: SDK 2025/01/28 10:04:53 DEBUG attempting waiter request, attempt count: 1
 
Jan 28 10:05:24 localhost nodeadm[2833]: SDK 2025/01/28 10:05:24 DEBUG retrying request EC2/DescribeInstances, attempt 2
 
Jan 28 10:05:55 localhost nodeadm[2833]: SDK 2025/01/28 10:05:55 DEBUG retrying request EC2/DescribeInstances, attempt 3
 
[...]

PR #2384, opened 2025-08-19
WIP: feat(nodeadm): add support to configure http-proxy for al2023/nodeadm

shvbsle:
**Issue #, if available:**
https://github.com/awslabs/amazon-eks-ami/issues/2128

**Description of changes:**
Nodeadm has two phases: 1. Config and 2. Run. Nodeadm has a set of outbound API calls (IMDS, AWS API calls) that could potentially need to be routed through a proxy. Currently, setting proxy configurations would happen after the config phase  in user-data. This PR adds support to configure http-proxy in `NodeConfigSpec`. 

TODOs:

- [ ] Add test cases for init.go
- [ ] Add test cases for proxy_env.go
- [ ] Add test cases for validation.go

**Testing Done**

### DIFF
--- a/nodeadm/api/v1alpha1/nodeconfig_types.go
+++ b/nodeadm/api/v1alpha1/nodeconfig_types.go
@@ -33,6 +33,7 @@ type NodeConfigSpec struct {
 	Containerd ContainerdOptions `json:"containerd,omitempty"`
 	Instance   InstanceOptions   `json:"instance,omitempty"`
 	Kubelet    KubeletOptions    `json:"kubelet,omitempty"`
+	Proxy      ProxyOptions      `json:"proxy,omitempty"`
 	// FeatureGates holds key-value pairs to enable or disable application features.
 	FeatureGates map[Feature]bool `json:"featureGates,omitempty"`
 }
@@ -85,6 +86,19 @@ type ContainerdOptions struct {
 // InstanceOptions determines how the node's operating system and devices are configured.
 type InstanceOptions struct {
 	LocalStorage LocalStorageOptions `json:"localStorage,omitempty"`
+}
+
+// ProxyOptions configures HTTP proxy settings for nodeadm's outbound network connections.
+// These settings are used when nodeadm makes AWS API calls during node initialization.
+type ProxyOptions struct {
+	// HTTPProxy specifies the HTTP proxy server URL for non-TLS requests.
+	HTTPProxy string `json:"httpProxy,omitempty"`
+
+	// HTTPSProxy specifies the HTTP proxy server URL for TLS requests.
+	HTTPSProxy string `json:"httpsProxy,omitempty"`
+
+	// NoProxy specifies hosts and domains that should bypass the proxy.
+	NoProxy []string `json:"noProxy,omitempty"`
 }
 
 // LocalStorageOptions control how [EC2 instance stores](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/InstanceStorage.html)
--- a/nodeadm/cmd/nodeadm/init/init.go
+++ b/nodeadm/cmd/nodeadm/init/init.go
@@ -67,6 +67,10 @@ func (c *initCmd) Run(log *zap.Logger, opts *cli.GlobalOptions) error {
 	}
 	log.Info("Loaded configuration", zap.Reflect("config", nodeConfig))
 
+	// Set proxy environment variables early so all AWS SDK calls use them
+	log.Info("Configuring proxy environment variables..")
+	setProxyEnvironmentVariables(nodeConfig.Spec.Proxy)
+
 	log.Info("Enriching configuration..")
 	if err := enrichConfig(log, nodeConfig); err != nil {
 		return err
--- a/nodeadm/cmd/nodeadm/init/proxy_env.go
+++ b/nodeadm/cmd/nodeadm/init/proxy_env.go
@@ -1,0 +1,67 @@
+package init
+
+import (
+	"os"
+	"strings"
+
+	"github.com/awslabs/amazon-eks-ami/nodeadm/internal/api"
+)
+
+// setProxyEnvironmentVariables configures HTTP_PROXY, HTTPS_PROXY, and NO_PROXY
+// environment variables based on the provided ProxyOptions.
+// This allows AWS SDK v2 to automatically use proxy configuration.
+func setProxyEnvironmentVariables(proxyOpts api.ProxyOptions) {
+	// Set HTTP_PROXY if configured
+	if proxyOpts.HTTPProxy != "" {
+		os.Setenv("HTTP_PROXY", proxyOpts.HTTPProxy)
+	}
+
+	// Set HTTPS_PROXY if configured
+	if proxyOpts.HTTPSProxy != "" {
+		os.Setenv("HTTPS_PROXY", proxyOpts.HTTPSProxy)
+	}
+
+	// Build NO_PROXY list from user-specified patterns
+	noProxyList := buildNoProxyList(proxyOpts.NoProxy)
+	if len(noProxyList) > 0 {
+		os.Setenv("NO_PROXY", strings.Join(noProxyList, ","))
+	}
+}
+
+// Create NO_PROXY list that includes:
+// 1. User-specified NoProxy patterns
+// 2. Standard localhost
+func buildNoProxyList(userNoProxy []string) []string {
+	noProxyList := []string{
+		"localhost",
+		"127.0.0.1",
+	}
+
+	for _, pattern := range userNoProxy {
+		pattern = strings.TrimSpace(pattern)
+		if pattern == "" {
+			continue
+		}
+
+		exists := false
+		for _, existing := range noProxyList {
+			if existing == pattern {
+				exists = true
+				break
+			}
+		}
+
+		if !exists {
+			noProxyList = append(noProxyList, pattern)
+		}
+	}
+
+	return noProxyList
+}
+
+// This can be used for cleanup or testing purposes.
+func clearProxyEnvironmentVariables() {
+	os.Unsetenv("HTTP_PROXY")
+	os.Unsetenv("HTTPS_PROXY")
+	os.Unsetenv("NO_PROXY")
+}
--- a/nodeadm/internal/api/types.go
+++ b/nodeadm/internal/api/types.go
@@ -32,6 +32,7 @@ type NodeConfigSpec struct {
 	Containerd   ContainerdOptions `json:"containerd,omitempty"`
 	Instance     InstanceOptions   `json:"instance,omitempty"`
 	Kubelet      KubeletOptions    `json:"kubelet,omitempty"`
+	Proxy        ProxyOptions      `json:"proxy,omitempty"`
 	FeatureGates map[Feature]bool  `json:"featureGates,omitempty"`
 }
 
@@ -94,6 +95,12 @@ const (
 
 type InstanceOptions struct {
 	LocalStorage LocalStorageOptions `json:"localStorage,omitempty"`
+}
+
+type ProxyOptions struct {
+	HTTPProxy  string   `json:"httpProxy,omitempty"`
+	HTTPSProxy string   `json:"httpsProxy,omitempty"`
+	NoProxy    []string `json:"noProxy,omitempty"`
 }
 
 type LocalStorageOptions struct {
--- a/nodeadm/internal/api/validation.go
+++ b/nodeadm/internal/api/validation.go
@@ -1,6 +1,10 @@
 package api
 
-import "fmt"
+import (
+	"fmt"
+	"net/url"
+	"strings"
+)
 
 func ValidateNodeConfig(cfg *NodeConfig) error {
 	if cfg.Spec.Cluster.Name == "" {
@@ -20,5 +24,34 @@ func ValidateNodeConfig(cfg *NodeConfig) error {
 			return fmt.Errorf("CIDR is missing in cluster configuration")
 		}
 	}
+
+	// TODO: Add detailed proxy configuration validation
+	if err := validateProxyOptionsBasic(cfg.Spec.Proxy); err != nil {
+		return fmt.Errorf("proxy configuration validation failed: %w", err)
+	}
+
+	return nil
+}
+
+// TODO: Expand this to include detailed URL validation, NoProxy pattern validation etc
+func validateProxyOptionsBasic(proxy ProxyOptions) error {
+	if proxy.HTTPProxy != "" {
+		if _, err := url.Parse(proxy.HTTPProxy); err != nil {
+			return fmt.Errorf("invalid HTTP_PROXY URL: %w", err)
+		}
+	}
+
+	if proxy.HTTPSProxy != "" {
+		if _, err := url.Parse(proxy.HTTPSProxy); err != nil {
+			return fmt.Errorf("invalid HTTPS_PROXY URL: %w", err)
+		}
+	}
+
+	for i, pattern := range proxy.NoProxy {
+		if strings.TrimSpace(pattern) == "" {
+			return fmt.Errorf("empty NO_PROXY pattern at index %d", i)
+		}
+	}
+
 	return nil
 }


nodeadm/api/v1alpha1/nodeconfig_types.go
```diff
@@ -33,6 +33,7 @@ type NodeConfigSpec struct {
 	Containerd ContainerdOptions `json:"containerd,omitempty"`
 	Instance   InstanceOptions   `json:"instance,omitempty"`
 	Kubelet    KubeletOptions    `json:"kubelet,omitempty"`
+	Proxy      ProxyOptions      `json:"proxy,omitempty"`
```
cartermckinnon: I think we probably want to move this into `InstanceOptions`

nodeadm/cmd/nodeadm/init/proxy_env.go
```diff
@@ -0,0 +1,67 @@
...
+	"os"
+	"strings"
+
+	"github.com/awslabs/amazon-eks-ami/nodeadm/internal/api"
+)
+
+// setProxyEnvironmentVariables configures HTTP_PROXY, HTTPS_PROXY, and NO_PROXY
+// environment variables based on the provided ProxyOptions.
+// This allows AWS SDK v2 to automatically use proxy configuration.
+func setProxyEnvironmentVariables(proxyOpts api.ProxyOptions) {
```
cartermckinnon: We want users to be able to set a proxy for `nodeadm`, but we also want to set the proxy for everything else on the system (otherwise folks would need to handle that with a user data script). So these values need to end up in `/etc/environment`.

I wonder if we should add a generic way to pass things to `/etc/environment`, and then just ensure that those variables are utilized by `nodeadm` (including the proxy-related ones).

e.g.

```
instance:
  environment:
    HTTP_PROXY: http://example.com
    NO_PROXY: amazon.com
```

That would be a lot more flexible; we could feasibly respect well-known AWS SDK config variables, etc. in addition to the proxy stuff.

shvbsle: addressed in latest revisions

nodeadm/cmd/nodeadm/init/environment.go
cartermckinnon, line 1: This should probably go in `internal/system/environment.go`

nodeadm/cmd/nodeadm/init/environment.go
```diff
@@ -0,0 +1,81 @@
...
+		if err := os.Setenv(key, value); err != nil {
+			log.Warn("Failed to set environment variable", zap.String("key", key), zap.Error(err))
+		}
+		log.Info("Set environment variable", zap.String("key", key), zap.String("value", value))
+	}
+
+	return nil
+}
+
+func writeSystemdEnvironmentConfig(log *zap.Logger, envVars map[string]string) error {
```
cartermckinnon: I think we have a utlity method to write out a file + do the mkdir stuff? check `internal/util`

nodeadm/cmd/nodeadm/init/init.go
```diff
@@ -67,6 +67,12 @@ func (c *initCmd) Run(log *zap.Logger, opts *cli.GlobalOptions) error {
 	}
 	log.Info("Loaded configuration", zap.Reflect("config", nodeConfig))
 
+	// Set environment variables early so all AWS SDK calls and system processes use them
+	log.Info("Configuring system environment variables..")
+	if err := handleSystemEnvironmentVariables(log, nodeConfig.Spec.Instance); err != nil {
```
cartermckinnon: I think we could implement this as a `SystemAspect`, take a look at that interface

nodeadm/internal/aws/imds/imds.go
```diff
@@ -3,16 +3,39 @@ package imds
...
 	"github.com/aws/aws-sdk-go-v2/feature/ec2/imds"
 )
 
 var Client *imds.Client
 
+// This function is a wrapper around the default `http.ProxyFromEnvironment` function
+// which cachces the proxy variables in environment when first invoked thus, preventing subsequent
+// configuration attempts of http proxy via derived from user-data. Since, the first outbound
+// API call in nodeadm is to the IMDS for fetching user-data, we bypass caching.
+// Ref: https://github.com/golang/go/blob/master/src/net/http/transport.go#L499
```
cartermckinnon: can you swap this for a permalink?

nodeadm/internal/aws/imds/imds.go
```diff
@@ -3,16 +3,39 @@ package imds
...
 
 	"github.com/aws/aws-sdk-go-v2/aws/retry"
+	awshttp "github.com/aws/aws-sdk-go-v2/aws/transport/http"
 	"github.com/aws/aws-sdk-go-v2/feature/ec2/imds"
 )
 
 var Client *imds.Client
 
+// This function is a wrapper around the default `http.ProxyFromEnvironment` function
+// which cachces the proxy variables in environment when first invoked thus, preventing subsequent
```
cartermckinnon: so is the goal of this to prevent the caching when we use the IMDS client, so that our later usage of e.g. `ec2.Client` re-checks the env vars?

nodeadm/cmd/nodeadm/init/environment.go
```diff
@@ -0,0 +1,81 @@
...
+	if err := file.Chmod(0644); err != nil {
+		return fmt.Errorf("failed to set systemd config file permissions: %w", err)
+	}
+
+	if _, err := fmt.Fprintln(file, "[Manager]"); err != nil {
+		return fmt.Errorf("failed to write systemd config header: %w", err)
+	}
+
+	for key, value := range envVars {
+		if _, err := fmt.Fprintf(file, "DefaultEnvironment=\"%s=%s\"\n", key, value); err != nil {
```
fletcherw: will this work if the user wants to set an environment variable that has a `"` in it? If no, is there a simple way to escape this so it would work?

nodeadm/cmd/nodeadm/init/init.go
fletcherw: optional: could be nice to just pass in nodeConfig.Spec.Instance.environment since you don't use the localStorage field that I can see.

nodeadm/internal/aws/imds/imds.go
```diff
@@ -3,16 +3,39 @@ package imds
...
+// configuration attempts of http proxy via derived from user-data. Since, the first outbound
+// API call in nodeadm is to the IMDS for fetching user-data, we bypass caching.
```
fletcherw: seems like it might be fragile, wouldn't this break if we added another outbound HTTP call that didn't go to localhost?

nodeadm/internal/aws/imds/imds.go
shvbsle: The goal is to prevent (or delay) caching only till we get the desired environment variables from userdata. 
Currently the flow is like so:

1. Nodeadm initializes a config-provider object
2. The config-provider object is used to fetch the user-data
   - As part of this fetch, the default http client is created and it caches whatever available proxy related env variables are available. This happens due to a `sync.Once` [here](https://github.com/golang/go/blob/master/src/net/http/transport.go#L957)
3. The user-data is parsed to evaluate env variables. 
4. For all subsequent calls, even if the nodeadm process runs `os.Setenv`, any proxy-related environment variables will be ignored by the client due to caching at 2. 
   - This can be mitigated if you simply reinitialize the client. However, maintaining two clients seems a little dirty to me

This code effectively moves the cache at the end of 3.

nodeadm/internal/aws/imds/imds.go
shvbsle: > outbound HTTP call that didn't go to localhost?

Could you elaborate more? Do you mean any out bound call like `169.254.*` (but not IMDS)? Any such call would have to exist before nodeadm's init phase invocation and iiuic, there is no such call/step.

Currently, in nodeadm's init phase, the very first outbound API call is to IMDS to get userdata [here](https://github.com/awslabs/amazon-eks-ami/blob/main/nodeadm/internal/configprovider/userdata.go#L39) which happens right after the config provider is [initialized](https://github.com/awslabs/amazon-eks-ami/blob/main/nodeadm/cmd/nodeadm/init/init.go#L60). 

let me know if I understood your concern well. We also have an option to reinitialzie the client after configuring env variables but that might be a little messy

fletcherw: Right now, I agree this will work, I'm just thinking about what will happen if we modify nodeadm in the future and it makes a second outbound call that does initialize the proxy. I think it's okay how you have it now as long as a test would catch this breakage.

shvbsle: Ack. I agree with the concern here. I'll add a test-case to catch this

nodeadm/internal/aws/imds/imds.go
shvbsle: Addressed in latest revision

nodeadm/cmd/nodeadm/init/init.go
shvbsle: Addressed in latest revision

nodeadm/cmd/nodeadm/init/environment.go
shvbsle: Addressed in latest revision

nodeadm/cmd/nodeadm/init/environment.go
shvbsle: Addressed this in the latest revision. Will shortly add some test-cases

shvbsle:
/ci
+workflow:k8s_versions 1.33

nodeadm/internal/system/environment.go
```diff
@@ -0,0 +1,101 @@
...
+	// Note that EnvironmentFile= and Environment= directives will take precedence over DefaultEnvironment=.
+	// Reference: systemd.exec(5) and systemd-system.conf(5) man pages
+	// https://www.freedesktop.org/software/systemd/man/systemd.exec.html#Environment
+	if err := a.writeSystemdEnvironmentConfig(envVars); err != nil {
+		return fmt.Errorf("failed to write systemd environment config: %w", err)
+	}
+
+	for key, value := range envVars {
+		if err := os.Setenv(key, value); err != nil {
+			log.Warn("Failed to set environment variable", zap.String("key", key), zap.Error(err))
```
fletcherw: missing a `continue` here?

nodeadm/internal/aws/imds/imds.go
fletcherw: nit:
```suggestion
// which caches the proxy variables in environment when first invoked, thus preventing subsequent
```

nodeadm/internal/system/environment.go
```diff
@@ -0,0 +1,101 @@
...
+
+	return nil
+}
+
+func (a *environmentAspect) generateSystemdConfig(envVars map[string]string) string {
+	var builder strings.Builder
+	builder.WriteString("[Manager]\n")
+
+	for key, value := range envVars {
+		escapedValue := a.escapeSystemdValue(value)
```
fletcherw: there are some other ways this could break, for example if `key` contains '=', but since it's user-specified config I don't know how hard we need to try.

nodeadm/internal/system/environment.go
shvbsle: Ack. Fixed

nodeadm/internal/aws/imds/imds.go
shvbsle: Ty. Fixed in latest revision

nodeadm/cmd/nodeadm/init/init.go
shvbsle: addressed in latest revision

nodeadm/internal/system/environment.go
```diff
@@ -28,32 +30,55 @@ func (a *environmentAspect) Name() string {
 }
 
 func (a *environmentAspect) Setup(cfg *api.NodeConfig) error {
-	if len(cfg.Spec.Instance.Environment) == 0 {
+	envOpts := cfg.Spec.Instance.Environment
+
+	// Check if any environment configuration exists
+	// Handle cases where maps might be nil (missing section) or empty
+	if (envOpts.Default == nil || len(envOpts.Default) == 0) &&
```
fletcherw: nit: `len()` of nil map is 0.

also, you check again in configureEnvironment, so the only effect of this is to print "No variables to configure". Could maybe be simplified.

nodeadm/internal/system/environment.go
```diff
@@ -0,0 +1,163 @@
...
+// configureEnvironment makes environment variables available system-wide and to specific services.
+func (a *environmentAspect) configureEnvironment(envOpts api.EnvironmentOptions) error {
+	log := zap.L()
+
+	// Configure systemd system.conf.d for all services
+	// Nodeadm will use the lowest precedence directive i.e. DefaultEnvironment= to configure environment
+	// Note that EnvironmentFile= and Environment= directives will take precedence over DefaultEnvironment=.
+	// Reference: systemd.exec(5) and systemd-system.conf(5) man pages
+	// https://www.freedesktop.org/software/systemd/man/systemd.exec.html#Environment
+	if envOpts.Default != nil && len(envOpts.Default) > 0 {
```
ndbaker1: this would suffice?
```suggestion
	if len(envOpts.Default) > 0 {
```

nodeadm/internal/system/environment.go
```diff
@@ -0,0 +1,163 @@
...
+			if err := os.Setenv(key, value); err != nil {
+				log.Warn("Failed to set environment variable", zap.String("key", key), zap.Error(err))
+				continue
+			}
+			log.Info("Set default environment variable", zap.String("key", key), zap.String("value", value))
+		}
+	}
+
+	// Configure kubelet-specific environment variables
+	if envOpts.SystemdKubelet != nil && len(envOpts.SystemdKubelet) > 0 {
```
ndbaker1:
```suggestion
	if len(envOpts.SystemdKubelet) > 0 {
```

nodeadm/internal/system/environment.go
```diff
@@ -0,0 +1,163 @@
...
+		if err := a.writeServiceDropinConfig("kubelet", kubeletServiceDropinPath, envOpts.SystemdKubelet); err != nil {
+			return fmt.Errorf("failed to write kubelet environment config: %w", err)
+		}
+	}
+
+	// Configure containerd-specific environment variables
+	if envOpts.SystemdContainerd != nil && len(envOpts.SystemdContainerd) > 0 {
```
ndbaker1:
```suggestion
	if len(envOpts.SystemdContainerd) > 0 {
```

nodeadm/test/e2e/cases/nodeadm-environment-config/run.sh
```diff
@@ -0,0 +1,88 @@
...
+source /helpers.sh
+
+mock::aws
+mock::kubelet 1.27.0
+wait::dbus-ready
+
+# Test 1 - Basic environment variable configuration
+nodeadm init --skip run --config-source file://config-basic-env.yaml
+assert::file-contains /etc/systemd/system.conf.d/environment.conf 'DefaultEnvironment="DUMMY_VAR=HELLO WORLD"'
+assert::file-contains /etc/systemd/system.conf.d/environment.conf 'DefaultEnvironment="HTTP_PROXY=http://example-proxy:8080"'
```
ndbaker1: can we have the expected `environment.conf` files in directories?  
if not its ok, but i generally try to template the test cases as the fixtures + expected results if you have a lot of them and then use [some kind of loop](https://github.com/awslabs/amazon-eks-ami/blob/9d57b9d2d0ff3949f7d9494c12b2696048e2bdcc/nodeadm/test/e2e/cases/kubelet-config-dir/run.sh#L13)

nodeadm/internal/system/environment.go
```diff
@@ -0,0 +1,163 @@
...
+		zap.L().Info("No environment variables to configure")
+		return nil
+	}
+
+	return a.configureEnvironment(envOpts)
+}
+
+// configureEnvironment makes environment variables available system-wide and to specific services.
+func (a *environmentAspect) configureEnvironment(envOpts api.EnvironmentOptions) error {
+	log := zap.L()
```
ndbaker1: nit: it can go both ways, but i think the precedent in this project is to just use the global logger wherever needed
```go
zap.L().Info("...")
```

nodeadm/internal/api/types.go
```diff
@@ -94,6 +94,13 @@ const (
 
 type InstanceOptions struct {
 	LocalStorage LocalStorageOptions `json:"localStorage,omitempty"`
+	Environment  EnvironmentOptions  `json:"environment,omitempty"`
```
ndbaker1, line 97: i think a pure string map interface is ok here as well, but i'll let you have the call.

its common in projects to document environment variable prefixes such as `SYSTEMD_kubelet` if they wanted to set an environment variable for only `kubelet.service`. 

this is nice because it can be extended without concrete types, and can be logged when ignored/not-handled.

nodeadm/internal/system/environment.go
```diff
@@ -0,0 +1,163 @@
...
+}
+
+func (a *environmentAspect) Setup(cfg *api.NodeConfig) error {
+	envOpts := cfg.Spec.Instance.Environment
+
+	// Check if any environment configuration exists
+	// Handle cases where maps might be nil (missing section) or empty
+	if (envOpts.Default == nil || len(envOpts.Default) == 0) &&
+		(envOpts.SystemdKubelet == nil || len(envOpts.SystemdKubelet) == 0) &&
+		(envOpts.SystemdContainerd == nil || len(envOpts.SystemdContainerd) == 0) {
```
ndbaker1: nit: dont need the nil checks here either

nodeadm/internal/system/environment.go
```diff
@@ -0,0 +1,163 @@
...
+		if err := a.writeSystemdEnvironmentConfig(envOpts.Default); err != nil {
+			return fmt.Errorf("failed to write systemd environment config: %w", err)
+		}
+
+		for key, value := range envOpts.Default {
+			if err := os.Setenv(key, value); err != nil {
+				log.Warn("Failed to set environment variable", zap.String("key", key), zap.Error(err))
+				continue
+			}
+			log.Info("Set default environment variable", zap.String("key", key), zap.String("value", value))
```
ndbaker1: remind me, we need this before we call any of the proxy-able endpoints right?

nodeadm/internal/system/environment.go
```diff
@@ -93,6 +118,43 @@ func (a *environmentAspect) generateSystemdConfig(envVars map[string]string) str
...
+		zap.String("path", dropinPath),
+		zap.Int("count", len(envVars)))
+
+	content := a.generateServiceDropinConfig(envVars)
+
+	if err := util.WriteFileWithDir(dropinPath, []byte(content), 0644); err != nil {
+		return fmt.Errorf("failed to write service drop-in config: %w", err)
+	}
+
+	log.Info("Reloading systemd configuration for service drop-in", zap.String("service", serviceName))
```
fletcherw: Do you need to reload config for every file you write? Or can you just reload once at the end after writing all the config files? (Maybe making sure not to reload if it's not needed)

nodeadm/internal/system/environment.go
ndbaker1: +1 here, is there a more standard sanitizer we can use?

also keep in mind the additional escapes that [DefaultEnvironment](https://www.freedesktop.org/software/systemd/man/latest/systemd-system.conf.html#DefaultEnvironment=) can do with `%`.

nodeadm/internal/system/environment.go
```diff
@@ -0,0 +1,163 @@
...
+	if err := exec.Command("systemctl", "daemon-reload").Run(); err != nil {
+		return fmt.Errorf("failed to reload systemd configuration: %w", err)
+	}
+
+	return nil
+}
+
+func (a *environmentAspect) generateSystemdConfig(envVars map[string]string) string {
+	var builder strings.Builder
+	builder.WriteString("[Manager]\n")
```
ndbaker1: just wondering do you want to use a template for this?

ndbaker1 (commented):
really close, just want to align on a couple more things

nodeadm/internal/system/environment.go
fletcherw: good point since it's quite similar to the generateDropinConfig, apart from the words used.

nodeadm/internal/system/environment.go
shvbsle: That is right. This is needed by nodeadm-config process can make calls to EC2 APIs that could go through the proxy.

nodeadm/internal/system/environment.go
shvbsle: Ack will skip the nil checks in the revision

nodeadm/internal/system/environment.go
shvbsle: Fair. In latest revision will be calling it only once

nodeadm/internal/system/environment.go
shvbsle: Ack. Handled in the latest revision

nodeadm/internal/system/environment.go
shvbsle: Ack. Handled in the latest revision

nodeadm/internal/system/environment.go
shvbsle: Ack. Handled in the latest revision

nodeadm/internal/system/environment.go
shvbsle: Ack. Handled in the latest revision

nodeadm/internal/system/environment.go
shvbsle: Ack. Handled in the latest revision

nodeadm/test/e2e/cases/nodeadm-environment-config/run.sh
shvbsle: I think I'm going to stick to the current approach to roll this feature out. Each test case can generate up to three different files and combination of these varies by test-case. Many test cases verify that the files should not be created (empty configs, service-only configs etc.). Also we need `assert::file-contains`, `assert::file-contains-literal` (for quote escaping), and file existence checks - harder to template cleanly.

nodeadm/internal/api/types.go
shvbsle: I'll stick to the current approach. The type safety and explicit structure makes the API clearer for the users. We can always extend it later if we need to support additional services

nodeadm/internal/system/environment.go
```diff
@@ -0,0 +1,154 @@
...
+	// Configure containerd-specific environment variables
+	if len(envOpts.SystemdContainerd) > 0 {
+		if err := a.writeServiceDropinConfig("containerd", containerdServiceDropinPath, envOpts.SystemdContainerd); err != nil {
+			return fmt.Errorf("failed to write containerd environment config: %w", err)
+		}
+	}
+
+	// Reload systemd configuration once after all config files are written
+	zap.L().Info("Reloading systemd configuration")
+	if err := exec.Command("systemctl", "daemon-reload").Run(); err != nil {
```
ndbaker1: would `.CombinedOutput` give you more details *if* you needed them?

nodeadm/internal/system/environment.go
```diff
@@ -0,0 +1,154 @@
...
+type environmentAspect struct{}
+
+func (a *environmentAspect) Name() string {
+	return environmentAspectName
+}
+
+func (a *environmentAspect) Setup(cfg *api.NodeConfig) error {
+	envOpts := cfg.Spec.Instance.Environment
+
+	// Check if any environment configuration exists. If nothing to configure, don't invoke systemctl daemon-reload
```
ndbaker1: do you forsee the daemon-reload causing issues, or just to avoid the call time?

nodeadm/internal/system/environment.go
```diff
@@ -0,0 +1,154 @@
...
+	return nil
+}
+
+func (a *environmentAspect) writeSystemdEnvironmentConfig(envVars map[string]string) error {
+	if len(envVars) == 0 {
+		return nil
+	}
+
+	log := zap.L()
+	log.Info("Writing environment variables to systemd system.conf.d", zap.Int("count", len(envVars)))
```
ndbaker1: similar to previous comment, can we keep occurrences of this uniform?
```suggestion
	zap.L().Info("Writing environment variables to systemd system.conf.d", zap.Int("count", len(envVars)))
```

nodeadm/test/e2e/helpers.sh
```diff
@@ -47,6 +47,23 @@ function assert::file-contains() {
   fi
 }
 
+function assert::file-contains-literal() {
```
ndbaker1: didn't realize this helper before, do we really need it since the regex should be a superset of literals?

ndbaker1: i try to avoid adding overlapping functionality so that its easier to do it in a standard way across all tests

nodeadm/internal/system/environment.go
shvbsle: agreed that systemd specifiers could cause issues. I'm going to add an escape sequence for specifiers. 

On standard sanitizer that we can use, the closest thing I found was `systemd-escape` and a go lib that provides a similar functionality. However this is to escape strings for usage in systemd unit names so doesn't fit our usecase well. Couldn't find any other existing sanitizer for cleaning up environment variables
https://pkg.go.dev/github.com/coreos/go-systemd/v22/unit#UnitNameEscape

nodeadm/test/e2e/helpers.sh
shvbsle: `assert::file-contains` treats the pattern as a regex that I have in my test data. I just needed a simpler function.

nodeadm/internal/system/environment.go
shvbsle: Oop. Missed two places. Fixing in revision

nodeadm/internal/system/environment.go
shvbsle: I don't foresee daemon-reload causing functional issues. It does scan and parse all unit files and rebuilds systemd's internal state so would be nice to avoid any unnecessary call time.

nodeadm/internal/system/environment.go
shvbsle: This is a good idea. I tried writing some toy code and an invalid systemd configuration and I do get more than just the exit status so will be useful for debugging

nodeadm/internal/system/environment.go
shvbsle: Gotcha. I wasn't familiar with go templates. I have replaced this manual generation of configs with go templates. I've switched to using templates with a FuncMap approach. Something like this:

```go
const systemdConfigTemplate = `[Manager]
{{range $key, $value := .}}DefaultEnvironment="{{$key}}={{escape $value}}"
{{end}}`

var templateFuncs = template.FuncMap{
    "escape": escapeSystemdValue,
}
```

shvbsle:
I'm defining the following system aspect before the config phase is invoked:
```go
	preConfigAspects := []system.SystemAspect{
		system.NewEnvironmentAspect(),
	}
```

This is because, `enrichConfig` calls the AWS EC2 API to get instance details and that call could potentially go through the proxy.

nodeadm/internal/system/environment.go
shvbsle: Update: Switching to this strconv.Quote:
https://pkg.go.dev/strconv#Quote

nodeadm/internal/system/environment.go
```diff
@@ -0,0 +1,171 @@
...
+	if len(envOpts.Default) == 0 && len(envOpts.SystemdKubelet) == 0 && len(envOpts.SystemdContainerd) == 0 {
+		zap.L().Info("No environment variables to configure")
+		return nil
+	}
+
+	return a.configureEnvironment(envOpts)
+}
+
+// configureEnvironment makes environment variables available system-wide and to specific services.
+func (a *environmentAspect) configureEnvironment(envOpts api.EnvironmentOptions) error {
```
shvbsle: separation into configureInstanceEnvironment vs configureNodeadmEnvironment

nodeadm/api/v1alpha1/nodeconfig_types.go
```diff
@@ -85,6 +85,19 @@ type ContainerdOptions struct {
...
+	Environment  EnvironmentOptions  `json:"environment,omitempty"`
+}
+
+// EnvironmentOptions configures environment variables for the system and systemd services.
+type EnvironmentOptions struct {
+	// Default environment variables applied system-wide to all systemd services
+	Default map[string]string `json:"default,omitempty"`
+
+	// SystemdKubelet environment variables applied specifically to the kubelet service
+	SystemdKubelet map[string]string `json:"systemd_kubelet,omitempty"`
```
shvbsle: Make this more general. Should support any systemd unit. Assume the convention that the cx only passes the systemd unit name ex. `kubelet` and not `kubelet.service`. We rely on the assumption that the dropin path will be generated based on `<unit-name>.service.d`

nodeadm/cmd/nodeadm/init/init.go
```diff
@@ -67,6 +67,24 @@ func (c *initCmd) Run(log *zap.Logger, opts *cli.GlobalOptions) error {
...
+	preConfigAspects := []system.SystemAspect{
+		system.NewEnvironmentAspect(),
+	}
+
+	// We set up any environment variables before nodeadm's config phase.
+	// This let's nodeadm respect any environment variables that may be critical for
+	// the node's setup. For example, prior to config phase, we make calls to EC2's API to
+	// get instance details which could pass through an HTTP(s) proxy.
+	log.Info("Setting up pre-config aspects...")
+	for _, aspect := range preConfigAspects {
```
shvbsle: Rename this to something like `initSystemAspects` and separate out the env config for nodeadm and systemd environments

nodeadm/test/e2e/cases/nodeadm-environment-config/run.sh
shvbsle: Will switch to `assert::files-equal()` in next revision

nodeadm/test/e2e/helpers.sh
shvbsle: Will switch to `assert::files-equal()` once I move my test-cases to .conf files

nodeadm/internal/system/environment.go
ndbaker1: Im leaning towards `configureBootstrapEnvironment`? I dont think many things in the repo know that its called `nodeadm` besides the commands?

nodeadm/test/e2e/helpers.sh
shvbsle: addressed in latest revision

nodeadm/internal/api/types.go
shvbsle: Addressed in latest revision. Switching to 
```go
type EnvironmentOptions map[string]map[string]string
``` 

to allow for any service configuration

nodeadm/cmd/nodeadm/init/init.go
```diff
@@ -84,7 +100,11 @@ func (c *initCmd) Run(log *zap.Logger, opts *cli.GlobalOptions) error {
 	}
 	defer daemonManager.Close()
 
-	aspects := []system.SystemAspect{
+	configAspects := []system.SystemAspect{
+		system.NewInstanceEnvironmentAspect(),
+	}
+
+	runAspects := []system.SystemAspect{
```
fletcherw: nit: should this be defined on line 140 just before it's used?

nodeadm/internal/system/environment.go
```diff
@@ -0,0 +1,177 @@
...
+
+	"go.uber.org/zap"
+
+	"github.com/awslabs/amazon-eks-ami/nodeadm/internal/api"
+	"github.com/awslabs/amazon-eks-ami/nodeadm/internal/util"
+)
+
+const (
+	instanceEnvironmentAspectName = "instance-environment"
+	nodeamdEnvironmentAspectName  = "nodeadm-environment"
```
fletcherw: nit: is `nodeadmd` a typo?

nodeadm/internal/system/environment.go
```diff
@@ -0,0 +1,177 @@
...
+				dropinPath := fmt.Sprintf("%s/%s.service.d/environment.conf", serviceDropinPathBase, serviceName)
+				if err := a.writeServiceDropinConfig(serviceName, dropinPath, envVars); err != nil {
+					return fmt.Errorf("failed to write %s environment config: %w", serviceName, err)
+				}
+			}
+		}
+	}
+
+	// Reload systemd configuration once after all config files are written
+	zap.L().Info("Reloading systemd configuration")
```
fletcherw: this will still trigger daemon reload if all the maps within EnvironmentOptions are empty, it's alright if you don't fix this though.

nodeadm/internal/system/environment.go
```diff
@@ -0,0 +1,177 @@
...
+
+	return a.configureInstanceEnvironment(envOpts)
+}
+
+// configureInstanceEnvironment makes environment variables available system-wide and to specific services.
+func (a *instanceEnvironmentAspect) configureInstanceEnvironment(envOpts api.EnvironmentOptions) error {
+
+	zap.L().Info("All envOpts: ", zap.Any("=", envOpts))
+	for serviceName, envVars := range envOpts {
+		if len(envVars) > 0 {
```
fletcherw: nit: invert if condition:
```
if len(envVars) == 0 {
  continue
}
```
to reduce indentation

nodeadm/internal/system/environment.go
```diff
@@ -0,0 +1,177 @@
...
+	configData := make([]struct{ Key, Value string }, len(keys))
+	for i, k := range keys {
+		configData[i] = struct{ Key, Value string }{k, envVars[k]}
+	}
+
+	tmpl := template.Must(template.New("envConfig").Funcs(templateFuncs).Parse(templateStr))
+	var buf bytes.Buffer
+	if err := tmpl.Execute(&buf, configData); err != nil {
+		zap.L().Error("Failed to execute environment config template", zap.Error(err))
+		return ""
```
fletcherw: why log the error here vs returning it and logging at a higher level
by returning `""` here you will write empty configs when you get an error.

fletcherw (approved):

shvbsle:
CI is flaky but hoping its smoother now

/ci 

edit: nvm: doesnt work after merging

ndbaker1:
@shvbsle can you add example usage docs for this change?

PR #2390, opened 2025-08-21
feat(al2023): enable graceful shutdown in 1.34+

cartermckinnon:
**Description of changes:**

Enables graceful node shutdown for 1.34+.

EC2 documents that an instance is allowed "several minutes" to shut down before it is forcefully terminated: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/how-ec2-instance-stop-start-works.html

The exact amount of time may change, so we're interpreting "several minutes" conservatively as "3 minutes". (I confirmed this is acceptable with the EC2 folks).

Some amount of time is needed for the rest of the OS to perform an orderly shutdown after kubelet completes its process, so we will limit the total `kubelet` shutdown grace period to 2 minutes and 30 seconds, with 30 seconds of that being allocated to critical pods.

### DIFF
--- a/nodeadm/internal/kubelet/config.go
+++ b/nodeadm/internal/kubelet/config.go
@@ -50,33 +50,35 @@ func (k *kubelet) writeKubeletConfig(cfg *api.NodeConfig) error {
 // KubeletConfiguration types:
 // https://pkg.go.dev/k8s.io/kubelet/config/v1beta1#KubeletConfiguration
 type kubeletConfig struct {
-	Address                  string                           `json:"address"`
-	Authentication           k8skubelet.KubeletAuthentication `json:"authentication"`
-	Authorization            k8skubelet.KubeletAuthorization  `json:"authorization"`
-	CgroupDriver             string                           `json:"cgroupDriver"`
-	CgroupRoot               string                           `json:"cgroupRoot"`
-	ClusterDNS               []string                         `json:"clusterDNS"`
-	ClusterDomain            string                           `json:"clusterDomain"`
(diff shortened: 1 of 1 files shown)


stevehipwell:
@cartermckinnon how does the few minutes relate to spot instances? Is this outside of the 120s warning before the termination command?

shvbsle (approved):

nodeadm/internal/kubelet/config.go
```diff
@@ -217,6 +219,15 @@ func (ksc *kubeletConfig) withVersionToggles(cfg *api.NodeConfig) {
 	if semver.Compare(cfg.Status.KubeletVersion, "v1.34.0") >= 0 {
 		ksc.FeatureGates["MutableCSINodeAllocatableCount"] = true
 	}
+	// Enable graceful node shutdown in 1.34+
+	if semver.Compare(cfg.Status.KubeletVersion, "v1.34.0") >= 0 {
+		ksc.ShutdownGracePeriod = &metav1.Duration{
```
shvbsle, line 224: nit: could have moved these under the same condition that we see at line 219

cartermckinnon:
@stevehipwell spot instances are terminated with the same flow, so it'll work the same.

nodeadm/internal/kubelet/config.go
cartermckinnon: Felt nice to keep them separate 🤷 but you're right

sidewinder12s:
@cartermckinnon Can this be enabled in 1.32/33 or was there a reason this wasn't enabled on the lower versions beyond not introducing significant changes in those older versions?

cartermckinnon:
No reason other than the behavior change 👍

PR #2393, opened 2025-08-22
Add soci snapshotter info to eks-log-collector

HusainZafar:
**Issue #, if available:**

**Description of changes:**
If containerd is setup to use soci snapshotter, collect the snapshotter related information as part of the eks-log-collector script

**Testing Done**

```
# Output from the extracted zip file when soci-snapshotter was configured. Otherwise, this directory will not be created
> ls containerd/soci-snapshotter/
config.toml  soci-snapshotter-log.txt  soci-snapshotter-status.txt
```

When soci snapshotter is not used, the new changes are ignored 
```
Trying to Collect Containerd snapshotter information...
Trying to archive gathered information...
```

### DIFF
--- a/log-collector-script/linux/eks-log-collector.sh
+++ b/log-collector-script/linux/eks-log-collector.sh
@@ -848,6 +848,26 @@ get_containerd_info() {
   fi
 
   ok
+
+  try "collect containerd snapshotter information"
+  get_soci_snapshotter_info
+
+  ok
+}
+
+get_soci_snapshotter_info() {
+  if ! systemctl is-active --quiet soci-snapshotter 2> /dev/null; then
+    return
+  fi
+
+  try "Collect soci snapshotter information"
+
+  mkdir -p "${COLLECT_DIR}"/containerd/soci-snapshotter
+  systemctl status soci-snapshotter > "${COLLECT_DIR}"/containerd/soci-snapshotter/soci-snapshotter-status.txt 2>&1
+  timeout 75 journalctl -u soci-snapshotter --since "${DAYS_10}" > "${COLLECT_DIR}"/containerd/soci-snapshotter/soci-snapshotter-log.txt 2>&1 || echo -e "\tTimed out, ignoring \"soci-snapshotter log output \" "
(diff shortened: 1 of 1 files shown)


log-collector-script/linux/eks-log-collector.sh
```diff
@@ -848,6 +848,34 @@ get_containerd_info() {
...
 
   ok
+
+  try "collect containerd snapshotter information"
+  get_soci_snapshotter_info
+
+  ok
+}
+
+get_soci_snapshotter_info() {
```
fletcherw, line 858: I don't think the comments in this function are helpful. They are mostly 'what' comments that explain what the code is doing (which can be inferred by just reading the code), rather than pointing out complexities or reasons why you are doing something. I recommend deleting them.

log-collector-script/linux/eks-log-collector.sh
```diff
@@ -848,6 +848,34 @@ get_containerd_info() {
...
+
+  # Create soci-snapshotter directory within containerd folder
+  mkdir -p "${COLLECT_DIR}"/containerd/soci-snapshotter
+
+  # Collect service status and logs
+  systemctl status soci-snapshotter > "${COLLECT_DIR}"/containerd/soci-snapshotter/soci-snapshotter-status.txt 2>&1
+  timeout 75 journalctl -u soci-snapshotter --since "${DAYS_10}" > "${COLLECT_DIR}"/containerd/soci-snapshotter/soci-snapshotter-log.txt 2>&1 || echo -e "\tTimed out, ignoring \"soci-snapshotter log output \" "
+
+  # Collect configuration files if they exist
+  if [[ -f /etc/soci-snapshotter-grpc/config.toml ]]; then
```
fletcherw: Some of the `cp` calls in this file check if the file exists first, some do not. Is it necessary, or is it okay for the `cp` to just fail if the file doesn't exist?

log-collector-script/linux/eks-log-collector.sh
HusainZafar: Removed the comments

log-collector-script/linux/eks-log-collector.sh
HusainZafar: Removed the check, verified that the redirection `2>/dev/null` ensures the errors are ignored

fletcherw (approved):

PR #2397, opened 2025-08-26
fix(nodeadm): resolve config race with systemd-networkd

mselim00:
**Issue #, if available:**

**Description of changes:**
Enforces a necessary ordering between the `nodeadm` phases as they relate to the networking components. Ensures that `nodeadm-boot-hook` runs before `systemd-networkd-wait-online` so that service can be used by later `nodeadm` phases as a source of truth on `nodeadm`'s desired network configuration. Restarts instead of reloading `systemd-networkd` to force synchronicity, the latter leaves open the possibility of routes being created/destroyed during `nodeadm-config` or later services.

The forced ordering may increase boot times by a few seconds, but it's a worthwhile trade-off for the time being to ensure network stability during the various `nodeadm` phases. In the future, we can potentially look to cut this down by either finding a way to discover the primary network interface without a network call (and therefore start the boot hook before networkd). We may also be able to save a little bit of time by making a reload call instead of a restart and waiting until the status matches the desired, but there does not seem to be a clean way to do this at the moment.

**Testing Done**

### DIFF
--- a/templates/al2023/runtime/rootfs/etc/systemd/system/nodeadm-boot-hook.service
+++ b/templates/al2023/runtime/rootfs/etc/systemd/system/nodeadm-boot-hook.service
@@ -4,6 +4,13 @@ Documentation=https://github.com/awslabs/amazon-eks-ami
 # this unit must precede all other nodeadm units because it sets the
 # expectation for OS networking.
 Before=nodeadm-config.service
+# must follow networkd as it requires network availability for reading
+# interface information from IMDS. Assumes the IMDS client used will
+# retry enough to capture eventual network readiness
+After=systemd-networkd.service
+# precede systemd-networkd-wait-online so future nodeadm services can rely on
+# that service as a source of truth on the configurations made by this service
+Before=systemd-networkd-wait-online.service
 
 [Service]
 Type=oneshot
--- a/nodeadm/cmd/nodeadm-internal/boot-hook/boot_hook.go
+++ b/nodeadm/cmd/nodeadm-internal/boot-hook/boot_hook.go
@@ -5,8 +5,11 @@ import (
 	"go.uber.org/zap"
 
 	"github.com/awslabs/amazon-eks-ami/nodeadm/internal/cli"
+	"github.com/awslabs/amazon-eks-ami/nodeadm/internal/daemon"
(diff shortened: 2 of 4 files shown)


templates/al2023/runtime/rootfs/etc/systemd/system/nodeadm-boot-hook.service
```diff
@@ -4,6 +4,13 @@ Documentation=https://github.com/awslabs/amazon-eks-ami
 # this unit must precede all other nodeadm units because it sets the
 # expectation for OS networking.
 Before=nodeadm-config.service
+# must follow networkd as it requires network availability for reading
+# interface information from IMDS. Assumes the IMDS client used will
+# retry enough to capture eventual network readiness
+After=systemd-networkd.service
+# precede systemd-networkd-wait-online so future nodeadm services can rely on
+# that service as a source of truth on the configurations made by this service
+Before=systemd-networkd-wait-online.service
```
ndbaker1, line 13: do you think you could collect some statistics on where these services usually land timing-wise on a typical boot?

mselim00: Sample on a `t3.medium` on an AMI before this change
```
0s - boot
6s - nodeadm-boot-hook, systemd-networkd, systemd-networkd-wait-online start
7s - nodeadm-boot-hook exists, systemd-networkd-wait-online exits, nodeadm-config starts
8s - nodeadm-config exits
```

I can try to collect some data on multi-card instance types, but there'll naturally be a lot of variance in this. A few seconds will likely be added by forcing `config` to start after `systemd-networkd-wait-online`

mselim00: After this PR, on a p4d.24xlarge with 4 EFA + ENA interfaces

```
0s  - boot
9s  - systemd-networkd, nodeadm-boot-hook start
10s - systemd-networkd restarted
11s - nodeadm-boot-hook exit, systemd-networkd-wait-online start and exit, nodeadm-config starts
12s - nodeadm-config exits
```

After merging we'll be able to collect some more statistics on registration times, but in this case it was really just a one second increase observed between these two cases (and there are a lot of other variables involved).

PR #2399, opened 2025-08-27
Add log collection for gmsa on Windows

KlwntSingh:
**Issue #, if available:**

**Description of changes:**
1. log collector will collect logs for both gMSA logs in file and Windows events.

**Testing Done**
1. Script is tested on older AMI without gmsa event logs but with file logs.
2. Script is tested on new AMI with gmsa event logs but no file logs.

### DIFF
--- a/log-collector-script/windows/eks-log-collector.ps1
+++ b/log-collector-script/windows/eks-log-collector.ps1
@@ -52,6 +52,7 @@ Function create_working_dir{
         Write-Host "Creating temporary directory"
         New-Item -type directory -path $info_system -Force >$null
         New-Item -type directory -path $info_system\eks -Force >$null
+        New-Item -type directory -path $info_system\gmsa\filelogs -Force >$null
         New-Item -type directory -path $info_system\docker -Force >$null
         New-Item -type directory -path $info_system\containerd -Force >$null
         New-Item -type directory -path $info_system\firewall -Force >$null
@@ -260,7 +261,24 @@ Function get_eks_logs{
         Write-Host "OK" -foregroundcolor "green"
     }
     catch{
-        Write-Error "Unable to collect ECS Agent logs"
+        Write-Error "Unable to collect EKS logs"
+        Break
+    }
+}
+
+Function get_gmsa_logs{
+    try {
+        Write-Host "Collecting gMSA logs"
+        if (Test-Path "C:\ProgramData\Amazon\gmsa-plugin\*") {
+            copy C:\ProgramData\Amazon\gmsa-plugin\* $info_system\gmsa\filelogs\
+        }
+        if (Get-WinEvent -ListProvider gMSA -ErrorAction SilentlyContinue) {
+            Get-WinEvent -ProviderName gMSA | Export-CSV $info_system/gmsa/eventlogs.csv
+        }
+        Write-Host "OK" -foregroundcolor "green"
+    }
+    catch{
+        Write-Error "Unable to collect gMSA logs"
         Break
     }
 }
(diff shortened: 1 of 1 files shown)


shashank-boyapally:
lgtm!

shashank-boyapally (approved):

log-collector-script/windows/eks-log-collector.ps1
```diff
@@ -260,7 +261,24 @@ Function get_eks_logs{
...
+}
+
+Function get_gmsa_logs{
+    try {
+        Write-Host "Collecting gMSA logs"
+        if (Test-Path "C:\ProgramData\Amazon\gmsa-plugin\*") {
+            copy C:\ProgramData\Amazon\gmsa-plugin\* $info_system\gmsa\filelogs\
+        }
+        if (Get-WinEvent -ListProvider gMSA -ErrorAction SilentlyContinue) {
+            Get-WinEvent -ProviderName gMSA | Export-CSV $info_system/gmsa/eventlogs.csv
```
shvbsle, line 276: Is the mixed use of backslashes and forward slashes intended (and does it work)?  

Line 273:
```
$info_system\gmsa\filelogs\
```

Line 276
```ps1
$info_system/gmsa/eventlogs.csv
```

log-collector-script/windows/eks-log-collector.ps1
```diff
@@ -260,7 +261,24 @@ Function get_eks_logs{
...
+        Write-Error "Unable to collect EKS logs"
+        Break
+    }
+}
+
+Function get_gmsa_logs{
+    try {
+        Write-Host "Collecting gMSA logs"
+        if (Test-Path "C:\ProgramData\Amazon\gmsa-plugin\*") {
+            copy C:\ProgramData\Amazon\gmsa-plugin\* $info_system\gmsa\filelogs\
```
shvbsle, line 273: Is there a reason to use `copy` and not `Copy-Item` which is used elsewhere in the script?

shvbsle: Referring to `Copy-Item` directive in this code-block:
https://github.com/awslabs/amazon-eks-ami/blob/main/log-collector-script/windows/eks-log-collector.ps1#L356-L361

Example
```ps1
Copy-Item "$env:SystemDrive\Windows\System32\Winevt\Logs\Application.evtx" -Destination $info_system\events
```

log-collector-script/windows/eks-log-collector.ps1
KlwntSingh: Yes, both Backward and forward slashes works. Windows has 2 shells for commands. Choice of slash is enforced by prompt.
Backward slashes works for Windows cmd commands. Forward slashes works for Powershell commands.
The whole script is executed from powershell prompt but it is backward compatible to run cmd commands. 

These changes work but we can definitely make the script consistent to follow powershell(since this is new and more powershell). This will be work for future though.

log-collector-script/windows/eks-log-collector.ps1
KlwntSingh: Copy is cmd and Copy-Item is powershell. We are using them interchangeably to do the same thing.

shvbsle: Gotcha. What are your thoughts on sticking to just `Copy-Item` through out the script given that this is mostly PowerShell? Would that break anything in the log-collector script? I'd be happy to send a refactor PR later today (in the spirit of maintaining the highest standards)

KlwntSingh: First of all, thank you for the feedback and willingness to ensure the highest standards are met.

I think we should move to Copy-Item. It's better to have a refactor as part of the larger goal to completely transition to PowerShell. This will involve analyzing and identifying other commands that can be moved to PowerShell.

There's a high chance that the change to Copy-Item won't break anything, but I would still want to test it before merging. Testing will involve running it on Windows 2019, 2022, and 2025. Windows team can track this goal as part of regular operational improvement.

PR #2324, opened 2025-07-11
feat(nodeadm): move network configuration earlier in init lifecycle

ndbaker1:
**Issue #, if available:**

**Description of changes:**

A potential race condition with `amazon-ec2-net-utils` can result in networking interfaces besides the primary being configured during the instance's boot, which can confuse networking operations from `nodeadm` during bootstrap.

the issue can only manifest when networking calls (eg. to `ec2.DescribeInstances`) are made before `nodeadm` has a chance to update `systemd-networkd` with a drop-in that forces the network config to match against primary ENI. This may sound like a severe issue today, but can only realistically occur when the following are true:
1. interfaces besides the primary take an extended period of time to become healthy
1. `nodeadm` is configured to use EC2PrivateDNSName as the node name. the [ec2 instance ID as node name](https://github.com/awslabs/amazon-eks-ami/blob/e541592913ed33413d813987dc4357b8e6160a59/nodeadm/doc/examples.md#using-instance-id-as-node-name-experimental) feature removes the need to make `ec2.DescribeInstances` calls to discover the private dns name of the instance.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

**Testing Done**

before changes, spun up a `m6in.32xlarge` with 2 EFA w/ ENA interface (one per NIC), and didn't provide userdata, resulting in `nodeadm-config` failing. this then lets us observe the normal ec2 net utils behavior, where additional routes get configured. running nodeadm manually we can then check that these routes get removed.

```
[root@ip-192-168-185-208 bin]# journalctl -u nodeadm-config
Jul 11 06:29:19 localhost systemd[1]: Starting nodeadm-config.service - EKS Nodeadm Config...
Jul 11 06:29:19 localhost nodeadm[14344]: {"level":"info","ts":1752215359.662831,"caller":"init/init.go:51","msg":"Checking user is root.."}
Jul 11 06:29:19 localhost nodeadm[14344]: {"level":"info","ts":1752215359.6638741,"caller":"init/init.go:59","msg":"Loading configuration..","configSource":"imds://user-data"}
Jul 11 06:29:20 localhost nodeadm[14344]: {"level":"fatal","ts":1752215360.6839051,"caller":"nodeadm/main.go:36","msg":"Command failed","error":"unexpected EOF","stacktrace":"main.main\n\t>
Jul 11 06:29:20 localhost systemd[1]: nodeadm-config.service: Main process exited, code=exited, status=1/FAILURE
[...]

PR #2419, opened 2025-09-08
feat(nodeadm): standardize systemd networking for all interfaces

ndbaker1:
**Issue #, if available:**

**Description of changes:**

continuation of https://github.com/awslabs/amazon-eks-ami/pull/2179.

changes in this PR:

1. the book hook has been removed. there is no longer a need to special case the loading of the interfaces because we are initializing interfaces with the correct route metrics, avoiding a regression to the issue addressed in https://github.com/awslabs/amazon-eks-ami/pull/2324.
2. similar to AL2 multi-card setup, we assume that all interfaces present at boot time should be managed by the system. however, this PR sets up for a future "managed-interface broker" concept. see PR comments for details.
3. the imds client used for fetching instance metadata has been configured to retry on `404`s to be more resilient to eventual consistency.
4. the networkd configurations in the worker installation script were moved into the rootfs bundle.

> [!NOTE]
> keep in mind that no mainstream Kubernetes CNI today leverages systemd-networkd, because routes are configured directly via kernel APIs.
> if they were to do so in the future, they would need to create entirely new `.network` files

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

**Testing Done**

*[See this guide for recommended testing for PRs.](../doc/CONTRIBUTING.md#testing-changes) Some tests may not apply. Completing tests and providing additional validation steps are not required, but it is recommended and may reduce review time and time to merge.*

PR #2432, opened 2025-09-12
fix(al2023): reload networkd and verify configurations

mselim00:
**Issue #, if available:**

**Description of changes:**
Switches back to a `networkctl reload` call in `nodeadm-boot-hook` as first introduced in https://github.com/awslabs/amazon-eks-ami/pull/2324. https://github.com/awslabs/amazon-eks-ami/pull/2397 has switched this out in favor of a `systemctl restart systemd-networkd`, with the aim of forcing synchronicity of the link reconfiguration to avoid a race condition where later `nodeadm` phases would fail to make outbound network calls. The `restart` had the unintended consequence of leaving routes configured for unmanaged links, which caused a period of network unreachability following `nodeadm` success, until the respective DHCP leases eventually expired.

This PR uses `networkctl reload` combined with `list` in order to ensure that only the primary interface is configured and all others become unmanaged. This is intended to supplement what `systemd-networkd-wait-online` will check after completion of `nodeadm-boot-hook`. `wait-online` only checks that all links are in a final state, i.e. `degraded`, `unmanaged`, or `configured`. This service more specifically checks that the primary link is `configured` and all other links are `unmanaged`, as is the intention of our systemd drop-in configuration. This should guarantee the desired outcome and a safe networking environment since, from testing, `systemd-networkd` will not mark a `configured` link as `unmanaged` until after it has removed all its routes (this is also why `wait-online` is insufficient for the use-case). 

**Testing Done**

Ran 740 `p4d.24xlarge` with 4 EFA + ENA interfaces configured (one on each network card) using the 1.33 AMI built from the CI run to make sure they always join the cluster and become ready. In specific, I spun up an ASG of 4 p4d.24xlarge and ran a pod that deleted any node that took less than 15 minutes to become ready (measured from instance launch time). Failed instances would therefore be leftover, but in this case every instance launched was ready in time to be cleaned up.

In testing, the boot hook consistently took under 3 seconds to execute (including polling the link states).

### DIFF
--- a/nodeadm/internal/system/networking.go
+++ b/nodeadm/internal/system/networking.go
@@ -4,16 +4,32 @@ import (
 	"bytes"
 	"context"
 	_ "embed"
+	"encoding/json"
 	"fmt"
 	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
 	"text/template"
+	"time"
 
 	"github.com/awslabs/amazon-eks-ami/nodeadm/internal/aws/imds"
 	"github.com/awslabs/amazon-eks-ami/nodeadm/internal/util"
 	"go.uber.org/zap"
 )
 
+type NetworkctlInterface struct {
+	Name                string `json:"Name"`
+	AdministrativeState string `json:"AdministrativeState"`
+}
+
+type NetworkctlList struct {
+	Interfaces []NetworkctlInterface `json:"Interfaces"`
+}
+
 const (
+	networkDeviceDir   = "/sys/class/net"
+	macAddressFileName = "address"
 	// the ephemeral networkd config directory, reset on reboot
 	administrationNetworkDir = "/run/systemd/network"
 	// the name of ec2 network configuration setup by amazon-ec2-net-utils:
@@ -39,34 +55,41 @@ var (
 // To address this issue temporarily, we use drop-ins to alter configuration of `80-ec2.network` after boot to make it match against primary ENI only.
 // TODO: there are limitations on current solutions as well, and we should figure long term solution for this:
 //  1. the altNames for ENIs(a new feature in AL2023) were setup by amazon-ec2-net-utils via udev rules, but it's disabled by eks.
-//
-// Returns true if the applied configuration requires a reload/restart of systemd-networkd
-func EnsureEKSNetworkConfiguration() (bool, error) {
+func EnsureEKSNetworkConfiguration() error {
 	primaryENIMac, err := imds.GetProperty(context.TODO(), "mac")
 	if err != nil {
-		return false, fmt.Errorf("failed to get MAC from IMDS: %w", err)
+		return fmt.Errorf("failed to get MAC from IMDS: %w", err)
 	}
 	networkCfgDropInDir := fmt.Sprintf("%s/%s.d", administrationNetworkDir, ec2NetworkConfigurationName)
 	eksPrimaryENIOnlyConfPathName := fmt.Sprintf("%s/%s", networkCfgDropInDir, eksPrimaryENIOnlyConfName)
 	if exists, err := util.IsFilePathExists(eksPrimaryENIOnlyConfPathName); err != nil {
-		return false, fmt.Errorf("failed to check eks_primary_eni_only network configuration existance: %w", err)
+		return fmt.Errorf("failed to check eks_primary_eni_only network configuration existance: %w", err)
 	} else if exists {
 		zap.L().Info("eks_primary_eni_only network configuration already exists, skipping configuration")
-		return false, nil
+		return nil
 	}
-
 	eksPrimaryENIOnlyConfContent, err := generateEKSPrimaryENIOnlyConfiguration(primaryENIMac)
 	if err != nil {
-		return false, fmt.Errorf("failed to generate eks_primary_eni_only network configuration: %w", err)
+		return fmt.Errorf("failed to generate eks_primary_eni_only network configuration: %w", err)
 	}
 	zap.L().Info("writing eks_primary_eni_only network configuration")
 	if err := os.MkdirAll(networkCfgDropInDir, networkConfDropInDirPerms); err != nil {
-		return false, fmt.Errorf("failed to create network configuration drop-in directory %s: %w", networkCfgDropInDir, err)
+		return fmt.Errorf("failed to create network configuration drop-in directory %s: %w", networkCfgDropInDir, err)
 	}
 	if err := os.WriteFile(eksPrimaryENIOnlyConfPathName, eksPrimaryENIOnlyConfContent, networkConfFilePerms); err != nil {
-		return false, fmt.Errorf("failed to write eks_primary_eni_only network configuration: %w", err)
+		return fmt.Errorf("failed to write eks_primary_eni_only network configuration: %w", err)
 	}
-	return true, nil
+	if err := reloadNetworkConfigurations(); err != nil {
+		return fmt.Errorf("failed to reload network configurations: %w", err)
+	}
+	primaryLinkName, err := getLinkNameByMacAddress(primaryENIMac)
+	if err != nil {
+		return fmt.Errorf("failed to determine the name of primary link: %w", err)
+	}
+	if err := ensurePrimaryOnlyConfiguration(primaryLinkName); err != nil {
+		return fmt.Errorf("failed to ensure primary ENI only configuration: %w", err)
+	}
+	return nil
 }
 
 // eksPrimaryENIOnlyTemplateVars holds the variables for eksPrimaryENIOnlyConfTemplate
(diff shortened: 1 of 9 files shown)


mselim00:
/ci
+test --instance-types=c6in.32xlarge --efa

nodeadm/internal/system/networking.go
```diff
@@ -86,3 +104,75 @@ func generateEKSPrimaryENIOnlyConfiguration(primaryENIMac string) ([]byte, error
...
+		return err
+	}
+	primaryLinkName, err := getInterfaceNameByMacAddress(primaryLinkMacAddress)
+	if err != nil {
+		return err
+	}
+	ctx, cancel := context.WithTimeout(context.Background(), 1*time.Minute)
+	defer cancel()
+	// number of retries should be greater than the time allows, just ensures we do not get stuck
+	for range 240 {
```
ndbaker1: if we have a context variable already i think we just loop
```suggestion
	for {
```

nodeadm/internal/system/networking.go
```diff
@@ -86,3 +104,75 @@ func generateEKSPrimaryENIOnlyConfiguration(primaryENIMac string) ([]byte, error
...
+			} else {
+				secondariesUnmanaged = (secondariesUnmanaged && (link.AdministrativeState == "unmanaged"))
+			}
+		}
+		if primaryManaged && secondariesUnmanaged {
+			return nil
+		}
+		if err := smithytime.SleepWithContext(ctx, 250*time.Millisecond); err != nil {
+			return err
+		}
```
ndbaker1: just me, but i prefer to just check the ctx.Done() vs time.After in a select than use a wrapper library

nodeadm/internal/system/networking.go
```diff
@@ -86,3 +111,80 @@ func generateEKSPrimaryENIOnlyConfiguration(primaryENIMac string) ([]byte, error
...
+	defer cancel()
+	// number of retries should be greater than the time allows, just ensures we do not get stuck
+	for range 240 {
+		var primaryConfigured bool
+		var rawListOutput bytes.Buffer
+		var networkctlOutput NetworkctlList
+		secondariesUnmanaged := true // assume no secondaries unless we find one
+		zap.L().Info("checking link states...")
+		cmd := exec.Command("networkctl", "list", "--json=pretty")
+		cmd.Stdout = &rawListOutput
```
ndbaker1: Curious if you care about seeing the real output in stdout, because you could use a multiwriter and pass both your buffer and os.Stdout

mselim00: the full output is pretty chunky so I skipped it in favor of verbose logging with explicit link states. here's an example of the short/non-pretty json output for a t3 with 2 interfaces:
```
{"Interfaces":[{"Index":1,"Name":"lo","Type":"loopback","Flags":65609,"FlagsString":"up,loopback,running,lower-up","KernelOperationalState":0,"KernelOperationalStateString":"unknown","MTU":65536,"MinimumMTU":0,"MaximumMTU":4294967295,"AdministrativeState":"unmanaged","OperationalState":"carrier","CarrierState":"carrier","AddressState":"off","IPv4AddressState":"off","IPv6AddressState":"off","OnlineState":null,"Addresses":[{"Family":10,"Address":[0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,1],"PrefixLength":128,"Scope":254,"ScopeString":"host","Flags":640,"FlagsString":"permanent,no-prefixroute","ConfigSource":"foreign","ConfigState":"configured"},{"Family":2,"Address":[127,0,0,1],"PrefixLength":8,"Scope":254,"ScopeString":"host","Flags":128,"FlagsString":"permanent","ConfigSource":"foreign","ConfigState":"configured"}]},{"Index":2,"Name":"ens5","AlternativeNames":["enp0s5"],"Type":"ether","Driver":"ena","Flags":69699,"FlagsString":"up,broadcast,running,multicast,lower-up","KernelOperationalState":6,"KernelOperationalStateString":"up","MTU":9001,"MinimumMTU":128,"MaximumMTU":9216,"HardwareAddress":[6,230,233,7,70,55],"PermanentHardwareAddress":[6,230,233,7,70,55],"BroadcastAddress":[255,255,255,255,255,255],"IPv6LinkLocalAddress":[254,128,0,0,0,0,0,0,4,230,233,255,254,7,70,55],"AdministrativeState":"configured","OperationalState":"routable","CarrierState":"carrier","AddressState":"routable","IPv4AddressState":"routable","IPv6AddressState":"degraded","OnlineState":"online","NetworkFile":"/usr/lib/systemd/network/80-ec2.network","RequiredForOnline":true,"RequiredOperationalStateForOnline":["degraded","routable"],"RequiredFamilyForOnline":"any","ActivationPolicy":"up","LinkFile":"/usr/lib/systemd/network/99-default.link","Path":"pci-0000:00:05.0","Vendor":"Amazon.com, Inc.","Model":"Elastic Network Adapter (ENA)","DNS":[{"Family":2,"Address":[172,31,0,2],"ConfigSource":"DHCPv4","ConfigProvider":[172,31,32,1]}],"SearchDomains":[{"Domain":"us-west-2.compute.internal","ConfigSource":"DHCPv4","ConfigProvider":[172,31,32,1]}],"DNSSettings":[{"LLMNR":"no","ConfigSource":"static"},{"MDNS":"no","ConfigSource":"static"},{"DNSDefaultRoute":true,"ConfigSource":"static"}],"Addresses":[{"Family":2,"Address":[172,31,46,43],"Broadcast":[172,31,47,255],"PrefixLength":20,"Scope":0,"ScopeString":"global","Flags":0,"FlagsString":null,"PreferredLifetimeUsec":54076810570,"ValidLifetimeUsec":54076810570,"ConfigSource":"DHCPv4","ConfigState":"configured","ConfigProvider":[172,31,32,1]},{"Family":10,"Address":[254,128,0,0,0,0,0,0,4,230,233,255,254,7,70,55],"PrefixLength":64,"Scope":253,"ScopeString":"link","Flags":128,"FlagsString":"permanent","ConfigSource":"foreign","ConfigState":"configured"}],"Routes":[{"Family":2,"Destination":[0,0,0,0],"DestinationPrefixLength":0,"Gateway":[172,31,32,1],"PreferredSource":[172,31,46,43],"Scope":0,"ScopeString":"global","Protocol":16,"ProtocolString":"16","Type":1,"TypeString":"unicast","Priority":1024,"Table":254,"TableString":"main(254)","Preference":0,"Flags":0,"FlagsString":"","ConfigSource":"DHCPv4","ConfigState":"configuring,configured","ConfigProvider":[172,31,32,1]},{"Family":2,"Destination":[172,31,0,2],"DestinationPrefixLength":32,"Gateway":[172,31,32,1],"PreferredSource":[172,31,46,43],"Scope":0,"ScopeString":"global","Protocol":16,"ProtocolString":"16","Type":1,"TypeString":"unicast","Priority":1024,"Table":254,"TableString":"main(254)","Preference":0,"Flags":0,"FlagsString":"","ConfigSource":"DHCPv4","ConfigState":"configuring,configured","ConfigProvider":[172,31,32,1]},{"Family":2,"Destination":[172,31,32,1],"DestinationPrefixLength":32,"PreferredSource":[172,31,46,43],"Scope":253,"ScopeString":"link","Protocol":16,"ProtocolString":"16","Type":1,"TypeString":"unicast","Priority":1024,"Table":254,"TableString":"main(254)","Preference":0,"Flags":0,"FlagsString":"","ConfigSource":"DHCPv4","ConfigState":"configuring,configured","ConfigProvider":[172,31,32,1]}]},{"Index":3,"Name":"ens6","AlternativeNames":["enp0s6"],"Type":"ether","Driver":"ena","Flags":69699,"FlagsString":"up,broadcast,running,multicast,lower-up","KernelOperationalState":6,"KernelOperationalStateString":"up","MTU":9001,"MinimumMTU":128,"MaximumMTU":9216,"HardwareAddress":[6,202,120,5,20,205],"PermanentHardwareAddress":[6,202,120,5,20,205],"BroadcastAddress":[255,255,255,255,255,255],"IPv6LinkLocalAddress":[254,128,0,0,0,0,0,0,4,202,120,255,254,5,20,205],"AdministrativeState":"unmanaged","OperationalState":"degraded","CarrierState":"carrier","AddressState":"degraded","IPv4AddressState":"off","IPv6AddressState":"degraded","OnlineState":null,"LinkFile":"/usr/lib/systemd/network/99-default.link","Path":"pci-0000:00:06.0","Vendor":"Amazon.com, Inc.","Model":"Elastic Network Adapter (ENA)","Addresses":[{"Family":10,"Address":[254,128,0,0,0,0,0,0,4,202,120,255,254,5,20,205],"PrefixLength":64,"Scope":253,"ScopeString":"link","Flags":128,"FlagsString":"permanent","ConfigSource":"foreign","ConfigState":"configured"}]}]}
```

nodeadm/internal/system/networking.go
mselim00: ig that would also make it clearer/more obvious how the context cancellation is being handled. will do that in the next commit

nodeadm/internal/system/networking.go
mselim00: I agree it's not really doing anything, but it's super critical this service exits since it blocks network-online.target and basically everything in the boot. the explicit range is sort of an additional safeguard to ensure that it exits in case the context handling is messed up or something, though it really shouldn't change any functionality in the current state

mselim00:
/ci
+test --instance-types=c6in.32xlarge --efa

nodeadm/internal/system/networking.go
ndbaker1: Was the 1 second an overestimate of the 240? I think this is a timed based operation so the context is really the only correct measurement

mselim00: 240 was an overestimate of the number of loops based on `context_duration / loop_time = ( 60 / ( 0.250 + execution_time ))`

but yeah you're right, that was just a bandaid, I'll replace it with an actual test case

mselim00: made that change. also made the retry a bit more aggressive by reducing the cooldown to 50 instead of 250 ms

mselim00:
\<cancelled ci>

mselim00:
\<cancelled ci>

mselim00:
/ci
+test --instance-types=c6in.32xlarge --efa

nodeadm/test/e2e/helpers.sh
```diff
@@ -124,6 +124,11 @@ function wait::server-responding() {
 
 function mock::imds() {
   local CONFIG_PATH=${1:-/etc/aemm-default-config.json}
+  # set mac dynamically to better emuluate a real run, assumes eth0 exists
+  TMP_CONFIG=$(mktemp)
+  PRIMARY_MAC_ADDR=$(cat /sys/class/net/eth0/address)
+  jq --arg mac "$PRIMARY_MAC_ADDR" '.metadata.values.mac = $mac' "$CONFIG_PATH" > "$TMP_CONFIG"
+  mv "$TMP_CONFIG" "$CONFIG_PATH"
```
ndbaker1: rather than applying this to the imds mock config, i think setting the mac like i do [here](https://github.com/awslabs/amazon-eks-ami/pull/2419/files#diff-4e6b2093c28ee08abf74654d2f51f9ad665a860a885445725967ec9c3c6bcc87R49) is easier. we wont have to update the other imds paths that way

ndbaker1 (approved):
one comment about the e2e environment, otherwise lgtm

nodeadm/test/e2e/helpers.sh
mselim00: sounds good, made that change

PR #2229, opened 2025-04-24
fix(nodeadm): calculate max pods on default network card

mselim00:
**Issue #, if available:**

**Description of changes:**

At the moment there's a gap between the max pods value from [eni-max-pods.txt](https://github.com/awslabs/amazon-eks-ami/blob/main/nodeadm/internal/kubelet/eni-max-pods.txt) and the [fallback calculator](https://github.com/awslabs/amazon-eks-ami/blob/main/nodeadm/internal/kubelet/eni_max_pods.go#L53) for multicard instance types.

Both use the equation `max # network interfaces * (max # ipv4 addresses per interface -1) +2` for the max pods value, but the text file is calculated using the max number of network interfaces for the default network card, whereas the fallback uses the total maximum number of network interfaces for the instance. To take an example, consider `p4de.24xlarge`

```
$ aws ec2 describe-instance-types --region=us-west-2 --instance-type=p4d.24xlarge | jq '.InstanceTypes[].NetworkInfo'
{
  "NetworkPerformance": "4x 100 Gigabit",
  "MaximumNetworkInterfaces": 60,
  "MaximumNetworkCards": 4,
  "DefaultNetworkCardIndex": 0,
  "NetworkCards": [
    {
      "NetworkCardIndex": 0,
      "NetworkPerformance": "100 Gigabit",
      "MaximumNetworkInterfaces": 15,
      "BaselineBandwidthInGbps": 100,
      "PeakBandwidthInGbps": 100
    },
    {
      "NetworkCardIndex": 1,
      "NetworkPerformance": "100 Gigabit",
      "MaximumNetworkInterfaces": 15,
      "BaselineBandwidthInGbps": 100,
      "PeakBandwidthInGbps": 100
    },
    {
      "NetworkCardIndex": 2,
      "NetworkPerformance": "100 Gigabit",
      "MaximumNetworkInterfaces": 15,
      "BaselineBandwidthInGbps": 100,
      "PeakBandwidthInGbps": 100
    },
    {
      "NetworkCardIndex": 3,
      "NetworkPerformance": "100 Gigabit",
      "MaximumNetworkInterfaces": 15,
      "BaselineBandwidthInGbps": 100,
      "PeakBandwidthInGbps": 100
    }
  ],
  "Ipv4AddressesPerInterface": 50,
  "Ipv6AddressesPerInterface": 50,
  "Ipv6Supported": true,
  "EnaSupport": "required",
  "EfaSupported": true,
  "EfaInfo": {
[...]

PR #2446, opened 2025-09-26
feat(nodeadm): support custom max pods expressions

### LINKED ISSUE #782: The max pods bootstrap logic is incorrect especially when prefixes are used
**What happened**:
When bootstrapping a new node, the memory reservation calculation uses the values in _/etc/eks/eni-max-pods.txt_ without any option to customize this for the new prefix mode.

This behaviour has always been wrong as there is the `USE_MAX_PODS` variable to disable using the values in _/etc/eks/eni-max-pods.txt_ as the node max pod limit but which has no effect on the memory reservation. With the prefix logic the current behaviour isn't even close enough to be safe; e.g. a m5.large instance couldn't have more than 29 pods without prefixes so a custom max limit would be lower, with prefixes the recommended max pods would be 110 so required reserves would need to be significantly higher and this logic would be well off.

The memory reservation should respect the `USE_MAX_PODS` variable. A new variable should be added with a new lookup file to support prefix mode. I'd suggest `MAX_PODS_MODE` and defaulting to `eni` with `prefix` being an option. it would be even better if `MAX_PODS` could be added to support custom max pods with the correct resource requests.

n/a

I'd be happy to open a PR to fix this if it would actually be reviewed.

**Environment**:
n/a

### LINKED ISSUE #953: Max Pods Calculator Kaizen
AWS EKS has the notion of calculating ENIs per Instance, which once calculated, provide you with the number of Pods which can be run per Instance.

Mapping is calculated from AWS EC2 API using the following formula:

- First IP on each ENI is not used for pods
- +2 for the pods that use host-networking (AWS CNI and kube-proxy)

**Number (#) of ENI * (# of IPv4 per ENI - 1) + 2**

👉🏼 https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-eni.html#AvailableIpPerENI

[max-pods-calculator.sh](https://github.com/awslabs/amazon-eks-ami/blob/master/files/max-pods-calculator.sh) helps as a Max Pod Calculator, however, it calculates on a binary outcome e.g. [two outcomes](https://github.com/awslabs/amazon-eks-ami/blob/master/files/max-pods-calculator.sh#L149-L163): **(1)** `MAX_POD_CEILING_FOR_LOW_CPU` and **(2)** `MAX_POD_CEILING_FOR_HIGH_CPU`, which is based on if the `CPU_COUNT` is higher or lower than **30**.

- ENI and IP calculation takes into account when using Pod Security Groups
- Example use cases and support for Pod Security Group users/customers
- Max Pods can be overridden explicitly

**Why is this needed**:

The calculation does not take into account two things:

1. ENI / IP calculation when using Pod Security Groups
2. Ability to override `MAX_POD_*` given a explicit setting [(existing PR](https://github.com/awslabs/amazon-eks-ami/pull/759) somewhat related)

It is needed so that *all* AWS EKS customers, whether they use Pod Security Groups or not, have feature parity in the use of this script to provide the most accurate output of ENI, IP and CPU allocation based on the subjective use of AWS.

### LINKED ISSUE #1698: AL2023 should use actual max pods to calculate kube-reserved

I'd like to see AL2023 (AL2 doesn't have the correct inputs) calculate the `kube-reserved` value from the actual max pods if provided and not always use the hypothetical instance value.

**Why is this needed**:

Running a calculation for `kube-reserved` based on max pods makes sense but using the wrong value for max pods in this calculation when there is a user defined value is part of the structured API doesn't make any sense. If I've set my node's max pods value to 110 then I expect the `kube-reserved` value to reflect this.

CC @cartermckinnon

### LINKED ISSUE #2045: feat(nodeadm): set kubelet's maxPods to support VPC CNI prefix delegation
A similar issue https://github.com/awslabs/amazon-eks-ami/issues/1902 has been closed by its author, however, it would be good to get a direct answer from the maintainers.

In AL2 we used the `/etc/eks/max-pods-calculator.sh` script to calculate max pods with the prefix delegation and to pass this number to Karpenter. However, it seems that this script is not provided in AL2023. What are the recommended alternatives now? Should we resort to fetching the script from this repo at boot time? This does not seem right

### LINKED ISSUE #2201: bug(AL2023): nodeadm: max-pods calculation doesn't account for custom networking in AL2023
**What happened**:
ln al2023 the max-pods calculation in nodeadm doesn't take into account custom networking configurations (CNI_CUSTOM_NETWORKING_ENABLED) which causes incorrect pod calculations. 

The max-pods [calculation code](https://github.com/awslabs/amazon-eks-ami/blob/main/nodeadm/internal/kubelet/eni_max_pods.go#L66) for al2023 should consider custom networking configuration and exclude the primary ENI from the calculation when custom networking is enabled, similar to how it worked in AL2 with the [max-pods-calculator.sh](https://github.com/awslabs/amazon-eks-ami/blob/b8db09326bbb215ba7407c25df4403e1de0824d4/templates/al2/runtime/max-pods-calculator.sh#L73) script.

If I understand it correctly the shell script does this:
Assume: 3 enis, 12 IPs per eni:
In Al2 shell script it would be: 2 * (12-1) + 2 =24
In AL2023 it is currently this: 3 *(12-1) + 2 = 35

Additionally, the current implementation is missing other features from AL2's calculator:
- prefix delegation support 
- max ENI configuration support
- network card-specific handling

**Environment**:
- AWS Region:
- Instance Type(s):
- Cluster Kubernetes version:
- Node Kubernetes version:
- AMI Version:

### LINKED ISSUE #2276: feat(nodeadm): enable configuring maxPods with an adjustable offset
 I would like to have an option in the node config to adjust the number of maximum pods base on the calculated value.
Example:
```yaml
apiVersion: node.eks.aws/v1alpha1
kind: NodeConfig
spec:
  kubelet:
    config:
      maxPodsOffset: 1
```
This would increase the calculated maxPods by 1.

**Why is this needed**:

When you deploy daemonsets using host network, this do not consume a new IP on nodes. We may want to account for this and increase the maximum number of pods allowed on a node, since the limit is typically based on the number of available IPs on the instance.
You can't simply set the maxPods value because you don't always know in advance which instance type will be launched (for example, when using Karpenter).

**additional info**
We currently do this using a userData script:
```bash
#!/bin/sh -e
cd /etc/kubernetes/kubelet/config.json.d
jq <00-nodeadm.conf >10-maxpods.conf '{"apiVersion": "kubelet.config.k8s.io/v1beta1", "kind": "KubeletConfiguration", "maxPods": (.maxPods+1)}'
```

mselim00:
**Issue #, if available:**
Fixes #2201 
Resolves #2276 
Closes #782 
Closes #953 
Fixes #1698 
Closes #2045 

**Description of changes:**

Adds a new kubelet configuration to the `NodeConfig` to allow insertion of a `MaxPodsExpression` which is evaluated using [cel-go](https://github.com/google/cel-go) to an int32 for use in the final kubelet configuration. The behavior is largely unchanged if an expression is not set, including the lack of the use of the cel evaluator. 

There are a few internal changes to how the `maxPods` value is calculated, but these are tested in all the relevant ways to ensure they do not become external. The most obvious of these changes is that the pre-computed value is split up into its constituent parts (# ENIs and # IPs) to be extracted from a JSON lines file and computed at runtime. 

There's one included change in the default behavior, which is to calculate the standard max pods value on the default network card rather than for all network cards on the instance. This only impacts users using an instance type not in the pre-existing `eni_max_pods.txt` file or the new `instance-info.jsonl` as this calculation is only done as a fallback. This change is a fix to prior inconsistent behavior to better line up with the values in the file cache. Some more details in https://github.com/awslabs/amazon-eks-ami/pull/2229, which is superseded by this PR.

**Testing Done**


nodeadm/api/v1alpha1/nodeconfig_types.go
```diff
@@ -84,7 +84,8 @@ type ContainerdOptions struct {
 
 // InstanceOptions determines how the node's operating system and devices are configured.
 type InstanceOptions struct {
-	LocalStorage LocalStorageOptions `json:"localStorage,omitempty"`
+	LocalStorage      LocalStorageOptions `json:"localStorage,omitempty"`
+	MaxPodsExpression string              `json:"maxPodsExpression,omitempty"`
```
shvbsle: From what I recall from the review meeting: Move this to `KubeletOptions` config

nodeadm/internal/kubelet/eni_max_pods.go
```diff
@@ -1,67 +1,119 @@
...
 // default value from kubelet
 // https://kubernetes.io/docs/reference/config-api/kubelet-config.v1beta1/#kubelet-config-k8s-io-v1beta1-KubeletConfiguration
 const defaultMaxPods = 110
 
-//go:embed eni-max-pods.txt
-var eniMaxPods string
-
-var MaxPodsPerInstanceType map[string]int
+//go:embed instance-info.jsonl
+var cachedInstanceInfoBytes []byte
```
ndbaker1, line 31: Converting this in `init` to a map still feels more natural to me

nodeadm/cmd/nodeadm-tools/instance-info/instance_info.go
ndbaker1, line 1: this wasn't exactly what i was thinking by a nodeadm tools package. if you just want to use the types/functions in the package then you can do that anywhere in the project, i was thinking something like `nodeadm/tools/instance-info/main.go`

see https://github.com/golang-standards/project-layout/tree/master/tools for conventions

nodeadm/internal/kubelet/config.go
```diff
@@ -252,11 +252,11 @@ func (ksc *kubeletConfig) withCloudProvider(cfg *api.NodeConfig, flags map[strin
 func (ksc *kubeletConfig) withDefaultReservedResources(cfg *api.NodeConfig, resources system.Resources) {
 	ksc.SystemReservedCgroup = ptr.String("/system")
 	ksc.KubeReservedCgroup = ptr.String("/runtime")
-	if maxPods, ok := MaxPodsPerInstanceType[cfg.Status.Instance.Type]; ok {
-		// #nosec G115 // known source from ec2 apis within int32 range
-		ksc.MaxPods = int32(maxPods)
+	if instanceInfo, err := GetInstanceInfo(context.TODO(), cfg.Status.Instance.Region, cfg.Status.Instance.Type); err != nil {
+		zap.L().Warn("Failed to gather instance info, falling back to default", zap.Error(err))
```
ndbaker1: "gather" is an interesting choice?
```suggestion
		zap.L().Warn("Failed to retrieve instance info, falling back to default", zap.Error(err))
```

nodeadm/internal/kubelet/eni_max_pods.go
```diff
@@ -1,67 +1,119 @@
...
-	lines := strings.Split(eniMaxPods, "\n")
-
-	for _, line := range lines {
-		line = strings.TrimSpace(line)
-		if line == "" || strings.HasPrefix(line, "#") {
+func GetInstanceInfo(ctx context.Context, awsRegion string, instanceType string) (util.InstanceInfo, error) {
+	// try to read it from the cached file first
+	cachedInfoReader := bytes.NewReader(cachedInstanceInfoBytes)
+	s := bufio.NewScanner(cachedInfoReader)
+	for s.Scan() {
```
ndbaker1:
```suggestion
	for s := bufio.NewScanner(bytes.NewReader(cachedInstanceInfoBytes)); s.Scan(); {
```

i typically try hold back purely styling nits, but i enforce this pattern to avoid polluting variable scopes

.github/workflows/sync-eni-max-pods.yaml
```diff
@@ -23,6 +23,12 @@ jobs:
           repository: awslabs/amazon-eks-ami
           ref: refs/heads/main
           path: amazon-eks-ami/
+        run: |
+          #!/usr/bin/env bash
+          set -o errexit
+          cd amazon-eks-ami/
+          make generate-instance-info
```
ndbaker1, line 30: im glad we're owning this ourselves now, can you also remove `nodeadm/internal/kubelet/eni-max-pods.txt` and remove that `cp` from the workflow flow?

nodeadm/doc/examples.md
```diff
@@ -156,3 +156,43 @@ spec:
...
+
+```
+---
+apiVersion: node.eks.aws/v1alpha1
+kind: NodeConfig
+spec:
+  cluster: ...
+  kubelet:
+    maxPodsExpression: "((default_enis - 1) * (ips_per_eni - 1)) + 2"
+```
```
ndbaker1:
````suggestion
```yaml
---
apiVersion: node.eks.aws/v1alpha1
kind: NodeConfig
spec:
  cluster: ...
  kubelet:
    maxPodsExpression: "((default_enis - 1) * (ips_per_eni - 1)) + 2"
```

⚠️ **Note**: Values set for `maxPods` in the `kubelet` config will take precedence over the result of the `maxPodsExpression`
````

File: nodeadm/internal/kubelet/eni_max_pods.go [hunk 7]
```diff
@@ -1,67 +1,119 @@
...
 		}
+	}
+	zap.L().Warn("Could not find instance info locally, attempting API request...", zap.String("instanceType", instanceType), zap.String("region", awsRegion))
+	cfg, err := config.LoadDefaultConfig(ctx, config.WithRegion(awsRegion))
+	if err != nil {
+		return util.InstanceInfo{}, err
+	}
+	ec2Client := &util.EC2Client{Client: ec2.NewFromConfig(cfg)}
+	resp, err := util.GetInstanceInfo(ctx, ec2Client, instanceType)
+	return resp, err
```
ndbaker1:
```suggestion
	return util.GetInstanceInfo(ctx, ec2Client, instanceType)
```

nodeadm/internal/kubelet/eni_max_pods.go
```diff
@@ -1,67 +1,119 @@
...
+		return -1, fmt.Errorf("failed to create environment for custom max pods expression: %w", err)
 	}
-	ec2Client := &util.EC2Client{Client: ec2.NewFromConfig(cfg)}
-	eniInfo, err := util.GetEniInfoForInstanceType(ec2Client, instanceType)
+	ast, issues := env.Compile(expression)
+	if issues != nil {
+		if issues.Err() != nil {
+			return -1, fmt.Errorf("failed to compile custom max pods expression: %w", issues.Err())
+		} else {
+			zap.L().Info("encountered non-fatal issues compiling max pods expression", zap.String("issues", issues.String()))
```
ndbaker1:
```suggestion
			zap.L().Warn("encountered non-fatal issues compiling max pods expression", zap.String("issues", issues.String()))
```

nodeadm/internal/kubelet/eni_max_pods.go
```diff
@@ -1,67 +1,119 @@
...
 		}
-
-		parts := strings.Fields(line)
-		if len(parts) != 2 {
-			continue
+		if instanceInfo.InstanceType == instanceType {
+			return instanceInfo, nil
 		}
+	}
+	zap.L().Warn("Could not find instance info locally, attempting API request...", zap.String("instanceType", instanceType), zap.String("region", awsRegion))
```
ndbaker1: nit: trying to use the same terminology as the [vpc cni](https://github.com/aws/amazon-vpc-cni-k8s/blob/a49107121e4c0337372dbe18d65f15d761973ebf/pkg/awsutils/awsutils.go#L1667)

```suggestion
	zap.L().Warn("Could not find instance info locally, making EC2 API call...", zap.String("instanceType", instanceType), zap.String("region", awsRegion))
```

nodeadm/internal/kubelet/eni_max_pods.go
mselim00: I'm not dead set on this approach but I wasn't really a fan of the init approach for 2 main reasons:
1.  unnecessary memory consumption - the embedding itself was about `18K`, now with the added data it's ~`81K`. additionally storing that as a map means more doubling the memory consumption for this one thing, and this file is likely to grow over time too (both in length and width as we add more instance types or details per instance type)
2. unnecessary computation - the init block runs unconditionally, even if the result's not needed or used (e.g. for `nodeadm-run`). this new approach only processes the data when needed and only up until the last point it's needed (i.e. it short circuits on discovery of an answer). this point may be pretty negligible, but I think these decisions add up across the project

imo `init` blocks only make sense when it's known that importers of the package cannot meaningfully utilize the package without the functionality fulfilled and that functionality cannot reasonably be done lazily

.github/workflows/sync-eni-max-pods.yaml
mselim00: I was planning to keep the VPC CNI generated one for this PR to utilize as sanity check in a test (computed value from instanceInfo = old val) and then clean it up in a follow up. if you're convinced it's consistent then I can do that cleanup now 😅

nodeadm/cmd/nodeadm-tools/instance-info/instance_info.go
mselim00: oh cool, didn't know about this standard. made that change, that's much more intuitive

nodeadm/internal/kubelet/config.go
mselim00: I was writing the same thing over and over and wanted to change it up a little lol. made that change in a diff commit though

nodeadm/doc/examples.md
mselim00: the `yaml` formatting's neat, we should do this for the others too

.github/workflows/sync-eni-max-pods.yaml
ndbaker1: i did a quick check, transforming the jsonl to txt with this to do a diff
```bash
jq -r '"\(.instanceType) \(.defaultMaxENIs * (.ipv4AddressesPerInterface - 1) + 2)"'
```
only to realize the diff i got was just the manually added entries from https://github.com/aws/amazon-vpc-cni-k8s/blob/a49107121e4c0337372dbe18d65f15d761973ebf/scripts/gen_vpc_ip_limits.go#L202-L374

nodeadm/internal/kubelet/eni_max_pods.go
ndbaker1: that's true, the fact we only call it once changes the dynamic. LGTM

nodeadm/internal/kubelet/eni_max_pods.go
```diff
@@ -1,67 +1,116 @@
...
-
-		parts := strings.Fields(line)
-		if len(parts) != 2 {
-			continue
+		if instanceInfo.InstanceType == instanceType {
+			return instanceInfo, nil
 		}
+	}
+	zap.L().Warn("Could not find instance info locally, making EC2 API call...", zap.String("instanceType", instanceType), zap.String("region", awsRegion))
+	cfg, err := config.LoadDefaultConfig(ctx, config.WithRegion(awsRegion))
```
ndbaker1, line 46: just confirming, but in general region shouldn't be needed here right

nodeadm/api/v1alpha1/nodeconfig_types.go
```diff
@@ -68,6 +68,11 @@ type KubeletOptions struct {
 	// Flags are [command-line `kubelet` arguments](https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/).
 	// that will be appended to the defaults.
 	Flags []string `json:"flags,omitempty"`
+
+	// MaxPodsExpression is a CEL expression used to compute a max pods value for
+	// the kubelet configuration. Any MaxPods value set in the config take precedence
```
fletcherw:
```suggestion
	// the kubelet configuration. Any MaxPods value set in Config takes precedence
```
nit: to clarify we're taking about the Config field above.

and I guess this needs to be applied in a couple other places as well.

nodeadm/internal/kubelet/eni_max_pods.go
```diff
@@ -1,67 +1,116 @@
...
-		if err != nil {
-			continue
+// Calcaultes a max pods value based on the provided instanceInfo and customExpression.
+// If a custom expression is not set, the default behavior should align with AL2,
+// which essentially is:
+//
+//	# of ENI on default network card * (# of IPv4 per ENI - 1) + 2
+//
+// TODO: isolate this into a public-facing package for external use by other projects
+func CalcMaxPods(instanceInfo util.InstanceInfo, customExpression string) int32 {
```
fletcherw, line 61: The logic is fairly nested. If you use early returns, it reads in a linear style which is easier to understand. It's not crucial for this function, but mentioning this since it's general good practice.

```
defaultMaxPods := calculateStandardMaxPods(instanceInfo)
if len(customExpression) == 0 {
  return defaultMaxPods
}

zap.L().Info("Applying custom max pods expression", zap.String("expression", customExpression))
customVal, err := evaluateCustomMaxPodsExpression(customExpression, instanceInfo, defaultMaxPods)
if err != nil {
	zap.L().Warn("Failed to evaluate custom expression, using standard max pods value", zap.Error(err))
    return defaultMaxPods
}

return customVal, nil
```

nodeadm/internal/kubelet/eni_max_pods.go
```diff
@@ -1,67 +1,116 @@
...
-		return defaultMaxPods
+		return -1, fmt.Errorf("failed to create environment for custom max pods expression: %w", err)
 	}
-	ec2Client := &util.EC2Client{Client: ec2.NewFromConfig(cfg)}
-	eniInfo, err := util.GetEniInfoForInstanceType(ec2Client, instanceType)
+	ast, issues := env.Compile(expression)
+	if issues != nil {
+		if issues.Err() != nil {
+			return -1, fmt.Errorf("failed to compile custom max pods expression: %w", issues.Err())
+		} else {
```
fletcherw: nit: else block can be run unconditionally if the preceding if block returns. (i.e. delete the surrounding `else { [...] }`

nodeadm/internal/kubelet/eni_max_pods_test.go
```diff
@@ -0,0 +1,325 @@
...
+			ipsPerENI:       9999,
+			standardMaxPods: 7,
+			expectedValue:   9,
+		},
+		{
+			// false variable references should error
+			expression:          "default_enis + fake_variable",
+			expectErr:           true,
+			expectedErrContents: "failed to compile custom max pods expression",
+		},
```
fletcherw, line 141: add tests for when custom expression returns negative value or 0?

nodeadm/internal/kubelet/eni_max_pods.go
```diff
@@ -1,67 +1,116 @@
...
+		ipsPerENIVar:   instanceInfo.Ipv4AddressesPerInterface,
+		maxPodsVar:     standardMaxPods,
+	})
+	castVal := rawVal.ConvertToType(cel.IntType)
+	if types.IsError(castVal) {
+		return -1, fmt.Errorf("could not interpret result %q from evaluation of custom max pods expression as an integer: %w", rawVal.Value(), err)
+	}
+	if int64Value, castOk := castVal.Value().(int64); !castOk {
+		return -1, fmt.Errorf("could not cast %v from evaluation of custom max pods expression to an integer", int64Value)
+	} else if int64Value > math.MaxInt32 || int64Value <= 0 {
```
fletcherw, line 108: do we want to enforce a higher minimum value for max pods? If a customer configures maxPods == 1, will that leave their node unusable since daemonsets will consume all the available slots?

nodeadm/internal/util/ec2.go
```diff
@@ -25,21 +31,95 @@ func (c *EC2Client) DescribeInstanceTypes(ctx context.Context, params *ec2.Descr
...
+		}
+	}
+	return getSortedInstanceTypeInfoFromMap(infoByInstanceType), nil
+}
+
+func getSortedInstanceTypeInfoFromMap(infoByInstanceType map[string]InstanceInfo) []InstanceInfo {
+	var instanceTypesInfo []InstanceInfo
+	for _, instanceTypeInfo := range infoByInstanceType {
+		instanceTypesInfo = append(instanceTypesInfo, instanceTypeInfo)
+	}
```
fletcherw, line 92: optional nit: can be written as `instanceTypesInfo := slices.Collect(maps.Values(infoByInstanceType))`

nodeadm/internal/util/ec2.go
fletcherw, line 1: What do you think about using the `InstanceType` typedef instead of `string` when referring to instance types in this file?

nodeadm/internal/kubelet/eni_max_pods.go
mselim00: I don't think it is, I'm just carrying over this logic to avoid a surprised gotcha

nodeadm/internal/kubelet/eni_max_pods.go
mselim00: good call, that's much more readable

nodeadm/internal/kubelet/eni_max_pods.go
mselim00: I wanted to keep it more inline with how the `maxPods` value in the kubelet config is handled, and it only excludes negative or zero values ([(1)](https://github.com/kubernetes/kubernetes/blob/93e18b57f3e34ca80f01e7719e216e5b324231e7/pkg/kubelet/apis/config/validation/validation.go#L123), [(2)](https://github.com/kubernetes/kubernetes/blob/93e18b57f3e34ca80f01e7719e216e5b324231e7/pkg/kubelet/apis/config/v1beta1/defaults.go#L194))

not likely to happen but in theory someone could build a custom AMI with all those daemons built in as system daemons and actually want `maxPods` to be 1 🤷

nodeadm/internal/util/ec2.go
mselim00: this just seems simpler/more intuitive to me

nodeadm/internal/util/ec2.go
mselim00: we need to support strings anyway for callers in the kubelet package to avoid importing the ec2 types, so I think it's just simpler to use strings throughout

nodeadm/internal/kubelet/eni_max_pods.go
fletcherw: sounds good.

mselim00:
hopefully the CI's not still flaky

/ci

.github/workflows/sync-eni-max-pods.yaml
mselim00: good catch, I added in those overrides to ensure consistent behavior. also temporarily added a test to do the same and removed it, along with `eni-max-pods.txt` in https://github.com/awslabs/amazon-eks-ami/pull/2446/commits/2311f7e09604d5494c3b3fab2f4c6aa530ebcb14

shvbsle (approved):
LGTM

ndbaker1 (approved):

Issue #2443, opened 2025-09-24
Nvidia Grid Driver support in AL2023_x86_64_NVIDIA


xjiaqing:
**What would you like to be added**:
Nvidia Grid Driver

**Why is this needed**:

From the [Available drivers by instance type](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/install-nvidia-driver.html#nvidia-driver-instance-type) table in [document](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/install-nvidia-driver.html#nvidia-GRID-driver), it can be seen that the NVIDIA Grid driver is required on g6f/gr6f.

Currently, when customers use g6f nodes in the EKS cluster, they need to manually install NVIDIA Grid Driver, which brings some operational difficulties to customers and hinders our promotion of g6f instances to them.

Ref: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/install-nvidia-driver.html

bryantbiggs:
possibly a duplicate of #2323

shvbsle:
EKS has added support for Nvidia GRID drivers in AL2023 Nvidia x86_64 AMIs since `v20250610` release:
https://github.com/awslabs/amazon-eks-ami/releases/tag/v20250610

Customers can verify this by spinning up a `gr6f.4xlarge` instance with latest AMI `amazon-eks-node-al2023-x86_64-nvidia-1.33-v20250915` and a fractional GPU `NVIDIA L4-12Q` will be detected:
```
sh-5.2$ nvidia-smi
Thu Sep 25 17:54:10 2025
+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 570.172.08             Driver Version: 570.172.08     CUDA Version: 12.8     |
|-----------------------------------------+------------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id          Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |           Memory-Usage | GPU-Util  Compute M. |
|                                         |                        |               MIG M. |
|=========================================+========================+======================|
|   0  NVIDIA L4-12Q                  On  |   00000000:35:00.0 Off |                    0 |
| N/A   N/A    P0            N/A  /  N/A  |       0MiB /  12288MiB |      0%      Default |
|                                         |                        |                  N/A |
+-----------------------------------------+------------------------+----------------------+

+-----------------------------------------------------------------------------------------+
[...]

PR #2442, opened 2025-09-23
feat(al2023): dynamically generate nvidia-open-supported-devices* text file from runfiles

shvbsle:
**Issue #, if available:**

**Description of changes:**
Instead of manually creating the `../templates/al2023/runtime/gpu/nvidia-open-supported-devices-<version>.txt` file everytime, we use the hack script to generate the version specific nvidia-open-supported-devices.txt file dynamically.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

**Testing Done**

1. Generated an AMI:
```bash
make -d k8s=1.33 os_distro=al2023 enable_accelerator=nvidia enable_efa=false > amibuild.log
```

2. Manually booted an instance and confirmed that  `nvidia-open-supported-devices` text file is generated at the expected location:
```
[root@ip-xx-xx-xx-xx bin]# head /etc/eks/nvidia-open-supported-devices-570.txt
# This file was generated from supported-gpus/supported-gpus.json contained in NVIDIA-Linux-x86_64-570.172.08.run
# The supported-gpus.json file is provided under the following license terms:
#
# Copyright (c) 2020 NVIDIA Corporation
#
# This software is provided 'as-is', without any express or implied
# warranty. In no event will the authors be held liable for any damages
# arising from the use of this software.
#
# Permission is granted to anyone to use this software for any purpose,
[root@ip-xx-xx-xx-xx bin]# nvidia^C
[root@ip-xx-xx-xx-xx bin]# cat /etc/eks/nvidia-open-supported-devices-570.txt | wc -l
333
```
3. Confirmed on a g6.xlarge that dkms module shows `nvidia-open` kernel-module loaded
```
[root@ip-xx-xx-xx-xx bin]# dkms status
nvidia/570.172.08, 6.12.40-64.114.amzn2023.x86_64, x86_64: installed
```

5. ci passing

*[See this guide for recommended testing for PRs.](../doc/CONTRIBUTING.md#testing-changes) Some tests may not apply. Completing tests and providing additional validation steps are not required, but it is recommended and may reduce review time and time to merge.*

PR #2450, opened 2025-10-01
fix: install grid drivers from ec2 grid runfile

shvbsle:
**Issue #, if available:**
https://github.com/awslabs/amazon-eks-ami/issues/2443

**Description of changes:**
This switches the `nvidia-open-grid` kmod installation to EC2 runfile [docs](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/install-nvidia-driver.html#nvidia-GRID-driver) which includes the grid driver licensing daemon. We qualified the Al2023 AMIs to support vGPU instances since 06/10 release. We build and archive the nvidia-grid drivers by passing `GRID_BUILD=1 GRID_BUILD_CSP=1` flag to dkms conf during the build:
```bash
  sudo sed -i "s/MAKE\[0\]=\"'make'/MAKE\[0\]=\"'make' GRID_BUILD=1 GRID_BUILD_CSP=1 /g" /usr/src/nvidia-open-grid-$NVIDIA_OPEN_VERSION/dkms.conf
```

However, this does not install the licensing daemon so after 24 hours, the driver performance is degraded. The customers would expect to see:
```
CUDA error: device doesn't have valid Grid license\nCUDA kernel errors might be asynchronously reported at some other API call, so the stacktrace below might be incorrect.\nFor debugging consider passing CUDA_LAUNCH_BLOCKING=1\nCompile with `TORCH_USE_CUDA_DSA` to enable device-side assertions.\n
```

**Testing Done**
1. Build an al2023 accelarated AMI
```
make -d k8s=1.33 os_distro=al2023 enable_accelerator=nvidia enable_efa=false
```

2. Booted a g6f.xlarge instance and confirmed that:
- dkms loads `nvidia-open-grid` kernel module
- nvidia-smi detects the correct vGPU
- License status = Licensed
- dkms archives contain three unique kernel module archives (`nvidia`, `nvidia-open`, `nvidia-open-grid`)

```
[root@ip-172-31-15-151 NVIDIA-Linux-x86_64-570.172.08-grid-aws]# dkms status
nvidia-open-grid/570.172.08, 6.12.40-64.114.amzn2023.x86_64, x86_64: built

[root@ip-172-31-15-151 NVIDIA-Linux-x86_64-570.172.08-grid-aws]# nvidia-smi
Wed Oct  1 20:08:13 2025
+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 570.172.08             Driver Version: 570.172.08     CUDA Version: 12.8     |
|-----------------------------------------+------------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id          Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |           Memory-Usage | GPU-Util  Compute M. |
|                                         |                        |               MIG M. |
|=========================================+========================+======================|
|   0  NVIDIA L4-12Q                  On  |   00000000:35:00.0 Off |                    0 |
| N/A   N/A    P0            N/A  /  N/A  |       0MiB /  12288MiB |      0%      Default |
|                                         |                        |                  N/A |
+-----------------------------------------+------------------------+----------------------+

+-----------------------------------------------------------------------------------------+
| Processes:                                                                              |
|  GPU   GI   CI              PID   Type   Process name                        GPU Memory |
|        ID   ID                                                               Usage      |
|=========================================================================================|
|  No running processes found                                                             |
+-----------------------------------------------------------------------------------------+
[root@ip-172-31-15-151 NVIDIA-Linux-x86_64-570.172.08-grid-aws]# nvidia-smi -q | grep 'vGPU Software' -A 2
    vGPU Software Licensed Product
        Product Name                      : NVIDIA RTX Virtual Workstation
        License Status                    : Licensed (Expiry: N/A)
        
```

### DIFF
--- a/templates/al2023/provisioners/install-nvidia-driver.sh
+++ b/templates/al2023/provisioners/install-nvidia-driver.sh
@@ -138,20 +138,51 @@ function archive-open-kmods() {
 
 function archive-grid-kmod() {
   local MACHINE
+  local EXTRACT_DIR="${WORKING_DIR}/NVIDIA-GRID-extract"
+  local NVIDIA_GRID_RUNFILE_NAME=""
   MACHINE=$(uname -m)
   if [ "$MACHINE" != "x86_64" ]; then
     return
   fi
-  echo "Archiving GRID kmods"
-  NVIDIA_OPEN_VERSION=$(ls -d /usr/src/nvidia-open-grid-* | sed 's/.*nvidia-open-grid-//')
-  sudo sed -i 's/PACKAGE_NAME="nvidia-open"/PACKAGE_NAME="nvidia-open-grid"/g' /usr/src/nvidia-open-grid-$NVIDIA_OPEN_VERSION/dkms.conf
-  sudo sed -i "s/MAKE\[0\]=\"'make'/MAKE\[0\]=\"'make' GRID_BUILD=1 GRID_BUILD_CSP=1 /g" /usr/src/nvidia-open-grid-$NVIDIA_OPEN_VERSION/dkms.conf
-  sudo dkms build -m nvidia-open-grid -v $NVIDIA_OPEN_VERSION
-  sudo dkms install nvidia-open-grid/$NVIDIA_OPEN_VERSION
 
+  echo "Archiving NVIDIA GRID kernel modules for major version ${NVIDIA_DRIVER_MAJOR_VERSION}"
+  NVIDIA_GRID_RUNFILE_NAME=$(aws s3 ls --recursive s3://ec2-linux-nvidia-drivers/ | \
+      grep "NVIDIA-Linux-x86_64-${NVIDIA_DRIVER_MAJOR_VERSION}" | \
+      sort -k1,2 | \
+      tail -1 | \
+      awk '{print $4}')
+  
+  if [[ -z "$NVIDIA_GRID_RUNFILE_NAME" ]]; then
+      echo "ERROR: No GRID driver found for major version ${NVIDIA_DRIVER_MAJOR_VERSION}"
+      return 1
+  fi
+
+  echo "Found GRID runfile: ${NVIDIA_GRID_RUNFILE_NAME}"
+  local GRID_RUNFILE_LOCAL_NAME=$(basename "${NVIDIA_GRID_RUNFILE_NAME}")
+  
+  echo "Downloading GRID driver runfile..."
+  aws s3 cp "s3://ec2-linux-nvidia-drivers/${NVIDIA_GRID_RUNFILE_NAME}" "${WORKING_DIR}/${GRID_RUNFILE_LOCAL_NAME}"
+  chmod +x "${WORKING_DIR}/${GRID_RUNFILE_LOCAL_NAME}"
+  echo "Extracting NVIDIA GRID driver runfile..."
+  sudo "${WORKING_DIR}/${GRID_RUNFILE_LOCAL_NAME}" --extract-only --target "${EXTRACT_DIR}"
+  
+  pushd "${EXTRACT_DIR}"
+
+  # When building the kernel module rename the package to `nvidia-open-grid` to maintain unique archive names
+  sudo sed -i 's/PACKAGE_NAME="nvidia"/PACKAGE_NAME="nvidia-open-grid"/g' kenel-open/dkms.conf
+  echo "Installing NVIDIA GRID kernel modules..."
+  sudo ./nvidia-installer \
+      --dkms \
+      --kernel-modules-type open \
+      --silent || sudo cat /var/log/nvidia-installer.log
+  
   sudo kmod-util archive nvidia-open-grid
   sudo kmod-util remove nvidia-open-grid
   sudo rm -rf /usr/src/nvidia-open-grid*
+  popd
+  # Clean up downloaded runfiles
+  sudo rm -rf "${EXTRACT_DIR}"
+  sudo rm "${WORKING_DIR}/${GRID_RUNFILE_LOCAL_NAME}"
 }
 
 function archive-proprietary-kmod() {


templates/al2023/provisioners/install-nvidia-driver.sh
```diff
@@ -138,20 +134,52 @@ function archive-open-kmods() {
...
-  sudo dkms build -m nvidia-open-grid -v $NVIDIA_OPEN_VERSION
-  sudo dkms install nvidia-open-grid/$NVIDIA_OPEN_VERSION
+
+  echo "Archiving NVIDIA GRID kernel modules for major version ${NVIDIA_DRIVER_MAJOR_VERSION}"
+  NVIDIA_GRID_RUNFILE_NAME=$(aws s3 ls --recursive s3://ec2-linux-nvidia-drivers/ \
+    | grep "NVIDIA-Linux-x86_64-${NVIDIA_DRIVER_MAJOR_VERSION}" \
+    | sort -k1,2 \
+    | tail -1 \
+    | awk '{print $4}')
+
```
mselim00, line 186: since we need the same version of the user space components across the 3 install methods, and GRID seems to be the main bottleneck in terms of adoption, can we install GRID first using the latest release for `NVIDIA_DRIVER_MAJOR_VERSION` and then install open + proprietary for that full version?

templates/al2023/provisioners/install-nvidia-driver.sh
```diff
@@ -138,20 +134,52 @@ function archive-open-kmods() {
...
+    --kernel-module-type open \
+    --silent || sudo cat /var/log/nvidia-installer.log
 
   sudo kmod-util archive nvidia-open-grid
   sudo kmod-util remove nvidia-open-grid
   sudo rm -rf /usr/src/nvidia-open-grid*
+  popd
+  # Clean up downloaded runfiles
+  sudo rm -rf "${EXTRACT_DIR}"
+  sudo rm "${WORKING_DIR}/${GRID_RUNFILE_LOCAL_NAME}"
```
mselim00: if possible, after installing grid first, can we clean up all the user space stuff we don't need (that's not GRID specific) and then leave that to be installed by the open install? this is a nice to have but it may not be necessary

templates/al2023/provisioners/install-nvidia-driver.sh
shvbsle: Good call out. I've addressed this in this commit: https://github.com/awslabs/amazon-eks-ami/pull/2450/commits/6310265db059606274bbf2a78e883e80841b89b3

Added some additional checks to ensure that across all three nvidia kmods, we maintain version consistency

templates/al2023/provisioners/install-nvidia-driver.sh
shvbsle: ~~I don't expect the userspace components of the GRID driver to introduce any breaking change however, this is definitely a good to have. I've addressed this in the following commit: https://github.com/awslabs/amazon-eks-ami/pull/2450/commits/df8fc74fe777dcadfa124858f7d1020ad5d2b458~~

~~Thankfully, the nvidia-runfile comes with an uninstall script that does the clean up for me.~~

See [thread](https://github.com/awslabs/amazon-eks-ami/pull/2450#discussion_r2403383917)

templates/al2023/provisioners/install-nvidia-driver.sh
```diff
@@ -138,36 +155,101 @@ function archive-open-kmods() {
...
+    return 1
+  fi
+
   if is-isolated-partition; then
     sudo dnf -y install "kmod-nvidia-latest-dkms-${NVIDIA_DRIVER_MAJOR_VERSION}.*"
   else
     sudo dnf -y module install nvidia-driver:${NVIDIA_DRIVER_MAJOR_VERSION}-dkms
   fi
+
+  NVIDIA_PROPRIETARY_VERSION=$(kmod-util module-version nvidia)
```
ndbaker1, line 240:
```suggestion
  local NVIDIA_PROPRIETARY_VERSION
  NVIDIA_PROPRIETARY_VERSION=$(kmod-util module-version nvidia)
```

templates/al2023/provisioners/install-nvidia-driver.sh
```diff
@@ -138,36 +155,101 @@ function archive-open-kmods() {
...
+
+  if [[ "$NVIDIA_PROPRIETARY_VERSION" != "$NVIDIA_DRIVER_FULL_VERSION" ]]; then
+    echo "ERROR: NVIDIA proprietary driver version ($NVIDIA_PROPRIETARY_VERSION) does not match GRID driver version ($NVIDIA_DRIVER_FULL_VERSION)"
+    echo "All NVIDIA drivers must be on the same version. GRID driver determines the version."
+    return 1
```
ndbaker1: shouldn't we be checking the return codes if we do this?

templates/al2023/provisioners/install-nvidia-driver.sh
```diff
@@ -138,36 +155,101 @@ function archive-open-kmods() {
...
+  sudo ./nvidia-installer --uninstall --silent || {
+    echo "ERROR: Failed to uninstall GRID driver userspace components"
+    sudo cat /var/log/nvidia-installer.log
+    return 1
+  }
+
+  popd
+  # Clean up downloaded runfiles
+  sudo rm -rf "${EXTRACT_DIR}"
+  sudo rm "${WORKING_DIR}/${GRID_RUNFILE_LOCAL_NAME}"
```
ndbaker1: why do we need to cleanup 2 things here? can't we just pull the runfile into a temp workdir, extract it there are well, and then remove the entire dir?

templates/al2023/provisioners/install-nvidia-driver.sh
```diff
@@ -138,36 +155,101 @@ function archive-open-kmods() {
...
+
+  echo "Downloading GRID driver runfile..."
+  aws s3 cp "s3://ec2-linux-nvidia-drivers/${NVIDIA_GRID_RUNFILE_NAME}" "${WORKING_DIR}/${GRID_RUNFILE_LOCAL_NAME}"
+  chmod +x "${WORKING_DIR}/${GRID_RUNFILE_LOCAL_NAME}"
+  echo "Extracting NVIDIA GRID driver runfile..."
+  sudo "${WORKING_DIR}/${GRID_RUNFILE_LOCAL_NAME}" --extract-only --target "${EXTRACT_DIR}"
+
+  pushd "${EXTRACT_DIR}"
+
+  # When building the kernel module rename the package to `nvidia-open-grid` to maintain unique archive names
```
ndbaker1:
```suggestion
  # When building the kernel module rename the package to `nvidia-open-grid` to maintain unique dkms archive names
```

templates/al2023/provisioners/install-nvidia-driver.sh
```diff
@@ -138,36 +155,101 @@ function archive-open-kmods() {
...
+    --kernel-module-type open \
+    --silent || sudo cat /var/log/nvidia-installer.log
 
   sudo kmod-util archive nvidia-open-grid
   sudo kmod-util remove nvidia-open-grid
   sudo rm -rf /usr/src/nvidia-open-grid*
+
+  # Uninstall the GRID driver userspace components since we only needed to archive the kernel modules
+  # This leaves a clean state for the open driver installation to handle userspace components
+  sudo ./nvidia-installer --uninstall --silent || {
```
ndbaker1: does `nvidia-gridd` not come with the userspace binaries?

templates/al2023/provisioners/install-nvidia-driver.sh
```diff
@@ -108,22 +121,26 @@ function archive-open-kmods() {
   # The open kernel module name changed from nvidia-open to nvidia in 570.148.08
   # Remove and re-add dkms module with the correct name. This maintains the current install and archive behavior
   NVIDIA_OPEN_VERSION=$(kmod-util module-version nvidia)
```
ndbaker1, line 133:
```suggestion
  local NVIDIA_OPEN_VERSION
  NVIDIA_OPEN_VERSION=$(kmod-util module-version nvidia)
```

templates/al2023/provisioners/install-nvidia-driver.sh
```diff
@@ -108,22 +121,26 @@ function archive-open-kmods() {
...
   sudo kmod-util archive nvidia-open
 
-  # Copy the source files to a new directory for GRID driver installation
-  sudo mkdir /usr/src/nvidia-open-grid-$NVIDIA_OPEN_VERSION
-  sudo cp -R /usr/src/nvidia-open-$NVIDIA_OPEN_VERSION/* /usr/src/nvidia-open-grid-$NVIDIA_OPEN_VERSION
-
   KMOD_MAJOR_VERSION=$(sudo kmod-util module-version nvidia-open | cut -d. -f1)
   SUPPORTED_DEVICE_FILE="${WORKING_DIR}/gpu/nvidia-open-supported-devices-${KMOD_MAJOR_VERSION}.txt"
-  sudo mv "${SUPPORTED_DEVICE_FILE}" /etc/eks/
+  sudo mv --context "${SUPPORTED_DEVICE_FILE}" /etc/eks/
```
ndbaker1: https://github.com/awslabs/amazon-eks-ami/pull/2442 might naturally work if we end up moving everything to the runfiles

ndbaker1 (commented):
nice, just have a couple of nits/questions

templates/al2023/provisioners/install-nvidia-driver.sh
```diff
@@ -24,6 +24,13 @@ function rpm_install() {
 
 echo "Installing NVIDIA ${NVIDIA_DRIVER_MAJOR_VERSION} drivers..."
 
+# The AL2023 GPU AMI currently builds and archives the following nvidia kernel modules
+# in /var/lib/dkms-archive: nvidia, nvidia-open, nvidia-open-grid. To maintain the stability
+# of the AMI, we want to ensure that all three kernel modules (and also the userspace modules)
+# are on the same NVIDIA driver version. Currently, the script installs the NVIDIA GRID drivers
+# first and decides the full NVIDIA driver version that the AMI will adhere to
+NVIDIA_DRIVER_FULL_VERSION=""
```
mselim00: I think we should set the version here and then re-use it in forming the GRID path. This makes it a bit clearer how the version was determined, and it's also necessary for the arm64 build since there's a short circuit on the grid driver install

I don't expect that two minor versions of the same driver would be maintained at the same time, but to be safe, it might be better to do a strict version comparison rather than incorporating the time, e.g. :
```suggestion
NVIDIA_DRIVER_FULL_VERSION=$(aws s3 ls --recursive s3://ec2-linux-nvidia-drivers/ \
  | grep -Eo "(NVIDIA-Linux-x86_64-)$NVIDIA_DRIVER_MAJOR_VERSION\.[0-9]+\.[0-9]+(-grid-aws\.run)" \
  | cut -d'-' -f4 \
  | sort -V \
  | tail -1)
```

templates/al2023/provisioners/install-nvidia-driver.sh
```diff
@@ -138,36 +155,101 @@ function archive-open-kmods() {
...
-  sudo sed -i "s/MAKE\[0\]=\"'make'/MAKE\[0\]=\"'make' GRID_BUILD=1 GRID_BUILD_CSP=1 /g" /usr/src/nvidia-open-grid-$NVIDIA_OPEN_VERSION/dkms.conf
-  sudo dkms build -m nvidia-open-grid -v $NVIDIA_OPEN_VERSION
-  sudo dkms install nvidia-open-grid/$NVIDIA_OPEN_VERSION
+
+  echo "Archiving NVIDIA GRID kernel modules for major version ${NVIDIA_DRIVER_MAJOR_VERSION}"
+  NVIDIA_GRID_RUNFILE_NAME=$(aws s3 ls --recursive s3://ec2-linux-nvidia-drivers/ \
+    | grep "NVIDIA-Linux-x86_64-${NVIDIA_DRIVER_MAJOR_VERSION}" \
+    | sort -k1,2 \
+    | tail -1 \
+    | awk '{print $4}')
```
mselim00, line 185: with the other change, this could now just get the latest pushed runfile for the full rendered guest driver version. this may be handy if there's a different artifact pushed for two versions of the grid driver, perhaps
```suggestion
  NVIDIA_GRID_RUNFILE_NAME=$(aws s3 ls --recursive s3://ec2-linux-nvidia-drivers/ \
    | grep "NVIDIA-Linux-x86_64-${NVIDIA_DRIVER_FULL_VERSION}" \
    | sort -k1,2 \
    | tail -1 \
    | awk '{print $4}')
```

Also, with this approach, it may be worth creating a variable for the bucket name

templates/al2023/provisioners/install-nvidia-driver.sh
```diff
@@ -138,36 +155,101 @@ function archive-open-kmods() {
...
+  if [[ -z "$NVIDIA_GRID_RUNFILE_NAME" ]]; then
+    echo "ERROR: No GRID driver found for major version ${NVIDIA_DRIVER_MAJOR_VERSION}"
+    return 1
+  fi
+
+  echo "Found GRID runfile: ${NVIDIA_GRID_RUNFILE_NAME}"
+  local GRID_RUNFILE_LOCAL_NAME
+  GRID_RUNFILE_LOCAL_NAME=$(basename "${NVIDIA_GRID_RUNFILE_NAME}")
+
+  NVIDIA_DRIVER_FULL_VERSION=$(echo "${GRID_RUNFILE_LOCAL_NAME}" | sed -n 's/NVIDIA-Linux-x86_64-\([0-9]\+\.[0-9]\+\.[0-9]\+\)-grid-aws\.run/\1/p')
```
mselim00: move this log line up earlier (to the other comment)

templates/al2023/provisioners/install-nvidia-driver.sh
```diff
@@ -138,36 +155,101 @@ function archive-open-kmods() {
...
   sudo kmod-util remove nvidia-open-grid
   sudo rm -rf /usr/src/nvidia-open-grid*
+
+  # Uninstall the GRID driver userspace components since we only needed to archive the kernel modules
+  # This leaves a clean state for the open driver installation to handle userspace components
+  sudo ./nvidia-installer --uninstall --silent || {
+    echo "ERROR: Failed to uninstall GRID driver userspace components"
+    sudo cat /var/log/nvidia-installer.log
+    return 1
+  }
```
mselim00: is gridd and so on retained in this? were they captured in the dkms archive?

also is the return 1 necessary (and is it missing from the original ./nvidia-installer execution, if so)

templates/al2023/provisioners/install-nvidia-driver.sh
```diff
@@ -108,22 +121,26 @@ function archive-open-kmods() {
   # The open kernel module name changed from nvidia-open to nvidia in 570.148.08
   # Remove and re-add dkms module with the correct name. This maintains the current install and archive behavior
   NVIDIA_OPEN_VERSION=$(kmod-util module-version nvidia)
+
+  # Sanity check to have consistent NVIDIA driver versions
+  if [[ "$NVIDIA_OPEN_VERSION" != "$NVIDIA_DRIVER_FULL_VERSION" ]]; then
+    echo "ERROR: NVIDIA open driver version ($NVIDIA_OPEN_VERSION) does not match GRID driver version ($NVIDIA_DRIVER_FULL_VERSION)"
+    echo "All NVIDIA drivers must be on the same version."
+    return 1
+  fi
```
mselim00, line 140: can you use `$NVIDIA_DRIVER_FULL_VERSION` in the `dnf` install instead? otherwise, this will error if there's a newer package version than the latest driver verison in the GRID installer bucket

templates/al2023/provisioners/install-nvidia-driver.sh
```diff
@@ -138,36 +155,101 @@ function archive-open-kmods() {
...
+
   sudo kmod-util archive nvidia
   sudo kmod-util remove nvidia
   sudo rm -rf /usr/src/nvidia*
 }
 
-archive-open-kmods
 archive-grid-kmod
+archive-open-kmods
 archive-proprietary-kmod
```
mselim00, line 255: nit: this may not be necessary after some of the comments

templates/al2023/provisioners/install-nvidia-driver.sh
shvbsle: Ack. I had added the `--context` flag to get the `/configure-selinux.sh` script to work before the change was merged. I'll revert these flags. Addressed in this commit: https://github.com/awslabs/amazon-eks-ami/pull/2450/commits/4abe246b2f4768fef3bbdb08830e573908d66dce

mselim00 (commented):
looking good, thanks, just a few more comments

templates/al2023/provisioners/install-nvidia-driver.sh
shvbsle: Addressed in https://github.com/awslabs/amazon-eks-ami/pull/2450/commits/4abe246b2f4768fef3bbdb08830e573908d66dce

templates/al2023/provisioners/install-nvidia-driver.sh
shvbsle: Addressed in https://github.com/awslabs/amazon-eks-ami/pull/2450/commits/4abe246b2f4768fef3bbdb08830e573908d66dce

templates/al2023/provisioners/install-nvidia-driver.sh
shvbsle: Addressed in https://github.com/awslabs/amazon-eks-ami/pull/2450/commits/4abe246b2f4768fef3bbdb08830e573908d66dce. Using exit 1 instead

templates/al2023/provisioners/install-nvidia-driver.sh
shvbsle: Aah yes that is why I reverted it [here](https://github.com/awslabs/amazon-eks-ami/pull/2450/commits/2dbad34842d9512620eef3a88693865e1b9839da). You are right, we'd have to rely on the user-space libraries from the grid runfile to have `nvidia-gridd`.

templates/al2023/provisioners/install-nvidia-driver.sh
shvbsle: Going to keep this ordering. See this [thread](https://github.com/awslabs/amazon-eks-ami/pull/2450#discussion_r2403383917)

templates/al2023/provisioners/install-nvidia-driver.sh
mselim00: I am a bit wary of the surface area of using the grid user space bits on the non-vGPU types vs the inverse. If we use the open bits on GRID, we theoretically currently just need to test g6f and gr6f for confidence

if this is what we’re doing, can we do some sanity checks to make sure the necessary common bits are actually the same?

templates/al2023/provisioners/install-nvidia-driver.sh
mselim00: Disregard the first part of this, hadn’t seen Nick already made the point in https://github.com/awslabs/amazon-eks-ami/pull/2450#discussion_r2403383917

templates/al2023/provisioners/install-nvidia-driver.sh
shvbsle: addressed here: https://github.com/awslabs/amazon-eks-ami/pull/2450/commits/28a5db691d6e38396bc6c9692262b8b4acbc39df

templates/al2023/provisioners/install-nvidia-driver.sh
shvbsle: This is a good-call out. I've addressed it in this commit: https://github.com/awslabs/amazon-eks-ami/pull/2450/commits/306cd0ad65632cf890b54dfc611df564332022ee

By default `dnf module install nvidia-driver` lists only the major-driver version in the streams. To fetch the module by the full driver version, we'll query dnf by the kmod name.

templates/al2023/provisioners/install-nvidia-driver.sh
shvbsle: addressed in this commit: https://github.com/awslabs/amazon-eks-ami/pull/2450/commits/2169daffa4fc8eebb5e53eab47301bdfdc7a51f0

templates/al2023/provisioners/install-nvidia-driver.sh
```diff
@@ -94,11 +111,14 @@ else
...
 
 function archive-open-kmods() {
+  local KMOD_NVIDIA_OPEN_DKMS_VERSION
   echo "Archiving open kmods"
+
   if is-isolated-partition; then
     sudo dnf -y install "kmod-nvidia-open-dkms-${NVIDIA_DRIVER_MAJOR_VERSION}.*"
   else
-    sudo dnf -y module install nvidia-driver:${NVIDIA_DRIVER_MAJOR_VERSION}-open
+    KMOD_NVIDIA_OPEN_DKMS_VERSION=$(sudo dnf module provides -q kmod-nvidia-open-dkms-${NVIDIA_DRIVER_FULL_VERSION}-1.* | grep Module | cut -d':' -f2-6)
```
mselim00: I don't think we need the `-1` in the version, and `awk` might be a bit more clear than the `cut`.`NVIDIA_OPEN_MODULE` might be a little clearer

```suggestion
    NVIDIA_OPEN_MODULE=$(sudo dnf module provides -q kmod-nvidia-open-dkms-${NVIDIA_DRIVER_FULL_VERSION}* | grep Module | cut -d':' -f2-6)
```

also an example of what the expected output `module provides` output looks like might be helpful here too for understanding the `cut`

templates/al2023/provisioners/install-nvidia-driver.sh
```diff
@@ -131,43 +156,104 @@ function archive-open-kmods() {
...
 }
 
 function archive-grid-kmod() {
   local MACHINE
+  local NVIDIA_GRID_RUNFILE_NAME
+  local GRID_INSTALLATION_TEMP_DIR
+  local EXTRACT_DIR
+
+  GRID_INSTALLATION_TEMP_DIR=$(mktemp -d)
+  EXTRACT_DIR="${GRID_INSTALLATION_TEMP_DIR}/NVIDIA-GRID-extract"
```
mselim00, line 173: we can re-use `WORKING_DIR` here instead of creating a new parent temporary directory

templates/al2023/provisioners/install-nvidia-driver.sh
```diff
@@ -131,43 +156,104 @@ function archive-open-kmods() {
...
+    echo "ERROR: No GRID driver found for driver version ${NVIDIA_DRIVER_FULL_VERSION} in EC2 S3 bucket"
+    exit 1
+  fi
+
+  echo "Found GRID runfile: ${NVIDIA_GRID_RUNFILE_NAME}"
+  local GRID_RUNFILE_LOCAL_NAME
+  GRID_RUNFILE_LOCAL_NAME=$(basename "${NVIDIA_GRID_RUNFILE_NAME}")
+
+  echo "Setting NVIDIA driver full version to: ${NVIDIA_DRIVER_FULL_VERSION} (from GRID driver)"
+
```
mselim00: can we move this log line up to where we set `NVIDIA_DRIVER_FULL_VERSION`?

templates/al2023/provisioners/install-nvidia-driver.sh
```diff
@@ -131,43 +156,105 @@ function archive-open-kmods() {
...
 }
 
 function archive-proprietary-kmod() {
+  local KMOD_NVIDIA_LATEST_DKMS_VERSION
   echo "Archiving proprietary kmods"
+
+  if [[ -z "${NVIDIA_DRIVER_FULL_VERSION:-}" ]]; then
+    echo "ERROR: NVIDIA_DRIVER_FULL_VERSION not set. GRID driver must be installed first."
+    exit 1
+  fi
```
mselim00: nit: this check's a bit redundant to the later consistency check

templates/al2023/provisioners/install-nvidia-driver.sh
```diff
@@ -131,43 +156,105 @@ function archive-open-kmods() {
...
+
   if is-isolated-partition; then
     sudo dnf -y install "kmod-nvidia-latest-dkms-${NVIDIA_DRIVER_MAJOR_VERSION}.*"
   else
-    sudo dnf -y module install nvidia-driver:${NVIDIA_DRIVER_MAJOR_VERSION}-dkms
+    KMOD_NVIDIA_LATEST_DKMS_VERSION=$(sudo dnf module provides -q kmod-nvidia-latest-dkms-${NVIDIA_DRIVER_FULL_VERSION}-1.* | grep Module | cut -d':' -f2-6)
```
mselim00: same comments as for `nividia-open`

templates/al2023/provisioners/install-nvidia-driver.sh
mselim00: discussed offline, not important

templates/al2023/provisioners/install-nvidia-driver.sh
shvbsle: just removed this since an equivalent log line exists at the top

templates/al2023/provisioners/install-nvidia-driver.sh
shvbsle: Added in comments for the latest commit and used awk `awk -F' : ' '{print $2}'` to get the module name for readability.
```bash
    # Output of `sudo dnf module provides -q kmod-nvidia-open-dkms-570.172.08* | grep Module` is:
    # Module   : nvidia-driver:570-open:20251009011129:f132e61741:x86_64
    NVIDIA_OPEN_MODULE=$(sudo dnf module provides -q kmod-nvidia-open-dkms-${NVIDIA_DRIVER_FULL_VERSION}* | grep Module | awk -F' : ' '{print $2}')
    sudo dnf -y module install ${NVIDIA_OPEN_MODULE}
```

mselim00: nice thanks,  fwiw `awk '{ print $ 3}'` should also work

mselim00:
just want to sanity check the CI on this for another PR (these results are a no-op for this change)

/ci

mselim00:
/ci

shvbsle:
/ci

shvbsle:
CI seems flaky but going to try once more:

/ci

PR #2459, opened 2025-10-08
Fix duplicate `--max-pods` arguments in KUBELET_EXTRA_ARGS

guessi:
**Issue #:** [eksctl#7946](https://github.com/eksctl-io/eksctl/issues/7946)

**Description of changes:**

- Deduplicate `--max-pods` arguments before writing kubelet config
- Keep only the last `--max-pods` value while preserving other args
- Ensure with/without/duplicate `--max-pods` passing would work
- Add test case to verify the fix

**Testing Done**

```bash
% bash test-harness.sh -c cases/duplicate-max-pods.sh
[+] Building 2.4s (22/22) FINISHED
...
=================================================================================================================
-> Executing Test Case: duplicate-max-pods.sh
✅ ✅ duplicate-max-pods.sh Tests Passed! ✅ ✅
=================================================================================================================
✅ ✅ All Tests Passed! ✅ ✅
```

### DIFF
--- a/templates/al2/runtime/bootstrap.sh
+++ b/templates/al2/runtime/bootstrap.sh
@@ -615,6 +615,13 @@ Environment='KUBELET_ARGS=$KUBELET_ARGS'
 EOF
 
 if [[ -n "$KUBELET_EXTRA_ARGS" ]]; then
+  # Extract the last --max-pods value if present, then remove all --max-pods arguments
+  LAST_MAX_PODS=$(echo "$KUBELET_EXTRA_ARGS" | grep -oE -- '--max-pods=[0-9]+' | tail -1 || true)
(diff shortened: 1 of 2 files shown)


mselim00:
Thanks for opening this @guessi!

What is the behavior without this change? Does `kubelet` not already always use the last flag provided by default? If not, given that AL2 and this script are late in their lifecycle, we're unlikely to make a behavior change like this as existing users may depend on it continuing to behave the same way. If this is mainly intended to reduce clutter in the file, I would instead question why several flags were set in the first place and avoid code complexity.

guessi:
@mselim00 actually, it should have no impact without this change, it's simply to avoid user get confused for duplicated value set.

I understand it's running late for AL2 branch (the Optimized AMI), but consider AL2 (the pure AL2) might still be used by community for a while until June 2026. I think it's worth to have it fixed.

ndbaker1:
im of the opinion that flag parsing is doing it's job here and its actually better to leave the remnants around just so its more clear who tried to set what, *especially* if the values aren't in agreement

fletcherw:
Closing this since AL2 support for EKS is end of life, and there's no actual functional breakage.

PR #2461, opened 2025-10-08
feat: allow parallel image pull on more instance types

### LINKED ISSUE #2452: Enable Parallel Pull on smaller instance sizes
Parallel Pull is currently only enabled for bigger instance types, however, for our use case we would like to be able to enable this on smaller instance sizes.

**Why is this needed**:
We have some nodes that are dedicated to a particular workload, so a new node is created and only pods for this specific workload can be scheduled onto these nodes. The container image for this workload is big, 6GB compressed / 16GB uncompressed! The node we're using has 4 vCPUs and 16GB of memory, but I believe the requirements for parallel pull are >= 8 vCPUs and >= 8GB of memory if I have read [this code](https://github.com/fletcherw/amazon-eks-ami/blob/main/nodeadm/internal/containerd/config.go#L184) correctly.

Because this node is essentially doing nothing until this workload is running, we're happy to have the CPU and memory used to pull the image as quickly as possible as it's not doing much at this time anyway.

Therefore, allowing parallel pull on smaller instance types would be great for us. I'd be happy for this to only be enabled where an extra setting is given to show that we're doing this purposely and understand the risks.

fletcherw:
Lower the threshold to include most xlarge instance types.

fixes #2452

### DIFF
--- a/nodeadm/internal/containerd/config.go
+++ b/nodeadm/internal/containerd/config.go
@@ -181,6 +181,8 @@ func UseSOCISnapshotter(cfg *api.NodeConfig, resources system.Resources) bool {
 		return false
 	}
 
-	return totalMemory >= 8*gibibyte && totalCPUMillicores >= 8000
-
+	// This should be usable on most xlarge instance types.
+	// Lower the memory threshold since an instance will have less available RAM shown in the kernel than its specs.
+	// e.g. in my test a c6a.2xlarge with 16GiB of memory showed 15.75GiB available.
+	return totalMemory >= 7*gibibyte && totalCPUMillicores >= 4000
 }
--- a/nodeadm/internal/containerd/containerd_config_test.go
+++ b/nodeadm/internal/containerd/containerd_config_test.go
@@ -1,6 +1,7 @@
 package containerd
 
 import (
+	"fmt"
 	"testing"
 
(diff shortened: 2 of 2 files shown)


nodeadm/internal/containerd/config.go
```diff
@@ -181,6 +181,8 @@ func UseSOCISnapshotter(cfg *api.NodeConfig, resources system.Resources) bool {
 		return false
 	}
 
-	return totalMemory >= 8*gibibyte && totalCPUMillicores >= 8000
-
+	// This should be usable on most xlarge instance types.
+	// Lower the memory threshold since an instance will have less available RAM shown in the kernel than its specs.
+	// e.g. in my test a c6a.2xlarge with 16GiB of memory showed 15.75GiB available.
+	return totalMemory >= 7*gibibyte && totalCPUMillicores >= 4000
```
ndbaker1, line 187: A 12.5% drop from the previous bar makes sense i guess, i'm not really aware of a precedent for this from ec2 besides just planning on 10% max being unavailable. e.g. https://repost.aws/questions/QUp9cS4T2TTRuThcUhv4XZLQ/memory-capacity-analysis-of-aws-ec2-instances-discrepancies-and-solutions-needed

PR #2471, opened 2025-10-15
fix(al2023): dynamic proxy function to bypass default caching by http.ProxyFromEnvironment

shvbsle:
**Issue #, if available:**

**Description of changes:**
Continuation of the al2023 http proxy feature. The following commit https://github.com/awslabs/amazon-eks-ami/pull/2384/commits/b799e3af333d38abd342960e24d54a6142779182 was left out after syncing changes from the main branch. This function is a wrapper around the default `http.ProxyFromEnvironment` function which cachces the proxy variables in environment when first invoked thus, preventing subsequent configuration attempts of http proxy via derived from user-data. Since, the first outbound API call in nodeadm is to the IMDS for fetching user-data, we bypass caching.

**Testing Done**
Built an AMI with these changes and launched the node in a fully private subnet with security groups only accessing the proxy-server. Confirmed that the node joins the cluster. Added a few unit-tests.

### DIFF
--- a/nodeadm/internal/aws/imds/imds_test.go
+++ b/nodeadm/internal/aws/imds/imds_test.go
@@ -30,19 +30,22 @@ func TestDynamicProxyFuncBehavior(t *testing.T) {
 			envVars: map[string]string{
 				"HTTPS_PROXY": "http://example-proxy:8080",
 			},
-			testURL: "http://169.254.169.254/latest/user-data",
+			testURL:       "http://169.254.169.254/latest/user-data",
+			expectedProxy: "",
 		},
 		{
 			name: "imds_ipv6_no_proxy",
 			envVars: map[string]string{
 				"HTTPS_PROXY": "http://example-proxy:8080",
 			},
-			testURL: "http://[fd00:ec2::254]/latest/user-data",
+			testURL:       "http://[fd00:ec2::254]/latest/user-data",
+			expectedProxy: "",
 		},
 		{
-			name:    "no_env_vars",
-			envVars: map[string]string{},
-			testURL: "https://ec2.amazonaws.com",
+			name:          "no_env_vars",
+			envVars:       map[string]string{},
+			testURL:       "https://ec2.amazonaws.com",
+			expectedProxy: "",
 		},
 	}
 


nodeadm/internal/aws/imds/imds_test.go
```diff
@@ -0,0 +1,106 @@
...
+	expectedProxy := os.Getenv("EXPECTED_PROXY")
+
+	req, _ := http.NewRequest("GET", testURL, nil)
+	proxyURL, err := dynamicProxyFunc(req)
+
+	if err != nil {
+		t.Fatalf("dynamicProxyFunc returned error: %v", err)
+	}
+
+	if shouldUseProxy {
```
fletcherw: Lines 95-106 are more or less `assert.Equal(t, expectedProxy, proxyURL.String())` no? If you remove `shouldUseProxy` and just pass `nil` to expectedProxy

nodeadm/internal/aws/imds/imds.go
```diff
@@ -40,7 +59,13 @@ type IMDSClient interface {
 }
 
 func New(retry404s bool, fnOpts ...func(*imds.Options)) *imds.Client {
+	// Create HTTP client with dynamic proxy function
+	httpClient := awshttp.NewBuildableClient().WithTransportOptions(func(tr *http.Transport) {
+		tr.Proxy = dynamicProxyFunc
+	})
+
```
ndbaker1, line 66: do we still need some kind of e2e test for all of this? it doesn't give me that much confidence knowing that just the proxier for a specific http client is updated.

nodeadm/internal/aws/imds/imds_test.go
```diff
@@ -0,0 +1,106 @@
...
+			shouldUseProxy: false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			// Testing the dynamicProxyFunc can invoke http.ProxyFromEnvironment() function
+			// which freezes the proxy environment variables for the entire process due to sync.Once caching.
+			// Hence, each test-case is run in a separate process to avoid caching.
+			cmd := exec.Command("go", "test", "-run", "TestSingleProxyCase", "./")
```
fletcherw, line 57: I don't love this nested test run, but I feel like your only alternative would be somehow hackily resetting the sync.Once within net/http so it's fine.

fletcherw (approved):

nodeadm/internal/aws/imds/imds.go
fletcherw: Agreed we should do this but not sure we should block the fix on it.

nodeadm/internal/aws/imds/imds_test.go
shvbsle: Ack. Addressed in this commit: https://github.com/awslabs/amazon-eks-ami/pull/2471/commits/62235304e9de5f5b0bbf55e34680d038cc344f1d

nodeadm/internal/aws/imds/imds.go
shvbsle: Discussed in person. To de-risk the feature going ahead with this after manual tests. Following up this PR with a full e2e test with a local proxy where we make sure the requests go to that proxy.

PR #2486, opened 2025-10-28
fix(al2023/neuron): provide neuron driver v2.21.x on inf1

mselim00:
**Issue #, if available:**

N/A but related PR: https://github.com/awslabs/amazon-eks-ami/pull/2579

**Description of changes:**

This updates how the `neuron` driver is loaded so that a `v2.21.x` driver version is always loaded on `inf1` instance types, since this is the last major version that supports them: https://awsdocs-neuron.readthedocs-hosted.com/en/latest/about-neuron/announcements/neuron2.x/announce-eos-neuron-driver-support-inf1.html. The newest version is kept added in the dkms tree at build time to offer load time optimization at boot for newer instance types.

To minimize perceptible impacts of this change, a new `neuron-kmod-load` service is introduced to be solely responsible for loading the module, which ensures consistent behavior between new boots on snapshots or reboots of the same instance. This is important because prior to this change, the latest driver would be loaded in both scenarios (on the inf1 and then in the new instance booted with the snapshot), but with this `v2.21.x` would be loaded on `inf1`. The `dkms` service included as part of the package and `systemd-modules-load` would try to reload the same version on the next instance type based on the module configurations, so this change disables those and allows `neuron-kmod-load` to correctly pick the module for the context. Building an AMI on an `inf1` for use on another instance type is not necessarily supported or guaranteed to work out, but it is a possible set up that someone may be currently using.

**Testing Done**

Verified on an `inf1.2xlarge`:

<details><summary>1. v2.21.x driver version loaded before and after boot</summary>

```
$ dkms status
aws-neuronx-legacy/2.21.37.0, 6.1.156-177.286.amzn2023.x86_64, x86_64: installed
efa-nv-peermem/1.2.2, 6.1.156-177.286.amzn2023.x86_64, x86_64: installed
efa/2.17.3, 6.1.156-177.286.amzn2023.x86_64, x86_64: installed (Original modules exist)

$ lsmod | grep 'neuron'
neuron                356352  0

$ cat /sys/module/neuron/version
2.21.37.0
```

</details>

Verified on a trn1:

<details><summary>1. Latest driver version loaded before and after boot</summary>

```
$ dkms status
aws-neuronx/2.24.7.0, 6.1.156-177.286.amzn2023.x86_64, x86_64: installed
efa-nv-peermem/1.2.2, 6.1.156-177.286.amzn2023.x86_64, x86_64: installed
efa/2.17.3, 6.1.156-177.286.amzn2023.x86_64, x86_64: installed (Original modules exist)

$ lsmod | grep 'neuron'
neuron                389120  0

$ cat /sys/module/neuron/version
2.24.7.0
```

</details> 

<details><summary>2. Ran nccl-tests with EFA, e.g.:</summary>

```
               size(B)    count(elems)    type    time:avg(us)    algbw(GB/s)    busbw(GB/s)
                     8               2    fp32           874.7           0.00           0.00
                    16               4    fp32          774.42           0.00           0.00
                    32               8    fp32           888.2           0.00           0.00
                    64              16    fp32          640.91           0.00           0.00
                   128              32    fp32           67.23           0.00           0.00
                   256              64    fp32           68.16           0.00           0.01
                   512             128    fp32           65.54           0.01           0.02
                  1024             256    fp32         1132.38           0.00           0.00
                  2048             512    fp32          771.05           0.00           0.01
                  4096            1024    fp32          890.24           0.00           0.01
                  8192            2048    fp32          769.13           0.01           0.02
                 16384            4096    fp32           899.7           0.02           0.04
                 32768            8192    fp32         1234.57           0.03           0.05
                 65536           16384    fp32         1010.24           0.06           0.13
                131072           32768    fp32          667.05           0.20           0.39
                262144           65536    fp32          901.34           0.29           0.57
                524288          131072    fp32          200.42           2.62           5.15
               1048576          262144    fp32          121.47           8.63          16.99
               2097152          524288    fp32          204.22          10.27          20.22
               4194304         1048576    fp32          338.93          12.38          24.36
               8388608         2097152    fp32          555.49          15.10          29.73
              16777216         4194304    fp32          784.57          21.38          42.10
              33554432         8388608    fp32         1073.27          31.26          61.55
              67108864        16777216    fp32         2102.19          31.92          62.85
             134217728        33554432    fp32         3974.83          33.77          66.48
             268435456        67108864    fp32         7591.04          35.36          69.62
             536870912       134217728    fp32         14551.4          36.89          72.64
            1073741824       268435456    fp32         28696.8          37.42          73.66
            2147483648       536870912    fp32        56722.37          37.86          74.54
        Avg bus bandwidth:      21.4182GB/s
```

</details> 

To test the boot times, I booted 4 `inf1.xlarge` and 4 `inf2.xlarge` and measured the execution time for `neuron-kmod-load.service` for both linear time and execution on a CPU core (the combined total). These are not enough data points for anything conclusive, but there are indications this may save about 1s of CPU time for `inf2` and no substantive difference in linear time, which may allow other boot-time processes to execute faster.

<details><summary>linear times</summary>

inf1
data: 21311ms, 25740ms, 17678ms, 21651ms
average: 21595ms

inf2
data: 22277ms, 21491ms, 18336ms, 25427ms
average: 21882.75ms

diff of averages (inf1 - inf2): -287.75ms

</details> 

<details><summary>cpu times</summary>

inf1
data: 23310ms, 23079ms, 21952ms ,22194ms
average: 22633.75ms

inf2
data: 21404ms, 22407ms, 21579ms, 21948ms
average: 21834.5ms

diff of averages (inf1 - inf2): -799.25 ms

</details> 

### DIFF
--- a/templates/al2023/provisioners/install-neuron-driver.sh
+++ b/templates/al2023/provisioners/install-neuron-driver.sh
@@ -37,4 +37,76 @@ sudo dnf -y install \
 
 sudo dnf versionlock 'kernel*'
 
-sudo dnf install -y aws-neuronx-dkms aws-neuronx-tools
+################################################################################
+### Install helpers ############################################################
+################################################################################
+
+sudo mv ${WORKING_DIR}/gpu/kmod-util /usr/bin/
+
+################################################################################
+### Archive 2.21 driver ########################################################
+################################################################################
+
+# Install and archive v2.21.x of the neuron driver as a legacy driver for inf1 support
+# https://awsdocs-neuron.readthedocs-hosted.com/en/latest/about-neuron/announcements/neuron2.x/announce-eos-neuron-driver-support-inf1.html
+sudo dnf install -y aws-neuronx-dkms-2.21*
+LEGACY_VERSION=$(kmod-util module-version aws-neuronx)
+
+sudo dkms remove "aws-neuronx/$LEGACY_VERSION" --all
+# Rename to aws-neuronx-legacy so the correct version can be easily loaded at runtime
+sudo sed -i 's/PACKAGE_NAME=.*/PACKAGE_NAME="aws-neuronx-legacy"/' /usr/src/aws-neuronx-${LEGACY_VERSION}/dkms.conf
+sudo mv /usr/src/aws-neuronx-${LEGACY_VERSION} /usr/src/aws-neuronx-legacy-${LEGACY_VERSION}
+# Disable "dkms autoinstall" from loading this module, this avoids automatic loading on boot by dkms.service
+# and allows neuron-kmod-load.service to always load the correct version for the machine instead
+sudo sed -i s/^AUTOINSTALL=.*/AUTOINSTALL=no/g /usr/src/aws-neuronx-legacy-${LEGACY_VERSION}/dkms.conf
+sudo dkms add -m aws-neuronx-legacy -v $LEGACY_VERSION
+sudo dkms build -m aws-neuronx-legacy -v $LEGACY_VERSION
+sudo dkms install -m aws-neuronx-legacy -v $LEGACY_VERSION
+
+# Archive, remove, and uninstall the driver
+sudo kmod-util archive aws-neuronx-legacy
+sudo kmod-util remove aws-neuronx-legacy
+sudo dnf remove -y aws-neuronx-dkms
+
+################################################################################
+### Archive new driver #########################################################
+################################################################################
+
+# This one should be installed last because it's what most and all new instance types
+# will use, so we kepe it loaded to offer boot time optimization for most uses
+sudo dnf install -y aws-neuronx-dkms
+DRIVER_VERSION=$(kmod-util module-version aws-neuronx)
+
+# Disable "dkms autoinstall" from loading this module, this avoids automatic loading on boot by dkms.service
+# and allows neuron-kmod-load.service to always load the correct version for the machine instead
+sudo sed -i s/^AUTOINSTALL=.*/AUTOINSTALL=no/g "/var/lib/dkms/aws-neuronx/${DRIVER_VERSION}/source/dkms.conf"
+
+# Archive but do not remove the driver
+sudo kmod-util archive aws-neuronx
+
+# Versionlock the new one to avoid unintentional installs of the driver
+sudo dnf versionlock aws-neuronx-dkms
+
+################################################################################
+### Install other dependencies #################################################
+################################################################################
+
+sudo dnf install -y aws-neuronx-tools
+
+################################################################################
+### Prepare for boot ###########################################################
+################################################################################
+
+# Disable automatic module loading by systemd based on the modules load config file
+# Writing an empty configuration ensures that even a lower priority config (e.g. in /usr/lib) is not picked up
+sudo rm -f /etc/modules-load.d/neuron.conf
+sudo ln -s /dev/null /etc/modules-load.d/neuron.conf
+
+sudo mv ${WORKING_DIR}/gpu/neuron-kmod-load.sh /etc/eks/
+sudo mv ${WORKING_DIR}/gpu/neuron-kmod-load.service /etc/systemd/system/neuron-kmod-load.service
+sudo systemctl daemon-reload
+sudo systemctl enable neuron-kmod-load.service
+
+# The postremove script used by the neuron module calls rmmod instead of the standard modrobe -r,
+# which can causes the module to still show up in lsmod and fail later checks in the validate.sh provisioner
+sudo rmmod neuron
--- a/templates/al2023/runtime/gpu/kmod-util
+++ b/templates/al2023/runtime/gpu/kmod-util
@@ -29,12 +29,16 @@ function load() {
   local MODULE_NAME="${1}"
   log "unpacking: ${MODULE_NAME}"
   local MODULE_ARCHIVE="${DKMS_ARCHIVE_DIR}/${MODULE_NAME}/*.tar.gz"
-  ${DKMS} ldtarball ${MODULE_ARCHIVE}
+  if ! ${DKMS} ldtarball ${MODULE_ARCHIVE}; then
+    # It may be meaningful to call load even when the module is already in-tree,
+    # since a module cannot be added twice but the modprobe only occurs on install
+    echo "WARNING: failed to load ${MODULE_NAME}"
+  fi
   log "unpacked: ${MODULE_NAME}"
   log "installing: ${MODULE_NAME}"
   local MODULE_VERSION
   MODULE_VERSION=$(module-version "${MODULE_NAME}")
-  ${DKMS} install -m "${MODULE_NAME}" -v "${MODULE_VERSION}"
+  ${DKMS} install --force -m "${MODULE_NAME}" -v "${MODULE_VERSION}"
   log "installed: ${MODULE_NAME}"
 }
 
--- a/templates/al2023/runtime/gpu/neuron-kmod-load.service
+++ b/templates/al2023/runtime/gpu/neuron-kmod-load.service
@@ -1,0 +1,12 @@
+[Unit]
+Description=Loading Neuron kernel modules
+After=network-online.target
+Wants=network-online.target
+
+[Service]
+Type=oneshot
+ExecStart=/etc/eks/neuron-kmod-load.sh
+RemainAfterExit=true
+
+[Install]
+WantedBy=multi-user.target
--- a/templates/al2023/runtime/gpu/neuron-kmod-load.sh
+++ b/templates/al2023/runtime/gpu/neuron-kmod-load.sh
@@ -1,0 +1,32 @@
+#!/usr/bin/env bash
+
+set -o errexit
+set -o pipefail
+set -o nounset
+
+readonly AMAZON_VENDOR_CODE="1d0f"
+# Based on https://github.com/aws-neuron/aws-neuron-driver/blob/fca19f7df31b44cbbffaf121230a66df6f59d118/neuron_device.h#L36-L39
+readonly INF1_DEVICE_IDS=("7064" "7065" "7066" "7067")
+
+NEURON_MODULE_NAME="aws-neuronx"
+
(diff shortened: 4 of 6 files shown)


templates/al2023/provisioners/install-neuron-driver.sh
```diff
@@ -37,4 +37,77 @@ sudo dnf -y install \
 
 sudo dnf versionlock 'kernel*'
 
-sudo dnf install -y aws-neuronx-dkms aws-neuronx-tools
+################################################################################
+### Install helpers ############################################################
+################################################################################
+
+sudo mv ${WORKING_DIR}/gpu/gpu-ami-util /usr/bin/
```
shvbsle: Curious to know why we need gpu-ami-util?

Looks like it just has nvidia-related stuff:
https://github.com/awslabs/amazon-eks-ami/blob/main/templates/al2023/runtime/gpu/gpu-ami-util#L3

templates/al2023/provisioners/install-neuron-driver.sh
```diff
@@ -37,4 +37,77 @@ sudo dnf -y install \
...
+sudo rm -rf /usr/src/aws-neuronx*
+
+# Do not remove the package and versionlock the current one to avoid unintentional installs of the driver
+sudo dnf versionlock aws-neuronx-dkms
+
+################################################################################
+### Install other dependencies #################################################
+################################################################################
+
+sudo dnf install -y aws-neuronx-tools
```
shvbsle, line 94: Does `aws-neuronx-tools` not have to be on the same version as that of `aws-neuronx-dkms`?

aws-neuronx-dkms ( can switch to legacy or latest depending on instance type)
aws-neuronx-tools (only latest)

templates/al2023/provisioners/install-neuron-driver.sh
```diff
@@ -37,4 +37,77 @@ sudo dnf -y install \
...
+
+################################################################################
+### Prepare for boot ###########################################################
+################################################################################
+
+# Disable automatic module loading by systemd based on the modules load config file
+# Writing an empty configuration ensures that even a lower priority config (e.g. in /usr/lib) is not picked up
+sudo rm -f /etc/modules-load.d/neuron.conf
```
shvbsle, line 102: what component creates the `/etc/modules-load.d/neuron.conf`? and whats in it? Is creating an empty one ok?

templates/al2023/provisioners/install-neuron-driver.sh
```diff
@@ -37,4 +37,77 @@ sudo dnf -y install \
...
+sudo ln -s /dev/null /etc/modules-load.d/neuron.conf
+
+sudo mv ${WORKING_DIR}/gpu/neuron-kmod-load.sh /etc/eks/
+sudo mv ${WORKING_DIR}/gpu/neuron-kmod-load.service /etc/systemd/system/neuron-kmod-load.service
+sudo systemctl daemon-reload
+sudo systemctl enable neuron-kmod-load.service
+
+# The postremove script used by the neuron module calls rmmod instead of the standard modrobe -r,
```
shvbsle, line 110: modeprobe -r should be safer since it does dependency check before unloading a module. Is there a way to increase verbosity and check why the module unloading fails with modprobe -r (i'd guess that some dependencies are broken)? Does modprobe not have an equivalent -f flag?

templates/al2023/runtime/gpu/neuron-kmod-load.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
```
shvbsle, line 1: where do you get this `.service` file from?

templates/al2023/provisioners/install-neuron-driver.sh
mselim00: as far as I'm aware, the `inf1` is continuing to be supported in the latest tools

templates/al2023/provisioners/install-neuron-driver.sh
mselim00: oops yeah this is a leftover of a different strategy I was trying out at first. will get this cleaned up

templates/al2023/provisioners/install-neuron-driver.sh
mselim00: there's a post install script for the module that creates it, the `/dev/null` symlink is the recommended approach for overriding package-provided configs https://www.freedesktop.org/software/systemd/man/latest/modules-load.d.html

templates/al2023/provisioners/install-neuron-driver.sh
mselim00: it fails because the module's in a weird limbo state where it's not actually loaded (so basic commands like `modinfo` fail or even a file system search for the same) but it shows as such. `rmmod -f` is the only way I could find to get out of that state

templates/al2023/runtime/gpu/neuron-kmod-load.service
mselim00: I created it to load the module on boot

templates/al2023/provisioners/install-neuron-driver.sh
mselim00: done in a new commit

shvbsle (dismissed):
lgtm

templates/al2023/runtime/gpu/neuron-kmod-load.service
```diff
@@ -0,0 +1,12 @@
...
+Description=Loading Neuron kernel modules
+After=network-online.target
```
fletcherw, line 3: why after network-online? Could this run earlier in boot? The script seems to not use network.

templates/al2023/runtime/gpu/neuron-kmod-load.sh
```diff
@@ -0,0 +1,32 @@
...
+  if [[ "$MATCHED_DEVICES" -gt 0 ]]; then
+    NEURON_MODULE_NAME="aws-neuronx-legacy"
+    break
+  fi
+done
+
+echo $NEURON_MODULE_NAME
+
+# For backwards compatibility, continue loading module unconditionally for now.
+# TODO: determine if this should be loaded only on instances with neuron devices, more similar to nvidia driver handling
```
fletcherw, line 24: imo if there's not an obvious reason to keep loading, I'd just take it out. Is there any reasonable way that a customer could be relying on the neuron drivers if there's no neuron device?

templates/al2023/runtime/gpu/neuron-kmod-load.sh
```diff
@@ -0,0 +1,32 @@
...
+kmod-util load "$NEURON_MODULE_NAME"
+
+# The neuron dkms module has a post install script that includes writing a modules-load configuration, we disable
+# automatic module loading by systemd based on the modules load config file to help ensure that it is always this
+# service that causes the module to be loaded. Writing an empty configuration ensures that even a lower priority
+# config of the same name (e.g. in /usr/lib) is not picked up
+sudo rm -f /etc/modules-load.d/neuron.conf
+sudo ln -s /dev/null /etc/modules-load.d/neuron.conf
```
fletcherw, line 32: seems kind of complex, why not just `touch /etc/modules-load.d/neuron.conf` to have an empty file? Is it so that it can't be written? And why does this need to be both here and in the install-kmod script?

templates/al2023/runtime/gpu/kmod-util
```diff
@@ -29,12 +29,16 @@ function load() {
...
+    # It may be meaningful to call load even when the module is already in-tree,
+    # since a module cannot be added twice but the modprobe only occurs on install
+    echo "WARNING: failed to load ${MODULE_NAME}"
+  fi
   log "unpacked: ${MODULE_NAME}"
   log "installing: ${MODULE_NAME}"
   local MODULE_VERSION
   MODULE_VERSION=$(module-version "${MODULE_NAME}")
-  ${DKMS} install -m "${MODULE_NAME}" -v "${MODULE_VERSION}"
+  ${DKMS} install --force -m "${MODULE_NAME}" -v "${MODULE_VERSION}"
```
fletcherw, line 41: why the change to force? does this need a comment?

templates/al2023/runtime/gpu/neuron-kmod-load.sh
mselim00: the file already exists, and I believe `touch` just updates the access times on an existing file but retains the contents. I did originally have an `rm` and `touch`, but then saw the symlin was the recommendation in the [manpage](https://www.freedesktop.org/software/systemd/man/latest/modules-load.d.html?__goaway_challenge=meta-refresh&__goaway_id=79e73ab2982a323adb31d60f63a79835&__goaway_referer=https%3A%2F%2Fwww.google.com%2F#Configuration%20Directories%20and%20Precedence)

the neuron dkms module also has a post-install script to write this file, so each time we `dkms install` it (at least each boot), this file gets re-written

templates/al2023/runtime/gpu/kmod-util
mselim00: sort of lumped this into the explanation for the ignoring a `dkms ldtarball` failure, I'll try to separate it out

the issue is that the module may already be added to the `dkms` tree, at which point `dkms install` would fail even if the module's not actually loaded

templates/al2023/runtime/gpu/neuron-kmod-load.service
mselim00: I'll double check but yeah I think you're right, no network use should be needed here. I'm looking into potentially switching this to a package-based approach where this may be needed though

templates/al2023/runtime/gpu/neuron-kmod-load.sh
mselim00: I agree but I would prefer to keep that separate from this. not really blocking, but one scenario could be if some user was doing some staging work (e.g. another AMI build) on a non-neuron instance type which includes some validation of the installed modules

mselim00:
Superseded by https://github.com/awslabs/amazon-eks-ami/pull/2579

PR #2489, opened 2025-10-30
fix(al2023): configure proxy before enriching config and e2e test

shvbsle:
**Issue #, if available:**
In the recent [release](https://github.com/awslabs/amazon-eks-ami/releases/tag/v20251023), the [PR to introduce support for drop-ins](https://github.com/awslabs/amazon-eks-ami/pull/2427) added a regression where, proxy is configured for nodeadm process **after** ec2 API calls.

**Description of changes:**
1. Moved `initAspects` inside `resolveConfig` and before `enrighConfig` calls happen
2. Moved moto-service to 0.0.0.0 host
3. Added an e2e regression test that brings up a http proxy server and captures the traffic that is reaching it.

**Testing Done**

### DIFF
--- a/nodeadm/cmd/nodeadm/init/init.go
+++ b/nodeadm/cmd/nodeadm/init/init.go
@@ -92,7 +92,6 @@ func (c *initCmd) Run(log *zap.Logger, opts *cli.GlobalOptions) error {
 			return err
 		}
 	}
-
 	log.Info("Loaded configuration", zap.Reflect("config", nodeConfig))
 
 	log.Info("Validating configuration..")
@@ -180,7 +179,6 @@ func (c *initCmd) resolveConfig(log *zap.Logger) (cfg *api.NodeConfig, isChanged
 		return nil, false, shouldEnrichConfig, err
 	}
 	nodeConfig, err := provider.Provide()
-
 	// if the error is just that no config is provided, then attempt to use the
 	// cached config as a fallback. otherwise, treat this as a fatal error.
 	if errors.Is(err, configprovider.ErrNoConfigInChain) && cachedConfig != nil {


nodeadm/test/e2e/infra/proxy.py
```diff
@@ -0,0 +1,52 @@
...
+        self.log_file.flush()
+
+
+def main():
+    parser = argparse.ArgumentParser(description='Minimal proxy server with traffic logging')
+    parser.add_argument('--output', required=True, help='Log file name (stored in /tmp)')
+    args = parser.parse_args()
+
+    log_path = f"/tmp/{args.output}"
+    print(f"Starting proxy server on 127.0.0.1:8080")
```
shvbsle: TODO: make the proxy port configurable

nodeadm/test/e2e/helpers.sh
```diff
@@ -199,7 +199,7 @@ export AWS_REGION=us-east-1
 
 # this is set regardless of whether the mock AWS API is started
 # because we don't want to inadvertently send requests to the real AWS API
-export AWS_ENDPOINT_URL=http://localhost:5000
+export AWS_ENDPOINT_URL=http://0.0.0.0:5000
```
shvbsle: Some lore behind why this is 0.0.0.0 and not localhost:
When running a test image, the mock API servers run on localhost. And by default, link-local addresses never take the proxy server path. However, in a real-node, AWS_ENDPOINT_URL wont be a link-local address.

Having the aws endpoint url running on 0.0.0.0 makes our tests slightly closer to a real world node.

There could be other ways to handle this that I haven't yet decided on:
1. I could override AWS_ENDPOINT_URL inside a specific test and leave it localhost here
2. Create an entry in `/etc/hosts`:
```
echo "127.0.0.1 fake-aws-endpoint" >> /etc/hosts
export AWS_ENDPOINT_URL=http://fake-aws-endpoint:5000
```

nodeadm/test/e2e/infra/requirements.txt
ndbaker1, line 1: good call adding this file 🤣

nodeadm/test/e2e/infra/Dockerfile
```diff
@@ -22,7 +22,8 @@ RUN if [[ "$CONTAINERD_VERSION" == *".*" ]]; then \
 RUN curl -O https://bootstrap.pypa.io/get-pip.py && \
     python3 get-pip.py && \
     rm get-pip.py
-RUN pip install --user 'moto[server]'
+COPY test/e2e/infra/requirements.txt /
+RUN pip install --user -r requirements.txt
```
ndbaker1, line 26: general nit on the dockerfile with this approach: if you can use the path `requirements.txt` this means the working directory is root `/`. so can we just start copying everything using `.` instead of `/` if you expect the requirements.txt` to be in the runnable workdir?

nodeadm/cmd/nodeadm/init/init.go
```diff
@@ -180,6 +177,11 @@ func (c *initCmd) resolveConfig(log *zap.Logger, opts *cli.GlobalOptions) (cfg *
 		return nil, false, err
 	}
 
+	log.Info("Setting up system init aspects...")
+	if err := c.setupAspects(log, nodeConfig, initAspects); err != nil {
+		return nil, false, err
+	}
+
```
ndbaker1, line 190: another nit: can we have this passed via a `fnOpts ...func(api.NodeConfig) error`? i just feel kinda icky putting in `[]system.SystemAspect` as part of the `resolveConfig` func signature

shvbsle: TIL. Really cool idea. Implemented this in the revision

nodeadm/test/e2e/cases/nodeadm-http-proxy/run.sh
```diff
@@ -0,0 +1,66 @@
...
+
+# Test 2 - Spin up a proxy-server and testing the traffic for nodeadm run phase
+echo "Starting HTTP Proxy Server"
+python3 /proxy.py --output nodeadm-run-trafic.log > /tmp/plogs.log &
+PROXY_SERVER_PID=$!
+echo "PROXY SERVER PID: ${PROXY_SERVER_PID}"
+echo "Sleeping for a few seconds to warm the proxy server"
+sleep 5
+
+nodeadm init --skip config
```
shvbsle, line 63: additional test-cases:
1. nodeadm init --skip config --configCache: expected to have no ec2 calls

nodeadm/test/e2e/infra/proxy.py
```diff
@@ -0,0 +1,75 @@
...
+        description='Minimal proxy server with traffic logging',
+        formatter_class=argparse.RawDescriptionHelpFormatter,
+        epilog="""
+Examples:
+  %(prog)s --output traffic.log
+  %(prog)s --output traffic.log --port 9000
+  %(prog)s -o traffic.log -p 9000
+        """
+    )
+    parser.add_argument('-o', '--output', required=True, help='Log file name (stored in /tmp)')
```
shvbsle: nit from Mazen:
1. `--output` -> `--output-path` (provide the output path) for general extensibility & flexibility

nodeadm/cmd/nodeadm/init/init.go
```diff
@@ -180,6 +176,15 @@ func (c *initCmd) resolveConfig(log *zap.Logger, opts *cli.GlobalOptions) (cfg *
 		return nil, false, err
 	}
 
+	if len(fnOpts) > 0 {
```
shvbsle: TODO: move before `cachedConfig`

nodeadm/cmd/nodeadm/init/init.go
```diff
@@ -180,6 +177,15 @@ func (c *initCmd) resolveConfig(log *zap.Logger, opts *cli.GlobalOptions) (cfg *
...
```
shvbsle: implement system aspect directly in here instead of fnOpts but leave comments wherever the function is called

nodeadm/test/e2e/cases/nodeadm-http-proxy/run.sh
```diff
@@ -0,0 +1,79 @@
...
+
+source /helpers.sh
+
+# This test checks if nodeadm's config and run phase are able to properly route ec2 traffic through an http proxy.
+# The testing container image contains the following servers:
+# 1. AWS_ENDPOINT_URL: Mocks the AWS EC2 API responses running at 0.0.0.0:5000
+# 2. AWS_EC2_METADATA_SERVICE_ENDPOINT: Mocks IMDS running at localhost:1338
+# 3. proxy server: Can be started as needed and is running at 8080.
+#
+# To enable http proxy on AL2023 AMIs pass HTTP_PRPOXY and HTTPS_PROXY in NodeConfig as shown below:
```
shvbsle: typo

nodeadm/cmd/nodeadm/init/init.go
```diff
@@ -180,6 +172,12 @@ func (c *initCmd) resolveConfig(log *zap.Logger, opts *cli.GlobalOptions) (cfg *
 		return nil, false, err
 	}
 
+	log.Info("Setting up nodeadm environment aspect...")
+	nodeadmEnvAspect := system.NewNodeadmEnvironmentAspect()
+	if err := nodeadmEnvAspect.Setup(nodeConfig); err != nil {
+		return nil, false, err
+	}
+
```
mselim00: I feel like the placement of this is a bit odd, it implicitly assumes that if the config is cached, we do not need to set up the environment aspect. this may be generally true in the current setup since `nodeadm-config` writes the cache and `nodeadm-config` runs `resolveConfig` so it would execute this aspect, but it's still not a natural implication. can we move this somewhere we're sure it'll always be ran at least once?

nodeadm/cmd/nodeadm/init/init.go
```diff
@@ -180,6 +172,12 @@ func (c *initCmd) resolveConfig(log *zap.Logger, opts *cli.GlobalOptions) (cfg *
 		return nil, false, err
 	}
 
+	log.Info("Setting up nodeadm environment aspect...")
+	nodeadmEnvAspect := system.NewNodeadmEnvironmentAspect()
+	if err := nodeadmEnvAspect.Setup(nodeConfig); err != nil {
+		return nil, false, err
+	}
```
mselim00: `resolveConfig` is (and sounds like) a read-only function, but this is a write operation. it seems unsafe from a maintenance perspective to make it do both. I'd be in favor of splitting up resolve and enrich config again if it means avoiding that, if it's not too complicated

nodeadm/internal/aws/imds/imds.go
```diff
@@ -26,8 +26,12 @@ func dynamicProxyFunc(req *http.Request) (*url.URL, error) {
 	// Link-local addresses for IMDS do not need to be going through a proxy
 	// The IMDS has two endpoints on an instance: IPv4 (169.254.169.254) and IPv6 ([fd00:ec2::254])
 	// https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-options.html
-	if req.URL.Host == "169.254.169.254" || req.URL.Host == "[fd00:ec2::254]" {
-		return nil, nil
+	hostname := req.URL.Hostname()
+	bypassHosts := []string{"169.254.169.254", "[fd00:ec2::254]", "localhost"}
```
mselim00, line 31: should `127.0.0.1` also be in here?

nodeadm/internal/aws/imds/imds.go
```diff
@@ -26,8 +26,12 @@ func dynamicProxyFunc(req *http.Request) (*url.URL, error) {
...
+	for _, host := range bypassHosts {
+		if hostname == host {
+			return nil, nil
+		}
 	}
```
mselim00: this could be done with a `slices` stdlib function:
```suggestion
	if slices.Contains(bypassHosts, host)  {
		return nil, nil
	}
```

nodeadm/test/e2e/cases/nodeadm-http-proxy/run.sh
```diff
@@ -0,0 +1,79 @@
...
+#         HTTPS_PROXY: "https://localhost:8080"
+#   cluster:
+#     apiServerEndpoint: https://example.com
+#     certificateAuthority: Y2VydGlmaWNhdGVBdXRob3JpdHk=
+#     cidr: 10.100.0.0/16
+#     name: my-cluster
+#
+# The above userdata is also base64 encoded into `http-proxy-imds-mock.json`
+mock::aws http-proxy-imds-mock.json
+mock::kubelet 1.27.0
```
mselim00: 1.27 is not officially supported, and I I generally default to the latest release version if the functionality is not kubelet version-bound
```suggestion
mock::kubelet 1.34.0
```

nodeadm/test/e2e/helpers.sh
```diff
@@ -79,7 +79,17 @@ function mock::kubelet() {
...
+  cat > /usr/bin/kubelet << 'SCRIPT'
+#!/usr/bin/env bash
+if [ "$1" = "--version" ]; then
+  echo "Kubernetes vVERSION_PLACEHOLDER"
+else
+  echo "Kubelet is running..."
+  sleep infinity
+fi
+SCRIPT
+  sed -i "s/VERSION_PLACEHOLDER/$VERSION/" /usr/bin/kubelet
```
mselim00: could this not be
```suggestion
  cat > /usr/bin/kubelet << SCRIPT
#!/usr/bin/env bash
if [ "$1" = "--version" ]; then
  echo "Kubernetes v$VERSION"
else
  echo "Kubelet is running..."
  sleep infinity
fi
SCRIPT
```
the `sed` seems over-complicated

nodeadm/test/e2e/helpers.sh
```diff
@@ -79,7 +79,17 @@ function mock::kubelet() {
     echo "Usage: mock::kubelet VERSION"
     exit 1
   fi
-  printf "#!/usr/bin/env bash\necho Kubernetes v%s\n" "$1" > /usr/bin/kubelet
+  local VERSION="$1"
```
mselim00:
```suggestion
  local VERSION
  VERSION="$1"
```
IIRC if `$1` does not exist, the current logic would likely still show a zero exit code since local variable creation succeeds

nodeadm/internal/aws/imds/imds.go
shvbsle: Not needed since we just want to bypass for imds. Added `localhost` because the test-suite uses `localhost` to mock imds

nodeadm/internal/aws/imds/imds.go
shvbsle: Nice. Sending  a revision. Much cleaner

nodeadm/cmd/nodeadm/init/init.go
shvbsle: [Ack addressed in latest revision](https://github.com/awslabs/amazon-eks-ami/pull/2489/commits/330b9e674f3a3270a5c37e4c0411197b506865b5)

nodeadm/cmd/nodeadm/init/init.go
shvbsle: Addressed in this commit:
https://github.com/awslabs/amazon-eks-ami/pull/2489/commits/330b9e674f3a3270a5c37e4c0411197b506865b5

PR #2548, opened 2025-12-05
feat: mount soci directory to instance store

fletcherw:
When instance store volumes are available, mount
/var/lib/soci-snapshotter-grpc to them as well, analogously to containerd. This will provide improved image pull performance.

**Testing Done**

Booted c5ad.large node in cluster, verified that /var/lib/soci-snapshotter-grpc is mounted to instance store when LocalStorageStrategy is enabled.

### DIFF
--- a/templates/shared/runtime/bin/setup-local-disks
+++ b/templates/shared/runtime/bin/setup-local-disks
@@ -22,6 +22,7 @@ print_help() {
   echo "--no-bind-kubelet disable bind mounting kubelet dir onto MD raid device"
   echo "--no-bind-containerd disable bind mounting containerd dir onto MD raid device"
   echo "--no-bind-pods-logs disable bind mounting /var/log/pods onto MD raid device"
+  echo "--no-bind-soci disable bind mounting SOCI container image dir onto MD raid device"
   echo "--no-bind-mounts disable all bind mounting onto MD raid device, only create and mount MD device"
   echo "-h, --help print this help"
 }
@@ -79,7 +80,7 @@ maybe_raid() {
   mount_unit_name="$(systemd-escape --path --suffix=mount "${array_mount_point}")"
   cat > "/etc/systemd/system/${mount_unit_name}" << EOF
   [Unit]
-  Description=Mount EC2 Instance Store NVMe disk RAID0
+  Description=Mount EC2 Instance Store NVMe disk RAID${raid_level}
   [Mount]
   What=UUID=${dev_uuid}
   Where=${array_mount_point}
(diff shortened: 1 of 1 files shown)


fletcherw:
/ci build
+workflow:k8s_versions 1.34

templates/shared/runtime/bin/setup-local-disks
```diff
@@ -115,6 +115,16 @@ EOF
...
 
+  ## Do this separately as path doesn't match systemd unit name.
+  if [[ "${BIND_SOCI}" == "true" ]]; then
+    if [[ "$(systemctl is-active 'var-lib-soci\x2dsnapshotter\x2dgrpc.mount')" != "active" ]]; then
+      if systemctl is-active soci-snapshotter > /dev/null 2>&1; then
+        prev_running+=" soci-snapshotter"
+      fi
+      needs_linked+=" /var/lib/soci-snapshotter-grpc"
+    fi
+  fi
```
mselim00: I think we can combine this with the other one by making the other one re-use the same `systemd-escape`'ing logic, something like this (maybe extracting mount_unit_name into a common func):
```
  for unit in "${BIND_MOUNTS[@]}"; do
    local mount_point mount_unit_name
    mount_point="/var/lib/${unit}"
    mount_unit_name="$(systemd-escape --path --suffix=mount "${mount_point}")"
    ## Check if the bind mount from the RAID already exists
    if [[ "$(systemctl is-active ${mount_unit_name})" != "active" ]]; then
      # Check if components that depend on the RAID are running and, if so, stop them
      if systemctl is-active "${unit}" > /dev/null 2>&1; then
        prev_running+=" ${unit}"
      fi
      needs_linked+=" $mount_point"
    fi
  done
```

templates/shared/runtime/bin/setup-local-disks
```diff
@@ -225,6 +236,7 @@ while [[ $# -gt 0 ]]; do
     --no-bind-mounts)
       BIND_KUBELET="false"
       BIND_CONTAINERD="false"
+      BIND_SOCI="false"
       BIND_VAR_LOG_PODS="false"
```
mselim00: should we add a dedicated `--no-bind-soci` as well?

fletcherw: I'd wait for a feature request for it. I'm trying not to extend setup-local-disks much since I'd like for us to rewrite it into go now that we're not held back by AL2.

mselim00: fair enough, would your thoughts be the same on accepting it as an environment variable instead?

templates/shared/runtime/bin/setup-local-disks
fletcherw: Done!

templates/shared/runtime/bin/setup-local-disks
```diff
@@ -159,6 +159,29 @@ EOF
...
+      echo "soci-snapshotter.service"
+      ;;
+    "/var/log/pods")
+      echo "kubelet.service"
+      ;;
+    *)
+      echo ""
+      ;;
+  esac
+}
```
mselim00, line 184: nit: maybe `BIND_MOUNTS` could be replaced with a map of mount point to unit name? kind of combines this switch case and the array logic

fletcherw: shell script maps are cursed, I'd prefer not to use them.

templates/shared/runtime/bin/setup-local-disks
fletcherw: added the flag.

templates/shared/runtime/bin/setup-local-disks
mselim00: very fair

Issue #2545, opened 2025-12-01
NOTICE: Amazon Linux 2 AMI end of support


mselim00:
As of November 26, 2025, the Amazon Linux 2 (AL2) EKS-optimized Amazon Machine Images (AMIs) have reached end of support. We will release the final official AL2-based AMIs in the week starting 12/8/2025. See [EKS AMI Deprecation FAQs](https://docs.aws.amazon.com/eks/latest/userguide/eks-ami-deprecation-faqs.html) for common questions on the deprecation and the [EKS AL2023 User Guide](https://docs.aws.amazon.com/eks/latest/userguide/al2023.html) for guidance on how to manage the migration to Amazon Linux 2023.

AL2-specific logic will be removed from the main branch following the final AMI release in the week of 12/8/2025. A snapshot of the repository will be created in the [al2 branch](https://github.com/awslabs/amazon-eks-ami/tree/al2) of this repository to enable users to manage security patches for their own custom-built images until Amazon Linux discontinues support (see [Amazon Linux 2 FAQs](https://aws.amazon.com/amazon-linux-2/faqs/)). Going forward, we will only review pull requests on this branch and issues related to it on a best-effort basis as needed to maintain that functionality, no new features will be introduced.

ketanxjs:
Hi @mselim00 

Thanks for the update, however on the website here it is mentioned the End of support for AL2 should be June 30, 2026.

https://aws.amazon.com/amazon-linux-2/faqs/

Can you clarify?

Thanks!

mselim00:
@ketanxjs that FAQ is for AL2 in general, EKS discontinued support for the distro earlier as described in https://docs.aws.amazon.com/eks/latest/userguide/eks-ami-deprecation-faqs.html

PR #2571, opened 2025-12-19
Add AWS profile option to max pods calculator script

zAbuQasem:
Updated the script to have a `--profile` option.

image


mselim00:
thanks for the contribution! unfortunately this project no longer offers new feature support for AL2, we have committed to only allowing changes necessary to preserve existing functionality. see https://github.com/awslabs/amazon-eks-ami/issues/2545 for more details

edit: for what it's worth, it also seems like the missing link here is assigning an instance profile for this machine. see https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_use_switch-role-ec2_instance-profiles.html for more details

PR #2573, opened 2025-12-24
fix(nodeadm): clarify missing cluster ID error for outposts

HusainZafar:
…st is enabled

**Issue #, if available:**

**Description of changes:**
The error message was misleading. Corrected the message

**Testing Done**

### DIFF
--- a/nodeadm/internal/api/validation.go
+++ b/nodeadm/internal/api/validation.go
@@ -17,7 +17,7 @@ func ValidateNodeConfig(cfg *NodeConfig) error {
 	}
 	if enabled := cfg.Spec.Cluster.EnableOutpost; enabled != nil && *enabled {
 		if cfg.Spec.Cluster.ID == "" {
-			return fmt.Errorf("CIDR is missing in cluster configuration")
+			return fmt.Errorf("cluster ID must be provided for outposts")
 		}
 	}
 	return nil


nodeadm/internal/api/validation.go
```diff
@@ -17,7 +17,7 @@ func ValidateNodeConfig(cfg *NodeConfig) error {
 	}
 	if enabled := cfg.Spec.Cluster.EnableOutpost; enabled != nil && *enabled {
 		if cfg.Spec.Cluster.ID == "" {
-			return fmt.Errorf("CIDR is missing in cluster configuration")
+			return fmt.Errorf("ID is mandatory in cluster configuration when EnableOutpost is true")
```
mselim00: one thing is we would want to be more consistent with
https://github.com/awslabs/amazon-eks-ami/blob/fbf782628d3a6c7d2e071171d1f926d2c33c2720/nodeadm/internal/kubelet/config.go#L172-L174

mselim00 (approved):
thanks!

nodeadm/internal/api/validation.go
HusainZafar: Isn't that a bit confusing since clusterId is not an actual field and requires an extra hop to understand what that is? wrt the current file, `in cluster configuration` is consistent. Alternatively, we should at least say `Spec.Cluster.ID cannot be empty...` or `Cluster.ID cannot be empty...`?

mselim00: I agree that `clusterId` is a bit confusing, I'd favor plain english that describes why we're failing over a JSON path though. something like "cluster ID must be provided for outposts"

HusainZafar: Updated to this

Issue #1839, opened 2024-06-06
amazon-ebs: fatal error: An error occurred (404) when calling the HeadObject operation: Key "1.30/2022-06-06/bin/linux/amd64/kubelet" does not exist


eravindar12:
**What happened**:

I am attempting to use the AL2023 base AMI for EKS optimized with the AMI Release v20240531. However, I am encountering a **"fatal error: An error occurred (404) when calling the HeadObject operation: Key "1.30/2022-06-06/bin/linux/amd64/kubelet" does not exist"** This appears to be a bug.

Do you have any suggestions for a workaround to fix this issue?

**Error details**:

```
2024-06-06T19:54:42Z:     amazon-ebs: Downloading binaries from: s3://amazon-eks
2024-06-06T19:54:43Z:     amazon-ebs: AWS cli present - using it to copy binaries from s3.
2024-06-06T19:54:44Z:     amazon-ebs: fatal error: An error occurred (404) when calling the HeadObject operation: Key "1.30/2022-06-06/bin/linux/amd64/kubelet" does not exist
2024-06-06T19:54:44Z: ==> amazon-ebs: Provisioning step had errors: Running the cleanup provisioner, if present...
2024-06-06T19:54:44Z: ==> amazon-ebs: Terminating the source AWS instance...
2024-06-06T19:55:44Z: ==> amazon-ebs: Cleaning up any extra volumes...
2024-06-06T19:55:45Z: ==> amazon-ebs: No volumes to clean up, skipping
2024-06-06T19:55:45Z: ==> amazon-ebs: Deleting temporary security group...
2024-06-06T19:55:45Z: ==> amazon-ebs: Deleting temporary keypair...
2024-06-06T19:55:45Z: Build 'amazon-ebs' errored after 2 minutes 2 seconds: Script exited with non-zero exit status: 1. Allowed exit codes are: [0]
==> Wait completed after 2 minutes 2 seconds
==> Some builds didn't complete successfully and had errors:
--> amazon-ebs: Script exited with non-zero exit status: 1. Allowed exit codes are: [0]
==> Builds finished but no artifacts were created.
make[1]: *** [Makefile:95: k8s] Error 1
```

templates/install-worker.sh

```
################################################################################
### Kubernetes #################################################################
################################################################################

sudo mkdir -p /etc/kubernetes/manifests
sudo mkdir -p /var/lib/kubernetes
sudo mkdir -p /var/lib/kubelet
[...]

Issue #1838, opened 2024-06-06
amazon-ebs: sudo: unable to execute /tmp/worker/awscli-install/aws/install: Permission denied


eravindar12:
**What happened**:

I am attempting to use the AL2023 base AMI for EKS optimized with the AMI Release v20240531. However, I am encountering a "permission denied" error. I have tried several different methods to resolve this issue, including using "{{.Vars}} bash '{{.Path}}'", but none have been successful. This appears to be a bug.

Do you have any suggestions for a workaround to fix this issue?

**Error Details:**
**amazon-ebs: sudo: unable to execute /tmp/worker/awscli-install/aws/install: Permission denied**

```
2024-06-06T21:18:33Z:     amazon-ebs: Installing awscli v2 bundle
2024-06-06T21:18:35Z:     amazon-ebs: sudo: unable to execute /tmp/worker/awscli-install/aws/install: Permission denied
2024-06-06T21:18:35Z: ==> amazon-ebs: Provisioning step had errors: Running the cleanup provisioner, if present...
2024-06-06T21:18:35Z: ==> amazon-ebs: Terminating the source AWS instance...
2024-06-06T21:19:35Z: ==> amazon-ebs: Cleaning up any extra volumes...
2024-06-06T21:19:36Z: ==> amazon-ebs: No volumes to clean up, skipping
2024-06-06T21:19:36Z: ==> amazon-ebs: Deleting temporary security group...
2024-06-06T21:19:36Z: ==> amazon-ebs: Deleting temporary keypair...
2024-06-06T21:19:36Z: Build 'amazon-ebs' errored after 2 minutes 4 seconds: Script exited with non-zero exit status: 1. Allowed exit codes are: [0]
==> Wait completed after 2 minutes 4 seconds
==> Some builds didn't complete successfully and had errors:
--> amazon-ebs: Script exited with non-zero exit status: 1. Allowed exit codes are: [0]
```

**code snippet**

template/al2023/install-workser.sh
```
################################################################################
### awscli #####################################################################
################################################################################

### isolated regions can't communicate to awscli.amazonaws.com so installing awscli through dnf
ISOLATED_REGIONS="${ISOLATED_REGIONS:-us-iso-east-1 us-iso-west-1 us-isob-east-1}"
if ! [[ ${ISOLATED_REGIONS} =~ $BINARY_BUCKET_REGION ]]; then
  # https://docs.aws.amazon.com/cli/latest/userguide/getting-started-install.html
[...]

PR #1556, opened 2024-01-03
Update generate-version-info.sh

varun-armis:
**Issue #, if available:**

It seems like install-worker makes kubelet and aws only accessible to the root user. When running this packer build as ec2-user, I get the following errors. 

```
2024-01-03T12:23:22-05:00: ==> amazon-ebs: Provisioning with shell script: /Users/varun.shivakumar/code/fedramp/ami/packer-al2/scripts/generate-version-info.sh
2024-01-03T12:23:23-05:00:     amazon-ebs: /home/ec2-user/script_9031.sh: line 19: /usr/bin/kubelet: Permission denied

2024-01-03T12:32:11-05:00: ==> amazon-ebs: Provisioning with shell script: /Users/varun.shivakumar/code/fedramp/ami/packer-al2/scripts/generate-version-info.sh
2024-01-03T12:32:13-05:00:     amazon-ebs: /home/ec2-user/script_1670.sh: line 26: aws: command not found
```

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

**Testing Done**

*[See this guide for recommended testing for PRs.](../doc/CONTRIBUTING.md#testing-changes) Some tests may not apply. Completing tests and providing additional validation steps are not required, but it is recommended and may reduce review time and time to merge.*

PR #2576, opened 2025-12-31
run binaries as a privileged user to avoid permission denied errors to support hardened AMIs

varun-armis:
**Issue #, if available:**
Related PRs/Issues: 

- https://github.com/awslabs/amazon-eks-ami/pull/1661
- https://github.com/awslabs/amazon-eks-ami/pull/1513
- https://github.com/awslabs/amazon-eks-ami/pull/1556
- https://github.com/awslabs/amazon-eks-ami/issues/1839
- https://github.com/awslabs/amazon-eks-ami/issues/1838

**Description of changes:**

Problem

When building a hardened AMI with SELinux enabled (enforcing), the amazon-eks-ami Packer build fails during provisioning. Several binaries installed or copied by the build are placed in /usr/bin with incorrect Unix ownership/permissions and without an executable SELinux file context.

As a result, these binaries cannot be executed during subsequent build steps, causing the Packer build to fail.

Solution

I am working on a separate branch to update the selinux contexts and permissions for these binaries but what works is just sudo-ing all the affected binaries. 

Unfortunately, we're not able to set ssh_username=root because there is a restriction to SSH into the machine as root in our hardened environment.

**Testing Done**

We have successfully built a working EKS AMI from this in both ARM and x86 using a base AL2023 hardened with https://github.com/ansible-lockdown/AMAZON2023-CIS

### DIFF
--- a/templates/al2023/provisioners/cache-pause-container.sh
+++ b/templates/al2023/provisioners/cache-pause-container.sh
@@ -5,5 +5,5 @@ set -o errexit
 set -o pipefail
 
 sudo systemctl start containerd
-cache-pause-container -i ${PAUSE_CONTAINER_IMAGE}
+sudo cache-pause-container -i ${PAUSE_CONTAINER_IMAGE}
 sudo systemctl stop containerd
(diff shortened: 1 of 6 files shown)


fletcherw:
I don't think adding `sudo` everywhere is the correct fix. Several of these binaries were missing the selinux executable context but those should have been fixed several months ago with https://github.com/awslabs/amazon-eks-ami/commit/89ce6578d9601510b28b28276c61ef556508b385

It's possible that others still have the issue though. Can you provide the reproduction steps for building the hardened base AMI and then building your EKS AMI?

fletcherw:
Closing since there's been no followup

PR #2604, opened 2026-01-26
feat: handle users calling nodeadm init directly

fletcherw:
Though it's not supported, users sometimes call nodeadm directly from user data.

This causes a race condition where the VPC CNI starts earlier than expected and ENIs that it attaches are marked as managed by systemd.

Have nodeadm's run phase write a marker file when it starts and use that as the marker file for managing ENIs to ensure this race doesn't happen.

**Testing:**

Boot nodes that directly call nodeadm init in user data and validate that the race doesn't happen.


nodeadm/internal/system/marker.go
```diff
@@ -0,0 +1,36 @@
...
+}
+
+func (a *markerAspect) Setup(*api.NodeConfig) error {
+	if _, err := os.Stat("/run/cloud-init/result.json"); os.IsNotExist(err) {
+		zap.L().Warn("cloud-init result file /run/cloud-init/result.json does not exist. Do not manually call nodeadm from user data")
+	}
+	if err := os.MkdirAll(markerDir, 0750); err != nil {
+		return err
+	}
+	return os.WriteFile(MarkerPath(), nil, 0600)
```
mselim00: there's a [WriteFileWithDir util](https://github.com/awslabs/amazon-eks-ami/blob/91446788b516b1ef0aa8ba25be7a0ef90d99d535/nodeadm/internal/util/files.go#L12) to simplify this a little this

nodeadm/internal/system/marker.go
```diff
@@ -0,0 +1,36 @@
...
+	return filepath.Join(markerDir, "init")
+}
+
+func (a *markerAspect) Name() string {
+	return "marker"
+}
+
+func (a *markerAspect) Setup(*api.NodeConfig) error {
+	if _, err := os.Stat("/run/cloud-init/result.json"); os.IsNotExist(err) {
+		zap.L().Warn("cloud-init result file /run/cloud-init/result.json does not exist. Do not manually call nodeadm from user data")
```
mselim00, line 30: nice add!

nodeadm/internal/system/marker.go
fletcherw: Done.

PR #2606, opened 2026-01-27
chore: move local-disks logic into golang

fletcherw:
Since AL2 support for EKS is EOL, move this logic into golang and add a real e2e test.

### DIFF
--- a/nodeadm/internal/system/local_disk.go
+++ b/nodeadm/internal/system/local_disk.go
@@ -1,12 +1,59 @@
 package system
 
 import (
+	"bytes"
+	_ "embed"
+	"errors"
+	"fmt"
 	"os"
 	"os/exec"
+	"path/filepath"
+	"slices"
+	"sort"
 	"strings"
+	"text/template"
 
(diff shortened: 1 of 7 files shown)


nodeadm/internal/system/local_disk.go
```diff
@@ -20,31 +67,328 @@ func (a *localDiskAspect) Name() string {
...
+
+	if info, err := os.Stat(mdConfigPath); err != nil || info.Size() == 0 {
+		args := []string{"--create", "--force", "--verbose", mdDevice,
+			fmt.Sprintf("--level=%d", level), fmt.Sprintf("--name=%s", mdName),
+			fmt.Sprintf("--raid-devices=%d", len(disks))}
+		args = append(args, disks...)
+		if err := runCmd("mdadm", args...); err != nil {
+			return fmt.Errorf("creating RAID array: %w", err)
+		}
+		out, err := exec.Command("mdadm", "--detail", "--scan").Output()
```
mselim00: should this also use `runCmd` to get a more useful error?

PR #2607, opened 2026-01-29
feat(nodeadm): pass nvidia gpu startup labels to kubelet 

fletcherw:
Adopting #2272.

Attempts to bridge an existing gap between EKS infra and upstream kubernetes NVIDIA componenets.

The NVIDIA GPU Operator install and leverages Node Feature Discovery to set PCI labels on nodes. If a PCI device matches the NVIDIA vendor ID (something like feature.node.kubernetes.io/pci-10de.present=true), the operator adds the high-level nvidia.com/gpu.present label, which has become the standard for scheduling within the NVIDIA kubernetes ecosystem.

ref: https://github.com/NVIDIA/gpu-operator/commit/e25291b86cf4542ac62d8635cda4bd653c4face3
anecdote: https://github.com/NVIDIA/gpu-operator/issues/78

This PR sets nvidia.com/gpu.present=true in a startup label for the node via the --node-labels cli flag, because we know can fetch the pci information ourselves during bootstrap.

Testing Done

`make test test-e2e`


nodeadm/internal/kubelet/config.go
```diff
@@ -278,22 +298,6 @@ func (ksc *kubeletConfig) withDefaultReservedResources(cfg *api.NodeConfig, reso
 	}
 }
 
-// withPodInfraContainerImage determines whether to add the
-// '--pod-infra-container-image' flag, which is used to ensure the sandbox image
-// is not garbage collected.
-//
-// TODO: revisit once the minimum supportted version catches up or the container
-// runtime is moved to containerd 2.0
-func (ksc *kubeletConfig) withPodInfraContainerImage(cfg *api.NodeConfig, flags map[string]string) error {
```
shvbsle, line 287: Do you know why we are removing this (is it because 1.29 reached end of support)?

nodeadm/internal/system/devices.go
```diff
@@ -0,0 +1,28 @@
...
+// are attached to the instance.
+func IsPCIVendorAttached(vendorId string) (bool, error) {
+	vendorPaths, err := filepath.Glob("/sys/bus/pci/devices/*/vendor")
+	if err != nil {
+		return false, err
+	}
+	for _, vendorPath := range vendorPaths {
+		vendorIdBytes, err := os.ReadFile(filepath.Clean(vendorPath))
+		if err != nil {
+			continue
```
shvbsle, line 23: is it worth adding a log here or are we ok silently continuing?

nodeadm/internal/system/devices.go
```diff
@@ -0,0 +1,28 @@
...
+
+// IsPCIVendorAttached returns whether any pcie devices with a given vendor id
+// are attached to the instance.
+func IsPCIVendorAttached(vendorId string) (bool, error) {
+	vendorPaths, err := filepath.Glob("/sys/bus/pci/devices/*/vendor")
+	if err != nil {
+		return false, err
+	}
+	for _, vendorPath := range vendorPaths {
+		vendorIdBytes, err := os.ReadFile(filepath.Clean(vendorPath))
```
shvbsle: Looks like `filepath.Glob` is gonna be cleaning up the path so you may not need filepath.Clean:
https://cs.opensource.google/go/go/+/refs/tags/go1.25.6:src/path/filepath/match.go;l=270

nodeadm/internal/kubelet/config.go
```diff
@@ -213,6 +213,26 @@ func (ksc *kubeletConfig) withOutpostSetup(cfg *api.NodeConfig) error {
...
+		}
+		if !ok {
+			continue
+		}
+		nodeLabel := fmt.Sprintf("%s=%s", nodeLabelKey, nodeLabelValue)
+		zap.L().Info("Adding node label", zap.String("label", nodeLabel))
+		nodeLabels = append(nodeLabels, nodeLabel)
+	}
+	if len(nodeLabels) > 0 {
+		flags["node-labels"] = strings.Join(nodeLabels, ",")
```
shvbsle, line 232: what are the contents of `flags["node-labels"]` when it reaches here? Can we make sure we are not overwriting existing value under the `node-label` key?

nodeadm/internal/kubelet/config.go
fletcherw: yeah that's why but I'll leave it out of this change.

nodeadm/internal/kubelet/config.go
fletcherw: we only set this value once. if a user set the node-labels flag it would appear later and be passed separately to the kubelet.

nodeadm/internal/system/devices.go
fletcherw: filepath.Clean does more than just remove the trailing separator, it also normalizes `..` and `.` in the path, but I think we can safely assume that glob is not going to return such a path.

nodeadm/internal/system/devices.go
fletcherw: good idea, done.

shvbsle:
Changes look good. I'm ok to approve once the e2e tests pass.

shvbsle (approved):
lgtm

fletcherw:
/ci
+workflow:k8s_versions 1.35

PR #2610, opened 2026-02-03
chore: create file system fake

fletcherw:
Create a file system abstraction and pass it into our resources library so that we can test the resources logic.

Greatly simplify the CPU count logic.

Update some unit tests to use the new file system abstraction.

### DIFF
--- a/nodeadm/internal/system/filesystem.go
+++ b/nodeadm/internal/system/filesystem.go
@@ -1,0 +1,171 @@
+package system
+
+import (
+	"io/fs"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+)
+
+type FileSystem interface {
+	Glob(pattern string) ([]string, error)
+	ReadFile(name string) ([]byte, error)
+	ReadDir(name string) ([]fs.DirEntry, error)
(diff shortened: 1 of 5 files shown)


nodeadm/internal/system/filesystem.go
```diff
@@ -0,0 +1,171 @@
...
+	return os.Stat(name)
+}
+
+const EmptyDirectoryMarker = "[empty directory]"
+
+type FakeFileSystem struct {
+	Files map[string]string
+}
+
+func (f FakeFileSystem) allPaths() map[string]bool {
```
shvbsle, line 44: If an explicit path is provided in FakeFileSystem.Files then this function would just sythesize the parent directories right? 

```go
// input to the func
{
"/sys/some/path",
}

// output of the function:
{
"/sys/some/path": true,
"/sys/some": true,
"/sys": true,
}
```

I believe this is whats happening. Correct me if I'm wrong. nit: Perhaps worth explaining in a docstring?

fletcherw: yes this is correct. we needed this so that Glob could handle matching on parent directories, not just files.

Issue #1667, opened 2024-02-14
Missing runtime metrics from cAdvisor


stevehipwell:
**What happened**:
The kubelet cAdvisor metrics are missing the runtime values meaning we can't see the node usage.

**What you expected to happen**:
I'd expect to see the runtime values as part of the kubelet metrics.

**How to reproduce it (as minimally and precisely as possible)**:
Run `kubectl get --raw "/api/v1/nodes/${NODE_NAME}/proxy/stats/summary" | jq -r '.node.systemContainers[] | .name'` and you'll see the following response.

```shell
kubelet
pods
```

**Anything else we need to know?**:
This appears to be broken because the `--runtime-cgroups=/runtime.slice/containerd.service` kubelet argument hasn't been set. I can confirm that if this argument has been set the outcome of the command above is a as follows.

```shell
kubelet
runtime
pods
```

**Environment**:
- AWS Region: n/a
- Instance Type(s): n/a
- EKS Platform version: n/a
- Kubernetes version: n/a
- AMI Version: v20240209
- Kernel: n/a
- Release information: n/a
```

<!-- If this is a security issue, please do not discuss on GitHub. Please report any suspected or confirmed security issues to AWS Security https://aws.amazon.com/security/vulnerability-reporting/ -->

stevehipwell:
CC @cartermckinnon

PR #2617, opened 2026-02-06
fix: set the runtime-cgroup flag

fletcherw:
Set runtime-cgroup flag on the kubelet so that cadvisor metrics will include resource usage for containers.

Fixes: https://github.com/awslabs/amazon-eks-ami/issues/1667

Tests: booted an instance with this flag set and confirmed that the metrics were available.


mselim00 (approved):
🚀

PR #2620, opened 2026-02-09
feat(nodeadm): add console logging support to nodeadm

chenwng:
**Issue #, if available:**

N/A

**Description of changes:**

This PR adds a global flag `--console` to enable sending nodeadm logs to the node console. This helps troubleshoot node not joining cluster issues and the access to node is limited.  The global flag `--console` defaults to false to be backward compatible.

**Testing Done**

### DIFF
--- a/templates/al2023/runtime/rootfs/etc/systemd/system/nodeadm-boot-hook.service
+++ b/templates/al2023/runtime/rootfs/etc/systemd/system/nodeadm-boot-hook.service
@@ -15,6 +15,7 @@ Before=systemd-networkd-wait-online.service
 [Service]
 Type=oneshot
 ExecStart=/usr/bin/nodeadm-internal boot-hook
+StandardError=journal+console
 
 [Install]
 WantedBy=multi-user.target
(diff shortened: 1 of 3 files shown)


fletcherw:
Hello! This is an interesting change, but I'm not sure we want to enable this by default. Can you say more about your issues with "This helps troubleshoot node not joining cluster issues and the access to node is limited." Do you have nodes where SSM or SSH access are not possible?

templates/al2023/runtime/rootfs/etc/systemd/system/nodeadm-boot-hook.service
```diff
@@ -14,7 +14,7 @@ Before=systemd-networkd-wait-online.service
 
 [Service]
 Type=oneshot
-ExecStart=/usr/bin/nodeadm-internal boot-hook
+ExecStart=/usr/bin/nodeadm-internal --console boot-hook
```
fletcherw, line 19: Please just do this by adding `StandardError=journal+console` immediately after ExecStart here and below.

chenwng:
Thanks for the suggestion. I've updated the PR. Please help review it.

fletcherw:
gosec failure is a false positive, proceeding to merge.

Issue #527, opened 2020-08-25
AWS_ACCESS_KEY_ID environment variable issue


danielkubat:
**What happened**:
When `AWS_ACCESS_KEY_ID` is not defined, script fails with `AWS_ACCESS_KEY_ID: unbound variable` error.

**What you expected to happen**:
Binaries should be fetch from `s3` according to condition:
https://github.com/awslabs/amazon-eks-ami/blob/88648213079b4c0f70fe69c479724519327d7ae1/scripts/install-worker.sh#L168

**How to reproduce it (as minimally and precisely as possible)**:
Don't define `AWS_ACCESS_KEY_ID` and run `make` from project root.

**Anything else we need to know?**:
Issue can be easily fixed by removing:
https://github.com/awslabs/amazon-eks-ami/blob/88648213079b4c0f70fe69c479724519327d7ae1/scripts/install-worker.sh#L4

because checking, if variables are defined is done in fucntion:
https://github.com/awslabs/amazon-eks-ami/blob/88648213079b4c0f70fe69c479724519327d7ae1/scripts/install-worker.sh#L13

willthames:
This is also a bad check - AWS_ACCESS_KEY_ID can be unset and the aws cli might still work because of instance profiles or AWS cli profiles or federated web identity (list probably nowhere near exhaustive). #296 works around it being unset but we could also just try `aws sts get-caller-identity` or something (or if getting it over HTTP is fine if the CLI doesn't work, just get it over HTTP all the time)

Issue #1772, opened 2024-04-25
Build fails when passing AWS_ACCESS_KEY_ID because it assumes that key has access to bucket 


mebays:
**What happened**: Running packer to build an image it would error on awscli commands then an access_key_id was passed in. The access_key is the key being used to create the instance with packer. I don't think it should be assumed that this key also has access to the s3 bucket. 

**What you expected to happen**: When using an AWS_ACCESS_KEY_ID and not trying to use a private bucket the build should default to using wget to obtain the binaries. 

**How to reproduce it (as minimally and precisely as possible)**: Use an AWS_ACCESS_KEY_ID that does not have permissions to a private bucket.  

**Anything else we need to know?**: I think the core issue would be relating to using aws s3 cp if an AWS_ACCESS_KEY_ID is set.  If a user is using a session token there is not an AWS_ACCESS_KEY_ID in the environment variables and it will default to wget. I would probably argue that is the norm on role based permission. So if access to a private bucket is needed it should probably be declared on the instance profile and triggered with something like PRIVATE_BUCKET vs assuming to use aws cli just based on the presence of an access_key_id.

fletcherw:
Fixed by #2621. If you have a private bucket, you need to provide credentials to the provisioners either by passing in env variables or by setting an iam_instance_role in packer. For a public bucket, the script will fall back to an unauthenticated pull if the authenticated pull fails.

Issue #762, opened 2021-09-21
Error fetching EKS binaries via aws cli from GovCloud


sl4mmy:
**What happened**:

We are trying to build custom AMIs for EKS inside an AWS GovCloud account (us-gov-west-1 Region).  Unfortunately, due to the way we currently have configured command line access into that account we always end up with `AWS_ACCESS_KEY_ID` exported in the environment of the `make(1)` process when running builds in this repo.  As a result, the test on line 207 of `scripts/install-worker.sh` always succeeds and we try to download the EKS binaries via the aws cli instead of `wget` (see also this comment: https://github.com/awslabs/amazon-eks-ami/pull/175#issuecomment-474510028).

Unfortunately, when downloading the binaries via the aws cli from within GovCloud we get the following error when line 209 of `scripts/install-worker.sh` is executed: `fatal error: An error occurred (400) when calling the HeadObject operation: Bad Request`

A larger snippet of the failure is shown below for additional context (I can also post the full build output, if necessary, but this should be easy enough for anyone to reproduce from another GovCloud account):
```
...
    amazon-ebs: Adding versionlock on: 0:docker-19.03.13ce-1.amzn2
    amazon-ebs: versionlock added: 1
    amazon-ebs: overlay
    amazon-ebs: br_netfilter
    amazon-ebs: net.bridge.bridge-nf-call-ip6tables = 1
    amazon-ebs: net.bridge.bridge-nf-call-iptables = 1
    amazon-ebs: net.ipv4.ip_forward = 1
    amazon-ebs: Downloading binaries from: s3://amazon-eks
    amazon-ebs: AWS cli present - using it to copy binaries from s3.
    amazon-ebs: fatal error: An error occurred (400) when calling the HeadObject operation: Bad Request
==> amazon-ebs: Provisioning step had errors: Running the cleanup provisioner, if present...
==> amazon-ebs: Terminating the source AWS instance...
==> amazon-ebs: Cleaning up any extra volumes...
==> amazon-ebs: No volumes to clean up, skipping
==> amazon-ebs: Deleting temporary security group...
==> amazon-ebs: Deleting temporary keypair...
Build 'amazon-ebs' errored after 5 minutes 9 seconds: Script exited with non-zero exit status: 1.Allowed exit codes are: [0]

==> Wait completed after 5 minutes 9 seconds

==> Some builds didn't complete successfully and had errors:
[...]

PR #2621, opened 2026-02-10
fix: make govcloud builds possible

fletcherw:
Make various fixes so that govcloud builds work:

* Make latest-binaries.sh handle fetching cross-partition
* Change s3 copy logic to not check the AWS_ACCESS_KEY_ID variable, which is a fundamentally flawed check, and instead just attempt a copy using the environment's credentials and fall back to --no-sign-request if the initial copy fails.
* update the cache-pause-container script to use the eks-distro public mirror

The net effect of this change is that by default, none of the provisioners rely on the caller's AWS credentials. I'm leaving in the logic that passes through the AWS credentials via environment variables since removing is a breaking change, but none of them are needed.

In a future major release, it would be nice to remove those entirely and require people to use instance roles if they need permissions in the provisioners (e.g. for private artifact buckets).

Fixes: #527, #762, #1772

Testing:
Builds in us-gov-west-1 and us-gov-east-1 with AWS credential env vars unset.
Build in us-gov-west-1 with AWS credential env vars set.
k8s 1.29 and 1.35 builds in us-west-2 with AWS credential env vars unset.

### DIFF
--- a/hack/latest-binaries.sh
+++ b/hack/latest-binaries.sh
@@ -4,16 +4,24 @@ set -o errexit
 set -o pipefail
 set -o nounset
 
-if [ "$#" -ne 1 ]; then
-  echo "usage: $0 KUBERNETES_MINOR_VERSION"
+if [ "$#" -ne 2 ]; then
+  echo "usage: $0 KUBERNETES_MINOR_VERSION REGION"
   exit 1
 fi
 
 MINOR_VERSION="${1}"
+REGION="${2}"
+
+S3_EXTRA_ARGS=""
+if [[ "${REGION}" == us-gov* || "${REGION}" == eusc* ]]; then
+  # Pull cross-partition from aws since there's no bucket in this partition.
+  S3_EXTRA_ARGS="--no-sign-request --region us-west-2"
+fi
 
 # retrieve the available "VERSION/BUILD_DATE" prefixes (e.g. "1.28.1/2023-09-14")
 # from the binary object keys, sorted in descending semver order, and pick the first one
-LATEST_BINARIES=$(aws s3api list-objects-v2 --bucket amazon-eks --prefix "${MINOR_VERSION}" --query 'Contents[*].[Key]' --output text | grep linux | cut -d'/' -f-2 | sort -Vru | head -n1)
+LATEST_BINARIES=$(aws s3api list-objects-v2 ${S3_EXTRA_ARGS} --bucket amazon-eks --prefix "${MINOR_VERSION}" --query 'Contents[*].[Key]' --output text \
+  | grep linux | cut -d'/' -f-2 | sort -Vru | head -n1)
 
 if [ "${LATEST_BINARIES}" == "None" ]; then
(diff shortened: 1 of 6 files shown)


hack/latest-binaries.sh
```diff
@@ -4,16 +4,24 @@ set -o errexit
...
+if [ "$#" -ne 2 ]; then
+  echo "usage: $0 KUBERNETES_MINOR_VERSION REGION"
   exit 1
 fi
 
 MINOR_VERSION="${1}"
+REGION="${2}"
+
+CROSS_PARTITION_ARGS=""
+if [[ "${REGION}" == us-gov* || "${REGION}" == eusc* ]]; then
```
mselim00, line 16: I'd like this to be more tied to connectivity. if it's not an isolated partition or China, we know we can allow the cross-partition

but I recognize this is a hack script and it's not so easy to make this determination here

hack/latest-binaries.sh
```diff
@@ -4,16 +4,24 @@ set -o errexit
...
-  echo "usage: $0 KUBERNETES_MINOR_VERSION"
+if [ "$#" -ne 2 ]; then
+  echo "usage: $0 KUBERNETES_MINOR_VERSION REGION"
   exit 1
 fi
 
 MINOR_VERSION="${1}"
+REGION="${2}"
+
+CROSS_PARTITION_ARGS=""
```
mselim00: nit: this name seems awkward in the invocation when it's not cross-partition, maybe something more generic like `S3_CLI_EXTRA_ARGS`

templates/al2023/variables-default.json
```diff
@@ -24,7 +24,7 @@
     "nvidia_driver_major_version": "580",
     "nvidia_repository_url": null,
     "nvidia_grid_runfile_bucket_name": "ec2-linux-nvidia-drivers",
-    "pause_container_image": "602401143452.dkr.ecr.us-west-2.amazonaws.com/eks/pause:3.10",
+    "pause_container_image": "public.ecr.aws/eks-distro/kubernetes/pause:3.10",
```
mselim00, line 27: also, we should fix the caching of this image - it's a different auth token and subcommand for ecr-public

mselim00 (approved):

templates/al2023/variables-default.json
fletcherw: Done.

hack/latest-binaries.sh
fletcherw: Done.

hack/latest-binaries.sh
fletcherw: I see your point but I think it's okay like this in the hack script.

fletcherw:
1.33 test failure was just a timeout.

bradwatsonaws:
@fletcherw is the real MVP!

fletcherw:
Thanks @bradwatsonaws ! Please let me know if you have any issues with the change.

PR #2624, opened 2026-02-12
fix: symlink drop-in directory for ec2-net-utils compatibility

### LINKED ISSUE #2623: ec2-net-utils secondary IP aliases not applied: drop-in directory name mismatch with 70-eks-{iface}.network
## Summary

On EKS AL2023 nodes, secondary private IPs assigned to the primary ENI via `AssignPrivateIpAddresses` API are not configured at the OS level. The `amazon-ec2-net-utils` package correctly detects the IPs from IMDS and generates the appropriate systemd-networkd drop-in configuration, but the drop-ins are placed in a directory that does not match the active network file created by `nodeadm`, so they are silently ignored by `systemd-networkd`.

## Environment

- **EKS version**: v1.35
- **AMI**: Amazon Linux 2023 (EKS-optimized, `eks_image_family: al2023`)
- **amazon-ec2-net-utils**: 2.7.1-1.amzn2023.0.1
- **systemd-networkd**: 252.23-11.amzn2023
- **nodeadm**: latest (from EKS AMI)

## Root Cause

`nodeadm`'s `udev-net-manager` creates an EKS-specific network file for the primary interface:

```go
// nodeadm/cmd/nodeadm-internal/udev/systemd.go:34-36
func eksNetworkPath(iface string) string {
    return filepath.Join("/run/systemd/network/", fmt.Sprintf("70-eks-%s.network", iface))
}
```

This produces `/run/systemd/network/70-eks-ens5.network`.

It also disables the default `80-ec2.network` via a drop-in that forces `[Match] Name=none`.

Meanwhile, `amazon-ec2-net-utils` (`setup-policy-routes` → `create_ipv4_aliases` in `lib.sh`) detects secondary IPs from IMDS and creates drop-in files in:

```bash
# amazon-ec2-net-utils/lib.sh - create_ipv4_aliases()
local drop_in_dir="${unitdir}/70-${iface}.network.d"
```

This produces `/run/systemd/network/70-ens5.network.d/ec2net_alias.conf`.

In `systemd-networkd`, drop-in files in `<name>.network.d/` only apply to `<name>.network`. Since the active network file is `70-eks-ens5.network`, the drop-in directory `70-ens5.network.d/` is completely ignored.

## Evidence

**Active network file:**
```
$ networkctl status ens5
Network File: /run/systemd/network/70-eks-ens5.network
```

**Drop-in created by ec2-net-utils (correct content, wrong location):**
```
$ cat /run/systemd/network/70-ens5.network.d/ec2net_alias.conf
[Address]
Address=10.60.195.43/32
AddPrefixRoute=false
[Address]
Address=10.60.200.9/32
AddPrefixRoute=false
[Address]
Address=10.60.203.231/32
AddPrefixRoute=false
```

**IPs on the interface (secondary IPs missing):**
```
$ ip addr show ens5
2: ens5: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 9001
    inet 10.60.197.31/20 scope global dynamic ens5
```

**Manual workaround confirms the diagnosis:**
```
$ ip addr add 10.60.200.9/32 dev ens5
$ curl http://<EIP-associated-to-10.60.200.9>
OK
```

## Impact

Any use case that relies on secondary private IPs being routable on EKS AL2023 nodes is broken:

- **Elastic IP association to secondary private IPs**: EIPs are unreachable from the internet because the kernel drops packets for IPs it doesn't know about
- **Multiple EIPs per node**: a common pattern for ingress/edge nodes that need to expose multiple public IPs on a single instance

Additionally, the `refresh-policy-routes@ens5.timer` is inactive (dead) by default, meaning even if the naming issue were fixed, secondary IPs added after boot would not be picked up until the timer is enabled.

## Suggested Fix

Since `nodeadm` renames the network file from `70-{iface}.network` to `70-eks-{iface}.network` (to avoid a naming conflict with ec2-net-utils), it should also create a symlink for the drop-in directory so that ec2-net-utils drop-ins are found by systemd-networkd:

```go
// After creating 70-eks-{iface}.network, add:
func ensureDropinCompat(iface string) error {
    unitdir := "/run/systemd/network"
    ec2netDir := filepath.Join(unitdir, fmt.Sprintf("70-%s.network.d", iface))
    eksDir := filepath.Join(unitdir, fmt.Sprintf("70-eks-%s.network.d", iface))
    os.MkdirAll(ec2netDir, 0755)
    return os.Symlink(ec2netDir, eksDir)
}
```

This would be called after `util.WriteFileWithDir(eksNetworkPath(c.iface), ...)` in `net_manager.go`.

This way:
- `ec2-net-utils` writes to `70-ens5.network.d/` (unchanged behavior)
- `systemd-networkd` reads `70-eks-ens5.network` and finds drop-ins via `70-eks-ens5.network.d/` → symlink → `70-ens5.network.d/`
- No changes needed in `ec2-net-utils`
- The `eks-` prefix is preserved to avoid the naming conflict

rtrompier:
## Summary

- Fixes ec2-net-utils drop-in files being silently ignored on EKS AL2023 nodes
- `nodeadm` creates `70-eks-{iface}.network` but ec2-net-utils writes drop-ins to `70-{iface}.network.d/` — the name mismatch causes systemd-networkd to never apply them
- Adds `ensureDropinCompat()` which creates a symlink `70-eks-{iface}.network.d → 70-{iface}.network.d` so drop-ins written by ec2-net-utils are discovered by systemd-networkd

## Problem

When `nodeadm` bootstraps an EKS node, it:
1. Creates `/run/systemd/network/70-eks-{iface}.network` via `eksNetworkPath()`
2. Disables the default `80-ec2.network` via `disableDefaultEc2Networking()`

Meanwhile, `amazon-ec2-net-utils` writes secondary IP alias configurations as drop-in files into `/run/systemd/network/70-{iface}.network.d/ec2net_alias.conf`.

Because systemd-networkd only reads drop-ins from `{unit-name}.d/`, drop-ins in `70-{iface}.network.d/` are never applied to `70-eks-{iface}.network`. This means:
- Secondary private IP aliases are never configured on the interface
- Any EIP associated with a secondary private IP becomes a black hole (unreachable)
- `refresh-policy-routes` runs successfully but its output is silently ignored

## Fix

After writing the EKS network file, create a symlink so that ec2-net-utils drop-ins are found:

```
70-eks-{iface}.network.d → 70-{iface}.network.d
```

The `ensureDropinCompat()` function:
1. Creates `70-{iface}.network.d/` (where ec2-net-utils writes)
2. Symlinks `70-eks-{iface}.network.d/` → `70-{iface}.network.d/` (where systemd-networkd reads)

This is idempotent and handles the case where the symlink already exists.

Fixes #2623

### DIFF
--- a/nodeadm/cmd/nodeadm-internal/udev/systemd.go
+++ b/nodeadm/cmd/nodeadm-internal/udev/systemd.go
@@ -50,7 +50,7 @@ func ensureDropinCompat(iface string) error {
 		return fmt.Errorf("failed to create ec2-net-utils drop-in dir: %w", err)
 	}
 	// Remove existing symlink or directory if present to ensure idempotency
-	os.Remove(eksDir)
+	os.RemoveAll(eksDir)
 	return os.Symlink(ec2netDir, eksDir)
 }
 


rtrompier:
Closing this — we ended up going with a DaemonSet that syncs secondary private IPs from IMDS to the host interface via netlink. Simpler and more reliable for our use case than patching the symlink.

The underlying issue (#2623) is still valid though.

PR #2639, opened 2026-02-25
Changes install-worker.sh to download *.gz artifacts.

cmattoon:
**Issue #, if available:** I noticed a broken build with the message `An error occurred (404) when calling the HeadObject operation: Key "1.35.0/2026-02-18/bin/linux/amd64/kubelet" does not exist.`

I checked the output of ./hack/latest-binaries.sh for 1.34 and 1.35 and ran `aws s3 ls` manually against both S3_PATHs and indeed there is no `kubelet`, only `kubelet.gz`.

The same problem occurs with `ecr-credential-provider` and the same fix is applied.

**Description of changes:**
This downloads `$binary.gz` and `$binary.gz.sha256`, checks the sha256sum on the `*.gz`, runs `gunzip`, then proceeds as always with the plain `$binary` file.

**Testing Done**

#WorksInMyPipeline

### DIFF
--- a/templates/al2023/provisioners/install-worker.sh
+++ b/templates/al2023/provisioners/install-worker.sh
@@ -166,7 +166,7 @@ AWS_DOMAIN=$(imds "/latest/meta-data/services/domain")
 S3_PATH="s3://$BINARY_BUCKET_NAME/$KUBERNETES_VERSION/$KUBERNETES_BUILD_DATE/bin/linux/$ARCH"
 
 BINARIES=(
-  kubelet
+    kubelet
 )
 for binary in "${BINARIES[@]}"; do
   FILES=(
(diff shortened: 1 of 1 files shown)


hectoralicea:
I found the same issue.  Looking for a quick solution to keep going

cmattoon:
I had to modify the component YAML (example [here](https://github.com/aws-samples/amazon-eks-custom-builds/blob/main/CloudFormation/imagebuilder-al2023.yaml)) to use my `GithubRepoUrl`, add a `GithubRepoRef` variable (not strictly required), then update the `git clone` step:

```
    - cd {{ HomePath }}
    - git clone {{ GithubRepoUrl }}
    - cd {{ GithubDir }} && git checkout {{ GithubRepoRef }}
    - cd {{ HomePath }}
```

Then call it with `GithubRepoUrl=https://github.com/cmattoon/amazon-eks-ami` and `GithubRepoRef=fix-gz-binaries`

mselim00:
thanks for putting this together this so quickly!

the unzipped versions of these artifacts should be available now. it's simpler to continue using those as long as they are, so I'm going to close this with that in mind, but hopefully anyone interested in this approach could pick it up from here

PR #2640, opened 2026-02-26
chore: suppress G602 for CPU to reserve

mselim00:
**Issue #, if available:**

**Description of changes:**
This array index is flagged because there's no implicit guarantee that `i + 1` exists in `cpuRanges` if `i` is based on the length of `cpuPercentageReservedForRanges`. This is informally guaranteed though by the nature of the function, so this just adds a suppression until this can be refactored in a way that avoids this.


fletcherw:
I think this is possibly a bug in gosec, assuming that a `for range` loop with indices is only indexing into the ranged container, but agree we should suppress.

mselim00:
I think you're probably right, I just saw https://github.com/securego/gosec/issues/1525 and https://github.com/securego/gosec/pull/1531 - but I don't think the latter would fix this this

PR #2650, opened 2026-03-10
fix: reserve EKS addon ports to prevent containerd port collisions

### LINKED ISSUE #2631: Reserve EKS add-on ports in AMI to prevent containerd CRI streaming server collision with managed components
Reserve port 9809 (and other EKS add-on component ports) in the EKS AMI by configuring net.ipv4.ip_local_reserved_ports in the system's sysctl configuration to prevent containerd's CRI streaming server from dynamically binding to ports required by EKS managed add-ons.

Specifically, add the following to /etc/sysctl.d/99-eks-reserved-ports.conf in EKS AMIs:

`net.ipv4.ip_local_reserved_ports = 9809,61678,10249,9153
`
Where:

9809: EFS CSI driver liveness probe port
61678: VPC CNI metrics port
10249: kube-proxy metrics port
9153: CoreDNS metrics port

**Why is this needed**:

Containerd's default configuration (stream_server_port in /etc/containerd/config.toml) allows the Go runtime to dynamically select any available port from the ephemeral range. During node initialization, containerd could randomly bound to port (possibly one of the EKS addon port like 9809) before the EFS CSI driver pod started, creating a race condition.

This could get us into a situation where addon fails to come up since the port is already in use.

`kubelet: SyncLoop (PLEG): event for pod "kube-system/efs-csi-node-c9phf" event={"Type":"ContainerDied"}
2026-02-11T12:41:34.532626906Z E0211 12:41:34.532608 1 main.go:175] "Failed to start http server" err="listen tcp [IP_ADDRESS]: bind: address already in use"
`
Node netstat output:

`tcp LISTEN 0 4096 [IP_ADDRESS] 0[IP_ADDRESS]0.0:* users:(("containerd",pid=7902,fd=12))
`

Shreyank031:
Reserve well-known ports used by EKS add-ons with **`hostNetwork=true`** via **`net.ipv4.ip_local_reserved_ports`** sysctl, preventing **containerd's CRI streaming server** from grabbing them via **dynamic ephemeral assignment**.

### **Issue #, if available:**

**Fixes #2631**

### **Description of changes**

Containerd's **CRI streaming server** uses **`stream_server_port=0`**, causing the kernel to assign **any available ephemeral port**. When that port collides with one required by an **EKS add-on pod running with `hostNetwork: true`**, the pod fails to bind and enters **`CrashLoopBackOff`**.

### **Observed in production:**

- Port **2703** (**eks-pod-identity-agent health check**) was grabbed by **containerd**
- → **eks-pod-identity-agent crashed**
- → **VPC CNI lost credentials → aws-node crashed**
- → **Node went NotReady**

This PR adds **`etc/sysctl.d/99-eks-reserved-ports.conf`** to the **AL2023 AMI rootfs**, configuring **`net.ipv4.ip_local_reserved_ports`** with ports used by **EKS add-on components that run with `hostNetwork: true`**.

### **Ports reserved**

| Port | Component |
|-----|-----------|
| **2703** | EKS Pod Identity Agent (health/liveness probe) |
| **9808** | Amazon EBS CSI Driver node plugin (health check) |
| **9809** | Amazon EFS CSI Driver node plugin (health check) |
| **10249** | kube-proxy (metrics) |
| **10250** | kubelet (API server) |
| **10256** | kube-proxy (health check) |
| **61678** | Amazon VPC CNI / aws-node (Prometheus metrics) |
| **61679** | Amazon VPC CNI / aws-node (introspection endpoint) |

### **Testing Done**

- **`make lint-code` passed** — no trailing whitespace, shellcheck and shfmt not applicable to `.conf` files  
- **`cd nodeadm && make test` passed**  
- **Port list verified against official EKS add-on manifests** — only add-ons confirmed to use **`hostNetwork: true`** are included  
- Follows the **same naming and format convention** as existing **`99-kubernetes-cri.conf`** and **`99-amazon.conf`** files in the same directory


templates/al2023/runtime/rootfs/etc/sysctl.d/67-eks-reserved-ports.conf
mselim00, line 1: can we make the prefix something in the `60`-`70` range?

since every port in here is either adjustable or for a non-required component, it's very conceivable that a user would want to override this file's contents, but the `99` prefix doesn't leave any clear room for that

Shreyank031: Renamed the file to 67-eks-reserved-ports.conf so users have clear room to override with any higher-numbered file. Thanks for the suggestion!

Shreyank031:
Thank you @mselim00  and @HusainZafar for the review and for merging this! Really appreciate the feedback and the quick turnaround.

PR #2653, opened 2026-03-11
chore: unpin containerd minor version

### LINKED ISSUE #2625: Upgrade containerd to 2.2+

Upgrade containerd to 2.2 to allow usage of subPath on OCI volume mount feature available in Kubernetes 1.35

**Why is this needed**:
It will allow us to remove some initContainer who are currently used as a dirty workaround to create a correct emptydir volume. 
It will improve the efficiency for our pod to start.

mselim00:
**Issue #, if available:**

Resolves https://github.com/awslabs/amazon-eks-ami/issues/2625

**Description of changes:**

Updating the variable defaults to start consuming the latest minor version in the 2.x branch as AL2023 has started including containerd2.2, and 2.1 reaches upstream EOS in 2 months: https://containerd.io/releases/.

**Testing Done**

### DIFF
--- a/templates/al2023/variables-default.json
+++ b/templates/al2023/variables-default.json
@@ -10,7 +10,7 @@
     "aws_session_token": "{{env `AWS_SESSION_TOKEN`}}",
     "binary_bucket_name": "amazon-eks",
     "binary_bucket_region": "us-west-2",
-    "containerd_version": "2.1.*",
+    "containerd_version": "2.*",
     "install_containerd_from_s3": "false",
     "creator": "{{env `USER`}}",
     "enable_accelerator": "",


Youssef-Beltagy (approved):

wwvela (approved):

mselim00:
the test failures were for cleanup, which can be traced to an issue with the test infrastructure itself that was recently fixed but not picked up by the runner. https://github.com/awslabs/amazon-eks-ami/pull/2655 should make it so that we use the latest version, so re-running for sanity

/ci

dacevedo12:
This update broke workloads with nixos-based containers. Is there anything that can be learned here to prevent this kind of regression or is it best left to the containerd project itself?

2rs2ts:
First off, I want to preface my message by saying I'm aware I lack a lot of context on the inner workings at AWS, I don't think this change was necessarily "bad" or "wrong" or "dumb," and I think nobody should feel ashamed about it; I really respect and appreciate the work you all do, and I'm just trying to provide some critical constructive feedback in the hopes that problems I and others have experienced are less likely to recur.

With that said, I believe unpinning the version is problematic. We don't want the presence of a package in the AL2023 repos to change the functionality and stability of this AMI, regardless of whatever safeguards come after the point where the package gets added to the AL repos and when a new AMI is released. I am just a user (who customizes the AMI with additional stuff my company needs internally) so I don't know what it's like for the AWS teams involved, but I imagine that there were probably _at least_ these many steps for getting a new package version out:

1. containerd project releases new version
2. Some team at AWS watches the post-release bugs come in/spends time vetting the software themselves in a limited capacity/adds any internal patches
3. This repeats until a version comes out that "seems all right"/meets some internal criteria for acceptability/*has* to get rushed out because of a security vulnerability (whether the patch is written by the containerd maintainers or by AWS developers)
4. Teams responsible for specific products (AL itself, EKS, ECS, probably some internal tools teams, etc.) get access to the package through an internal package repo that can be used for test builds
5. The team responsible for AL finishes its testing, decides to include the package in an OS release
6. Users now have a new AL2023 image to install; if there are issues, they may provide reports that might bubble up the support chain and cause the step above to be reverted before a new EKS image gets released
7. The team responsible for EKS finishes its testing, makes this repo use the new containerd package, and updates all relevant configuration to ensure it works correctly
8. A new EKS AMI is released within a few weeks

However, unpinning the minor version introduces or exacerbates the following problems (compared to the previous state of unpinning the patch version, or especially compared to my _ideal_ which is exact pinning):

- If steps 1-5 above happen twice before steps 6-8 finish once, the current build configuration will update the containerd version more than what might have been tested against. We all know that automated regression tests performed before cutting a release are never the be-all and end-all for checks we do as developers; I'm certain the EKS team has their own ways of detecting regressions that aren't codified and automated due to technical or development time constraints.
- It makes it more likely for any oversights or mistakes that happen in steps 1-5 to cause delays with step 7 or cause bad releases in step 8. We're human, so we all make mistakes, which is why "trust but verify" is so important and unpinning minor releases for software that's known not to strictly and correctly follow semver is dangerous.
- It reduces the timeframe that step 6 provides as a buffer. I know it's not ideal to crowdsource regression detection for cases that got overlooked like the nixos one mentioned in the post above mine, but I would be surprised if there was never a case in the history of the EKS AMI that a user of AL reported an issue in a package and that caused either the package to be pulled or a hotfixed version to be released in time for a new EKS AMI to not include the bugged software.
- It exacerbates the existing issue where users struggle to match the containerd version on their EKS servers with the containerd version on their development machines; in a real enterprise, it's never as simple as "just use AL for all your machines" or "just configure everything to be on `2.*`" because there are also multiple teams in each user enterprise who need to check for regressions in their own systems/products when an update lands. (I can speak from experience that the 1.7.29 -> 2.1.5 update caused a lot of problems and internal blockers.)
- It makes bisecting more inconvenient.

I'm saying this while having to make some assumptions about how things work over at AWS. I could be totally wrong about SOP and be over-estimating the potential for harm caused by this unpinning (I hesitate to say I'm under-estimating it, since the nixos regression happened and that's about as big of a regression that I can imagine realistically slipping through the cracks.) I *really* hope that despite having a lot to say to plead my case, it doesn't come across as harsh, and that you're willing to accept and think on it. Even if you never go back to pinning the minor version, so long as you do so while having heard me out and taken my concerns seriously, that's enough for me.

dacevedo12:
@2rs2ts AWS won't be as polite when charging you for compute, so I believe it's reasonable to expect better. We experienced several hours of unavailability because of this. All our fleet on EKS runs on nixos/nix container images so for us it was a widespread problem.

We, as paying AWS customers expect better handling of these updates to prevent regressions. I mean, containerd is the basis of it all when it comes to those workloads. Something that critical shouldn't be so liberally bumped in version without proper testing.

PR #2696, opened 2026-04-28
Add support for NVIDIA GDS, GDRCopy and FSxL

### LINKED ISSUE #2181: Support for NVIDIA GPUDirect Storage with EFA

Customers serving very large language models (>500GB) need a fast way to load the model into GPU memory as fast as possible. With the [launch](https://aws.amazon.com/about-aws/whats-new/2024/11/amazon-fsx-lustre-elastic-fabric-adapter-nvidia-gpudirect-storage/) of EFA support for FSx for Lustre, a throughput of 1200Gbps can be supported. If I am not mistaken this requires the inclusion of the the [GDS](https://docs.aws.amazon.com/fsx/latest/LustreGuide/configure-efa-clients.html#install-gds-on-client) driver in the AMI.

**Why is this needed**:

Loading large language models into GPU memory is a very time consuming process. Downloading it every time is not feasible. S3, EFS, EBS based options all have considerable performance (e.g. extra copies through a bounce buffer in the CPU’s memory). GPUDirect Storage enables a direct data path between storage and GPU memory

erezzarum:
**Issue #, if available:**

**Description of changes:**

Adds support for NVIDIA GDS (nvidia_fs), NVIDIA GDRCopy (gdrdrv) and install lustre-client by default.
This simplify setup of FSxLustre with GDS.

**Testing Done**

Build and deployed.


shvbsle:
/ci

erezzarum:
The defaults are currently not to build nvidiafs and gdrcopy, we need to enable it for testing and later for release.
Thanks!

erezzarum:
Updates:
1. The configure FSxLustre EFA script is now included in the AMI, this is relevant for non accelerated AMIs as well.
2. Created an experimental `featureGate` to trigger the script by nodeadm
3. Fixed an issue with the S3 calls during build when the instance doesn't have a role to allow calls to S3, fallback to use S3 without signed calls.
4. NVIDIA FS and GDRCopy are now by default enabled.

templates/al2023/provisioners/install-worker.sh
```diff
@@ -236,6 +236,15 @@ else
   fi
 fi
 
+################################################################################
+### Lustre #####################################################################
+################################################################################
+
+echo "Installing lustre-client from AL core repository"
```
shvbsle, line 243: I do get the intent behind including the client in the AMI but I don't think we should ship the `lustre-client` in the AMI at all. Two main reasons:
1. The nvidia-fs and gdrdrv kernel modules do not depend on lustre-client. `nvidia-fs` is designed to work with any filesystem and storage that supports it so it's not lustre specific.
2. The client installs both user-space and kernel-space modules which moves the CVE ownership to us. Every Lustre CVE becomes an EKS AMI CVE

I think user-data is a more appropriate place to bring this in. Something like so:
```yaml
MIME-Version: 1.0
Content-Type: multipart/mixed; boundary="BOUNDARY"

--BOUNDARY
Content-Type: text/x-shellscript; charset="us-ascii"

#!/bin/bash
sudo dnf install -y lustre-client

```

(if we were to bake this into the ami, a few more tension points would come up : a) this would install the client on all ami variants b) generally installs need to be partition aware to ensure builds work in china and air-gapped regions etc)

templates/al2023/provisioners/install-nvidia-driver.sh
```diff
@@ -220,7 +241,58 @@ function archive-proprietary-kmod() {
...
+}
+
+function install-nvidiafs-kmod() {
+  local NVIDIA_FS_DRIVER_INSTALLATION_TEMP_DIR
+
+  NVIDIA_FS_DRIVER_INSTALLATION_TEMP_DIR=$(mktemp -d)
+
+  echo "Installing NVIDIA FS driver"
+
+  curl -L https://github.com/NVIDIA/gds-nvidia-fs/archive/refs/tags/v${NVIDIA_FS_DRIVER_VERSION}.tar.gz | sudo tar -xz -C $NVIDIA_FS_DRIVER_INSTALLATION_TEMP_DIR
```
shvbsle, line 254: This curl calls to public internet endpoints with no partition handling. ISO regions are air-gapped so no access to docs.aws.amazon.com or github.com. These curls will fail and abort the build (set -o errexit). In China regions, github.com access is unreliable.

You should be able to find this in in Nvidia Cuda repo for al2023 (which is already configured earlier in the script):
https://developer.download.nvidia.com/compute/cuda/repos/amzn2023/x86_64/

I do see `nvidia-fs-dkms-2.28.2-1.x86_64.rpm ` available in the repo. This would make it partition aware.

templates/al2023/provisioners/install-nvidia-driver.sh
```diff
@@ -220,7 +241,58 @@ function archive-proprietary-kmod() {
...
+  sudo cp -r ${NVIDIA_FS_DRIVER_INSTALLATION_TEMP_DIR}/gds-nvidia-fs-${NVIDIA_FS_DRIVER_VERSION}/src /usr/src/nvidia-fs-${NVIDIA_FS_DRIVER_VERSION}
+  sudo sed -i 's/BUILD_DEPENDS\[0\]="nvidia"/BUILD_DEPENDS\[0\]="nvidia-open"/' /usr/src/nvidia-fs-${NVIDIA_FS_DRIVER_VERSION}/dkms.conf
+  sudo sed -i 's/\(NVFS_MAX_PEER_DEVS=\${NVFS_MAX_PEER_DEVS:-\)[0-9]\{1,5\}\(}\)/\1128\2/g' /usr/src/nvidia-fs-${NVIDIA_FS_DRIVER_VERSION}/dkms.conf
+  sudo sed -i 's/\(NVFS_MAX_PCI_DEPTH=\${NVFS_MAX_PCI_DEPTH:-\)[0-9]\{1,5\}\(}\)/\116\2/g' /usr/src/nvidia-fs-${NVIDIA_FS_DRIVER_VERSION}/dkms.conf
+  sudo dkms add -m nvidia-fs -v $NVIDIA_FS_DRIVER_VERSION
+  sudo dkms build -m nvidia-fs -v $NVIDIA_FS_DRIVER_VERSION
+  sudo dkms install -m nvidia-fs -v $NVIDIA_FS_DRIVER_VERSION
```
shvbsle, line 261: I would like to push back against loading any drivers in the dkms tree here. You will notice that all other drivers are archived in a special location and loaded at AMI boot time. Generally the archived modules are stored in `/var/lib/dkms-archive`. I built this PR and I don't see this module in the archive:

```
# ls /var/lib/dkms-archive/
nvidia  nvidia-open  nvidia-open-grid
```

The general idea is, you let kernel modules stay dormant in the archive and load only those modules on GPUs that actually use them. The problem with this approach right now is that these modules won't be relevant for nvidia instance types that don't support it. For reference it's worth looking at how the grid driver installation is handled which only loads the drivers for g6f type instances


You can use `kmod-util archive` to archive these modules

templates/al2023/provisioners/install-nvidia-driver.sh
```diff
@@ -220,7 +241,58 @@ function archive-proprietary-kmod() {
...
+  local MODULE_SUBDIR
+
+  EXTRACT_DIR=$(mktemp -d)
+  GDRCOPY_DRIVER_INSTALLATION_TEMP_DIR="${EXTRACT_DIR}/gdrcopy-${NVIDIA_GDRCOPY_DRIVER_VERSION}"
+  GDRCOPY_SRC_DIR="/usr/src/gdrdrv-${NVIDIA_GDRCOPY_DRIVER_VERSION}"
+  MODULE_SUBDIR="/kernel/drivers/misc/"
+
+  echo "Installing GDRCopy driver"
+
+  curl -L https://github.com/NVIDIA/gdrcopy/archive/refs/tags/v${NVIDIA_GDRCOPY_DRIVER_VERSION}.tar.gz | tar -xz -C $EXTRACT_DIR
```
shvbsle, line 279: This suffers from the same lack of partition awareness issue. See:
https://github.com/awslabs/amazon-eks-ami/pull/2696/changes#r3174611531

templates/al2023/provisioners/install-nvidia-driver.sh
```diff
@@ -220,7 +241,58 @@ function archive-proprietary-kmod() {
...
+  sudo cp ${GDRCOPY_DRIVER_INSTALLATION_TEMP_DIR}/packages/dkms.conf "${GDRCOPY_SRC_DIR}/dkms.conf"
+  sudo cp -r ${GDRCOPY_DRIVER_INSTALLATION_TEMP_DIR}/scripts $GDRCOPY_SRC_DIR/
+
+  sudo sed -i "s/@FULL_VERSION@/${NVIDIA_GDRCOPY_DRIVER_VERSION}/g" "${GDRCOPY_SRC_DIR}/dkms.conf"
+  sudo sed -i "s/@VERSION@/${NVIDIA_GDRCOPY_DRIVER_VERSION}/g" "${GDRCOPY_SRC_DIR}/dkms.conf"
+  sudo sed -i "s|@MODULE_LOCATION@|${MODULE_SUBDIR}|g" "${GDRCOPY_SRC_DIR}/dkms.conf"
+
+  sudo dkms add -m gdrdrv -v $NVIDIA_GDRCOPY_DRIVER_VERSION
+  sudo dkms build -m gdrdrv -v $NVIDIA_GDRCOPY_DRIVER_VERSION
+  sudo dkms install -m gdrdrv -v $NVIDIA_GDRCOPY_DRIVER_VERSION
```
shvbsle, line 295: Same pattern as flagged earlier. See: https://github.com/awslabs/amazon-eks-ami/pull/2696/changes#r3174655495

templates/al2023/runtime/gpu/nvidia-kmod-load.sh
```diff
@@ -81,6 +81,45 @@ function disable-gsp() {
...
 }
 
+function load-nvidiafs-module() {
+  echo "Checking for nvidia_fs kernel module"
+  if ! modinfo nvidia_fs &> /dev/null; then
+    echo "nvidia_fs kernel module not found, skipping"
+    return 0
+  fi
+  echo "nvidia_fs kernel module found, attempting to load"
+  if modprobe nvidia_fs; then
```
shvbsle, line 91: See: https://github.com/awslabs/amazon-eks-ami/pull/2696/changes#r3174655495

Consider loading them from archive using `kmod-util load <module_name>`. Also needs an additional check to ensure we load this only on (p4, p5 or supported instances)

templates/al2023/runtime/gpu/nvidia-kmod-load.sh
```diff
@@ -81,6 +81,45 @@ function disable-gsp() {
...
+}
+
+function load-gdrdrv-module() {
+  echo "Checking for gdrdrv kernel module"
+  if ! modinfo gdrdrv &> /dev/null; then
+    echo "gdrdrv kernel module not found, skipping"
+    return 0
+  fi
+  echo "gdrdrv kernel module found, attempting to load"
+  if modprobe gdrdrv; then
```
shvbsle, line 106: See: https://github.com/awslabs/amazon-eks-ami/pull/2696/changes#r3174655495

Consider loading them from archive using `kmod-util load <module_name>`. Also needs an additional check to ensure we load this only on (p4, p5 or supported instances)

templates/al2023/provisioners/install-nvidia-driver.sh
```diff
@@ -220,7 +241,58 @@ function archive-proprietary-kmod() {
...
   sudo kmod-util archive nvidia
   sudo kmod-util remove nvidia
-  sudo rm -rf /usr/src/nvidia*
+  sudo find /usr/src -maxdepth 1 -name 'nvidia*' ! -name 'nvidia-fs*' -exec rm -rf {} +
+}
+
+function install-nvidiafs-kmod() {
+  local NVIDIA_FS_DRIVER_INSTALLATION_TEMP_DIR
+
+  NVIDIA_FS_DRIVER_INSTALLATION_TEMP_DIR=$(mktemp -d)
```
shvbsle, line 250: This would create a dir in `/tmp.<something>` but not get cleaned up. (the packer cleanup steps only clears /tmp/worker dir)

templates/al2023/provisioners/install-nvidia-driver.sh
```diff
@@ -220,7 +241,58 @@ function archive-proprietary-kmod() {
...
+  echo "options nvidia_fs peer_stats_enabled=1 rw_stats_enabled=1" | sudo tee /etc/modprobe.d/nvidia-fs.conf
+}
+
+function install-gdrcopy-kmod() {
+  local EXTRACT_DIR
+  local GDRCOPY_DRIVER_INSTALLATION_TEMP_DIR
+  local GDRCOPY_SRC_DIR
+  local MODULE_SUBDIR
+
+  EXTRACT_DIR=$(mktemp -d)
```
shvbsle, line 272: This would create a dir in `/tmp.<something>` but not get cleaned up. (the packer cleanup steps only clears /tmp/worker dir)

doc/usage/al2023.md
```diff
@@ -35,6 +35,10 @@
 | `nvidia_driver_major_version` | To be used only when ```enable_accelerator = nvidia```. Driver version to install, depends on what is available in NVIDIA repository. |
 | `nvidia_repository_url` | YUM/DNF Repository override for the NVIDIA driver packages |
 | `nvidia_grid_runfile_bucket_name` | Bucket name for sourcing the Nvidia GRID runfile |
+| `enable_nvidia_fs_driver` |  |
+| `nvidia_fs_driver_version` |  |
```
shvbsle, line 39: Could we add descriptions here?

nodeadm/api/v1alpha1/nodeconfig_types.go
```diff
@@ -168,4 +168,14 @@ const (
 	// may result in faster image pull times. This flag will be ignored on
 	// instances with memory and vCPU below a certain threshold.
 	FastImagePull Feature = "FastImagePull"
+
+	// FSxLustreEFAClient enables EFA-optimized FSx for Lustre client configuration.
+	// When enabled, nodeadm will run the configure-efa-fsx-lustre-client script
+	// on each boot to configure EFA interfaces for the Lustre client.
+	FSxLustreEFAClient Feature = "FSxLustreEFAClient"
```
shvbsle, line 175: I think that the feature gates and nodeadm aspect are unnecessary. The AMI should provide kernel drivers (nvidia-fs, gdrdrv), and users should handle configuration (lustre-client install, EFA setup) via user data.

templates/al2023/variables-default.json
```diff
@@ -24,6 +24,10 @@
     "nvidia_driver_major_version": "580",
     "nvidia_repository_url": null,
     "nvidia_grid_runfile_bucket_name": "ec2-linux-nvidia-drivers",
+    "enable_nvidia_fs_driver": "true",
```
shvbsle, line 27: Consider disable by default in variables-default.json, enable only in variables-1.35.json, expand gradually to older k8s versions. I'd want to ensure that newer versions of k8s are stable (with fewer users) before rolling out the driver to all other k8s versions

shvbsle (commented):
Thanks for the PR. I've added some comments.

templates/al2023/provisioners/install-worker.sh
erezzarum: The lustre-client doesn't include any kernel-space modules, the kernel modules are already included in AL2023 in-tree kernel, it only lay down configurations and mount helpers.
The lustre-client comes from upstram AL2023 repo, so the case is for any package.

The idea behind this PR is to reduce the burden of customers to configure FSxLustre with EFA as well as several improvements.
Embedding it all in a userData increase startup time and is less customer friendly

templates/al2023/provisioners/install-nvidia-driver.sh
erezzarum: I will have to check if it will be simple to customize it, as we need to adjust `NVFS_MAX_PEER_DEVS` and `NVFS_MAX_PCI_DEPTH` to have decent performance with FSxLustre and GDS.

templates/al2023/provisioners/install-nvidia-driver.sh
erezzarum: Initially (not in the PR but in my testing) i went for this approach, this approach introduce latency in startup time, about 13s avg, this will impact all instances using the nvidia-open driver.

If i understand, the main logic behind using this approach (dkms archive) from the beginning was to include all drivers variant (open, proprietary and grid) in one AMI instead of creating 3 AMI variants.

templates/al2023/provisioners/install-nvidia-driver.sh
erezzarum: GDRCopy (gdrdrv) is not available as dkms module from al2023 repo.

templates/al2023/provisioners/install-nvidia-driver.sh
erezzarum: It's being cleaned up as it is a `/tmp` directory, but can change it.
We do the same for the grid drivers.
```
function archive-grid-kmod() {
  local NVIDIA_GRID_RUNFILE_KEY
  local GRID_INSTALLATION_TEMP_DIR
  local EXTRACT_DIR

  GRID_INSTALLATION_TEMP_DIR=$(mktemp -d)
  EXTRACT_DIR="${GRID_INSTALLATION_TEMP_DIR}/NVIDIA-GRID-extract"
```

erezzarum:
Closing this PR for now.
Will open a new PR for GDRCopy only.

PR #2707, opened 2026-05-07
feat: use igzip for faster container image layer decompression

### LINKED ISSUE #2410: feat(al2023): optimize GZIP
**What would you like to be added**: Would it be possible to implement similar decompression in the EKS AMI as was done in Bottlerocket?

https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/443

ref: https://aws.amazon.com/blogs/containers/introducing-seekable-oci-parallel-pull-mode-for-amazon-eks/

**Why is this needed**:

Appears to significantly improve image pull performance.

rcipdev:
**Issue #, if available:**

**Description of changes:**
Install isa-l-tools from SPAL and configure the SOCI to use igzip instead of unpigz for gzip decompression during container image layer unpacking.
It benefits two image pull paths:
Standard containerd pull — containerd auto-detects igzip in PATH and uses it for layer decompression. No config change needed.
SOCI/FastImagePull — SOCI uses the binary explicitly configured in its config file. Updated from /usr/bin/unpigz to /usr/bin/igzip
https://github.com/amazonlinux/amazon-linux-2023/issues/1091

**Testing Done**
Built AMIs for both x86_64 and arm64, launched clusters, and verified containerd uses igzip during image pulls by monitoring processes during ctr image pull.

### DIFF
--- a/templates/al2023/provisioners/install-worker.sh
+++ b/templates/al2023/provisioners/install-worker.sh
@@ -67,6 +67,10 @@ sudo dnf install -y \
   pigz \
   python3-dnf-plugin-versionlock
 
+# Install isa-l-tools from SPAL for faster gzip decompression (used by containerd and SOCI)
+sudo dnf install -y spal-release
+sudo dnf install -y isa-l-tools
+
 # we need to handle different kernel packages depending on the namespace
 # associated with the minor version.
 KERNEL_PACKAGE="kernel"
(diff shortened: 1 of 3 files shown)


templates/al2023/provisioners/install-worker.sh
```diff
@@ -67,6 +67,10 @@ sudo dnf install -y \
   pigz \
   python3-dnf-plugin-versionlock
 
+# Install isa-l-tools from SPAL for faster gzip decompression (used by containerd and SOCI)
+sudo dnf install -y spal-release
+sudo dnf install -y isa-l-tools
+
```
rcipdev, line 73: isa-l-tools is in the SPAL repo which isn't enabled by default on the base AMI. spal-release adds the repo config. They can't be combined into a single dnf install because dnf resolves all packages before installing any it would fail to find isa-l-tools since the SPAL repo doesn't exist yet at resolution time.

shvbsle:
merged this so ci should work https://github.com/awslabs/amazon-eks-ami/pull/2709

sidewinder12s:
There a reason this was closed or just age?

PR #2733, opened 2026-06-07
Improve start time for NVIDIA AMI

erezzarum:
**Issue #, if available:**

**Description of changes:**

* DKMS by default trigger `dracut` that takes long time, `dracut` is not needed for NVIDIA modules, resolved this by temporary removing `dracut` trigger when loading the driver through the `kmod-util` and then adding it back.
* Remove the motd available update, it takes time and not relevant for EKS AMI.
* Make `nvidia-kmod-load` service not rely on network, fetch instance type from DMI sysfs instead of IMDS. this makes the driver to be fully loaded into kernel before cloud-init starts to avoid race conditions with user provided `userData` that rely on NVIDIA kernel module to be fully loaded.
* Fix S3 calls in GRID driver by adding a fallback to unauthenticated anonymous S3 call.
* Fixed a bug that caused the `nvidia-kmod-load.service` failure if module is already loaded (e.g: on reboot).

**Testing Done**

Testing done on:
* p5.48xlarge
* p5e.48xlarge
* g7e.12xlarge
* g6e.2xlarge

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -78,6 +78,8 @@ jobs:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # 6.0.2
         with:
           ref: 'main'
+      # https://github.com/actions/checkout/issues/2437
+      - run: git config --global maintenance.auto false
       - uses: aws-actions/configure-aws-credentials@d979d5b3a71173a29b74b5b88418bfda9437d885 # 6.1.1
         with:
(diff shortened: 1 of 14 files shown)


shvbsle:
/ci

kummatty:
We do have an effort lined up with NVIDIA AMIs which provides most benefits called out in this PR as a bonus. 
I am curious though to see if you have benchmarks on how much boot time improvements each of these improvements brought in?

PR #2736, opened 2026-06-08
Add support for GDRCopy gdrdrv kernel driver

erezzarum:
**Issue #, if available:**

**Description of changes:**
Add GDRCopy gdrdrv kernel driver.

**Testing Done**

* p5.48xlarge
* p5e.48xlarge


templates/al2023/provisioners/install-nvidia-driver.sh
```diff
@@ -144,6 +144,10 @@ function archive-open-kmods() {
   SUPPORTED_DEVICE_FILE="${WORKING_DIR}/gpu/nvidia-open-supported-devices-${KMOD_MAJOR_VERSION}.txt"
   sudo mv "${SUPPORTED_DEVICE_FILE}" /etc/eks/
 
+  if [[ "$ENABLE_NVIDIA_GDRCOPY_DRIVER" == "true" ]]; then
```
wwvela: Should check both `ENABLE_NVIDIA_GDRCOPY_DRIVER` and `NVIDIA_GDRCOPY_DRIVER_VERSION` to make sure the version is not null or empty string before running `archive-gdrdrv-kmod`

templates/al2023/runtime/gpu/nvidia-kmod-load.sh
```diff
@@ -102,3 +125,8 @@ esac
 echo "options nvidia NVreg_CoherentGPUMemoryMode=driver" > /etc/modprobe.d/40-eks-nvidia-openrm.conf
 
 kmod-util load "${MODULE_NAME}"
+
+if [ "${MODULE_NAME}" = "nvidia-open" ]; then
+  kmod-util load gdrdrv
```
wwvela: here load the gdrdrv when model name is nvidia-open, if `ENABLE_NVIDIA_GDRCOPY_DRIVER` is not set to true and  gdrdrv kmod is not archived in `install-nvidia-driver.sh`, the glob won't match, dkms ldtarball will fail, does it have risk that the entire nvidia-kmod-load.sh script will exit due to errexit? More reasonable only install this when the gdrdrv archive exist

templates/al2023/variables-1.34.json
```diff
@@ -1,3 +1,5 @@
 {
+    "enable_nvidia_gdrcopy_driver": "true",
+    "nvidia_gdrcopy_driver_version": "2.5.2",
```
wwvela: Any reason we enable it by defualt in 1.34, 1.35? If need to enable it by default in some version+, it's should be added in `template.json`, and disable it explicitly in `variables-<version>.json`. the version variable files are used to set those non-defualt field injections for specific version.

wwvela:
/ci
+workflow:k8s_versions 1.35

wwvela:
/ci
+workflow:k8s_versions 1.32

wwvela:
> @wwvela the [workflow](https://github.com/awslabs/amazon-eks-ami/actions/runs/28480839625) that you [requested](https://github.com/awslabs/amazon-eks-ami/pull/2736#issuecomment-4848525467) has completed. 🎉
> AMI variant	Build	Test
> 1.32 / al2023	failure ❌	skipped ⏭️

```
Error: 

2 errors occurred:
	* required variable not set: nvidia_gdrcopy_driver_version
	* required variable not set: enable_nvidia_gdrcopy_driver
```

Seems the 1.32 AMi build failed with this issue when using default template

templates/al2023/provisioners/install-nvidia-driver.sh
```diff
@@ -223,6 +227,40 @@ function archive-proprietary-kmod() {
...
+  local MODULE_SUBDIR
+
+  EXTRACT_DIR=$(mktemp -d)
+  GDRCOPY_DRIVER_INSTALLATION_TEMP_DIR="${EXTRACT_DIR}/gdrcopy-${NVIDIA_GDRCOPY_DRIVER_VERSION}"
+  GDRDRV_SRC_DIR="/usr/src/gdrdrv-${NVIDIA_GDRCOPY_DRIVER_VERSION}"
+  MODULE_SUBDIR="/kernel/drivers/misc/"
+
+  echo "Archiving gdrdrv kmod"
+
+  curl -L https://github.com/NVIDIA/gdrcopy/archive/refs/tags/v${NVIDIA_GDRCOPY_DRIVER_VERSION}.tar.gz | tar -xz -C $EXTRACT_DIR
```
shvbsle: Since this wont work in the air-gapped regions. We will maintain the `gdrdrv` in an s3 bucket and then we'd need to enhance this logic for partition awareness.

templates/al2023/variables-1.34.json
erezzarum: In a previous PR i was told we should first enable new features in newer versions, then later enable them by default on all versions.

erezzarum:
Resolved open comments.
This should be merged along with #2733 or we risk adding extra 13s latency to startup time, because `dracut` is running.

erezzarum:
AL2023 added support for GDRCopy packages.
Updated.
Tested on: g7e.12xlarge, p5.48xlarge

templates/al2023/provisioners/install-nvidia-driver.sh
erezzarum: AL2023 now ships GDRCopy

shvbsle:
/ci
+workflow:k8s_versions 1.32

mselim00:
/ci
+workflow:k8s_versions 1.34

templates/al2023/variables-default.json
```diff
@@ -24,6 +24,8 @@
     "nvidia_driver_major_version": "580",
     "nvidia_repository_url": null,
     "nvidia_grid_runfile_bucket_name": "ec2-linux-nvidia-drivers",
+    "enable_nvidia_gdrcopy_driver": "false",
+    "nvidia_gdrcopy_driver_version": "false",
```
wwvela: We set this default `nvidia_gdrcopy_driver_version` to null instead of false 
```suggestion
    "nvidia_gdrcopy_driver_version": null,
```

templates/al2023/variables-1.34.json
wwvela: Current setting also works but there are 2 potential issue:
* the future version like 1.37 will use default value which disable the nvidia gdrcopy driver
* when backport old versions, you will need to adding more and more override instead of removing the override

You can disable it in older version with override, like set below in lower versions like in `variables-1.33.json` to `variables-1.30.json`
```
    "enable_nvidia_gdrcopy_driver": "false",
    "nvidia_gdrcopy_driver_version": null,
```
And set variables-default.json with enabled setting which means the 1.34+ without explicit override will using default values here.
```
    "enable_nvidia_gdrcopy_driver": "true",
    "nvidia_gdrcopy_driver_version":"2.5.2",
```
So when backporting, we can remove the override setting in specific version config file 1 by 1 instead of adding more override.

wwvela:
/ci
+workflow:k8s_versions 1.34

wwvela:
/ci
+workflow:k8s_versions 1.32

wwvela (approved):
Tests pass. changes LGTM

PR #2742, opened 2026-06-17
fix: maxPods for ipv6 clusters

sumukha-radhakrishna:
**Issue #, if available:** NA

**Description of changes:**

For IPv6 clusters, the IP count per ENI is not a constraint on pod density since each pod gets a unique IPv6 address from the vast address space. The previous ENI-based formula (ENIs * (IPs_per_ENI - 1) + 2) unnecessarily limited maxPods on IPv6 clusters.

This Change fixes by setting defaultMaxPods for ipv6 clusters.

EKS/IPv6 is only supported in prefix mode, If you run ipv6 on EKS then the VPC CNI assigns a /80 IPv6 prefix which give 2⁴⁸ addresses.

https://docs.aws.amazon.com/eks/latest/best-practices/ipv6.html#_overview
https://docs.aws.amazon.com/eks/latest/best-practices/prefix-mode-linux.html

**Testing Done**

Unit tests.

### DIFF
--- a/nodeadm/internal/kubelet/eni_max_pods.go
+++ b/nodeadm/internal/kubelet/eni_max_pods.go
@@ -57,16 +57,18 @@ func GetInstanceInfo(ctx context.Context, awsRegion string, instanceType string)
 //
 //	# of ENI on default network card * (# of IPv4 per ENI - 1) + 2
 //
-// For IPv6 clusters, the IP count per ENI is not a constraint, so the default
-// max pods value (110) is used instead of the ENI-based calculation.
+// For IPv6 clusters, the IP count per ENI is not a constraint, so the result
+// is max(110, ENI-formula) — small instances get at least 110 while large
+// instances keep their higher ENI-based value.
 //
 // TODO: isolate this into a public-facing package for external use by other projects
 func CalcMaxPods(instanceInfo util.InstanceInfo, customExpression string, isIPv6 bool) int32 {
 	var standardMaxPods int32
+	eniMaxPods := calculateStandardMaxPods(instanceInfo)
 	if isIPv6 {
-		standardMaxPods = defaultMaxPods
+		standardMaxPods = max(defaultMaxPods, eniMaxPods)
 	} else {
-		standardMaxPods = calculateStandardMaxPods(instanceInfo)
+		standardMaxPods = eniMaxPods
 	}
 	if len(customExpression) == 0 {
 		return standardMaxPods
--- a/nodeadm/internal/kubelet/eni_max_pods.go
+++ b/nodeadm/internal/kubelet/eni_max_pods.go
@@ -57,9 +57,17 @@ func GetInstanceInfo(ctx context.Context, awsRegion string, instanceType string)
 //
 //	# of ENI on default network card * (# of IPv4 per ENI - 1) + 2
 //
+// For IPv6 clusters, the IP count per ENI is not a constraint, so the default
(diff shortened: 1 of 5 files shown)


nodeadm/internal/kubelet/eni_max_pods.go
```diff
@@ -57,9 +57,17 @@ func GetInstanceInfo(ctx context.Context, awsRegion string, instanceType string)
...
+// For IPv6 clusters, the IP count per ENI is not a constraint, so the default
+// max pods value (110) is used instead of the ENI-based calculation.
+//
 // TODO: isolate this into a public-facing package for external use by other projects
-func CalcMaxPods(instanceInfo util.InstanceInfo, customExpression string) int32 {
-	standardMaxPods := calculateStandardMaxPods(instanceInfo)
+func CalcMaxPods(instanceInfo util.InstanceInfo, customExpression string, isIPv6 bool) int32 {
+	var standardMaxPods int32
+	if isIPv6 {
+		standardMaxPods = defaultMaxPods
```
haouc: Should we scale the defaultMaxPods based on instance type? I wonder how we put right number for smaller instances regarding other resources, cpu/mem?

sumukha-radhakrishna: maxPods is a hard ceiling. if we cap it based on cpu/mem, we will be making an opinionated assumption about pod resource usage that blocks the highdensity lightweight pod use case entirely. imo we should let scheduler handle the cpu/mem constraints

haouc (dismissed):
lgtm

nodeadm/internal/kubelet/eni_max_pods_test.go
```diff
@@ -57,14 +58,29 @@ func TestCalcMaxPods(t *testing.T) {
...
+		{
+			// IPv6 should return default max pods (110) regardless of ENI/IP count
+			defaultENIs:   3,
+			ipsPerENI:     6,
+			isIPv6:        true,
+			expectedValue: 110,
+		},
+		{
+			// IPv6 with custom expression should use 110 as the standard base
+			customExpression: "max_pods",
```
shvbsle, line 77: Suggestion: test-coverage for checking if `customExpression` evaluates using the still-populated ENI vars when ipv6 is enabled

```
{
    customExpression: "(default_enis * (ips_per_eni - 1)) + 2",
    defaultENIs:      3,
    ipsPerENI:        6,
    isIPv6:           true,
    expectedValue:    17, 
},
```

nodeadm/internal/kubelet/eni_max_pods.go
shvbsle: I get that this helps small instances but consider `m5.24xlarge` which has max-pods of `737` in IPv4 mode but in IPv6 this will have `110`. Isn't this going to have the opposite effect to what you intended with this PR i.e. larger instances have much lower density?

This would not be backward compatible for users who are expecting `737` max pods on ipv6 clusters and using a large instance. 

Maybe a better heuristic could be:
```
max(calculateStandardMaxPods(instanceInfo), defaultMaxPods)
```

sumukha-radhakrishna: Wait I thought the maxPods maxValue was 110, TIL (shouldve read the calc) it could go over 110. Yeah we need to calculate for ipv4 and get the max value. Good point, lm make the change

PR #2746, opened 2026-06-23
fix: persist kernel args to GRUB_CMDLINE_LINUX so they survive boot config regerneration

### LINKED ISSUE #2741: bug: nvidia images has incorrect kernel boot options config
Hi, we recently start ugrade from al2023@v20251007 to AMI al2023@v20260505 for our GPU nodes
And found that new images lost important boot config options:

## Config we use:
AMI: ami-04f68437f7b40d7f0 (amazon/amazon-eks-node-al2023-x86_64-nvidia-1.35-v20260527)
instance-id: i-070b2d2ef1b2990ef

## numa_balanging is not disabled

kernel.numa_balancing=1, but is should be disabled for GPU nodes, otherwise this result in up to 15% performance cost on D2H, H2D workload because of pagefault pentaly. 

##  cstate  max is not properly restricted
ctate limit boot options was lost, so it use default value of 8, which is fatal for NCCL latency, it spikes ~100% for small packets. see:
```
cat /proc/cmdline                                                            
BOOT_IMAGE=(hd0,gpt1)/boot/vmlinuz-6.12.88-119.157.amzn2023.x86_64 root=UUID=c2bed23b-0468-4c02-af90-9d133d635d2e ro rd.driver.blacklist=nouveau console=tty0 console=tt
yS0,115200n8 nvme_core.io_timeout=4294967295 rd.emergency=poweroff rd.shell=0 selinux=1 security=selinux quiet
sh-5.2$ 
sh-5.2$ sudo grep .  /sys/module/processor/parameters/max_cstate /sys/module/processor/parameters/max_cstate
/sys/module/processor/parameters/max_cstate:8
/sys/module/processor/parameters/max_cstate:8
```

I do not understand how it is actually happens, because https://github.com/awslabs/amazon-eks-ami/blob/main/templates/al2023/provisioners/limit-c-states.sh seems to be correct, but this boot options lost, BTW https://github.com/awslabs/amazon-eks-ami/blob/main/templates/al2023/provisioners/set-clocksource.sh also has no effect. may be it is grubby issue? 

# Open question for the team:
c-state issue can be detected via trivial nccl-test for small block size ~ 256b, so I am wondering do you have a regression tests which executed on GPU platforms?

kummatty:
**Issue #, if available:**

Fixes #2741 

**Description of changes:**

On AL2023 NVIDIA AMIs, the kernel command-line args set by `set-clocksource.sh (clocksource=tsc tsc=reliable) and limit-c-states.sh (intel_idle.max_cstate=1 processor.max_cstate=1)` were missing at boot.  Standard and Neuron AMIs were unaffected. 

These scripts applied the args with `grubby --update-kernel=ALL`, which edits only the BLS boot entries (/boot/loader/entries/*.conf). The NVIDIA build script pulls in `nvidia-kmod-common` as a driver dependency, and its `%post` scriptlet runs `nvidia-boot-update` , which regenerates the boot entries and drops args that existed only in the boot entries. Standard/Neuron builds never regenerate, so their grubby-only edits survived.
Ref: https://github.com/NVIDIA/yum-packaging-nvidia-kmod-common/blob/main/nvidia-boot-update

This change sets kernel args both on the current boot entries (via grubby) and in GRUB_CMDLINE_LINUX in `/etc/default/grub`, so they survive a later boot-config regeneration. The call is idempotent and merges into any existing GRUB_CMDLINE_LINUX value.

**Testing Done**

Built v1.35 AMIs for both standard and NVIDIA variants

Standard:
```
cat /proc/cmdline
BOOT_IMAGE=(hd0,gpt1)/boot/vmlinuz-6.12.92-122.166.amzn2023.x86_64 root=UUID=4b9941ea-b69d-42ce-a01a-2acd80dbd5fa ro console=tty0 console=ttyS0,115200n8 nvme_core.io_timeout=4294967295 rd.emergency=poweroff rd.shell=0 selinux=1 security=selinux quiet clocksource=tsc tsc=reliable intel_idle.max_cstate=1 processor.max_cstate=1
```

NVIDIA
```
cat /proc/cmdline
BOOT_IMAGE=(hd0,gpt1)/boot/vmlinuz-6.12.92-122.166.amzn2023.x86_64 root=UUID=4b9941ea-b69d-42ce-a01a-2acd80dbd5fa ro clocksource=tsc tsc=reliable intel_idle.max_cstate=1 processor.max_cstate=1 rd.driver.blacklist=nouveau console=tty0 console=ttyS0,115200n8 nvme_core.io_timeout=4294967295 rd.emergency=poweroff rd.shell=0 selinux=1security=selinux quiet
```

### DIFF
--- a/templates/al2023/runtime/bin/set-kernel-arg
+++ b/templates/al2023/runtime/bin/set-kernel-arg
@@ -1,0 +1,36 @@
+#!/usr/bin/env bash
+#
+# usage: set-kernel-arg "ARG[ ARG...]"
+#
+# Adds kernel command-line argument(s) to the current boot entries and to
+# GRUB_CMDLINE_LINUX in /etc/default/grub, so they also survive a regeneration
+# of the boot config. Idempotent.
+
+set -o pipefail
+set -o nounset
+set -o errexit
+
+if [ "$#" -eq 0 ]; then
+  echo >&2 "usage: set-kernel-arg \"ARG[ ARG...]\""
+  exit 1
+fi
+
+KERNEL_ARGS="$*"
+
+# Apply to the current boot entries.
+sudo grubby --update-kernel=ALL --args="${KERNEL_ARGS}"
+
+. /etc/default/grub
+CMDLINE="${GRUB_CMDLINE_LINUX:-}"
+# Append each arg that isn't already present.
+for arg in ${KERNEL_ARGS}; do
+  grep -qFw -- "${arg}" <<< "${CMDLINE}" || CMDLINE+=" ${arg}"
+done
+CMDLINE="${CMDLINE# }"
+
+# Replace the existing GRUB_CMDLINE_LINUX line, or add it if absent.
+if grep -q '^GRUB_CMDLINE_LINUX=' /etc/default/grub; then
+  sudo sed -i "s|^GRUB_CMDLINE_LINUX=.*|GRUB_CMDLINE_LINUX=\"${CMDLINE}\"|" /etc/default/grub
+else
+  echo "GRUB_CMDLINE_LINUX=\"${CMDLINE}\"" | sudo tee -a /etc/default/grub > /dev/null
+fi
--- a/templates/al2023/provisioners/limit-c-states.sh
+++ b/templates/al2023/provisioners/limit-c-states.sh
@@ -8,6 +8,4 @@ if [ "$ENABLE_EFA" != "true" ]; then
 fi
 
 echo "Limiting deeper C-states"
-sudo grubby \
-  --update-kernel=ALL \
-  --args="intel_idle.max_cstate=1 processor.max_cstate=1"
+set-kernel-arg "intel_idle.max_cstate=1 processor.max_cstate=1"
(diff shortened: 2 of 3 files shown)


kummatty:
/ci
+workflow:k8s_versions 1.35

templates/al2023/runtime/bin/set-kernel-arg
```diff
@@ -0,0 +1,38 @@
...
+  exit 1
+fi
+
+KERNEL_ARGS="$*"
+GRUB_DEFAULTS="/etc/default/grub"
+
+# Apply to the current boot entries.
+sudo grubby --update-kernel=ALL --args="${KERNEL_ARGS}"
+
+# shellcheck disable=SC1090
```
fletcherw: I don't think you'll need the shellcheck disable if you add
`# shellcheck source=/etc/default/grub`
or inline the constant entirely.

Not a big deal though.

fletcherw (approved):

kummatty:
/ci
+workflow:k8s_versions 1.35
+test --instance-types=m5.large

kummatty:
/ci
+workflow:k8s_versions 1.35
+test --instance-types=m5.large

kummatty:
/ci
+workflow:k8s_versions 1.35
+test --instance-types=m5.large

Issue #2122, opened 2025-01-23
Conflict between EKS Best Practices and latest AL2023 AMI.


dawgware:
While upgrading one of our EKS clusters to 1.31 we experienced an issue with new 1.31 nodes not joining the cluster.  
In the logs we saw the following for containerd:

```
Jan 23 03:55:48 ip-10-0-130-143.ec2.internal systemd[1]: Started containerd.service - containerd container runtime.
Jan 23 03:55:50 ip-10-0-130-143.ec2.internal containerd[4125]: time="2025-01-23T03:55:50.855808769Z" level=info msg="RunPodSandbox for &PodSandboxMetadata{Name:efs-csi-node-5fdcs,Uid:dc3211ea-b057-4c13-8d05-598cabb16988,Namespace:kube-system,Attempt:0,}"
Jan 23 03:55:50 ip-10-0-130-143.ec2.internal containerd[4125]: time="2025-01-23T03:55:50.859534903Z" level=info msg="trying next host" error="failed to do request: Head \"https://localhost/v2/kubernetes/pause/manifests/latest\": dial tcp 127.0.0.1:443: connect: connection refused" host=localhost
Jan 23 03:55:50 ip-10-0-130-143.ec2.internal containerd[4125]: time="2025-01-23T03:55:50.861665120Z" level=error msg="RunPodSandbox for &PodSandboxMetadata{Name:efs-csi-node-5fdcs,Uid:dc3211ea-b057-4c13-8d05-598cabb16988,Namespace:kube-system,Attempt:0,} failed, error" error="failed to get sandbox image \"localhost/kubernetes/pause\": failed to pull image \"l>
Jan 23 03:55:50 ip-10-0-130-143.ec2.internal containerd[4125]: time="2025-01-23T03:55:50.861694974Z" level=info msg="stop pulling image localhost/kubernetes/pause:latest: active requests=0, bytes read=0"
```

Checking the logs of one of the running 1.30 nodes we found the following:
```
Jan 22 04:25:28 ip-10-0-128-200.ec2.internal systemd[1]: Started containerd.service - containerd container runtime.
Jan 22 04:25:30 ip-10-0-128-200.ec2.internal containerd[4224]: time="2025-01-22T04:25:30.245168728Z" level=info msg="PullImage \"602401143452.dkr.ecr.us-east-1.amazonaws.com/eks/pause:3.5\""
[...]

PR #2755, opened 2026-07-06
Fix stale mdadm.conf blocking RAID rebuild

maitreya526:
**Issue #, if available:**
Related to https://github.com/awslabs/amazon-eks-ami/issues/2122

**Description of changes:**
After an EC2 instance stop/start, instance-store NVMe disks are wiped but `/.aws/mdadm.conf` persists on the EBS root volume. The `maybe_raid()` function in `setup-local-disks` only checks if the config file is non-empty before deciding to skip array creation. This causes the RAID0 rebuild to be skipped, leaving the node broken (bind-mounts for /var/lib/containerd, /var/lib/kubelet, /var/log/pods fail).

This change adds an additional condition to verify the array actually exists before skipping creation:

```
# Before:
if [[ ! -s "${md_config}" ]]; then

# After:
if [[ ! -s "${md_config}" ]] || ! mdadm --detail --scan 2>/dev/null | grep -q "ARRAY"; then
```

**Testing Done**
Reproduced the bug on an AMI without the fix to confirm the failure. 
Built AMI with the fix (c5d.large, us-east-1, k8s 1.29):
- First boot: setup-local-disks raid0 creates RAID0 successfully
- Stop/start: script detects missing array and rebuilds it successfully
- Reboot: array persists, no unnecessary rebuild (no regression)

### DIFF
--- a/templates/al2023/runtime/bin/setup-local-disks
+++ b/templates/al2023/runtime/bin/setup-local-disks
@@ -45,7 +45,7 @@ maybe_raid() {
   local array_mount_point="${MNT_DIR}/0"
   mkdir -p "$(dirname "${md_config}")"
 
-  if [[ ! -s "${md_config}" ]]; then
+  if [[ ! -s "${md_config}" ]] || ! mdadm --detail --scan 2>/dev/null | grep -q "ARRAY"; then
     mdadm --create --force --verbose \
       "${md_device}" \
       --level="${raid_level}" \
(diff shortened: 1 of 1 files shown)


wwvela:
/ci
+workflow:k8s_versions 1.35
+test --instance-types=c5d.large

wwvela (approved):
Changes LGTM. Lets merge after the ci test passed

shvbsle (approved):

PR #2759, opened 2026-07-09
feat: add multi-EBS volume storage support for AL2023 nodes

maitreya526:
**Issue #, if available:**
Related to https://github.com/awslabs/amazon-eks-ami/issues/2122

**Description of changes:**
Adds support for mounting dedicated EBS volumes to container runtime directories (e.g., ```/var/lib/containerd```,``` /var/lib/kubelet```) on AL2023 EKS nodes via NodeConfig.

Customers on instance types without instance-store NVMe (C7a, M7a, R7a) had no supported way to offload containerd from the root volume. This change adds a new ```storage.volumes``` field to the NodeConfig API:

```
spec:
  instance:
    storage:
      volumes:
        - device: "/dev/xvdb"
          mountTargets:
            - /var/lib/containerd
          fsType: xfs
```
What's included:
- New API types: ```StorageOptions```, ```VolumeMount``` in v1alpha1 and internal API
- New system aspect ```ebs_storage.go``` that formats, copies data, and mounts via systemd mount units
- Registered before ```LocalDiskAspect``` in init.go
- Coexistence with NVMe RAID handled naturally — ```setup-local-disks``` checks ```systemctl is-active``` for each path and skips if already mounted
- Generated deepcopy, conversion, and API docs updated
- e2e test cases added for containerd, kubelet, and coexistence scenarios
- Updated `config-cache` expected files to include new `storage` field

**Testing Done**
Built AMI (c5d.large, us-east-1). Instance launched with second EBS: ```--block-device-mappings '[{"DeviceName":"/dev/xvdb","Ebs":{"VolumeSize":50,"VolumeType":"gp3"}}]'```

EBS-only (containerd):
- User-data: ```storage.volumes[0].device: /dev/xvdb, mountTargets: [/var/lib/containerd]```, ```fsType: xfs```
- ```mountpoint /var/lib/containerd``` → ```is a mountpoint```
- ```df -h /var/lib/containerd``` → ```/dev/nvme1n1 50G```
- ```systemctl status var-lib-containerd.mount``` → ```active (mounted), enabled```
- ```systemctl status containerd``` → ```active (running)```

EBS-only (kubelet):
- User-data: ```storage.volumes[0].device: /dev/xvdb, mountTargets: [/var/lib/kubelet]```, ```fsType: xfs```
- ```mountpoint /var/lib/kubelet``` → ```is a mountpoint```
- ```df -h /var/lib/kubelet``` → ```/dev/nvme1n1 50G```
- ```systemctl status var-lib-kubelet.mount ```→ ```active (mounted), enabled```

Coexistence (EBS containerd + NVMe RAID):
- User-data: ```localStorage.strategy: RAID0 + storage.volumes[0].mountTargets: [/var/lib/containerd]```
- ```df -h /var/lib/containerd``` → ```/dev/nvme1n1 50G (EBS)```
- ```df -h /var/lib/kubelet``` → ```/dev/md127 47G (NVMe RAID)```
- ```mdadm --detail --scan``` → ```ARRAY /dev/md/kubernetes present```
- ```setup-local-disks``` detected active systemd mount, skipped containerd automatically

Stop/start:
- ```systemctl is-active var-lib-containerd.mount``` → ```active``` after stop/start
- ```systemctl is-active var-lib-kubelet.mount``` → active after stop/start

Reboot:
- ```systemctl is-active var-lib-containerd.mount``` → ```active``` after reboot

End-to-end (nodeadm full boot with EKS cluster):
- Launched node with user-data into a test cluster (us-east-1, k8s 1.31)
- nodeadm parsed ```storage.volumes``` from user-data, ran EBS aspect, mounted successfully
- Confirmed with both EBS-only and coexistence configs


maitreya526:
/ci
+workflow:k8s_versions 1.35
+test --instance-types=c5d.large

maitreya526:
/ci
+workflow:k8s_versions 1.35
+test --instance-types=m5.xlarge

maitreya526:
/ci
+workflow:k8s_versions 1.35
+test --instance-types=m5.2xlarge

nodeadm/test/e2e/cases/ebs-storage-kubelet/run.sh
```diff
@@ -0,0 +1,27 @@
...
+
+source /helpers.sh
+
+mock::aws
+wait::dbus-ready
+mock::kubelet 1.35.0
+
+# Create a loopback device to simulate an EBS volume
+dd if=/dev/zero of=/tmp/ebs-disk.img bs=1M count=100
+LOOP_DEV=$(losetup -f --show /tmp/ebs-disk.img)
```
fletcherw, line 15: I had a lot of trouble with e2e test reliability using loopback devices, have you seen any issues?

nodeadm/doc/api.md
```diff
@@ -165,3 +166,28 @@ _Appears in:_
...
+
+VolumeMount specifies an EBS volume device and where it should be mounted.
+
+_Appears in:_
+- [StorageOptions](#storageoptions)
+
+| Field | Description |
+| --- | --- |
+| `device` _string_ | Device is the block device path (e.g., "/dev/xvdb"). |
+| `mountTargets` _string array_ | MountTargets is the list of directories to bind-mount onto this volume.<br />For v1, only /var/lib/containerd is supported. |
```
fletcherw: Drop the "For v1" imo.

nodeadm/internal/system/ebs_storage.go
```diff
@@ -0,0 +1,237 @@
...
+	return "ebs-storage"
+}
+
+func (a *ebsStorageAspect) Setup(cfg *api.NodeConfig) error {
+	if len(cfg.Spec.Instance.Storage.Volumes) == 0 {
+		zap.L().Info("No EBS storage volumes configured")
+		return nil
+	}
+
+	for _, vol := range cfg.Spec.Instance.Storage.Volumes {
```
fletcherw, line 40: what happens if you attempt to mount the same path twice (duplicate volume or mountTarget entries). Does this fail in a reasonable way?

nodeadm/internal/system/ebs_storage.go
```diff
@@ -0,0 +1,237 @@
...
+func (a *ebsStorageAspect) Name() string {
+	return "ebs-storage"
+}
+
+func (a *ebsStorageAspect) Setup(cfg *api.NodeConfig) error {
+	if len(cfg.Spec.Instance.Storage.Volumes) == 0 {
+		zap.L().Info("No EBS storage volumes configured")
+		return nil
+	}
+
```
fletcherw, line 30: how do you know this is an EBS volume? Is there a way to make this robust in case the user specifies a `device` that's not EBS?

nodeadm/internal/system/ebs_storage.go
```diff
@@ -0,0 +1,237 @@
...
+	for _, target := range vol.MountTargets {
+		if err := a.setupMountTarget(log, devUUID, target, fsType); err != nil {
+			return fmt.Errorf("failed to setup mount target %s: %w", target, err)
+		}
+	}
+	return nil
+}
+
+func (a *ebsStorageAspect) setupMountTarget(log *zap.Logger, devUUID string, target string, fsType string) error {
```
fletcherw, line 92: seems like this works for any path, not just containerd? Is that intended?

nodeadm/doc/api.md
fletcherw: these aren't really bind mounts, and I think only a single target is supported since you're directly mounting "device" at the target and an additional target would fail with EBUSY.

nodeadm/test/e2e/cases/ebs-storage-kubelet/run.sh
maitreya526: I did initially hit ```Device or resource busy``` when hard coding ```/dev/loop0```. When I switched to ```losetup -f --show```, I did not face any issue. If there might be other reliability issues in this implementation, I can take another approach.

nodeadm/internal/system/ebs_storage.go
maitreya526: I have a ```isUnitActive``` function which checks if a particular path is already mounted and if it is, it just skips silently(the function catches duplicates and reboots). I can gracefully handle it so that the cx know that a path is duplicated if required.

nodeadm/internal/system/ebs_storage.go
maitreya526: Good point, currently it is not verifying for an EBS volume specifically. I will add a check using ```lsblk -o MODEL ```. Does that work?

nodeadm/internal/system/ebs_storage.go
maitreya526: Yes, I tested it with both ```/var/lib/containerd``` and ```/var/lib/kubelet```and both work. The API uses a mountTargets list so customers can target different directories. Coexistence with NVMe RAID is handled by ```setup-local-disks``` checking ```systemctl is-active``` for each path before bind-mounting, so there's no conflict regardless of which path is used.

nodeadm/doc/api.md
maitreya526: Okay yeah, makes sense. I'll change ```mountTargets []string``` to ```mountTarget string``` and fix the doc wording. Customers who need multiple paths on EBS would use multiple volumes (one device per target).

maitreya526:
/ci
+workflow:k8s_versions 1.35
+test --instance-types=m5.2xlarge

erezzarum:
I would add +1 for a feature built on top of this.
Can we add the ability to stripe multiple EBS volumes as RAID0 and mount by targets?
This will allow striping multiple EBS volumes to achieve higher throughput and IOPs at zero cost.

maitreya526:
> I would add +1 for a feature built on top of this. Can we add the ability to stripe multiple EBS volumes as RAID0 and mount by targets? This will allow striping multiple EBS volumes to achieve higher throughput and IOPs at zero cost.

Makes sense, good idea for a follow up. Can be added without breaking the current API.
Created an issue regarding the same https://github.com/awslabs/amazon-eks-ami/issues/2767

fletcherw (approved):

PR #2769, opened 2026-07-19
Add RTX PRO 4500 Blackwell Server Edition (0x2C3A) to 580 open-kmod supported devices

### LINKED ISSUE #2768: AL2023 NVIDIA AMI: g7 instance family GPU (10de:2c3a) missing from nvidia-open supported-devices list — proprietary kmod loaded, GPU unusable
## What happened

On EKS nodes from the **g7 instance family**, the AL2023 NVIDIA AMI loads the **proprietary** NVIDIA kernel module instead of the open one. The g7 family's GPU (Blackwell, PCI ID `10de:2c3a`) is only supported by the open kernel modules, so the driver fails to initialize, `nvidia-smi` reports `No devices were found`, the NVIDIA k8s device plugin registers 0 GPUs, and the node never advertises `nvidia.com/gpu` — GPU pods stay `Pending` with `Insufficient nvidia.com/gpu`.

## Environment

- AMI: `amazon-eks-node-al2023-x86_64-nvidia-1.36-v20260709` (driver 580.159.03)
- Instance family: **g7** (any size; reproduced in us-east-1)
- Kubernetes: EKS 1.36, nodes provisioned by Karpenter (`al2023@latest` alias)

## Root cause

`/etc/eks/nvidia-kmod-load.sh` decides the kmod flavor by matching the GPU's PCI device ID against `/etc/eks/nvidia-open-supported-devices-580.txt`. That file is missing `0x2C3A` — the device ID of the g7 family's GPU (it identifies as "NVIDIA RTX PRO 4500 Blackwell Server Edition") — while the retail Blackwell siblings are present:

```
$ grep -c 2C3A /etc/eks/nvidia-open-supported-devices-580.txt
0
$ grep 2C3 /etc/eks/nvidia-open-supported-devices-580.txt
0x2C31 NVIDIA RTX PRO 4500 Blackwell
0x2C33 NVIDIA RTX PRO 4000 Blackwell SFF Edition
0x2C34 NVIDIA RTX PRO 4000 Blackwell
0x2C38 NVIDIA RTX PRO 5000 Blackwell Generation Laptop GPU
0x2C39 NVIDIA RTX PRO 4000 Blackwell Generation Laptop GPU
```

So `devices-support-open()` returns false and the script falls back to the proprietary kmod, which refuses the GPU:

```
NVRM: The NVIDIA GPU 0000:2f:00.0 (PCI ID: 10de:2c3a)
NVRM: installed in this system requires use of the NVIDIA open kernel modules.
NVRM: GPU 0000:2f:00.0: RmInitAdapter failed! (0x22:0x56:897)
```

## Verification that the open kmod works

Swapping flavors on a live g7 node fully fixes it:

```
$ kmod-util remove nvidia && kmod-util load nvidia-open && modprobe nvidia
$ nvidia-smi
NVIDIA-SMI 580.159.03   Driver Version: 580.159.03   CUDA Version: 13.0
GPU 0: NVIDIA RTX PRO 4500 Blackwell...  32623MiB
```

```
NVRM: loading NVIDIA UNIX Open Kernel Module for x86_64  580.159.03
```

After this the device plugin advertises `nvidia.com/gpu: 1` and GPU workloads run normally (verified with CUDA inference).

## Suggested fix

Add `0x2C3A` (and any other g7-family GPU device IDs) to the generated `nvidia-open-supported-devices-*.txt`, or make `nvidia-kmod-load.sh` prefer the open kmod for device IDs the proprietary driver explicitly rejects.

## Workaround we're using

EC2NodeClass/userData script that appends `0x2C3A` to the supported-devices file before/after `nvidia-kmod-load.service` and swaps the kmod if the proprietary one was already installed.

harryzsh:
## Description

Fixes #2768

The 580 supported-devices list (`templates/al2023/runtime/gpu/nvidia-open-supported-devices-580.txt`) is missing the **RTX PRO 4500 Blackwell Server Edition** (`10de:2c3a`) used by the EC2 **g7** instance family. This looks like an enumeration gap rather than an intentional exclusion:

- the workstation variant `0x2C31 NVIDIA RTX PRO 4500 Blackwell` **is** in the 580 list
- so are the laptop variants `0x2C38` / `0x2C39`
- the **595 list in this repo already contains this exact entry** (`0x2C3A NVIDIA RTX PRO 4500 Blackwell Server Edition`)

## Impact without this fix

On g7 instances, `nvidia-kmod-load.sh` misses the allowlist and falls back to the proprietary kmod. Blackwell requires the open kernel modules, so every GPU fails to initialize:

```
NVRM: The NVIDIA GPU 0000:57:00.0 (PCI ID: 10de:2c3a)
NVRM: installed in this system requires use of the NVIDIA open kernel modules.
NVRM: GPU 0000:57:00.0: RmInitAdapter failed! (0x22:0x56:897)
```

`nvidia-smi` reports "No devices were found" while the node otherwise joins the cluster looking healthy — the device plugin registers 0 GPUs and all GPU pods stay Pending.

## Change

One line: add `0x2C3A` to the 580 list, entry text copied verbatim from the 595 list, inserted in device-ID order between `0x2C39` and `0x2C58`.

## Testing

Verified on `g7.48xlarge` with `amazon-eks-node-al2023-x86_64-nvidia-1.34-v20260714` (EKS 1.34, us-east-1): appending `0x2C3A` to the on-node list before `nvidia-kmod-load.service` runs results in the open kmod loading cleanly — all 8 GPUs visible, `modinfo nvidia` → `license: Dual MIT/GPL`, node Ready in ~3 minutes with no reboots.


HusainZafar:
We generate this file using the script `hack/generate-nvidia-open-supported-devices.sh` by running a command like `./generate-nvidia-open-supported-devices.sh NVIDIA-Linux-x86_64-580.173.02.run`. If you've made this change manually, it will be overwritten by the next script run. 

I ran that script locally, it generated a diff for file `templates/al2023/runtime/gpu/nvidia-open-supported-devices-580.txt`. However the diff does not include `0x2C3A`

Open question: 
1. Why is `0x2C3A` not marked as supported in 580 by Nvidia?
2. Does this GPU really need to be marked as supported despite Nvidia not claiming so? If that is the case, we will have to potentially update the script to make an exception, which sounds more like a hack

We need to understand the answer to the above questions before we make any changes to the file

HusainZafar:
Closing this PR since the original issue is marked as not planned

PR #2773, opened 2026-07-23
docs: clarify drop in file overrides

fletcherw:
Explain how to use NodeConfig drop ins for overriding config via cloud-init or for configuring nodes via custom AMIs.


mselim00 (approved):
thanks for adding this!

PR #2777, opened 2026-07-29
feat(nodeadm): add more variables to maxPodsExpression

joshuakguo:
**Issue #, if available:**

**Description of changes:**

adding more variables (vcpus and physical_memory_mib) to nodeadm maxPodsExpression CEL expression

**Testing Done**

unit tests

make test


nodeadm/internal/kubelet/eni_max_pods_test.go
```diff
@@ -300,5 +332,14 @@ func TestInstanceInfoLoadable(t *testing.T) {
...
 		// we expect at least 2 pods for the host networking ones
 		assert.Greater(t, calculateStandardMaxPods(instanceInfo), int32(1))
+		// a modern instance type must carry the vcpus/memory_mib keys the CEL env now exposes;
+		// legacy supplemented types (addInstanceTypeSupplements) legitimately carry zeros, so
+		// we assert against a known type rather than every line
+		if instanceInfo.InstanceType == "m5.large" {
+			sawVCpusAndMemory = true
+			assert.Equal(t, int32(2), instanceInfo.VCpus)
+			assert.Equal(t, int64(8192), instanceInfo.MemoryMiB)
+		}
```
mselim00: these validations are meant to be bare minimum, let's just assert that VCpus and MemoryMiB are nonzero rather than testing the values for a specific instance type

nodeadm/internal/kubelet/eni_max_pods.go
```diff
@@ -21,6 +21,8 @@ const (
 	defaultENIsVar = "default_enis"
 	ipsPerENIVar   = "ips_per_eni"
 	maxPodsVar     = "max_pods"
+	vcpusVar       = "vcpus"
+	memoryMiBVar   = "memory_mib"
```
mselim00: I think `physical_memory_mib` or something like that may be better future proofing - the value we're inserting here is the total amount of physical RAM, the total memory perceivable from the system (`MemTotal`) will be less than this and might be something we want to inject in the future

joshuakguo: changed it

nodeadm/internal/kubelet/eni_max_pods_test.go
joshuakguo: added asserts

nodeadm/internal/kubelet/eni_max_pods_test.go
```diff
@@ -300,5 +331,11 @@ func TestInstanceInfoLoadable(t *testing.T) {
 		assert.Greater(t, instanceInfo.Ipv4AddressesPerInterface, int32(0))
 		// we expect at least 2 pods for the host networking ones
 		assert.Greater(t, calculateStandardMaxPods(instanceInfo), int32(1))
+		// legacy supplemented types (addInstanceTypeSupplements) predate the vcpus/memory
+		// keys and carry zeros for both; every EC2-sourced entry must have them nonzero
+		if instanceInfo.VCpus != 0 || instanceInfo.PhysicalMemoryMiB != 0 {
+			assert.Greater(t, instanceInfo.VCpus, int32(0))
+			assert.Greater(t, instanceInfo.PhysicalMemoryMiB, int64(0))
+		}
```
mselim00: we have the instance type in this data, so instead of guarding this against a non-zero check, let's specifically enumerate the instance types where we want to tolerate zero values and check this for all other ones. if a new instance type is added with VCpus = 0, we should fail. the list of instance types should only shrink

joshuakguo: added the instance types that we know as zero into a hard coded check

mselim00 (approved):
lgtm, thanks!

PR #2783, opened 2026-08-14
feat(nvidia): scaffold multi-version NVIDIA driver tree build

kummatty:
**Description of changes:**

Initial scaffolding for supporting multiple nvidia driver versions in a single AMI (lts, pb).
The NVIDIA LTS and PB versions will be provided as separate variables, which will conditionally invoke a new provisioner.

In this PR, the new provisioner only does basic scaffolding such as 
1. Configuring the NVIDIA AL repository.
2. Resolves full nvidia driver version for each tree.
3. Creates a separate dir tree for the two versions to be baked into the AMI


templates/al2023/provisioners/install-nvidia-driver-trees.sh
```diff
@@ -0,0 +1,126 @@
...
+  MAJOR="${NVIDIA_TREE_MAJOR[${TREE}]}"
+  NVIDIA_TREE_VERSION[${TREE}]=$(resolve-tree-version "${MAJOR}")
+  echo "Tree ${TREE}: major version ${MAJOR} resolved to ${NVIDIA_TREE_VERSION[${TREE}]}"
+
+  TREE_DIR="${NVIDIA_TREE_ROOT}/${TREE}"
+  sudo mkdir -p "${TREE_DIR}"
+  echo "${NVIDIA_TREE_VERSION[${TREE}]}" | sudo tee "${TREE_DIR}/.version" > /dev/null
+  # A simple marker to act as conditional for systemd services
+  sudo touch "${TREE_DIR}/.tree-${TREE}"
+done
```
mselim00, line 116: instead of the array, map, and this for loop, I think this would be cleaner + more readable extracted as a function and called explicitly for each version we want to handle

templates/al2023/provisioners/install-nvidia-driver-tree.sh
```diff
@@ -0,0 +1,126 @@
...
+
+# To ensure proper functionality, we need to enforce that all three kernel modules are on the same NVIDIA driver version
+# so they are compatible with the same userspace components.
+# If one of the open module or the grid driver runfile version are older, we use that version for all installations.
+function resolve-tree-version() {
+  local MAJOR="${1}"
+  local CUDA_VERSION
+  local GRID_VERSION
+
+  CUDA_VERSION=$(dnf repoquery "${NVIDIA_MODULE_HOTFIXES}" --latest=1 --queryformat "%{version}" "kmod-nvidia-open-dkms-${MAJOR}*")
```
mselim00: nit: I think `NVIDIA_MODULE_HOTFIXES` makes it a bit harder to read, I don't think this variable is needed

templates/al2023/provisioners/install-nvidia-driver-tree.sh
```diff
@@ -0,0 +1,126 @@
...
+################################################################################
+### Resolve versions and create the driver trees ###############################
+################################################################################
+
+# To ensure proper functionality, we need to enforce that all three kernel modules are on the same NVIDIA driver version
+# so they are compatible with the same userspace components.
+# If one of the open module or the grid driver runfile version are older, we use that version for all installations.
+function resolve-tree-version() {
+  local MAJOR="${1}"
+  local CUDA_VERSION
```
mselim00:
```suggestion
  local OPEN_VERSION
```

this variable is used to store the open driver's version, not CUDA

templates/al2023/provisioners/install-nvidia-driver-tree.sh
```diff
@@ -0,0 +1,126 @@
...
+
+if [ "$ENABLE_ACCELERATOR" != "nvidia" ]; then
+  exit 0
+fi
+
+# A driver-tree build requires both tree versions. Without them,
+# install-nvidia-driver.sh performs the single-version build instead.
+if [ -z "${NVIDIA_DRIVER_LTS_VERSION:-}" ] || [ -z "${NVIDIA_DRIVER_PB_VERSION:-}" ]; then
+  exit 0
+fi
```
mselim00: talked offline but I don't think we need the added complexity/lines of code that maintaining "backwards compatibility" with install-nvidia-driver.sh brings, this is work is confined to a branch branch that will eventually directly overwrite install-nvidia-driver.sh in mainline

mselim00 (approved):

templates/al2023/provisioners/install-nvidia-driver-trees.sh
kummatty: I'll change this more, as I start plugging the rest of the installs into this file (kernel modules + userspace programs). I am not clear on the final look yet to commit to this.

templates/al2023/template.json
```diff
@@ -260,20 +261,17 @@
     {
       "type": "shell",
       "remote_folder": "{{ user `remote_folder`}}",
-      "script": "{{template_dir}}/provisioners/install-nvidia-driver.sh",
+      "script": "{{template_dir}}/provisioners/install-nvidia-driver-trees.sh",
```
mselim00, line 264: nit: `-trees` is an implementation detail, we can preserve the original name

mselim00 (approved):
thanks!

PR #2784, opened 2026-08-18
feat(nvidia): installing prebuilt kernel modules to trees

kummatty:
**Description of changes:**

Builds the NVIDIA kernel modules into the driver trees created in the previous PR https://github.com/awslabs/amazon-eks-ami/pull/2783

A new sourced library, `helpers/nvidia/nvidia-kmods.sh`, builds and harvests the kernel modules for a single tree, and is invoked once per driver version. 

In this PR:

- Installs the kernel module build dependencies (`kernel-devel`, `headers`, `modules-extra`, `dkms`) and version-locks the kernel.
- Builds the open and proprietary flavors from their dkms source RPMs, without installing them onto the build host.
- Builds the GRID flavor from the runfile with a direct `make`, and validates the module set it produces.
- Builds `gdrdrv` against the open driver source, harvested into the open flavor only.
- Harvests each flavor's modules into `/opt/nvidia/<tree>/flavors/<flavor>/lib/modules/<kernel>/extra`, and keeps the kmod RPMs in `/opt/nvidia/<tree>/.rpms` for registration in the rpmdb at boot.

Build time checks
- Open and proprietary - Sources `dkms.conf` to ensure built module names are copied as produced
- Grid - Since a `dkms.conf` file is not available and a `.manifest` is brittle to parse, relying on a hard coded list of known driver modules. The build will fail if any module is not available from the known list after build and if the build produces more than the known modules.

### DIFF
--- a/templates/al2023/helpers/nvidia-kmods.sh
+++ b/templates/al2023/helpers/nvidia-kmods.sh
@@ -1,0 +1,184 @@
+#!/usr/bin/env bash
+#
+# Per-flavor builders for one driver tree's kernel modules, each harvesting into
+# <tree>/flavors/<flavor>/lib/modules/<kernel>/.
+
+KERNEL_RELEASE="$(uname -r)"
+readonly KERNEL_RELEASE
+
+ARCH="$(uname -m)"
+readonly ARCH
+
+# Build and harvests nvidia open flavor kernel modules
+function build-open-kmods() {
+  local TREE_DIR="${1}"
+  local DRIVER_VERSION="${2}"
+
+  echo "Building open kernel modules for NVIDIA ${DRIVER_VERSION}"
+  extract-kmod-source "kmod-nvidia-open-dkms-${DRIVER_VERSION}" "${TREE_DIR}"
+  sudo dkms add -m nvidia -v "${DRIVER_VERSION}"
+  sudo dkms build -m nvidia -v "${DRIVER_VERSION}"
+  harvest-dkms-modules nvidia "${DRIVER_VERSION}" "${TREE_DIR}/flavors/open"
+
+  # Built here, while this version's nvidia source is the only one in /usr/src.
+  build-gdrdrv "${TREE_DIR}" "${DRIVER_VERSION}"
+
+  sudo dkms remove -m nvidia -v "${DRIVER_VERSION}" --all
+  sudo rm -rf "/usr/src/nvidia-${DRIVER_VERSION}"
+}
+
+# Build and harvests nvidia proprietary flavor kernel modules
+function build-proprietary-kmods() {
+  local TREE_DIR="${1}"
+  local DRIVER_VERSION="${2}"
+
+  echo "Building proprietary kernel modules for NVIDIA ${DRIVER_VERSION}"
+  extract-kmod-source "kmod-nvidia-latest-dkms-${DRIVER_VERSION}" "${TREE_DIR}"
+  sudo dkms add -m nvidia -v "${DRIVER_VERSION}"
+  sudo dkms build -m nvidia -v "${DRIVER_VERSION}"
+  harvest-dkms-modules nvidia "${DRIVER_VERSION}" "${TREE_DIR}/flavors/proprietary"
+
+  sudo dkms remove -m nvidia -v "${DRIVER_VERSION}" --all
+  sudo rm -rf "/usr/src/nvidia-${DRIVER_VERSION}"
+}
+
+# Build and harvests gdrdrv kernel modules. Built only with the open nvidia flavor
+function build-gdrdrv() {
+  local TREE_DIR="${1}"
+  local DRIVER_VERSION="${2}"
+
+  if [ "${ENABLE_NVIDIA_GDRCOPY_DRIVER}" != "true" ] || [ -z "${NVIDIA_GDRCOPY_DRIVER_VERSION}" ]; then
+    return 0
+  fi
+
+  local GDRCOPY_VERSION="${NVIDIA_GDRCOPY_DRIVER_VERSION}"
+
+  echo "Building gdrdrv ${GDRCOPY_VERSION} against NVIDIA ${DRIVER_VERSION}"
+  extract-kmod-source "gdrcopy-kmod-${GDRCOPY_VERSION}" "${TREE_DIR}"
+  sudo dkms add -m gdrdrv -v "${GDRCOPY_VERSION}"
+  sudo dkms build -m gdrdrv -v "${GDRCOPY_VERSION}"
+  harvest-dkms-modules gdrdrv "${GDRCOPY_VERSION}" "${TREE_DIR}/flavors/open"
+
+  sudo dkms remove -m gdrdrv -v "${GDRCOPY_VERSION}" --all
+  sudo rm -rf "/usr/src/gdrdrv-${GDRCOPY_VERSION}"
+}
+
+# Build and harvest nvidia grid flavor kernel modules
+function build-grid-kmods() {
+  local TREE_DIR="${1}"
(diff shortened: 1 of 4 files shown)


templates/al2023/provisioners/nvidia/nvidia-kmods.sh
```diff
@@ -0,0 +1,196 @@
...
+  sudo dkms remove -m gdrdrv -v "${GDRCOPY_VERSION}" --all
+  sudo rm -rf "/usr/src/gdrdrv-${GDRCOPY_VERSION}"
+}
+
+# Build and harvest nvidia grid flavor kernel modules
+function build-grid-kmods() {
+  local TREE_DIR="${1}"
+  local DRIVER_VERSION="${2}"
+
+  if [ "${MACHINE}" != "x86_64" ]; then
```
mselim00: let's redefine MACHINE as a local variable, it's a bit difficult to track that the caller (of `nvidia-kmods.sh`) is supposed to set this as an env var

templates/al2023/provisioners/nvidia/nvidia-kmods.sh
```diff
@@ -0,0 +1,196 @@
...
+  for MODULE_NAME in "${NVIDIA_DRIVER_MODULES[@]}"; do
+    if [ ! -f "${KERNEL_OPEN_DIR}/${MODULE_NAME}.ko" ]; then
+      echo >&2 "ERROR: GRID build is missing ${MODULE_NAME}.ko"
+      return 1
+    fi
+  done
+  if [ "${#BUILT_MODULES[@]}" -ne "${#NVIDIA_DRIVER_MODULES[@]}" ]; then
+    echo >&2 "ERROR: GRID built ${#BUILT_MODULES[@]} modules, expected ${#NVIDIA_DRIVER_MODULES[@]}: ${BUILT_MODULES[*]##*/}"
+    return 1
+  fi
```
mselim00: conventionally, this should be in `validate.sh`

templates/al2023/helpers/nvidia-kmods.sh
```diff
@@ -0,0 +1,196 @@
...
+    echo >&2 "ERROR: GRID built ${#BUILT_MODULES[@]} modules, expected ${#NVIDIA_DRIVER_MODULES[@]}: ${BUILT_MODULES[*]##*/}"
+    return 1
+  fi
+
+  # Harvest the modules to the tree
+  local DEST_DIR="${TREE_DIR}/flavors/grid/lib/modules/${KERNEL_RELEASE}/extra"
+  local MODULE_PATH
+  sudo install -d "${DEST_DIR}"
+  for MODULE_PATH in "${BUILT_MODULES[@]}"; do
+    strip -g --strip-unneeded "${MODULE_PATH}"
```
mselim00, line 109: is this only intended for the space savings? if so, how much are we actually saving by doing this

templates/al2023/helpers/nvidia-kmods.sh
```diff
@@ -0,0 +1,196 @@
...
+  sudo mv "${DOWNLOAD_DIR}"/*.rpm "${TREE_DIR}/.rpms/"
+  sudo rm -rf "${DOWNLOAD_DIR}"
+}
+
+# Copy the modules a dkms.conf declares out of the dkms build tree and into a flavor
+# All dkms modules end up in /lib/modules/${KERNEL_RELEASE}/extra for AL2023
+# https://man.archlinux.org/man/extra/dkms/dkms.8.en#DEST_MODULE_LOCATION___=
+function harvest-dkms-modules() {
+  local DKMS_NAME="${1}"
+  local DKMS_VERSION="${2}"
```
mselim00:
```suggestion
  local MODULE_NAME="${1}"
  local MODULE_VERSION="${2}"
```
ok to add a `DKMS_` prefix, but the current naming confused me a bit at first

templates/al2023/provisioners/nvidia/nvidia-kmods.sh
```diff
@@ -0,0 +1,196 @@
...
+    if ! compgen -G "${DKMS_MODULE_DIR}/${MODULE_NAME}.ko*" > /dev/null; then
+      echo >&2 "ERROR: ${DKMS_NAME} declares ${MODULE_NAME}, missing from ${DKMS_MODULE_DIR}"
+      return 1
+    fi
+    sudo install -m 0644 "${DKMS_MODULE_DIR}/${MODULE_NAME}".ko* "${DEST_DIR}/"
+  done
+}
+
+# Print the module names a dkms.conf declares.
+# https://man.archlinux.org/man/extra/dkms/dkms.8.en#BUILT_MODULE_LOCATION___=
```
mselim00:
```suggestion
# https://man.archlinux.org/man/extra/dkms/dkms.8.en#BUILT_MODULE_NAME___=
```

templates/al2023/helpers/nvidia-kmods.sh
mselim00, line 1: this should be under `templates/al2023/runtime` (it's not a `provisioner`, it's used by a `provisioner`)

templates/al2023/helpers/nvidia-kmods.sh
kummatty: Did a quick test. It saves about ~75 MB per tree (with nvidia-uvm) being the chunky one.

This is also something a dkms operation gets by default (https://man.archlinux.org/man/extra/dkms/dkms.8.en#STRIP___=). Since we used `nvidia-installer --dkms` before, this maintains parity.
I got the `--strip-unneeded` from bottlerocket, even though it only saves about 0.3% of the total.

templates/al2023/helpers/nvidia-kmods.sh
kummatty: Discussed offline. Will move `nvidia-kmods` into another top level directory under `templates/al2023/`

templates/al2023/helpers/nvidia-kmods.sh
kummatty: Replacing this with PACKAGE_NAME and PACKAGE_VERSION. I'd like to refer to modules as the kernel modules each dkms package produces that we harvest.

templates/al2023/helpers/nvidia-kmods.sh
mselim00: ah ok I thought this would've been new, good to keep then

templates/al2023/helpers/nvidia-kmods.sh
```diff
@@ -0,0 +1,186 @@
...
+  aws s3 cp "s3://${EC2_GRID_DRIVER_S3_BUCKET%%/*}/${RUNFILE_KEY}" "${RUNFILE_PATH}"
+  chmod +x "${RUNFILE_PATH}"
+  "${RUNFILE_PATH}" --extract-only --target "${GRID_DIR}/source"
+
+  # IGNORE_CC_MISMATCH is required even when the compiler matches the kernel's.
+  make -C "${KERNEL_OPEN_DIR}" \
+    -j"$(nproc)" \
+    IGNORE_CC_MISMATCH=1 \
+    SYSSRC="/lib/modules/${KERNEL_RELEASE}/build" \
+    modules
```
mselim00: can you elaborate a bit more on setting `IGNORE_CC_MISMATCH`? if the runfile was previously also ignoring that mismatch, that'd be a good enough explanation

mselim00 (approved):

templates/al2023/helpers/nvidia-kmods.sh
kummatty: Inherited from BR. I believe we can remove it though, since we install the kernel-devel prior and also lock the kernel before any installs. 
Tested without it and build passes.

PR #2785, opened 2026-08-20
fix: wait for systemd restart to complete

fletcherw:
When calling EnsureRunning for a systemd unit, that was triggering an async restart on systemd and then immediately polling for the "Running" status on the unit. In certain cases (especially when SOCI was enabled), that check was racy, meaning that the status poll returned too early and dependent calls could fail.

Fix this by using the pre-existing status channel from the DBus API to block until the systemd operation completes.

As part of this change, also thread some contexts throughout the codebase to provide timeouts if needed.


mselim00:
/ci 
+test --nodeadm-feature-gates=FastImagePull=true

junpengdev (approved):

PR #2788, opened 2026-08-21
feat(nvidia): installing nvidia userspace programs into driver trees

kummatty:
**Description of changes:**

Installs the NVIDIA userspace into the driver trees created in #2783 

A new sourced library, `helpers/nvidia-userspace.sh`, downloads each tree's userspace RPMs and sorts them into packages shared between the driver versions and packages belonging to one version. The host cannot hold two versions of the same package, so the second group is extracted into the trees instead of being installed.

In this PR:

- Downloads the full userspace closure per tree, without installing it onto the build host.
- Partitions the RPMs into shared and version-specific sets (A test between 580 and 595 resulted in 199 shared and 19 per tree)
- Shared set is directly installed to host
- Specific set is stored inside the version specific tree
- Created a user for running nvidia-persistenced daemon as per the `%pre` scriptlet of the RPM.

`nvidia-imex` is purposely skipped in this CR. Will be adding it back later after collecting a few more details. 

Build time checks

- For each tree, ensure that the kernel modules, userspace programs within the tree all belong to the same version recorded for the tree

### DIFF
--- a/templates/al2023/helpers/nvidia-userspace.sh
+++ b/templates/al2023/helpers/nvidia-userspace.sh
@@ -1,0 +1,143 @@
+#!/usr/bin/env bash
+#
+# Userspace for the driver trees. Packages identical across driver versions are installed on
+# the host once. The rest go into their own tree, since the host cannot hold two versions of
+# the same package.
+
+readonly NVIDIA_USERSPACE_RPM_DIR="${WORKING_DIR}/nvidia-userspace-rpms"
+readonly NVIDIA_VERSION_SHARED_RPM_DIR="${NVIDIA_USERSPACE_RPM_DIR}/version-shared"
+
+readonly NVIDIA_USERSPACE_PACKAGES=(
+  libnvidia-fbc
+  nvidia-driver
+  nvidia-driver-cuda
+  nvidia-fabricmanager
+  nvidia-libXNVCtrl-devel
+  nvidia-persistenced
+  nvidia-settings
+  nvidia-xconfig
+  xorg-x11-nvidia
+)
+
+# Download one tree's full set of userspace rpms without installing any of them.
+function download-nvidia-userspace-rpms() {
+  local TREE="${1}"
+  local RPM_LIST="${NVIDIA_USERSPACE_RPM_DIR}/${TREE}"
+
+  local DRIVER_VERSION
+  DRIVER_VERSION=$(cat "${NVIDIA_TREE_ROOT}/${TREE}/.version")
+
+  local VERSIONED_PACKAGES=()
+  local PACKAGE
+  for PACKAGE in "${NVIDIA_USERSPACE_PACKAGES[@]}"; do
+    VERSIONED_PACKAGES+=("${PACKAGE}-${DRIVER_VERSION}")
+  done
+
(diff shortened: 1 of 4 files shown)


templates/al2023/helpers/nvidia-userspace.sh
```diff
@@ -0,0 +1,158 @@
...
+  local FIRMWARE_DIR="${NVIDIA_TREE_ROOT}/${TREE}/usr/lib/firmware/nvidia"
+
+  if [ ! -d "${FIRMWARE_DIR}" ]; then
+    echo >&2 "ERROR: tree ${TREE} carries no GSP firmware at ${FIRMWARE_DIR}"
+    return 1
+  fi
+
+  sudo install -d /usr/lib/firmware/nvidia
+  sudo cp -a "${FIRMWARE_DIR}/." /usr/lib/firmware/nvidia/
+}
```
mselim00: discussed offline, this was done due to the size of the directory (>~ 300MiB). let's consider doing this conditionally only for the version we select at boot

templates/al2023/provisioners/validate.sh
```diff
@@ -100,10 +100,74 @@ validate_nvidia_grid_modules() {
...
+validate_nvidia_tree_version() {
+  local tree=$1
+  local tree_dir="/opt/nvidia/${tree}"
+  local driver_version flavor extra_dir modules module_version
+
+  driver_version=$(cat "${tree_dir}/.version")
+
+  for flavor in "${NVIDIA_KMOD_FLAVORS[@]}"; do
+    # NVIDIA publishes no aarch64 GRID runfile, so that flavor is never built there.
+    if [ "${flavor}" == "grid" ] && [ "$(uname -m)" != "x86_64" ]; then
```
mselim00:
```suggestion
    if [ "${flavor}" == "grid" ] && [ "$(uname -m)" == "aarch64" ]; then
```
is a bit cleaner, it matches the comment to directly assert that we're skipping this on `aarch64` (avoids amd64/x86_64 confusion for example)

templates/al2023/provisioners/validate.sh
```diff
@@ -100,10 +100,74 @@ validate_nvidia_grid_modules() {
...
-  # NVIDIA publishes no aarch64 GRID runfile, so that flavor is never built there.
-  if [ "$(uname -m)" == "x86_64" ]; then
-    validate_nvidia_grid_modules lts
-    validate_nvidia_grid_modules pb
+  for tree in "${NVIDIA_TREES[@]}"; do
+    validate_nvidia_tree_version "${tree}"
+    validate_nvidia_supported_device_list "${tree}"
+
+    # NVIDIA publishes no aarch64 GRID runfile, so that flavor is never built there.
+    if [ "$(uname -m)" == "x86_64" ]; then
```
mselim00: in keeping with the other comment
```suggestion
    if [ "$(uname -m)" != "aarch64" ]; then
```

mselim00 (approved):

templates/al2023/helpers/nvidia-userspace.sh
kummatty: Checked size. Its actually ~100MB, which should be cheaper to copy during boot. So leaving it for boot for now. 
Will revisit in the future if found to cost more time.

PR #2792, opened 2026-08-26
feat(nvidia): add boot-time kmod load scripts

mselim00:
**Issue #, if available:**

**Description of changes:**
Adds two scripts for boot-time driver selection and loading. `resolve-nvidia-driver.sh` is responsible for selecting the appropriate NVIDIA driver flavor and major version for the current hardware, and creating a symbolic link to the `current` tree from the `lts`/`pb` so that `load-nvidia-kernel-modules.sh` can commit that one. `load-nvidia-kernel-modules.sh` will reflink the /lib/modules and /usr/lib/firmware for GSP from the appropriate driver version and flavor, utilizing a `driver-flavor` file that `resolve-nvidia-driver.sh` writes.

The resolver will abort if the current hardware does not have any attached NVIDIA devices, which ensures users can continue using EKS AMIs as a base for accelerated AMI builds on non-accelerated instance types. If a snapshot is created from an instance with an NVIDIA attached device, subsequent instances launched from that snapshot will use the same driver type, which is the same behavior as with the predcessor `nvidia-kmod-load.sh`.

### DIFF
--- a/templates/al2023/runtime/gpu/resolve-nvidia-driver.sh
+++ b/templates/al2023/runtime/gpu/resolve-nvidia-driver.sh
@@ -1,0 +1,140 @@
+#!/usr/bin/env bash
+
+set -x
+set -o errexit
+set -o pipefail
+set -o nounset
+
+NVIDIA_TREE_ROOT="${NVIDIA_TREE_ROOT:-/opt/nvidia}"
+EKS_CONFIG_DIR="${EKS_CONFIG_DIR:-/etc/eks}"
+MODPROBE_D_DIR="${MODPROBE_D_DIR:-/etc/modprobe.d}"
+DMI_PRODUCT_NAME_PATH="${DMI_PRODUCT_NAME_PATH:-/sys/devices/virtual/dmi/id/product_name}"
+
+readonly NVIDIA_VENDOR_ID="10de"
+readonly PCI_CLASS_CODES=(
+  "0300" # VGA controller; instance types like g3, g4
+  "0302" # 3D controller; instance types like p4, p5
+)
+
+# NVIDIA vGPU (GRID) device:subsystem tuples. Format: "DEVICE:SUBDEVICE" (lowercase, no 0x prefix).
+readonly GRID_REQUIRED_SUBDEVICES=(
+  "27b8:1733" # L4:L4-3Q
+  "27b8:1735" # L4:L4-6Q
+  "27b8:1737" # L4:L4-12Q
+)
+
+# Return 0 iff every attached NVIDIA GPU's device id is present in the given
+# supported-devices file. Returns 1 if any device id is absent from the file.
+function devices_supported_by() {
+  local support_file="${1}"
+  if [[ ! -f "${support_file}" ]]; then
+    echo >&2 "resolve: supported-devices file not found: ${support_file}"
+    exit 1
+  fi
+  local pci_class_code nvidia_device_id
+  for pci_class_code in "${PCI_CLASS_CODES[@]}"; do
+    for nvidia_device_id in $(lspci -n -mm -d "${NVIDIA_VENDOR_ID}::${pci_class_code}" | awk '{print $4}' | tr -d '"' | tr '[:lower:]' '[:upper:]'); do
+      if ! grep "^0x${nvidia_device_id}\s" "${support_file}"; then
+        return 1
+      fi
+    done
+  done
+  return 0
+}
+
+function device-requires-grid() {
+  local nvidia_grid_subdevice nvidia_subdevice
+  for nvidia_grid_subdevice in "${GRID_REQUIRED_SUBDEVICES[@]}"; do
+    for nvidia_subdevice in $(lspci -n -mm -d "${NVIDIA_VENDOR_ID}:" | awk '{print $4":"$7}' | tr -d '"'); do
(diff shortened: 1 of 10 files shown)


templates/al2023/runtime/gpu/resolve-nvidia-driver.sh
```diff
@@ -0,0 +1,147 @@
...
+  fi
+
+  if device-requires-grid; then
+    if [[ -d "${NVIDIA_TREE_ROOT}/${tree}/flavors/grid" ]]; then
+      # GRID is currently only supported on x86_64
+      flavor="grid"
+    else
+      flavor="open"
+      echo >&2 "warn: GRID-required device present but no grid subtree in ${tree}; using open"
+    fi
```
mselim00: maybe should exit 1?

the `device-requires-grid`, as the name implies, returns true iff the device should only work with the GRID driver, not necessarily if _can_ work with the GRID driver. some instance types, like those in the g6 family, can be supported by either the GRID or open drivers, but the function would currently return false for them

it doesn't seem clearly better to hard fail by exiting or try with the open driver, both are likely to end up with unusable GPU(s). the former gives a clearer, loud failure mode. the latter leaves "a chance" of creating a usable state, particularly if we add to the list where GRID is preferred but open is functional

templates/al2023/runtime/gpu/load-nvidia-kernel-modules.sh
```diff
@@ -0,0 +1,99 @@
...
+  else
+    echo >&2 "load: gdrdrv did not load (open flavor, optional)"
+  fi
+fi
+
+# rpm db registration can carry a big time penalty; to optimize first-boot time
+# we keep this at the end so it's out of the critical path for functionality
+readonly COMMITTED_SENTINEL="${EKS_NVIDIA_STATE_DIR}/driver-committed"
+if [[ ! -f "${COMMITTED_SENTINEL}" ]]; then
+  rpm -i --justdb --noscripts --nodeps --nodigest --nosignature --replacefiles "${TREE}"/.rpms/*.rpm
```
mselim00: we're currently extracting all RPMs (for all supported flavors) to a top-level `.rpms` dir in the tree. since the RPM for open and proprietary write to the same paths, we need the additional `--replacefiles` flag so the `rpm` cli tolerates this and just lets the last one win

https://github.com/awslabs/amazon-eks-ami/blob/64b317954fab39f026a83e69fb1006b95e4e4fcc/templates/al2023/helpers/nvidia-kmods.sh#L130-L132

if we keep common user-space RPMs top-level and move the flavor-specific stuff to the flavor sub-tree (should just be the driver), we should be able to drop that flag

templates/al2023/runtime/gpu/load-nvidia-kernel-modules.sh
```diff
@@ -0,0 +1,99 @@
...
+  echo >&2 "load: no ${FLAVOR_SUBTREE}/lib/modules/${KERNEL_VERSION}"
+  exit 1
+fi
+
+VERSION="$(cat "${TREE}/.version")"
+readonly VERSION
+
+readonly LOADED_SENTINEL="${TREE}/.loaded"
+if [[ ! -f "${LOADED_SENTINEL}" ]]; then
+  for dir in "${FLAVOR_SUBTREE}/lib/modules/${KERNEL_VERSION}"/*/; do
```
kummatty: We are explicit about the path `/lib/modules/{kernel}/extra` during build. Maybe we don't need this for loop?

templates/al2023/runtime/gpu/load-nvidia-kernel-modules.sh
kummatty: >  move the flavor-specific stuff to the flavor sub-tree (should just be the driver), we should be able to drop that flag

yes, lets do this

templates/al2023/runtime/gpu/resolve-nvidia-driver.sh
kummatty: The current AMI will fail load for grid if the driver does not exist right? exit 1 seems correct to me as well

mselim00: I realized resolve shouldn't worry about this at all - it can do it's job completely without worrying about this. load already will fail if it tries to load that flavor's tree but the tree's not present, that's where the failure should be

templates/al2023/runtime/gpu/load-nvidia-kernel-modules.sh
mselim00: replaced the loop with just /extra

templates/al2023/runtime/gpu/load-nvidia-kernel-modules.sh
mselim00: revised for that

templates/al2023/runtime/gpu/set-nvidia-clocks.service
```diff
@@ -11,4 +11,4 @@ ExecStart=/usr/bin/set-nvidia-clocks
 RemainAfterExit=true
 
 [Install]
-WantedBy=multi-user.target
+WantedBy=multi-user.target
```
kummatty, line 14: We may need this service to be ordered after setup as well

kummatty (approved):

PR #2793, opened 2026-08-26
feat: speed up node boot

fletcherw:
Speed up node boot by
1. prefetching nodeadm and nodeadm-internal binaries
2. stripping symbols from nodeadm and nodeadm-internal to reduce size (saves about 25MB for nodeadm and 6MB for nodeadm-internal).

### DIFF
--- a/nodeadm/Makefile
+++ b/nodeadm/Makefile
@@ -71,14 +71,18 @@ test: crds generate fmt vet ## Run go test against code.
 .PHONY: test-e2e
 # the test infra container needs a linux executable
 test-e2e: GOOS=linux
-test-e2e: build ## Run e2e tests.
+test-e2e: release ## Run e2e tests.
 	test/e2e/run.sh
 
 ##@ Build
 
 .PHONY: build
-build: ## Build binary.
+build: ## Build debug binary.
 	GOOS=$(GOOS) go build -o $(LOCALBIN)/ ./cmd/...
+
+.PHONY: release
+release: ## Build release binary
+	GOOS=$(GOOS) go build -trimpath -ldflags="-s -w" -o $(LOCALBIN)/ ./cmd/...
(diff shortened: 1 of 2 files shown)


nodeadm/Makefile
```diff
@@ -77,9 +77,13 @@ test-e2e: build ## Run e2e tests.
 ##@ Build
 
 .PHONY: build
-build: ## Build binary.
+build: ## Build debug binary.
 	GOOS=$(GOOS) go build -o $(LOCALBIN)/ ./cmd/...
 
+# Release build
```
fletcherw, line 83: I think we should add some e2e tests against the release binary as well (or maybe switch all tests to be against it? I'm not sure)

fletcherw:
benchmarked against a custom AMI built without the changes:
baseline: median: 67.5s, n=10
changed: median: 49.0s, n=10

I think the absolute numbers are not as relevant but there's clearly an improvement from the changes.

nodeadm/Makefile
mselim00, line 83:
```suggestion
.PHONY: release
```

mselim00: yeah I think we should switch e2e to use the release binary since that's what the AMI builds use

mselim00 (commented):
nice improvement! just some minor things, excited to see this rollout

Youssef-Beltagy:
/ci

I look forward to deploying this. 1.34 failed with a tcp timeout, so I'm retrying the workflow.

Youssef-Beltagy (approved):

nodeadm/Makefile
fletcherw: Done, I think.

mselim00 (approved):
lgtm 🚀

PR #2795, opened 2026-08-28
feat(nvidia): add versioned nvidia services at boot

mselim00:
**Issue #, if available:**

**Description of changes:**
Adds logic to support `nvidia-fabricmanager` and `nvidia-persistenced` services. For the first boot, the setup service copies their service files from their place in the version tree into `/etc/systemd/system`, started, and enabled. On subsequent boots, the setup service does not need to do anything for the services as `systemd` will automatically start them after its completion, thanks to the prior enable.

### DIFF
--- a/templates/al2023/runtime/gpu/setup-nvidia.sh
+++ b/templates/al2023/runtime/gpu/setup-nvidia.sh
@@ -96,3 +96,20 @@ if [[ ! -f "${COMMITTED_SENTINEL}" ]]; then
     echo "kernel=${KERNEL_VERSION}"
   } > "${COMMITTED_SENTINEL}"
 fi
+
+readonly DAEMONS_INSTALLED_SENTINEL="${TREE}/.daemons-installed"
+if [[ ! -f "${DAEMONS_INSTALLED_SENTINEL}" ]]; then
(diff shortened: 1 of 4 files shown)


templates/al2023/runtime/gpu/setup-nvidia.sh
```diff
@@ -96,3 +96,26 @@ if [[ ! -f "${COMMITTED_SENTINEL}" ]]; then
     echo "kernel=${KERNEL_VERSION}"
   } > "${COMMITTED_SENTINEL}"
 fi
+
+readonly DAEMONS_INSTALLED_SENTINEL="${TREE}/.daemons-installed"
+if [[ ! -f "${DAEMONS_INSTALLED_SENTINEL}" ]]; then
+  for daemon in nvidia-persistenced nvidia-fabricmanager; do
```
kummatty: I think we can still copy all the systemd services to the host, and enable/start only fb and persistenced. This will bring parity with the current setup.

kummatty (approved):

templates/al2023/runtime/gpu/setup-nvidia.sh
mselim00: true, made this change

PR #2798, opened 2026-09-01
feat(nvidia): installing nvidia-imex

kummatty:
**Description of changes:**

Installing `nvidia-imex` as part of userspace programs. The systemd service as part of the package, will be automatically copied over to the host based on the works of the previous PRs.

This PR also copies over config files in /etc/ for xorg, x11 and xdg along with imex.

Also fixing a bug in copying to `/lib/modules/extra`. Earlier it was being copied over to `/lib/modules/extra/extra`. It looked like the kernel modules were still being recognized even with the misformed path, but maintaining right pattern here as well as for the gsp firmware path.

Also, validate.sh now requires the env variable `ENABLE_NVIDIA_GDRCOPY_DRIVER` which was not passed in `template.json` causing build failures.

### DIFF
--- a/templates/al2023/helpers/nvidia-userspace.sh
+++ b/templates/al2023/helpers/nvidia-userspace.sh
@@ -12,6 +12,7 @@ readonly NVIDIA_USERSPACE_PACKAGES=(
   nvidia-driver
   nvidia-driver-cuda
   nvidia-fabricmanager
+  nvidia-imex
   nvidia-libXNVCtrl-devel
   nvidia-persistenced
   nvidia-settings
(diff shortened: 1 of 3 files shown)


mselim00 (approved):

templates/al2023/helpers/nvidia-userspace.sh
```diff
@@ -104,7 +105,14 @@ function extract-version-specific-rpms() {
...
   for RPM in "${VERSION_SPECIFIC_RPMS[@]}"; do
-    echo "  $(basename "${RPM}")"
+    RPM_FILENAME=$(basename "${RPM}")
+    # The flavor subtrees already stage the kmod rpms. A copy here would put both flavors'
+    # in one rpmdb registration, where they conflict over the same /usr/src paths.
+    if [[ "${RPM_FILENAME}" == kmod-nvidia-* ]]; then
+      echo "  skipping ${RPM_FILENAME}, owned by the flavor subtree"
+      continue
+    fi
+    echo "  ${RPM_FILENAME}"
```
mselim00, line 121: is this potentially a bit too generic? it's safe with the current package set we have, but take the `nvidia-kmod`package for example. if its name had been the flip, it'd be incorrectly caught in this pattern

I'm wondering if we could instead check if the full exact rpm is in any of the 3 flavor subtrees' directory and skip only those?

mselim00 (approved):

PR #2803, opened 2026-09-04
feat(nvidia): installing grid license server

kummatty:
**Description of changes:**

The GRID run file archive comes with kernel source as well as a userspace stack which includes both shared object libraries as well as its vGPU license server stack.
All of the shared object libraries have its counterparts in the rpm's we fetch for a versions userspace programs. 

Other files like gtk2, topologyd are inert or not needed in AL2023 AMI's. These leaves the license server stack.

In this PR we harvest the license server stack components
```
gridd.conf.template 
nvidia-gridd
nvidia-gridd.service 
```

These are technically userspace components that can exist in the host irrespective of the instance being vGPU, so placing them within the tree's existing paths where mount and copy mechanism's put them into the stack.

Two additional bug fixes,
1. Copy of /etc files during boot should skip if the files already exist. Since we are dealing with configuration files, there is a high chance a user could have modified it, which we would rewrite on a reboot if left unguarded.
2. restorecon failures should not be swallowed. 

One Refactor,
Collapsed a method which validated grid kernel boot modules into an existing generic method.

**Testing Done**

Tested the AMI on a g6f instance
```
nvidia-smi -q | grep -A3 'vGPU Software'
    vGPU Software Licensed Product
        Product Name                                   : NVIDIA RTX Virtual Workstation
        License Status                                 : Licensed (Expiry: N/A)
    GPU Recovery Action                                : None
```

Attached is an AI generated file which shows why each file in the .run archive was skipped

[grid-findings.md](https://github.com/user-attachments/files/31814662/grid-findings.md)

### DIFF
--- a/templates/al2023/provisioners/validate.sh
+++ b/templates/al2023/provisioners/validate.sh
@@ -76,28 +76,48 @@ fi
 #############################
 ### nvidia drivers #####
 #############################
-NVIDIA_DRIVER_MODULES=(nvidia nvidia-drm nvidia-modeset nvidia-peermem nvidia-uvm)
+NVIDIA_DRIVER_MODULES=(nvidia nvidia-drm nvidia-modeset nvidia-uvm)
 KERNEL_RELEASE=$(uname -r)
 
-# GRID is harvested by glob rather than from a dkms manifest, so confirm the tree
-# ended up with exactly the driver's modules.
-validate_nvidia_grid_modules() {
+validate_nvidia_boot_modules() {
   local tree=$1
-  local extra_dir="/opt/nvidia/${tree}/flavors/grid/lib/modules/${KERNEL_RELEASE}/extra"
-  local module_name harvested
+  local flavor=$2
+  local extra_dir="/opt/nvidia/${tree}/flavors/${flavor}/lib/modules/${KERNEL_RELEASE}/extra"
+  local expected=("${NVIDIA_DRIVER_MODULES[@]}")
+  local module_name
 
-  for module_name in "${NVIDIA_DRIVER_MODULES[@]}"; do
-    if [ ! -f "${extra_dir}/${module_name}.ko" ]; then
+  # gdrdrv is only harvested on the open flavor, and only when the build enabled it.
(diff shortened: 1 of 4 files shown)


templates/al2023/provisioners/validate.sh
```diff
@@ -76,28 +76,47 @@ fi
...
+  local tree=$1
+  local path
+
+  for path in "${NVIDIA_GRID_USERSPACE[@]}"; do
+    if [ ! -f "/opt/nvidia/${tree}/${path}" ]; then
+      echo "/opt/nvidia/${tree}/${path} is missing; the GRID runfile userspace was not harvested"
+      exit 1
+    fi
+  done
 }
```
mselim00, line 121: nit: seeing `usr/bin/nvidia-gridd`, etc. without the leading `/` looks odd, it should be ok to include it and bash should cannonicalize the dual slash that's expanded in the if condition to a single slash

templates/al2023/runtime/gpu/setup-nvidia.sh
```diff
@@ -39,21 +39,21 @@ VERSION="$(cat "${TREE}/.version")"
 readonly VERSION
 
 # copy /etc/ files including modprobe.d and configs
-cp -a --reflink=auto "${TREE}/etc/." /etc/
+cp -a -n --reflink=auto "${TREE}/etc/." /etc/
```
mselim00, line 44: what's the rationale with preferring what's already on the host? that should be a code comment

also, `-n`'s a deprecated flag, should this be `--update=none`?

mselim00 (approved):

templates/al2023/runtime/gpu/setup-nvidia.sh
kummatty: Added the comment. Since /etc/ contains configuration files that are meant to be updated, I didn't want us to replace any modified files at each reboot. 

Yes, -n is deprecated. But --update=none is available only from coreutils 9.3+. AL2023 is at 8.32

PR #2808, opened 2026-09-07
docs: document NVIDIA 595 AMI instance type restrictions

kappa8219:
Documents the EC2 AMI instance type specification command for custom NVIDIA 595 AMIs.

The command blocks P3, P3dn, and G6f instance families, which are incompatible with NVIDIA driver 595.

### DIFF
--- a/doc/usage/g7-ami.md
+++ b/doc/usage/g7-ami.md
@@ -25,6 +25,26 @@ This produces an AMI with NVIDIA 595 drivers suitable for G7 instances. The buil
 
 The build takes approximately 45-60 minutes and outputs the AMI ID upon completion.
 
+### Restricting incompatible instance types
+
+To prevent this AMI from launching on instance types that are incompatible with
+driver 595, set an [AMI instance type specification](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ami-allowed-instance-types.html)
+after the build completes. Replace `ami-xxxx` with the AMI ID produced by the
+build:
+
+```bash
+aws ec2 replace-image-instance-type-specification \
+  --region us-east-1 \
+  --image-id ami-xxxx \
+  --instance-type-specification '{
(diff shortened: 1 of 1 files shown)


kappa8219:
@sgundapu I enriched this doc a bit.

csplinter:
Support for G7 added in this AMI release https://github.com/awslabs/amazon-eks-ami/releases/tag/v20260917

The EKS AMIs now include two NVIDIA driver versions and select the correct one based on the instance type at first boot.

Additional details in the docs here: https://docs.aws.amazon.com/eks/latest/userguide/ml-eks-optimized-ami.html#eks-amis-nvidia-al2023

Let me know if this PR can be closed now. We will update that g7-ami.md as well.

kappa8219:
It works, both for g7 and g6f

G7
Screenshot From 2026-09-22 15-14-24

G6f:
image

so incompatibiltity mitigated with driver version switch for those 2 types.

What AI advises for better test coverage for new script in particular:

```
  What exists:
  - validate.sh (modified in the commit) — runs at AMI build time (packer provisioner). It validates that both lts and pb trees are correctly built: driver version files, all kernel
    module flavors (open, proprietary, grid), and GRID userspace. This is a solid sanity check.
  - e2e tests for containerd NVIDIA runtime config — but those test the containerd plugin, not driver resolution.
  
  What's missing:
  - No unit tests for resolve-nvidia-driver.sh — the critical new bash script that picks the driver tree/flavor at boot. It uses lspci and reads DMI product name, neither of which is
    mocked/stubbed in any test.
  - No CI jobs run against GPU instance types (the CI workflows only cover linting and nodeadm Go tests on PR auto-trigger).
  - The actual boot-time resolution logic (lts open → pb open → lts proprietary) is only tested when an AMI is actually deployed on a GPU node.
```

Issue #2417, opened 2025-09-05
feat(al2023): NVIDIA driver version control


jonathanmhurley:
**What would you like to be added**:
Add a feature to expand control options for NVidia driver version.
Currently, in order to control adoption of an NVidia driver version, it is necessary to statically configure an AMI from pre-update; this necessitates both code management as well as monitoring AMI releases to determine driver version.

One option might be to produce multiple AMIs, one per NVidia driver version, which would fit seamlessly into the standard management capabilities for EKS (e.g. Karpenter amiSelectorTerms and drift management) but could become cumbersome over time.

Another option might be to statically associate AMI tags to upstream NVidia driver branches (e.g. PB, NFB, etc) which would still produce multiple AMIs but a predictable number per AMI release and similarly fit within existing drift controls.

**Why is this needed**:
At change points for NVidia driver versions, it would be convenient to be able to delay or control adoption, especially for associated CUDA changes such as here: https://github.com/awslabs/amazon-eks-ami/pull/2357
This allows for phased testing, rollout, and rollback options without being statically associated to specific EKS AMI versions.

cartermckinnon:
We are unlikely to maintain “branches” of the AMI for different NVIDIA driver versions. This would significantly expand the matrix of what needs to be supported (and tested), and the deprecation story would become complicated (and would be out of our control).

Conceptually, the GPU Operator is a way to do this, though we don’t necessarily love the tradeoffs (such as much longer boot times). You’d also need to take on the testing.

We do consider CUDA compatibility when changing driver versions, I completely understand that a breaking change there could be a huge pain.

jonathanmhurley:
Thanks Carter!

To confirm - this is the operator that you mentioned, correct: https://docs.nvidia.com/datacenter/cloud-native/gpu-operator/latest/getting-started.html

cartermckinnon:
Yep, that’s it. I’m not aware if GPU-O properly supports Amazon Linux these days, it did not in the past.

bryantbiggs:
not natively, no. it may or may not work depending on which version of the operator and OS is used, usually resulting in glibc errors if the versions are incompatible. 

https://github.com/search?q=repo%3ANVIDIA%2Fgpu-operator+amazon+linux&type=issues

looks like support for AL2023 was abandoned without any details https://github.com/NVIDIA/gpu-operator/pull/1056

cartermckinnon:
I don’t think support is blocked at this point, the work just needs to be done.

[...]

PR #2809, opened 2026-09-07
feat: configure AMI instance type restrictions

kappa8219:
Adds optional `supported_instance_types` and `unsupported_instance_types` Makefile variables. After Packer creates the AMI, the build applies an EC2 AMI instance type specification with an allow-list, block-list, or both.

https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ami-allowed-instance-types.html

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,6 +72,7 @@ lint-code: ## Check the source files for syntax and format issues
 	hack/lint-space-errors.sh
 
 PACKER_BINARY ?= packer
+AWS_CLI ?= aws
 PACKER_TEMPLATE_DIR ?= templates/$(os_distro)
 PACKER_TEMPLATE_FILE ?= $(PACKER_TEMPLATE_DIR)/template.json
 PACKER_DEFAULT_VARIABLE_FILE ?= $(PACKER_TEMPLATE_DIR)/variables-default.json
@@ -104,6 +105,13 @@ validate: ## Validate packer config
 k8s: validate ## Build default K8s version of EKS Optimized AMI
 	@echo "Building AMI [os_distro=$(os_distro) kubernetes_version=$(kubernetes_version) arch=$(arch) $(if $(enable_accelerator),enable_accelerator=$(enable_accelerator))]"
 	$(PACKER_BINARY) build -timestamp-ui -color=false $(PACKER_ARGS) $(PACKER_TEMPLATE_FILE)
+ifneq ($(strip $(unsupported_instance_types)),)
+	@ami_id="$$(jq -er '.builds[] | select(.builder_type == "amazon-ebs") | .artifact_id | split(":")[1]' "$(ami_name)-manifest.json")"; \
+	$(AWS_CLI) ec2 replace-image-instance-type-specification \
+		--region "$(aws_region)" \
+		--image-id "$$ami_id" \
+		--instance-type-specification 'UnsupportedInstanceTypes=$(unsupported_instance_types)'
+endif
 
 .PHONY: lint-docs
 lint-docs: ## Lint the docs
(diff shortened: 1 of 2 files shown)


mselim00:
hey @kappa8219, can you add a bit more on what this PR is attempting to accomplish. namely, why do it here?

this API call is not something we would want to make in AMIs we build as part of official releases, and it's one that any custom AMI builder can make independently after invoking the build recipe. My push back on this is because this repository is mainly intended to supply a packer build recipe to get AMIs like the official released ones, but this PR adds something totally independent

kappa8219:
> hey @kappa8219, can you add a bit more on what this PR is attempting to accomplish. namely, why do it here?
> 
> this API call is not something we would want to make in AMIs we build as part of official releases, and it's one that any custom AMI builder can make independently after invoking the build recipe. My push back on this is because this repository is mainly intended to supply a packer build recipe to get AMIs like the official released ones, but this PR adds something totally independent

hello @mselim00, IMHO this adds posibility(without obligation to use) to mark AMIs eligable for specific EC2 types. Which is usefull to get error earlier. Ex. all ARM AMIs used by mistake with AMD instance will trigger EC2(s) start just to see "unsupported container image" spending money of customers and resources of cloud provider in vain.

Also see history with GPU AMIs splitting by driver version. #2417

mselim00:
> Ex. all ARM AMIs used by mistake with AMD instance will trigger EC2(s) start just to see "unsupported container image" spending money of customers and resources of cloud provider in vain.

The error returned in that case is at `EC2:RunInstances` API call time, the architecture is a first-class attribute of both the AMI and the Instance Type, so the EC2 API call is able to reject an invalid combination without launching an instance. In any case, attempting to enumerate all the supported or unsupported instance types by architecture for each image would be fragile endeavor, as new instance types are constantly released.

> Also see history with GPU AMIs splitting by driver version.

Our official NVIDIA AMIs have historically supported the full set of instance types supported by EC2. That issue discusses the potential for allowing users to choose a specific major version, but that was in the context of unlocking new features from newer versions, not basic functionality. There is an incompatibility with g7 instance types noted in [g7-ami.md](https://github.com/awslabs/amazon-eks-ami/blob/main/doc/usage/g7-ami.md), but the [develop/nvidia-driver](https://github.com/awslabs/amazon-eks-ami/tree/develop/nvidia-driver) feature branch is actively being worked on to address this.

> IMHO this adds posibility(without obligation to use) to mark AMIs eligable for specific EC2 types. 

I agree that it's optional and some users may have value for it. My disagreement is whether or not it makes more sense here or for users to call individually. For deciding that, I generally look to whether or not there is potential for it to be used in the official releases, and I do not see that being the case for this one at this time

Thank you for the contribution but, if there is no other motivation here, I would encourage you to add this to your existing process instead of this project. In the future, I would also encourage you to open a GitHub issue to discuss an issue prior to raising a PR

kappa8219:
> > There is an incompatibility with g7 instance types noted in [g7-ami.md](https://github.com/awslabs/amazon-eks-ami/blob/main/doc/usage/g7-ami.md), but the 

Not only G7, older types like G6F are already not supported with newer cuda driver(see the page). So it is diverged already, working on new driver will not change it.



> > enumerate all the supported or unsupported instance types by architecture for each image would be fragile endeavor

agree it is toil, but for now no other options seen

mselim00:
> older types like G6F are already not supported with newer cuda driver(see the page)

Can you share the page you're referring to on this? 

It's true that the newest NVIDIA driver versions do not support some of the oldest EC2 instance types. The feature branch I was referring to is actually aiming to address this by supporting multiple versions of the NVIDIA driver within a single AMI, so that the correct version for the hardware can be loaded dynamically.

kappa8219:
> > older types like G6F are already not supported with newer cuda driver(see the page)
> 
> Can you share the page you're referring to on this? 




You shared it first.

> 
> It's true that the newest NVIDIA driver versions do not support some of the oldest EC2 instance types. The feature branch I was referring to is actually aiming to address this by supporting multiple versions of the NVIDIA driver within a single AMI, so that the correct version for the hardware can be loaded dynamically


Will see if it will be possible, honestly amii that requires post start actions(even via cloudinit) to chose correct drivers sounds like a dirty fix.

PR #2804, opened 2026-09-04
refactor: Use Hashicorp's HCL2 for the AL2023 Template + added an init target that helps setup the amazon plugin

Youssef-Beltagy:
**Description of changes:**

1. This PR migrates AL2023's template to use Hashicorp's HCL2 format. HCL2 provides more flexibility than json files and allows us to break the template into parts (which will be helpful soon).
2. Added an init target that helps setup the amazon plugin
3. I removed the hacky aws environment variable passing

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

**Testing Done**

1. I manually reviewed the changes
2. I had Kiro with Claude 5.0 review the changes
4. I built an AMI before the changes and after. Then compared the generated `version-info.json` files and observed no difference.
5. I was able to use `packer inspect templates/al2023/` after the changes (it didn't work before).
 

### Packer Inspect after the Changes
```
Packer Inspect: HCL2 mode

> input-variables:

var.ami_component_description: ""
var.ami_description: "<unknown>"
var.ami_name: "<unknown>"
var.ami_regions: "<unknown>"
var.ami_users: "<unknown>"
var.arch: "<unknown>"
var.associate_public_ip_address: "<unknown>"
var.aws_region: "<unknown>"
var.binary_bucket_name: "<unknown>"
var.binary_bucket_region: "<unknown>"
var.containerd_version: "<unknown>"
var.creator: "ybeltagy"
var.custom_endpoint_ec2: "<unknown>"
var.enable_accelerator: "<unknown>"
var.enable_efa: "<unknown>"
var.enable_fips: "<unknown>"
var.enable_nvidia_gdrcopy_driver: "<unknown>"
var.encrypted: "<unknown>"
var.iam_instance_profile: "<unknown>"
var.install_containerd_from_s3: "<unknown>"
var.instance_type: "<unknown>"
var.kms_key_id: "<unknown>"
var.kubernetes_build_date: "<unknown>"
var.kubernetes_version: "<unknown>"
var.launch_block_device_mappings_volume_size: "<unknown>"
var.nodeadm_build_image: "<unknown>"
var.nvidia_driver_major_version: "<unknown>"
var.nvidia_gdrcopy_driver_version: "<unknown>"
var.nvidia_grid_runfile_bucket_name: "<unknown>"
var.nvidia_repository_url: "<unknown>"
var.pause_container_image: "<unknown>"
var.remote_folder: "<unknown>"
var.runc_version: "<unknown>"
[...]

PR #2810, opened 2026-09-08
fix: fixed test flakiness caused by gitlab.alpinelinux.org/alpine/go returning 418

Youssef-Beltagy:
This PR fixes the test flakiness we have been observing in the past few PRs:
- https://github.com/awslabs/amazon-eks-ami/pull/2804
- https://github.com/awslabs/amazon-eks-ami/pull/2793

I updated `GOPROXY` for github actions  to use `https://proxy.golang.org` before falling back to `direct`. This should reduce our direct calls to `gitlab.alpinelinux.org/alpine/go`.


fletcherw (approved):
nice

PR #2811, opened 2026-09-08
fix(nvidia): cp src labels were being applied to /etc

kummatty:
**Description of changes:**

A bug fix in how we copied `/${TREE}/etc` to `/etc`. We were applying the source label to `/etc` as well, which could break SELINUX usecases, since commands like `ldconfig` which writes to `/etc/ld.so.cache` etc will break.

Tested with selinux mode as enforcing.

### DIFF
--- a/templates/al2023/runtime/gpu/setup-nvidia.sh
+++ b/templates/al2023/runtime/gpu/setup-nvidia.sh
@@ -41,7 +41,8 @@ readonly VERSION
 # copy /etc/ files including modprobe.d and configs. this runs on every boot, and /etc holds
 # config files, so -n keeps it from replacing a config a user changed after the first boot.
 # TODO: -n is deperecated. Move to --update=none when coreutils 9.3+ is available.
-cp -a -n --reflink=auto "${TREE}/etc/." /etc/
+shopt -s dotglob
+cp -a -n --reflink=auto "${TREE}"/etc/* /etc/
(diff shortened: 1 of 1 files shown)


mselim00 (commented):
would this implied ordering actually be safe? wouldn't nvidia persistenced or fabric manager have dependencies on shared libraries supplied by the driver's installation that need to be loaded w/ `ldconfig`?

also, contrary to how it looks in code, it won't actually change anything about the start times. the daemons are "started" in the background with the `systemctl start --no-block` call, but because of their `After` ordering with `nvidia-setup` service, they actually shouldn't be started until the full `setup-nvidia.sh` script completes anyway

kummatty:
> would this implied ordering actually be safe? wouldn't nvidia persistenced or fabric manager have dependencies on shared libraries supplied by the driver's installation that need to be loaded w/ `ldconfig`?
> 
> also, contrary to how it looks in code, it won't actually change anything about the start times. the daemons are "started" in the background with the `systemctl start --no-block` call, but because of their `After` ordering with `nvidia-setup` service, they actually shouldn't be started until the full `setup-nvidia.sh` script completes anyway

Ah you're right. Forgot about the After clause. 
Let's target boot time regression as a separate effort. Amended this PR by adding only the copy label bug on /etc/

mselim00 (approved):

PR #2818, opened 2026-09-14
fix(nvidia): enable the correct module stream for loaded version

mselim00:
**Issue #, if available:**

**Description of changes:**

Configures the stream for the `nvidia-driver` dnf module to look at the `{major-ver}-open` stream of the selected driver major version. This is consistent with the behavior of existing AMI releases: 
https://github.com/awslabs/amazon-eks-ami/blob/2d5832cf02d6c0beda1f631b136631fa196ec251/templates/al2023/provisioners/install-nvidia-driver.sh#L49-L50

This ensures that any further `dnf` operations at boot are bound to the same major version stream that would have been used if the AMI only supported that single major version. The in-place editing of the file rather than using `dnf module enable` is intended to avoid the repository metadata requirements, since neither network access nor the presence of the dnf metadata cache can be guaranteed.


templates/al2023/runtime/gpu/install-nvidia-packages.sh
```diff
@@ -16,9 +16,19 @@ readonly FLAVOR
...
 shopt -s nullglob # the GRID flavor has no specific RPMs to expand to
 readonly RPMS_TO_REGISTER=("${TREE}"/.rpms/*.rpm "${FLAVOR_SUBTREE}"/.rpms/*.rpm)
 shopt -u nullglob
 if ((${#RPMS_TO_REGISTER[@]} > 0)); then
   rpm -i --justdb --noscripts --nodeps --nodigest --nosignature "${RPMS_TO_REGISTER[@]}"
 fi
+
+# enable the correct NVIDIA repo module stream for the chosen driver version
+# TODO: consider selecting the {ver}-dkms stream for the proprietary driver
+sed -i "s/DRIVER_TREE_VERSION_PLACEHOLDER/${TREE_VERSION}-open/g" /etc/dnf/modules.d/nvidia-driver.module
```
kummatty: .version is fully qualified. Let's replace with major version prefix.

kummatty (approved):

templates/al2023/runtime/gpu/install-nvidia-packages.sh
mselim00: good catch, fixed that

PR #2819, opened 2026-09-15
feat(nvidia): bake lts and pb driver trees, resolve one at boot

kummatty:
### Multi-version NVIDIA driver support for AL2023

Replaces the single-version NVIDIA driver install with two prebuilt driver trees baked into the AMI, one of which is selected at boot based on the attached hardware.

### Why

The AMI previously baked one driver major version and built its kernel module with DKMS on every boot. That meant one AMI could not serve both older GPUs still needing an LTS driver and newer GPUs needing a production-branch driver.

#### Build
Two self-contained RPM trees under /opt/nvidia/{lts,pb}, each a mirror of the parts of `/` a driver occupies:

- Each tree resolves one exact version.
- All kernel module flavors are prebuilt into the tree
- The userspace closure is partitioned: packages identical across both versions are installed on the host normally, the version-specific ones are extracted into their version specific RPM tree.

#### Boot

- `nvidia-driver-resolve.service` picks the tree and flavor based on the booted machine.
- `usr-{bin,lib64,share}.mount` overlay the tree's `/usr` over the base image
- `nvidia-setup.service` copies the flavor's modules and GSP firmware into place, loads the kernel modules and starts the systemd daemons.
- `nvidia-package-install.service` registers the tree's rpms in the rpm database 

### Breaking change

`nvidia_driver_major_version` is replaced by `nvidia_driver_lts_version` and `nvidia_driver_pb_version`, defaulting to 580 and 595. Both must be set together.

Individual reviews: #2783, #2784, #2788, #2798, #2799, #2803, #2811, #2814.

---


kummatty:
/ci
+test --instance-types=g6.8xlarge --nodes=2
+workflow:k8s_versions 1.36

mselim00 (approved):
lgtm 🚀

kappa8219:
> /ci +build enable_accelerator=nvidia enable_efa=true +test --instance-types=g6.8xlarge --nodes=2 +workflow:k8s_versions 1.36

Not so many tests for such important case.

PR #2823, opened 2026-09-16
fix(build,nvidia): only configure module stream for non-iso

mselim00:
**Issue #, if available:**

**Description of changes:**

ISO partitions are configured to use the AL NVIDIA repository whereas non-ISO utilize the NVIDIA CUDA repositories directly. The latter configures module streams per major version and open/proprietary flavor, but the former does not. We currently attempt to enable a module stream regardless of the enabled repository, this updates it to only do so for non-ISO partitions where the configuration is actually available.


Youssef-Beltagy (approved):

templates/al2023/provisioners/validate.sh
```diff
@@ -244,9 +244,12 @@ if [[ "$ENABLE_ACCELERATOR" == "nvidia" ]]; then
     fi
   done
 
-  if ! grep -q 'stream=DRIVER_TREE_VERSION_PLACEHOLDER' /etc/dnf/modules.d/nvidia-driver.module; then
-    echo "/etc/dnf/modules.d/nvidia-driver.module missing placeholder for driver version"
-    exit 1
+  if ! [[ $(imds /latest/meta-data/services/partition) =~ ^aws-iso ]]; then
+    # ISO partitions use the AL NVIDIA repo, which does not include module streams
+    if ! grep -q 'stream=DRIVER_TREE_VERSION_PLACEHOLDER' /etc/dnf/modules.d/nvidia-driver.module; then
```
Youssef-Beltagy, line 249: nit: can you avoid nesting if statements?

kummatty (approved):

